PR #2, opened 2016-02-23
Avoid custom aliases when using sed

djmattyg007:
The nature of bash completions means they're sourced into a user's
existing shell. This means they have to play nicely with any
customisations the user makes. The biggest source of problems are
undoubtedly when a user has a custom alias for a core command.

If you alias 'sed' to 'sed -r' to enable modern regex syntax by default,
it can sometimes cause conflicts. This is the case with the
"_known_hosts_real" helper function provided by the bash_completion
project to all completion functions. Fortunately, this is easy to work
around: just look up sed in the user's path and use the full path to the
sed binary. This avoids the conflict altogether.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -847,7 +847,7 @@ _mac_addresses()
     # - ifconfig on FreeBSD: ether
     # - ip link: link/ether
     COMPREPLY+=( $( \
-        { LC_ALL=C ifconfig -a || ip link show; } 2>/dev/null | sed -ne \
+        { LC_ALL=C ifconfig -a || ip link show; } 2>/dev/null | command sed -ne \
         "s/.*[[:space:]]HWaddr[[:space:]]\{1,\}\($re\)[[:space:]].*/\1/p" -ne \
         "s/.*[[:space:]]HWaddr[[:space:]]\{1,\}\($re\)[[:space:]]*$/\1/p" -ne \
         "s|.*[[:space:]]\(link/\)\{0,1\}ether[[:space:]]\{1,\}\($re\)[[:space:]].*|\2|p" -ne \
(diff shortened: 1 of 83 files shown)


scop:
There are some problems with this PR.

First, the single instance of sed invocation touched by this change is far from the only one. The word `sed` occurs in 97 files:
```shell
$ git grep -wFl sed | wc -l
97
```

Second, we don't want to use `which`, that's not a standard bash builtin or command that can be assumed present. We have the `type` builtin in bash for that purpose.

Third, there should be no need to look up `sed` from the path in the first place. Using `command sed` instead of plain `sed` should do the trick.

djmattyg007:
Thank you for your feedback! I've clearly still got lots to learn.

The reason I only changed this one usage of sed is becayse it was the only one I've come across that doesn't work properly when using `sed -r`. I'm more than happy to change all usages of sed if you'd like.

djmattyg007:
I've updated the patch to use `command`. Let me know if you'd like me to update all usages of `sed` to be run this way.

scop:
Yes, please do.

djmattyg007:
I've added another commit that changes as many usages of `sed` as I could find.

djmattyg007:
I've reverted the awk-related change and pushed up the change as requested. That was an accident, though I did notice that there were several usages of `awk` that were not being run through `command`. I was wondering if you'd like another pull request to change those over as well?

Also, while making these changes, I found this project:
https://github.com/GArik/bash-completion
Is it connected to this project? Which one is used mainly by Linux distributions?

scop:
Merged, thanks.

Wrt awk, no, I don't want that until someone shows that there is a real problem to be solved, and while at it shows that our awk usage in bash-completion is portable enough so that people on some (mostly non-GNU) platforms do not need to alias awk to gawk to get bash-completion to work. In fact a couple of years ago our sed usage wasn't portable (and I'm not sure it still is, but we've made good progress), and at that time I would have rejected the "command sed" changes as well, because in some setups it was necessary to alias sed to gsed. And quite frankly, I do think that to an extent you're shooting yourself in the foot and a bit asking for trouble with the sed aliasing you're describing and doing, but I think this is not the first time I've heard of it, so...

GArik/bash-completion is the personal fork of a bash-completion contributor, made from the old alioth.debian.org repository apparently a few years back. scop/bash-completion is the "official" one.

Issue #3, opened 2016-02-24
ssh: please don't complete hosts found after Hostname in .ssh/config


ukleinek:
When there is a host configuration in ~/.ssh/config that's probably because some options should be applied when connecting to this host. So it's only sensible to complete names after "Host", but not after "Hostname" because when using the latter the options are not applied.

This is also reported to the Debian bugtracker as [Bug 689585](https://bugs.debian.org/689585)

PR #7, opened 2016-02-25
Don't complete hostnames found after Hostname in ~/.ssh/config

ukleinek:
When connecting to a host listed after Hostname the options in the
config file are not applied, so don't complete these names accordingly.

If you still want some hosts completed, just add an other entry with
just Host <yourhostname> at the end of the config file.

This closes scop/bash-completion#3 and https://bugs.debian.org/689585


scop:
Thanks!

PR #9, opened 2016-02-29
xmllint, xmlwf, xsltproc: Complete on Mallard *.page files

shaunix:
I use xmllint and xsltproc on Mallard page files almost daily. It's unpleasant with bash refusing to complete on those files. This patch makes it pleasant again.


scop:
Thanks!

PR #12, opened 2016-03-04
unrar: complete all filenames for non-first option

Arthur2e5:
unrar(1) gives a command syntax of:

    unrar <command> [-<switch 1> -<switch N>] archive [files...] [path...]

Therefore, it's a bad idea to alway assume that the user is going to need RARs.


Arthur2e5:
Expect the CI test for the part thing to fail. I forgot to include non-part files.

Arthur2e5:
Ugh… The test suite believes that it needs `unrar`. I guess this is a bug?

scop:
Not sure what you mean by requiring unrar being a bug. It is required in order for the completion to be installed (and thus testable), even if we wouldn't be actually using the executable for anything in this case, at the moment.

The unnecessary stylistic changes I noted previously seem to be still present.

The test case is missing a `sync_after_int` between tests.

The test files should be pre-generated and committed as fixtures, not created on the fly (without cleanup etc). See examples of this in the tree, e.g. the 7z test and related fixture.

scop:
Also, could you elaborate on the `set -f` stuff in the part filename completion? I don't think we're doing it anywhere else.

And I seem to remember that stuff passed to `compgen -d` should /not/ be quoted like it is now in the PR. `_filedir` does it that way, and I think the reason can be found in commit messages somewhere if interested.

Quite frankly, the part completion gives me chills, seems like awful lot of quirky code for a pretty rare corner case. Need to think about this some.

Arthur2e5:
The problem with parts is that when people use parts, twenty parts for a file are really common. These extra parts might then flood the whole completion output. So anyway I think I should try to avoid this.

(Oh regarding parts… I should add some _filedir r01.

PR #13, opened 2016-03-08
Fixe lvm completion related with physical volumes

lhwang:
The current lvm completion has problems when executing the commands related with physical volumes.
These 3 patches correct completion of all lvm commands related with physical volumes.


scop:
I'm not a lvm user myself, but the changes look sane on surface, so applied. Thanks.

lhwang:
> I'm not a lvm user myself, but the changes look sane on surface, so applied. Thanks.
Thank you very much:)

Best,
Liuhua Wang

PR #16, opened 2016-03-23
[FIX] BSD sed won't cut off ^t anymore

peshay:
With the old sed command, leading 't' would be removed by BSD sed. This sed works with BSD as well as GNU sed.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1567,7 +1567,7 @@ _known_hosts_real()
 
     # append any available aliases from config files
     if [[ ${#config[@]} -gt 0 && -n "$aliases" ]]; then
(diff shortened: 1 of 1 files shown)


scop:
Looks like this PR changes more than it documents -- in addition to sed syntax tweaking, it essentially reverts #7. I assume that's not on purpose?

peshay:
Sorry, I just copied the sed from another bash_completion where it worked. I will create a new pull request where I just change a minor thing to make it work.

scop:
Ok. Please also consider creating a test case for our test suite that demonstrates the failure and the fix.

peshay:
Would love to, but have no idea how to ... 
I just recognized on MacOS X, that Host entries like

`Host tasleadingletter`

would complete to 

`ssh asleadingletter `

PR #17, opened 2016-03-23
[FIX] BSD sed won't cut off ^t anymore

peshay:
Hope this fits better now. I tried it with gsed and bsd sed, worked same for both.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1567,7 +1567,7 @@ _known_hosts_real()
 
     # append any available aliases from config files
     if [[ ${#config[@]} -gt 0 && -n "$aliases" ]]; then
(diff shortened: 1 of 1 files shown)


peshay:
About a test case, I have no idea how to tell Travis-CI to use a BSD sed. Also this is my first pull request on github, so sorry if this request is not as perfect, but I'm still learning.

scop:
No need or way to try to use BSD on Travis. If you have access to a *BSD box, you can run the test suite yourself there. But I do understand that the test suite quite obscure. PR #18 actually contained a good explanation that I could use to construct a test case for this, but I'm getting confused which of the versions of the change/PR you are suggesting us to merge. Please clarify.

Also, when updating an existing PR, please keep the history clean by amending your commits and squashing them into one that does the one thing, and force-pushing to the existing PR (or opening a new one). We generate change logs from commit messages and this PR alone has already 4 commits which would end up looking weird/messy in the changelog.

peshay:
Okay... I start all over again... :)

PR #18, opened 2016-03-24
Patch 2

peshay:


scop:
Hm, could you elaborate what this is about? It seems to duplicate #17 to some extent, and the _filedir_xspec change is something I did 5 years ago, not sure why it's here in the first place?

peshay:
I just clicked the edit button at the other pull request and then I clicked pull request again ... don't know how this other commit came in then.

PR #19, opened 2016-03-24
[FIX] improvement for BSD sed

peshay:
Currently with BSD sed on MacOS X an .ssh/config entry like this
`Host tfirst-letter`
would complete to
`ssh first-letter`
So only the letter "t" in the beginning will be cut off.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1567,7 +1567,7 @@ _known_hosts_real()
 
     # append any available aliases from config files
     if [[ ${#config[@]} -gt 0 && -n "$aliases" ]]; then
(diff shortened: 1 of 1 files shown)


scop:
I managed to reproduce the issue on FreeBSD, and your change fixes it. But I don't quite understand why it happens in the first place. The sed implementation should not affect this in any way, the intent is just to pass the big regexp to sed and we're doing some quoting tricks **shell** level to get it through; it's the equivalent of

```shell
$ echo 'foo'"bar"'foo'
foobarfoo
```

So I don't understand why escaping the double quotes fixes it -- as far as sed is concerned, it shouldn't matter (as those quotes shouldn't end up being passed to it at all, but maybe they are and that's the problem...?). I'm going to add a different fix which makes my eyes hurt less, testing that would be appreciated.

PR #20, opened 2016-03-24
[FIX] improvement for BSD sed

peshay:
Currently with BSD sed on MacOS X an .ssh/config entry like this
Host tfirst-letter
would complete to
ssh first-letter
So only the letter "t" in the beginning will be cut off.


scop:
Looks like a duplicate of #19

PR #21, opened 2016-04-03
Add ip_netns

hhktony:
Add ip_netns

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -912,13 +912,6 @@ _kernel_versions()
     COMPREPLY=( $( compgen -W '$( command ls /lib/modules )' -- "$cur" ) )
 }
 
-# This function completes on available netns
-#
-_available_netns()
-{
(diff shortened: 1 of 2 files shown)


scop:
What other completions would `_available_netns` be useful for? If there's nothing at the moment, please just embed it in the ip completion's related `add|delete|exec|pids` case statement, no need to make it a global eagerly loaded function nor really a function in the first place in its current form.

Also, the `ls` should probably accompanied with `2>/dev/null` to avoid console trashing on systems where the dir is not present.

Finally, the list of subcommands does not match the `ip` on my system (Fedora 23). I have `list-id` instead of `list`, and I have `set` in addition.
```
$ ip netns help
Usage: ip netns list
       ip netns add NAME
       ip netns set NAME NETNSID
       ip [-all] netns delete [NAME]
       ip netns identify [PID]
       ip netns pids NAME
       ip [-all] netns exec [NAME] cmd ...
       ip netns monitor
       ip netns list-id
```

PR #22, opened 2016-04-08
tipc: add tipc completions

rical:
tipc is a user space configuration tool for the TIPC protocol which is
part of the mainline Linux Kernel. The tipc tool itself is part of the
iproute2 package which is present in most common Linux distributions.

### DIFF
--- a/completions/tipc
+++ b/completions/tipc
@@ -80,7 +80,7 @@ _tipc()
     COMPREPLY=()
 
     # Flags can be placed anywhere in the commandline
-    case "${words[$cword]}" in
+    case "$cur" in
     -*)
         COMPREPLY=( $( compgen -W '-h --help' -- $cur ) )
         return 0
--- a/completions/tipc
+++ b/completions/tipc
@@ -19,7 +19,7 @@ _tipc_bearer() {
     local media
 
     if _tipc_media $optind; then
-        return 1
+        return
     fi
 
     for ((i = 0; i < $cword; i++)); do
@@ -83,7 +83,7 @@ _tipc()
     case "$cur" in
     -*)
         COMPREPLY=( $( compgen -W '-h --help' -- $cur ) )
-        return 0
+        return
         ;;
     esac
 
@@ -98,7 +98,7 @@ _tipc()
 
             if [[ $cword -lt $optind+4 ]]; then
                 _tipc_bearer $optind
-                return 0
+                return
             fi
 
             for ((i = 0; i < $cword; i++)); do
@@ -115,7 +115,7 @@ _tipc()
                     declare -a params=("domain" "priority")
                     ;;
                 *)
-                    return 0
+                    return
                     ;;
             esac
 
@@ -124,7 +124,7 @@ _tipc()
             # kown key as value
             for i in ${params[@]}; do
                 if [[ $prev == $i ]]; then
-                    return 0
+                    return
                 fi
             done
 
@@ -138,13 +138,13 @@ _tipc()
             done
 
             COMPREPLY=( $( compgen -W '${params[@]}' -- $cur) )
-            return 0
+            return
             ;;
         disable)
             let optind++
 
             _tipc_bearer $optind
-            return 0
+            return
             ;;
         get)
             let optind++
@@ -154,7 +154,7 @@ _tipc()
             elif [[ $cword -ge $optind+1 ]]; then
                 _tipc_bearer $(($optind + 1))
             fi
-            return 0
+            return
             ;;
         set)
             let optind++
@@ -164,15 +164,15 @@ _tipc()
             elif [[ $cword -ge $optind+2 ]]; then
                 _tipc_bearer $(($optind + 2))
             fi
-            return 0
+            return
             ;;
         list)
-            return 0
+            return
             ;;
         esac
 
         COMPREPLY=( $( compgen -W 'enable disable set get list' -- $cur ) )
-        return 0
+        return
         ;;
     link)
         let optind++
@@ -186,7 +186,7 @@ _tipc()
             elif [[ $cword -ge $optind+1 ]]; then
                 _tipc_link $(($optind + 1)) "peers"
             fi
-            return 0
+            return
             ;;
         set)
             let optind++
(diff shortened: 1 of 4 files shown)


scop:
Some observations based on a quick peek:

Could you add test cases for this to our test suite? See CONTRIBUTING.md and existing stuff in `test/`. You can use `./generate tipc` in the `test` dir to generate placeholder test suite files to fill.

The function names here are too generic. I suggest using a `_tipc_` prefix in them, and dropping e.g. the apparently useless `compgen` in them. Also, I'm not convinced that we should pollute the shell with this many functions that are being called from one place only. How about embedding at least the simplest ones directly inside `_tipc`?

The "have" stuff on the last line should be dropped, and the `# ex ...` line added. See existing completions for examples.

rical:
Thanks for the comments scop!

I submitted some changes. I tried merged as many functions as possible without destroying the readability altogether. Not sure about the framework stuff though, I generated the base but I looked around and didn't find any prior art which I could make sense of. Let me know if this suffices or how to proceed.

Cheers!

completions/tipc
```diff
@@ -0,0 +1,311 @@
...
+_tipc()
+{
+    local cur prev words cword optind
+    _init_completion || return
+
+    optind=1
+    COMPREPLY=()
+
+    # Flags can be placed anywhere in the commandline
+    case "${words[$cword]}" in
```
scop: Isn't this the same as simply `$cur`?

completions/tipc
```diff
@@ -0,0 +1,311 @@
...
+    local optind=$1
+    local filter=$2
+
+    if [[ $cword -eq $optind ]]; then
+        COMPREPLY=( $( compgen -W 'link' -- $cur) )
+    elif [[ $cword -eq $optind+1 ]]; then
+        # awk drops link state and last trailing :
+        local links=$(tipc link list | awk '{print substr($1, 0, length($1)-1)}')
+        if [[ $filter == "peers" ]]; then
+            links=$(echo "$links" | sed '/broadcast-link/d')
```
scop: This could be simplified to `command sed '/broadcast-link/d' <<<"$links"`?

scop:
The functions here are using 0/1 return values that aren't AFAICT used for anything. If that's true, I suggest removing them. (Yes, we do have bunch of completions that do those `return 0`s but I don't think we should add more if they serve no puropose.)

completions/tipc
rical: It is when completing the first sub-command. I.e. "tipc bearer" will switch correctly. When an additional command is passed "tipc bearer enable" $cur will be "enable", making this switch fail. I need to travers down the command tree looking at each passed subcommand. I didn't find any cleaner way to do that.

completions/tipc
rical: Done :+1:

scop: Careful, it should be `command sed`, not `sed`... :)

completions/tipc
scop: Note that I'm talking about the first `case` block (the one that completes to `-h` and `--help`). I understand the need to use `$optind` as the index in the one that follows this one, but for this the index is AFAICT always going to be `$cword` and the value at that index should be always the same as `$cur` unless I'm missing something.

completions/tipc
```diff
@@ -0,0 +1,306 @@
...
+        esac
+
+        COMPREPLY=( $( compgen -W 'list'  -- $cur ) )
+        return 0
+        ;;
+    esac
+
+    COMPREPLY=( $( compgen -W 'bearer link media nametable node socket' -- $cur ) )
+}
+
```
scop: Looks like a bit too much was removed from here, something like `&& complete -F _tipc tipc` is still needed (just the `$have` part was supposed to go away).

rical:
Not sure why this suddenly failed in the CI test? Looked briefly at the log and didn't see anything tipc related.

completions/tipc
```diff
@@ -0,0 +1,307 @@
+# tipc(8) completion                                       -*- shell-script -*-
+
+_tipc_media() {
+    local optind=$1
+
+    if [[ $cword -eq $optind ]]; then
+        COMPREPLY=( $( compgen -W 'media' -- $cur) )
+        return 0
```
scop: Could you replace all these `return 0` and `return 1` with plain `return` across the whole file as the return values do not seem to be used for anything?

completions/tipc
```diff
@@ -0,0 +1,307 @@
...
+
+_tipc_bearer() {
+    local optind=$1
+    local media
+
+    if _tipc_media $optind; then
+        return 1
+    fi
+
+    for ((i = 0; i < $cword; i++)); do
```
scop: I think this needs a `local i`

completions/tipc
```diff
@@ -0,0 +1,307 @@
...
+        if [ ${words[$i]} == 'media' ]; then
```
scop: Style: use `[[` instead of `[` for consistency

completions/tipc
```diff
@@ -0,0 +1,307 @@
...
+        enable)
+            local media params
+            let optind++
+
+            if [[ $cword -lt $optind+4 ]]; then
+                _tipc_bearer $optind
+                return 0
+            fi
+
+            for ((i = 0; i < $cword; i++)); do
```
scop: `local i`...?

completions/tipc
```diff
@@ -0,0 +1,307 @@
...
+            # that key Note that this would break if the user attempts to use a
+            # kown key as value
+            for i in ${params[@]}; do
+                if [[ $prev == $i ]]; then
+                    return 0
+                fi
+            done
+
+            # In order not to print already used options we remove them
+            for p in ${words[@]}; do
```
scop: `local p`...?

scop:
Yeah, the test suite is unfortunately not that stable. I've restarted the run.

rical:
Whats the status on this Scop?

scop:
Sorry, I've dropped the ball on this. Merged now, thanks.

rical:
Np, thanks for merging.

When could we expect this to be in a tagged release? I noticed Fedora 24 was running the latest 2.3 and Ubuntu 16.04 was running 2.1. How does the main distros normally handle new versions of this repo?

scop:
I suppose we've accumulated enough changes since 2.3, I'll look into tagging a release soonish.

For Fedora, I maintain the bash-completion package and do keep it up to date. I'm not sure about the status of Debian/Ubuntu package maintenance, but I've noticed that it has stated to lag behind somewhat recently.

PR #24, opened 2016-04-09
Fix mpv's completion (untested).

shlomif:
Previously, doing «mpv [tab]» in an interactive shell resulted in absolutely
nothing getting completed and "complete | grep -i mpv" returned nothing.
Doing "complete -r -D" made mpv complete on plain pathname completion.

Some testing done by changing the file locally on the /usr filesystem and
trying mpv in a new shell.

### DIFF
--- a/completions/mplayer
+++ b/completions/mplayer
@@ -286,6 +286,6 @@ _mplayer()
 
     return 0
 } &&
-complete -F _mplayer mplayer mplayer2 mencoder gmplayer kplayer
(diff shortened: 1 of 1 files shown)


scop:
Thanks for the report. The intent is to not do anything about mpv, see 00abd48e5b1d5d79fff46b7f791b2b90d1d6953b

shlomif:
Thanks!

PR #26, opened 2016-04-12
Support completing array variables and expansions

algorythmic:
* Complete `${arr[` with `${arr[idx1]} ${arr[idx2]}`, etc ..
* Complete `${arr[idx]` with closing curly brace
* Complete `${ar` with `${arr[` (nospace) if it is the only matching variable
* Complete `${arr[@` with `${arr[@]}` (also for `*`)
* Support completion of `${!var` and `${#var` forms for both regular and array variables

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -623,10 +623,40 @@ _split_longopt()
 #          False (> 0) if not.
 _variables()
 {
-    if [[ $cur =~ ^(\$\{?)([A-Za-z0-9_]*)$ ]]; then
-        [[ $cur == *{* ]] && local suffix=} || local suffix=
-        COMPREPLY+=( $( compgen -P ${BASH_REMATCH[1]} -S "$suffix" -v -- \
-            "${BASH_REMATCH[2]}" ) )
+    if [[ $cur =~ ^(\$(\{[!#]?)?)([A-Za-z0-9_]*)$ ]]; then
+        # Completing $var / ${var / ${!var / ${#var
+        if [[ $cur == \${* ]]; then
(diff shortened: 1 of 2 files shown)


algorythmic:
I wasn't sure about where to put the tests. `unit/_variables.exp` seemed the most appropriate but I can do something else with them if needed.

There was also something strange with `assert_complete_partial` -- it wouldn't recognize the string to remove from the expected output so I used `-expect-cmd-minus`.  Maybe the matching gets confused with the presence of variables in the string? In any case, it passes as expected like this.

scop:
Wow, this is impressive stuff, thanks! Just one tidbit suggestion if you wish to still tweak it in the future: maybe offer `@` and `*` in addition to array indexes in `${arr[` completion?

algorythmic:
I'd actually tried that out but ended up removing it as it seemed to make for somewhat messy output.

My thinking is that if you complete `${ARRAY[` it's pretty cool to see exactly what all the indexes are and how many there are.  If you want `@`, you can just type `@` instead of hitting Tab and _then_ typing `@`.

Changing the behavior is pretty easy if you do think that would be better:

```
- COMPREPLY+=( $( compgen -W '$(printf %s\\n "${!'${BASH_REMATCH[2]}'[@]}")' \
+ COMPREPLY+=( $( compgen -W '$(printf %s\\n "${!'${BASH_REMATCH[2]}'[@]}" "@" "*")' \
```

scop:
Yep, agreed, I'll leave it as it is now.

PR #28, opened 2016-04-14
man: Improve usage with alternate man implementations

somasis:
This commit sets a fallback MANPATH to use, for man implementations
such as busybox's man, which does not print the MANPATH when just
given `man -w`.

### DIFF
--- a/completions/man
+++ b/completions/man
@@ -58,9 +58,9 @@ _man()
 
     local manpath
     if [[ $OSTYPE == *@(darwin|linux|freebsd|cygwin)* ]] || _userland GNU; then
-        manpath=$( manpath 2>/dev/null || command man -w )
+        manpath=$( manpath 2>/dev/null || command man -w 2>/dev/null || echo "$MANPATH")
     else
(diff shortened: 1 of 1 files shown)


scop:
Hm, this implementation would leak the `MANPATH` variable into the environment in setups where it was not set. How about changing the `else` branch into another `if` that checks if `manpath` is empty, and if it is, then doing the fallback?

somasis:
That's a better idea. Amended.

scop:
Does this version still actually work with busybox environments? I'd assume those go to the first `if` branch and there's no longer a default path there. Thinking more about this, I suppose `$MANPATH` should actually be given more priority than it currently has, and if we always default to something, there's no longer a reason for the `comgen -c` fallback. What do you think about this: https://gist.github.com/scop/69f72ea2720ae2d54f278b102907faa1

somasis:
It does still work. But your patch seems a lot cleaner.

somasis:
Thanks :)

PR #31, opened 2016-04-22
Add EC2 host SSH completion

jeffbyrnes:
From my coworker @eherot who did the original work:

> Not sure if there's any interest in this in the broader project but we needed it internally so I figured I might as well push it up stream:
> 
> This function uses `aws ec2 describe-instances` to do command completion against instance name tags in the user's EC2 cluster, provided it is installed.

It’s up to the operator to have [properly configured the AWS CLI](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#config-settings-and-precedence).

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -172,6 +172,11 @@ _ec2_hosts()
     COMPREPLY+=( $( compgen -W \
         "$(cat "${cache_file}")" -- "${1}" ) )
 
+    # Refresh the cache in the background every time
+    aws ec2 describe-instances | \
+        grep 'TAGS\tName' | \
+        awk '{print $3}' > "${cache_file}" &
+
     return 0
 }
 
--- a/completions/ssh
+++ b/completions/ssh
@@ -172,10 +172,10 @@ _ec2_hosts()
     # Refresh the cache if:
     # - There is no cache
     # - It doesn't contain the search string
-    # - It's more than 5 minutes old
+    # - It's more than 1 day old
     if [ ! -f "${cache_file}" ] || \
         ! grep -q ${1} "${cache_file}" || \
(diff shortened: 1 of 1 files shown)


scop:
This stuff is fairly specific, but if it was less intrusive, I could consider it. But it violates an important principle we follow, from CONTRIBUTING.md: "Do not write to the file-system under any circumstances".

I don't know if the caching behavior is essential to this stuff to work sensibly. If it is, perhaps it could be somehow handled outside of bash completion, for example by updating a block of entries in ssh's config or known hosts files.

Some other remarks, there's maybe more but I'm not looking deeper now:
- What happens if the "properly configured the AWS CLI" part has not been done?
- The cache time comments need updating, one says 1 minute, another 5 minutes, and the reality in code is 1 day.
- `which` is not a command that we assume to be available, `type` should be used instead.
- The `-mmin` option to `find` is not portable. See portability related discussion and the POSIX doc link in CONTRIBUTING.md.

jeffbyrnes:
@scop I’ve updated a few things as per your suggestions, but am still trying to think up an alternate way to handle the caching.

Thanks for your patience & straightforward explanation, apologies for neglecting to read the contribution guidelines ahead of time.

As my cohort @eherot who did the initial work on this, it may not be suitable for the larger project. Using this without the caching is pretty painful, even with just the ~130 EC2 instances we have & the optimizations I’ve done so far.

jeffbyrnes:
This has lain dormant for a couple years now, and I lack write access to the repo it’s based off of, so I cannot rebase it even if it would be accepted.

The file cache is pretty necessary for this to work well, so I’m going to close this. Thanks!

PR #32, opened 2016-04-23
Update bash_completion

gene-pavlovsky:
Source compat completion dir: use shell globbing instead of external `ls` command. Cuts bash startup time on Cygwin by 40 ms.


scop:
Thanks, applied with improved commit title/message. (The first line should describe the actual change made.)

gene-pavlovsky:
Thanks!

PR #33, opened 2016-04-27
Fix aptitude package completion

deivid-rodriguez:
Without this, there's a very annoying delay when trying to autocomplete
packages for install, for example. I do it just like apt-get does, which
autocompletes fine without delay.

Refs: https://alioth.debian.org/tracker/index.php?func=detail&aid=315216&group_id=100114&atid=413095


scop:
Thanks, fixed another way by reusing the functionality from apt-cache completion

deivid-rodriguez:
Thank you, much better! :)

PR #34, opened 2016-04-27
Allow pathname completion after equals sign

deivid-rodriguez:
This is useful for things like:

```
./configure --install-dir=<TAB>
```

or

```
bin/rake test TEST=<TAB>
```

Refs: https://alioth.debian.org/tracker/index.php?func=detail&aid=315238&group_id=100114&atid=413095

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2023,9 +2023,9 @@ _completion_loader()
     __load_completion "$cmd" && return 124
 
     # Need to define *something*, otherwise there will be no completion at all.
-    complete -F _minimal -- "$cmd" && return 124
+    complete -o default -F _minimal -- "$cmd" && return 124
 } &&
-complete -D -F _completion_loader
+complete -D -o default -F _completion_loader
 
 # Function for loading and calling functions from dynamically loaded
 # completion files that may not have been sourced yet.


scop:
Note that that the `./configure` case is already handled by our specific `configure` completion.

Regarding the commit title, it is misleading -- the implementation is about falling back to the shell's default completions if our (default) one for commands we have no specific completion for does not produce anything. Not a general pathname/equals sign thing. That needs to be fixed somehow.

If a change like this would be applied, should the default to be used be `default`, or `bashdefault`? Or both? What combination of those correspond's to the shell's default completion if no function has been installed?

Also, what is the effect of `-o default` in the `complete -D` line?

deivid-rodriguez:
I've reopened another PR with better branch name and commit message.https://github.com/scop/bash-completion/pull/36

scop:
Will have a look later, but for future reference, no need to fiddle with new PR's for commit messages, just amend and force-push to your existing branch.

deivid-rodriguez:
Thanks, I know. I just figured the branch name was as confusing as the commit message. I took me about one more minute than amending and force-pushing :wink:

PR #35, opened 2016-04-28
tar: silence --version's stderr output

somasis:
Causes an unknown argument error to be output. Tested with busybox tar.


scop:
Thanks!

PR #36, opened 2016-04-28
Better completion defaults

deivid-rodriguez:
Falling back to the shell's default completions if our (default) one for
commands we have no specific completion for does not produce anything.

Supersedes https://github.com/scop/bash-completion/pull/34.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2023,9 +2023,9 @@ _completion_loader()
     __load_completion "$cmd" && return 124
 
     # Need to define *something*, otherwise there will be no completion at all.
-    complete -F _minimal -- "$cmd" && return 124
+    complete -o default -F _minimal -- "$cmd" && return 124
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
> If a change like this would be applied, should the default to be used be default, or bashdefault? Or both? What combination of those correspond's to the shell's default completion if no function has been installed?

I don't know... I remember I investigated this a while ago (when I submitted the patch to the old tracker), but I forgot now.

> Also, what is the effect of -o default in the complete -D line?

I don't know either... sorry.

I remember that this fixed my particular problem with completions after the equal sign, but I don't remember the rationale. I'll try to find time to investigate and answer your questions in the future.

deivid-rodriguez:
I just revised this and my original fix is no longer working for me, so I'm closing this. Regards!

PR #38, opened 2016-05-03
Please consider to add these changes

bitstreamout:
Those changes are forwarded from bug fixes for current used bash-completion 2.1 on openSUSE and SUSE Linux Enterprise.  This would help a lot here.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -52,9 +52,6 @@ shopt -s extglob progcomp
 # A lot of the following one-liners were taken directly from the
 # completion examples provided with the bash 2.04 source distribution
 
-# Make directory commands see only directories
-complete -d pushd
-
 # start of section containing compspecs that can be handled within bash
 
 # user commands see only users
@@ -1848,7 +1845,14 @@ _longopt()
             command sed -ne 's/.*\(--[-A-Za-z0-9]\{1,\}=\{0,1\}\).*/\1/p' | sort -u )" \
             -- "$cur" ) )
         [[ $COMPREPLY == *= ]] && compopt -o nospace
-    elif [[ "$1" == @(@(mk|rm)dir|chroot) ]]; then
+    elif [[ "$1" == pushd ]]; then
+        if shopt -q cdable_vars; then
+            compopt -o variable -o nospace
+        else
+            compopt -o nospace
+        fi
+        _cd
+    elif [[ "$1" == @(@(rmdir|mkdir)|chroot) ]]; then
         _filedir -d
     else
         _filedir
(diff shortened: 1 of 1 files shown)


bitstreamout:
I'm a bit puzzled about the failed check ... seems to be

 FAIL: ~part should complete to ~full<space> if non-existing dir

as this pull request changes ls do behave well for variables this might influence the expected behaviour.  The only question I have is: What is the expectzed behaviour here

 set test "~part should complete to ~full<space> if non-existing dir"
 # Create list of users, having non-existing home dir
 assert_bash_exec {for u in $(compgen -u); do \
     eval test -d ~$u || echo $u; unset u; done} {} /@ users
 find_unique_completion_pair $users part full
 assert_complete "~$full " "ls ~$part" $test -nospace

scop:
Thanks for submitting these, but please submit them as single individual pull requests. Some of these may need work, some are not acceptable and it's not good to keep the bits that could be applied waiting for others that are unrelated.

Some comments about two of them, haven't looked at the others yet:

rmdir, pushd etc: Not acceptable, because it would be a downgrade compared to the current functionality, long options would no longer be completed.

lvm: This has been already discussed somewhere with Liuhua Wang from SUSE (I'm unable to find the exact reference at the moment but it was probably #13 which has been modified and possibly original discussion lost): this is not acceptable, it tries to modify `/etc/lvm/lvm.conf` and writing to the filesystem is just something we don't do (see CONTRIBUTING.md), and this particular implementation would only work for root, has no error control etc.

bitstreamout:
Thanks a lot for your answer.  I'll try to find the time to split them into several pieces as the commits of my repostitory are. Question on rmdir, pushd ... what is about  importing _cd() into _longopt() for those commands to get the functionality working also for those commands.

bitstreamout:
OK now build has finished with success.  The advantage of using _cd() in comparison to plain _filedir() is that  CDPATH will be considered as well

scop:
Understood, but I don't think `rmdir`, `chroot`, or `mkdir` (which FWIW are not shell builtins) actually use `CDPATH` in any way -- only `pushd` and obviously `cd` do -- so I don't think it makes sense to use `_cd` for those three.

bitstreamout:
Hmm ... what is wrong to extend the usage of CDPATH on commands not being builtins?  This is (IMHO) one of the major features of bash complete programming to do such things.

scop:
If you set CDPATH to let's say /tmp, and there's a dir /tmp/foo, you have the rmdir completion use CDPATH, and you invoke rmdir in some other dir than /tmp, due to the CDPATH completion you will be offered "foo" among possibly other completions. Now, if you actually follow the completion suggestion and do "rmdir foo" in some other dir than /tmp, it will not end up trying to remove /tmp/foo but will give you an error message saying that foo does not exist. Same thing with chroot and mkdir.

So using CDPATH to complete commands that do not actually use CDPATH themselves will lead to completion suggestions that don't actually work when invoked as commands, and we don't want that.

bitstreamout:
OK ... now with rmdir, chroot, and mkdir not using `_cd' but `_filedir -d '.

scop:
The commit title or description no longer matches the whole commit's set of changes. I suggest splitting it further there are now bunch of entirely unrelated changes here. Actually I'm not sure if we want any of the other changes here besides the pushd ones in the first place:

Why `-o nospace`? I don't notice any behavior to the current one with e.g. rmdir. Dir completions (including `_filedir -d` and `_cd`) already implicitly have the nospace behavior as far as I can tell.

Why `-o dirnames`, isn't `_filedir -d` enough?

bitstreamout:
OK ... _filedir -d seems to be enough though.  Hopefully I can now get pushd change upstream

scop:
On closer inspection, pushd does not actually have long options so using `_longopt` for it doesn't make sense. I've just made it use `_cd` directly.

bitstreamout:
Perfect

PR #39, opened 2016-05-07
Support pod document files for perldoc

plambert:
Extend the perl helper to optionally include *.pod files, then include those only for perldoc completions.

For example, this allows the completion of `perl -MMoose::M` to continue to show only actual perl modules than can be loaded, while allowing the completion of `perldoc Moose::M` to include the Moose::Manual* documentation.

### DIFF
--- a/completions/perl
+++ b/completions/perl
@@ -15,6 +15,14 @@ _perlfunctions()
         -- "$cur" ) )
 }
 
+_perldocs()
+{
+    COMPREPLY=( $( compgen -P "$prefix" -W \
+        "$( ${1:-perl} ${BASH_SOURCE[0]%/*}/../helpers/perl perldocs $cur )" \
+        -- "$cur" ) )
+    __ltrim_colon_completions "$prefix$cur"
+}
+
 _perl()
 {
     local cur prev words cword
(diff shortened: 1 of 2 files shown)


completions/perl
```diff
@@ -15,6 +15,15 @@ _perlfunctions()
         -- "$cur" ) )
 }
 
+_perldocs()
+{
+    printf "%s\t%s\t%s\n" "$(date)" "_perldocs" "$cur" >> /tmp/perl-helper.log
```
scop: This looks like a security hole: allows appending to arbitrary files to which the completion user has write access through symlinks. But the specifics here don't matter as writing to files is something we don't do in completions in the first place, see CONTRIBUTING.md. So that stuff needs to go.

plambert:
Oh, how embarrassing, that was debugging code I'd left in only on accident.  I've removed it.

plambert:
It's not clear to me why this build failed; any idea what I'm missing here?

scop:
Our test suite isn't that stable, looks like the failure has nothing to do with your changes. But could you squash all your changes into one and force-push the result, that way it ends up neater in our change log which is generated from git logs?

plambert:
I think I figured out how to squash it.  Let's see if the tests pass.  Thanks for all your help, and especially for your efforts on bash-completion!

scop:
No problem, thank you for the contribution; I added a couple of related (non-functional) commits on top of yours. And I also learned that I could have done the squashing in the GitHub UI myself...

plambert:
Thanks; I saw your improvements and they look great.

I saw Github added squashing very recently; however, it was worth learning
it anyway, as I am working more and more with other, non-github repos.  Git
is a very simple underlying model which leads to some very interesting
structure and workflows that can be built on it.  Lots to learn!  :)

Thanks again for your help and your work on this project; it’s saving a lot
of people a lot of time!

Paul L.

PR #41, opened 2016-05-11
Expand variables to see if directories are handled

bitstreamout:
that is expand variables and command in backticks to get a usable
command line.  For the ls command and it various aliases allow to
handle files and directories as well.   For this avoid escaped
variables at completion time by switching complete options on demand.

Side effect is that if variables for directories are used, those
remain expanded on the command line.

This fixes openSUSE bugs #889319, #905348, and #940837
as well as SUSE Linux Enterprise bug #959299.

Signed-off-by: Werner Fink <werner@suse.de>


bitstreamout:
How should this work ?
 > FAIL: ~part should complete to ~full<space> if non-existing dir
 no user here has every reported this new behaviour ...

bitstreamout:
That must be an upstream FAIL as I do not have changed
FAIL: kill 1 should show completions at timeout
nor
FAIL: Tab should show complete process ids at timeout
and here on my local build host all checks had passed
Maybe I should simply drop an critical issue upstream to get the ls command including
its aliases to work as the users expected it.

bitstreamout:
Why does this behave different at travis, compare with

log/completion.log:/@kill 1
log/completion.log:/@kill 1PASS: kill 1 should show completions
log/completion.log:/@kill -s 
log/completion.log:/@kill -s PASS: kill -s  should show completions
log/completion.log:/@kill -
log/completion.log:/@kill -PASS: kill - should show completions
log/completion.log:/@renice 1PASS: Tab should show complete process ids
log/completion.log:/@renice -g PASS: Tab should show complete process group ids

bitstreamout:
Ping?

scop:
Pong. The test suite isn't that stable, it fails every now and then on its own. I restarted the run, and looks like it succeeded now.

Regarding the changes, this stuff looks quite hairy and it's likely that it'll take more than a little while for me to get to this. Others are welcome to beat me to it, naturally. It's not only because of the code (and lack of time on my part), but I'm having trouble understanding the title of the commit and the more verbose explanation of it. Also, some other things that caught my attention on the first look were that I don't find `_dollar` a very good name for a function that does what I guess it does, and the various fallbacks for ls and friends are inconsistent with what we do for other stuff (meaning that we usually don't fall back to those defaults, and I'm not sure why in the proposed cases we should). And we'll definitely want test cases for this functionality.

bitstreamout:
I've no problem to rename the function ... and maybe there are other and also smarter ways to fix the two issuses I've reported.  Neverthless those issuses had caused a lot of trouble here due to the bad response of the reporting users here.  In past I had written a working completion but with the package bash-completion I had taken quite some beating bug reports.  Therefore I had tried to port some of the features of my old completion stuff to bash completion.  Mainyl the usage of the builtin completion for e.g. wildcards as well as expansion of variables if useful.

bitstreamout:
An other way to do wildcard completion is to combine noglob with compgen -o bashdefault ... someting like

set -f
cmd=*.spec
set +f
COMPREPLY=($(set -f; compgen -o bashdefault -- "$cmd"))

Issue #44, opened 2016-05-20
Completions fail with 'unbound variable' error if 'nounset' option is set


siteshwar:
Tab completions fail with 'unbound variable' error if 'nounset' option is set :

```
# set -u
# ls /roo<TAB>bash: !ref: unbound variable
bash: !ref: unbound variable
bash: words[i]: unbound variable
```

```
bash-completion version 2.1.
```

scop:
Known problem, documented in README.md

scop:
On 2nd thought, let's leave this open, maybe people notice better that way that it's a known issue.

siteshwar:
Yes it's a known problem, but is there a planned fix or workaround for it ?

scop:
I'm not aware of anyone working on it. The workaround is not to use `nounset`.

deekej:
In case we dedicate somebody to work on this and fix it, will you accept the patch?

scop:
See discussion in #45

scop:
To facilitate looking into things by folks that are interested in this stuff, I pushed a new `nounset` branch which runs the test suite in `nounset` mode (failing spectacularly for now as expected), has #45 applied, and contains some initial bits to get the first unit test to succeed in that mode. Feel free to hack away and submit pull requests (as minimal/clean as possible) against that branch.

jwalt:
Just for the record, this is a different workaround with little clutter:

At the start of `__reassemble_comp_words_by_ref` and `_filedir`:

    trap "set -$-" RETURN; set +o nounset

It doesn't solve the problem that 3rd-party completion scripts exhibit the same incompatibility, but for me it works fine and could be expanded to cover other shell options as well.

mgoldshteyn:
Now I am running into this problem with the above fix:

    set: usage: set [-abefhkmnptuvxBCHP] [-o option-name] [--] [arg ...]
    -bash: set: -i: invalid option

Turning on xtrace with file and line enabled by changing $PS4 and pressing TAB in an attempt to expand the name of a dir in the current directory, I get the source of the errors as follows:
...
    +(/usr/share/bash-completion/bash_completion:17): _upvars(): eval 'words=("${@:3:2}")'
    ++(/usr/share/bash-completion/bash_completion:1): _upvars(): words=("${@:3:2}")
    +(/usr/share/bash-completion/bash_completion:18): _upvars(): shift 4
    +(/usr/share/bash-completion/bash_completion:7): _upvars(): ((  0  ))
[...]

PR #45, opened 2016-05-23
__reassemble_comp_words_by_ref: avoid triggering nounset on indirect references

Anvil:


Anvil:
Should fix #44

scop:
While this probably fixes one specific case when `nounset` is in effect, _lots_ of other cases across the codebase remain; it's just the standard coding practice in bash-completion to just go ahead and use variables and assume unset implies empty. I'm not convinced that we should be applying bits here and there, nor on the other hand if we should uglify all of the codebase by sprinkling those `:-`'s everywhere.

Anvil:
I understand this patch is way unsufficient and I dont blame you for not merging it. But I'm curious about your feelings about this nounset business. Would you want to see it fixed ? And If yes, *how* would you like to see it fixed ?

The way I see it, there are 3 ways to handle nounset:

1. Using "${foo-}" forms (not "${foo:-}") when a variable is possibly non-existent - this does not mean "everywhere".
2. Testing with [[ -v <identifier> ]] that variables exist and making the code bigger and uglier 
3. Ensuring all variables exists, which is rather complex compared to 1. or even 2.

scop:
I don't have anything against fixing this if it doesn't get too ugly. No, I don't have a clear definition for "too ugly" :) but I'm fairly convinced that 2) and 3) would be. Possibly also 1), I'm undecided. I'm not committing to this either, but one more imaginable alternative would be to have some kind of wrapper for all our completion functions that would automatically set `nounset` off at start and reset it to previous value at end.

If something gets done about this, the test suite should be modified to be run with `nounset` on right at start of the work, and ensured that it fails where appropriate.

Not quite sure what you mean with `${foo-}` vs `${foo:-}` here -- as far as unset variables go, I think they're the same thing (the former slightly prettier though).

Anvil:
I think i see your wrapper theory. Maybe it should be extended to a common sane, safe, feature-full environment every completion-author could rely on. 

(WRT: ${foo-} vs ${foo:-}: i was just thinking the former form was one char shorter than the later and possibly less ugly, but it's not important)

scop:
Yep, given a cleanly controllable completion environment, we wouldn't for example have to force `extglob` on everyone, and there's probably some more settings we'd want to ensure on/off as well.

Anvil:
There would be two things to put in place : allowing bash-completion to load correctly (without parsing issues) and letting it run in a safe environment.

There are many features that can alter bash parser behaviour and some other features that render code incompatible in different ways. I can see at least those:

1. extglob
2. posix
3. failglob
4. nounset
5. CDPATH
6. errexit(?)
7. braceexpand
8. noglob
9. nocaseglob/nocasematch...

... and the list goes unfortunately on. Maybe switching on and off all these features every time someone hits the tab key is not so much a good idea. Maybe a subset of the most uncommon or problematic features should be taken out of this list...

Also, *someone* should define if those features should be set to on or off in bash-completion environment.

scop:
Yes, the list is long, but just to point out: multiple options can be turned on/off with a single `set` invocation so no more than two of those should be needed per completion invocation (plus the other imaginable stuff that isn't controlled by `set`). But that's still not that good and I'm aware of these things, that's why I noted not committing to this approach either...

I added a new `nounset` branch and this PR is applied there, see #44 for discussion.

idallen:
I run all my interactive shells with nounset and failglob.
Yes, please fix those unset variable errors so that I don't have to do it myself every release.
In most cases it means using $(var-} instead of ${var} and in a few others it means declaring the variable as "local var=" rather than just "local var", e.g. "local -a toks=()", "local prefix=", etc.
Also fix the quoting to hide array brackets so that they don't trigger on failglob, e.g.
unset 'COMP_WORDS[i]' not unset COMP_WORDS[i]
Another example:
eval "$ref"='${!ref-}${COMP_WORDS[i]-}' not eval $2[$j]=\${!ref}\${COMP_WORDS[i]}

bitstreamout:
What is the reason why such fixes are not merged? Compare with https://github.com/scop/bash-completion/issues/44

scop:
Merged as 0b0323c514276c0143568b9e03646b65ee57e788

PR #46, opened 2016-05-23
mysql: enabling nullglob in case no charset/*.xml are available to avoid failglob triggering

Anvil:
When failglob is enabled, an error could raised by bash about the /usr/share/m{ariadb,ysql}/charsets/*.xml patterns if either mariadb or mysql is not installed. The proposed change should avoid the error.

### DIFF
--- a/completions/mysql
+++ b/completions/mysql
@@ -20,7 +20,11 @@ _mysql()
             return 0
             ;;
         --default-character-set)
-            local charsets=( $( printf '%s\n' /usr/share/m{ariadb,ysql}/charsets/*.xml ) )
+            local opt_save=$(shopt -p failglob nullglob )
+            shopt -u failglob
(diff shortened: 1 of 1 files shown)


scop:
Well, if mysql or mariadb are not installed, I suppose it's fairly unlikely for the code path in question to be executed... but yeah, why not. I've pushed an alternative change that is more in line with how we do shopt stuff nowadays. The nullglob equivalent was already handled.

Anvil:
Hmm, your code will produce one (possible two) empty value(s) in the array. Is that okay with the compgen right below ?

scop:
I don't see why it wouldn't be ok. Empties just blend with whitespace when the array is flattened to a string.

Anvil:
Juste asking !

PR #56, opened 2016-07-28
mn (Mininet network emulator): New completion.

PeterDaveHello:
bash-completion for the Mininet (mn) network emulator

### DIFF
--- a/completions/mn
+++ b/completions/mn
@@ -1,0 +1,65 @@
+# bash completion for mn                              -*- shell-script -*-
+
+_mn()
+{
+    local cur prev words cword
+    _init_completion || return
+
+    case $prev in
+        --switch)
+             COMPREPLY=($( compgen -W 'default ivs lxbr ovs ovsbr ovsk user
+                 ' -- "$cur" ) )
+             return
+             ;;
+        --host)
+             COMPREPLY=($( compgen -W 'block random' -- "$cur" ) )
+             return
+             ;;
+        --controller)
+             COMPREPLY=($( compgen -W 'default none nox ovsc ref remote
+                 ryu' -- "$cur" ) )
+             return
+             ;;
+        --link)
+             COMPREPLY=($( compgen -W 'default ovs tc' -- "$cur" ) )
+             return
+             ;;
+        --topo)
+             COMPREPLY=($( compgen -W 'linear minimal reversed single torus
+                 tree ' -- "$cur" ) )
+             return
+             ;;
+        --test)
+             COMPREPLY=( $( compgen -W 'cli build pingall pingpair iperf
+                 all iperfudp none' -- "$cur" ) )
+             return
+             ;;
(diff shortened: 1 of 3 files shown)


completions/mn
```diff
@@ -0,0 +1,65 @@
...
+_mn()
+{
+    local cur prev words cword
+    _init_completion || return
+
+    case $prev in
+        --switch)
+             COMPREPLY=($( compgen -W 'default ivs lxbr ovs ovsbr ovsk user
+                 ' -- "$cur" ) )
+             return 0
```
scop: Please change all these to simply `return` without a value for consistency with other completions.

PeterDaveHello: Done! Thanks.

completions/mn
```diff
@@ -0,0 +1,65 @@
...
+        --custom)
+             _filedir
+             return 0
+             ;;
+         -h|--help|-c|--clean|--version|--listenport)
+             return 0
+             ;;
+    esac
+
+    COMPREPLY=( $( compgen -W '-h -c -x -i -v
```
scop: Some of the short options have a long equivalent, please remove the short ones that do. I guess at least `-h`, `-c`, and `-v` have one, what about `-i` and `-x` (also make sure that they're handled by the `case` block above if appropriate)?

test/completion/mn.exp
```diff
@@ -0,0 +1 @@
+assert_source_completions mn
```
scop: Please add an actual test case. doc/testing.txt contains some instructions for generating a bare bones one.

scop:
Have you contacted mininet upstream to see if they would be interested in including the completion with mininet itself, instead of shipping it here?

Do the option argument lists need to be hardcoded, i.e. is there no command that could generate the valid values e.g. for `--switch`?

completions/mn
PeterDaveHello: @scop do you mean you don't want the short options be shown in the option list when a user is using the [tab] button?

PeterDaveHello:
@scop yeah I tried, so far no any response, since this project contains so many useful completions of different tools, so that's the reason why I sent a PR here, BTW, people in this project may be much more professional in this part.

Currently, I didn't see any commands that it can generate the support controller list, that's why I hard coded there, unless you would like to parse `mn -h`, it's not a good idea for my own opion, since it won'
t change too much, I think we could just use the hard coded list, what do you think?

completions/mn
scop: Yes, short options that have an equivalent long one should not be displayed. So if `-h` and `--help` are the same, offer only `--help`. That's what we do for all completions.

scop:
If there is a sane way/place to parse the arguments from, it should be done. If not, hardcode it, but in that case the completion may need more maintenance and may always be to some extent incorrect on systems that have `mn` with different accepted options.

PR #57, opened 2016-07-28
Use new container based infra on travis-ci

PeterDaveHello:


scop:
According to Travis' logs, we're already running on container based infra, so I don't think this change is necessary. Let me know if you have evidence that says otherwise. For example, from the latest master build log:

https://travis-ci.org/scop/bash-completion/builds/146825001
> This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.
> If you require sudo, add 'sudo: required' to your .travis.yml
> See https://docs.travis-ci.com/user/workers/container-based-infrastructure/ for details.

PeterDaveHello:
Okay, some old projects will need this additional config, while new projects don't :smile:

PeterDaveHello:
@scop would you please take a look at #56? Thanks.

PR #74, opened 2016-08-30
man: do not use MANPATH environment variable

### LINKED ISSUE #73: man: not all man pages are listed if MANPATH environment is set
If MANPATH environment is set, only man pages located under directories specified by MANPATH are shown.

e.g., 
```
$ echo $MANPATH
/usr/share/man/openmpi-x86_64:/usr/share/lmod/lmod/share/man::
$ man ma(tab)
(no output)
$ man MPI(tab)
MPI
MPI_Abort
(snip)

$ unset MANPATH
$ man ma(tab)
mac2unix
machinectl
(snip)
$ man MPI(tab)
(no output)
```

As explained in manpath.1
> If $MANPATH is set, manpath displays its value rather than determining it on the fly.  If $MANPATH is prefixed by a colon, then the value of the variable is appended to the list determined from the content of the  configuration  files.   If the colon comes at the end of the value in the variable, then the determined list is appended to the content of the variable.  If the value of the variable contains a double colon (::), then the determined list is inserted in the middle of the value, between the two colons.

I think in the completion for man should use the output of manpath command.

yuwata:
fixes #73.


completions/man
```diff
@@ -56,16 +56,15 @@ _man()
         return
     fi
 
-    local manpath="$MANPATH"
-    [[ -z $manpath ]] && \
-        manpath=$( manpath 2>/dev/null || command man -w 2>/dev/null )
+    local manpath=$( manpath 2>/dev/null || command man -w 2>/dev/null )
     [[ -z $manpath ]] && manpath="/usr/share/man:/usr/local/share/man"
```
scop: I think we should continue to support systems for which neither `manpath` or `man -w` return useful info. I.e. just swap the order around; don't just drop use of `$MANPATH`.

scop:
Also, I think this deserves some more test cases, see lib/completions/man.exp

yuwata:
Force pushed revised version. Sorry, I have no time to add test cases.

PR #75, opened 2016-09-03
Add completion for myrepos (mr)

benwebber:
This PR adds completion for [myrepos](https://myrepos.branchable.com/), a tool used to manage multiple local repositories at once.

I'd appreciate some input on the test suite in particular. I'm not sure how to test the following completions:

* `mr run <TAB>` should complete commands
* `mr bootstrap <TAB>` suggests standard input (`-`) to the user along with the normal list of files. Once the user starts typing a filename, `-` should not longer be suggested.

Thanks in advance for your feedback.

### DIFF
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -229,6 +229,7 @@ bashcomp_DATA = 2to3 \
 		mount \
 		mount.linux \
 		mplayer \
+		mr \
 		msynctool \
 		mtx \
 		munindoc \
--- a/completions/mr
+++ b/completions/mr
@@ -4,9 +4,9 @@ _mr() {
     local cur prev words cword
     _init_completion || return
 
-    local options='-c --config -d --directory -f --force --force-env -i
-    --interactive -j --jobs -k --insecure -m --minimal -n --no-recurse -q
-    --quiet -s --stats -t --trust-all -v --verbose'
+    local options='--config --directory --force --force-env --interactive
+    --jobs --insecure --minimal --no-recurse --quiet --stats --trust-all
+    --verbose'
 
     local commands='bootstrap checkout ci clean co commit config diff fetch
     grep help list log ls offline online push record register remember run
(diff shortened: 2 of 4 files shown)


scop:
Looks good! Some feedback:
- The `return` in the bootstrap stdin if block could be removed
- Option list should not be hardcoded, you can get it with `PERLDOC=-otext _parse_help "$1" help`. After doing this this I don't think it's necessary to be that strict about the exact list of options completed in tests.
- Similarly, command list should not be hardcoded, it can be parsed from `PERLDOC=-otext "$1" help`. After doing this this I don't think it's necessary to be that strict about the exact list of commands completed in tests.
- A simple `assert_complete_any "mr run "` in tests _should_ cover that part of the tests, but doesn't because there's too much output (that's a generic problem with our test suite at the moment). If you want, feel free to add that to the tests, commented out with an explanation.
- Bug: `mr bootstrap -` does not complete to `mr bootstrap - ` (with trailing space, GH seems to eat it). Testing for `-z $cur || $cur == -` should fix it. The currently existing `mr bootstrap` test already covers the stdin case -- if `-` was among suggestions there, the test would fail.
- Comment typo "optiions"
- `mr` should be added to `completions/Makefile.am`

benwebber:
Thanks for the detailed review. I especially appreciate your suggestion to remove the hard-coded options and commands. `_parse_help` is a useful helper.

scop:
Using full path to `cat` in `PERLDOC_PAGER` sounds like asking for trouble, just call it without the path? (I see we had the same issue in completions/perl, I just fixed it there.)

Please move options parsing to the block where the options are actually used to avoid unnecessary upfront slowdown. And you may want to use `PERLDOC_PAGER` there as well; for consistency with commands parsing if for nothing else.

scop:
Further, because the help output from where both commands and options are parsed from is the same, you could invoke the command once and store the output in a variable, then parse both commands and options from it where needed, like

- `local help=$( PERLDOC_PAGER=cat PERLDOC=-otext "$1" help )`
- `awk '/\[options\]/ { print $3 }' <<<"$help"`
- `_parse_help - <<<"$help"`

benwebber:
Thanks, back to you!

PR #78, opened 2016-10-05
[WIP] Improve p4 completion

zachwhaley:
- Supports Perforce 2015.2 commands
- Complete global options
- Complete command options, including multi-parameter options, like -dup
- Complete file revisions to an extent

I need to add testing.
Please provide feedback!

### DIFF
--- a/completions/p4
+++ b/completions/p4
@@ -1,51 +1,1824 @@
 # Perforce completion                                      -*- shell-script -*-
-# by Frank Cusack <frank@google.com>
+# for Perforce 2015.2
+# by Zach Whaley <zachbwhaley@gmail.com>
 
-_p4()
-{
+# Generate completion reply
+# Accepts 1 to 4 arguments:
+# 1. Space separated string of possible completion words
+# 2. A prefix to be added to each possible completion word
+# 3. Generate possible completion matches for this word
+# 4. A suffix to be added to each possible completion word. Can be used to prevent completing a single result
+__p4_complete() {
+    local IFS=$' \t\n' cur_="${3-$cur}"
+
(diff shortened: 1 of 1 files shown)


scop:
I'm afraid reviewing and feedback for changes this large will require help from someone who actually has access to and uses p4 for something. I'm not sure even how I would go about starting to test, assuming the test cases existed...

zachwhaley:
Yeah, I was afraid of that.  I know some of the p4 devs are active on Twitter.  I might see if any of them would want to test this.

In the mean time, I'll work on writing tests in y'alls test suite.

scop:
Just an idea, perhaps the p4 folks would be interested in adopting the completion with the product itself, or contribs of some sort if it has one, or something like that?

lizlam:
@zachwhaley Wow! This is very thorough and complete (no pun intended 😄 )  What do you need to get the pull request accepted?  Do you just need it tested?  Also I can post on the p4 community Slack channel and see if others can help.

scop:
My first preference would be to have this shipped somewhere outside bash-completion altogether, see my previous comment -- any thoughts on that? Having completions that we (== currently sadly "I") cannot sanely test automatically (nor possibly even manually!) and that have no active users maintaining them is not good.

zachwhaley:
That makes sense.

Thanks for the advice :smile:

scop:
Note that my opinion applies to the current p4 completion included with bash-completion as well; I'd gladly drop it if a better and maintained one was available somewhere else. If you do get yours included somewhere, let me know?

zachwhaley:
Thanks again!

I've got the script on GitHub currently, and I'll see what can be done with the Perforce folks.

https://github.com/zachwhaley/p4-completion

PR #79, opened 2016-10-06
ssh-keygen: support ed25519 keys

ckujau:
OpenSSH supports ed25519 since version 6.5 (Jan 2014, http://www.openssh.com/txt/release-6.5), this patch adds a completion for this key type.


scop:
Thanks!

Issue #70, opened 2016-08-24
Support Include in ssh config


scop:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=834695

PR #80, opened 2016-10-08
Add support for Include in ssh config (#70)

edlmo:
This patch adds support for the Include directive in _ssh_config_ files (issue #70).
It adds the included (readable) files to the `config` variable to be later processed by the `_known_hosts_real()` function.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1533,8 +1533,8 @@ _known_hosts_real()
     fi
 
     # "Include" keyword in config files
-    for i in ${config[@]}; do
-        _included_config_files $i
+    for i in "${config[@]}"; do
+        _included_config_files "$i"
     done
 
     # Known hosts files from configs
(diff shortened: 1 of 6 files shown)


edlmo:
Agree with the function name, it could be `_included_ssh_config_files`.

But I'm not so sure about moving that functionality inlined on `_known_hosts_real`. The ssh config allows to use the Include directive inside the included files... so the recursive function seems to be a good approach to me.

If you are ok with that I could change the function name and look into the  test cases.

PD: thanks to you for bash-completion ;)

scop:
Ah, I somehow missed that the includes and the function is recursive. In that case, agreed with the current approach.

edlmo:
I've added a test case for the ssh_config Include functionality (fixing some bugs along the way).

scop:
LGTM, thanks

PR #84, opened 2016-10-25
ant: parse targets in imported buildfiles

sustmi:
This PR adds support for `<import>` tags.
It is only basic support as it does not care about `basedir` setting.

### DIFF
--- a/completions/ant
+++ b/completions/ant
@@ -4,7 +4,8 @@ _ant_parse_targets()
 {
     local line basedir
 
-    basedir=$(dirname "$1")
+    # alternative of "dirname $1" in pure bash
+    basedir=${1%/*}
 
     # parse buildfile for targets
     while read -rd '>' line; do
--- a/completions/ant
+++ b/completions/ant
@@ -1,5 +1,30 @@
 # bash completion for ant and phing                        -*- shell-script -*-
 
+_ant_parse_targets()
+{
+    local line basedir
+
+    basedir=$(dirname "$1")
+
+    # parse buildfile for targets
+    while read -rd '>' line; do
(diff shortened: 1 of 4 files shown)


completions/ant
```diff
@@ -1,5 +1,30 @@
 # bash completion for ant and phing                        -*- shell-script -*-
 
+_ant_parse_targets()
+{
+    local line basedir
+
+    basedir=$(dirname "$1")
```
scop: Looks otherwise ok to me, but please don't use `dirname`. I think `basedir=${1%/*}` should work instead of it.

sustmi: To be honest I do not understand that syntax. I am not Bash guru though. What does it mean?

scop: It's about removing a matching suffix pattern. For details, see "Parameter Expansion" in the bash manual page.

sustmi: Wow, parameter expansion is a whole new world for me. What are the advantages of `${1%/*}` over `dirname`? Because I am afraid about its readability.

scop:
Your current $(...) needs a subshell invocation, and depends on the dirname executable, which we don't currently expect or require to be available. My proposal is a portable pure bash implementation, and readability is in the eye of the reader.

sustmi:
I replaced `dirname` with pure bash implementation and added a comment to preserve readability.

scop:
Thanks, applied with a followup tweak to the basedir param expansion that makes the test suite pass -- it didn't work for cases where the arg didn't have a slash. Comment removed because this pattern occurs frequently in our sources and there's little reason to single this particular one out.

sustmi:
Sorry, I did not run the tests locally and the CI is broken.
Thank you for helping me out. I am not used to such language constructs but I respect your decisions. ;)

PR #86, opened 2016-11-05
ccache: fix completing compiler's flags

rrthomas:
This fixes the ccache completion function so that completing the
compiler’s flags works correctly.

I took the fixes from the sudo completion code, which works fine.

### DIFF
--- a/completions/ccache
+++ b/completions/ccache
@@ -5,10 +5,14 @@ _ccache()
     local cur prev words cword split
     _init_completion -s || return
 
-    if [[ $COMP_CWORD -eq 1 && ${COMP_WORDS[COMP_CWORD]} != -* ]]; then
-        _command_offset 1
-        return
-    fi
+    local i
+    for (( i=1; i <= COMP_CWORD; i++ )); do
+        if [[ ${COMP_WORDS[i]} != -* ]]; then
+            _command_offset $i
+            return
+        fi
+        [[ ${COMP_WORDS[i]} == -[oFM] ]] && ((i++))
+    done
 
     case $prev in
         -h|--help|-V|--version|-F|--max-files|-M|--max-size)
--- a/test/lib/completions/ccache.exp
+++ b/test/lib/completions/ccache.exp
@@ -15,4 +15,29 @@ assert_complete_any "ccache -"
 sync_after_int
 
 
+set test "Tab should offer correct options from partial option"
+assert_complete_partial [list "--cleanup" "--clear"] "ccache" "--clea" $test
+sync_after_int
+
+
+set test "Tab should complete a command after ccache"
+assert_complete [list "stty"] "ccache stt" $test
+sync_after_int
+
+
+set test "Tab should complete a command after ccache --FLAG"
+assert_complete [list "stty"] "ccache --zero-stats stt" $test
+sync_after_int
+
+
+set test "Tab should complete a command's flags after ccache"
+assert_complete [list "--help"] "ccache ls --hel" $test
+sync_after_int
+
+
+set test "Tab should complete a command's flags after ccache --FLAG"
+assert_complete [list "--help"] "ccache --zero-stats ls --hel" $test
+sync_after_int
+
+
 teardown


completions/ccache
```diff
@@ -5,12 +5,15 @@ _ccache()
     local cur prev words cword split
     _init_completion -s || return
 
-    if [[ $COMP_CWORD -eq 1 && ${COMP_WORDS[COMP_CWORD]} != -* ]]; then
-        _command_offset 1
-        return
-    fi
+    for (( i=1; i <= COMP_CWORD; i++ )); do
```
scop: Most likely `local i` missing before this; it introduces a test suite failure.

completions/ccache
```diff
@@ -5,12 +5,15 @@ _ccache()
...
+        if [[ ${COMP_WORDS[i]} != -* ]]; then
+            _command_offset $i
+            return
+        fi
+        [[ ${COMP_WORDS[i]} == -@(o) ]] && ((i++))
+    done
 
-    case $prev in
+    case "$prev" in
```
scop: This change looks unrelated to the actual fix.

completions/ccache
```diff
@@ -26,8 +29,11 @@ _ccache()
 
     $split && return
 
-    COMPREPLY=( $( compgen -W '$( _parse_help "$1" )' -- "$cur" ) )
-    [[ $COMPREPLY == *= ]] && compopt -o nospace
+    if [[ "$cur" == -* ]]; then
+        COMPREPLY=( $( compgen -W '$( _parse_help "$1" )' -- "$cur" ) )
+        [[ $COMPREPLY == *= ]] && compopt -o nospace
+        return
+    fi
```
scop: Like the previous hunk, this seems unrelated to the actual fix, they look mostly like cosmetic changes to me. Or am I missing something? If not, please leave these out of the patch.

completions/ccache
```diff
@@ -5,12 +5,15 @@ _ccache()
...
-    if [[ $COMP_CWORD -eq 1 && ${COMP_WORDS[COMP_CWORD]} != -* ]]; then
-        _command_offset 1
-        return
-    fi
+    for (( i=1; i <= COMP_CWORD; i++ )); do
+        if [[ ${COMP_WORDS[i]} != -* ]]; then
+            _command_offset $i
+            return
+        fi
+        [[ ${COMP_WORDS[i]} == -@(o) ]] && ((i++))
```
scop: I suppose in addition to `-o`, `-F` and `-M` need this special treatment as well as they too take arguments.

completions/ccache
rrthomas: This looked wrong to me, but as you think it's cosmetic (i.e. prev cannot be empty), I've removed it.

completions/ccache
rrthomas: Added, thanks.

completions/ccache
rrthomas: Added, thanks.

completions/ccache
rrthomas: Without this, completion of the compiler's options don't work.

rrthomas:
Thanks very much for the review. I've fixed everything I believe needs fixing.

However, the existing test still fails, with the same error as before. I can't see what I've done wrong; I added `local i`, as in `sudo`.

Also, I can't see how to run a single test; that would be useful, as currently it is slow to recheck the tests for `ccache`.

I looked into the test suite, but it appears to be undocumented. I tried looking at similar tests, for `sudo`, but I can't find what I want:

1. A test that the flags of `ccache` complete correctly
2. A test that after `ccache`, a command can be completed.
2a. A Test that, after `ccache --SOME_FLAG`, a command can be completed.
3. A test that after a command, the flags of that command complete correctly.
3a. Equivalent of 2a.
4. [Failing test] A test that after `ccache`, shell built-ins are not completed. This currently fails e.g. with sudo: for example, you can complete `read`, but there is no `read` binary on my system.

rrthomas:
Update: by accident, I found `doc/testing.txt`. (Not an obvious location for documentation of the test system!) I can now run a single test. Now I shall try to work out why the test still fails even though I've declared the variable `local`.

rrthomas:
I found the source of the test failure: the test was loading `~/.bash_completion.d/ccache`, not `bash-completion/completions/ccache`. Looks like some extra environment cleaning is needed. However, this doesn't explain why the Travis checks fail.

rrthomas:
The latest Travis check does not seem to be testing the latest commit from my PR. I recall there being problems like this when history is rewritten.

rrthomas:
I've updated the PR with some tests now, all except for the failing test I mentioned above.

rrthomas:
Please correct me if I'm wrong; it looks as though I should now file two issues:

1. Something about `_command_offset` allowing built-ins as well as commands.
2. Bug in `make check` not ignoring completion files in one's home directory.

rrthomas:
Rerunning the checks and looking at the logs, the tests that fail are for `mr` and `ssh-keygen`, i.e. nothing I've changed (because I only changed the code for `ccache`.

rrthomas:
Ping?

rrthomas:
Ping again? Note that the failed checks are, as far as I can tell, bogus.

rrthomas:
Ping! I believe this PR is complete. Do let me know if you need anything else to merge it.

completions/ccache
scop: I tried without this, and completion of the compiler's options seem to work just fine. I mean e.g. `ccache gcc -<TAB>` completes gcc's options, both with and without this change. The test suite also passes with or without it. So I still cannot find a justification for this part of the change. Could you give me a specific reproducer what exactly does not work with it?

rrthomas: I had a play, and I think you're right: it does work without this change. I shall update the PR.

I guess the reason that the `if` is needed in the case of `sudo` is because there is another path that could be followed (there is some following code), but not in this case.

rrthomas:
PR updated as per your comments, @scop; thanks!

scop:
Thanks.

PR #91, opened 2016-11-25
Support for python gui source files

BartDeWaal:
Add support for Python gui source files

Especially in Windows land, the extension ".pyw" is used to indicate a
python file is intended to be launched without a console window. You can
still just run these from python.


scop:
Thanks!

PR #94, opened 2016-12-11
make the completion about ip smart

Sherlock-Holo:
In zsh, I can use ip a[tab] to complement, but in bash, I need to ip addr [tab], I think it which I change can help you to make it better.
Sorry my English is not good. 
Chinese:在zsh里面可以使用ip a [tab]进行补全，但是在bash这样做不行，只能打完addr再[tab]，所以我修改了一下ip的补全，希望可以帮到你更好的完善bash-completion

### DIFF
--- a/completions/ip
+++ b/completions/ip
@@ -62,7 +62,7 @@ _ip()
     [[ $subcmd == help ]] && return
 
     case $cmd in
-        l)
+        l*)
             case $subcmd in
                 add)
                     # TODO
--- a/completions/ip
+++ b/completions/ip
@@ -62,7 +62,7 @@ _ip()
     [[ $subcmd == help ]] && return
 
     case $cmd in
-        link)
+        l)
             case $subcmd in
                 add)
                     # TODO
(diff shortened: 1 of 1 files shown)


Artoria2e5:
This PR correctly recognizes `ip`'s behavior of allowing matching commands by its prefix as long as no ambiguity presents, therefore allowing completion of well-formed shorthand commands like `ip a` for `ip addr`.

scop:
While the idea sounds ok, I don't think the implementation is appropriate. My `ip` (from iproute 4.6.0) has two separate commands starting with a (`address` and `addrlabel`), and two separate ones starting with l (`link` and `l2tp`).

Sherlock-Holo:
I know but I think most of people use IP a to replace IP address

Ville Skyttä <notifications@github.com>于2016年12月12日周一 21:47写道：

scop:
If you just want to add support for `a`, I suggest changing the current `addr|address)` case to `a|addr|address)` (I just added `|address` in b53b287). I'm not willing to knowingly add incorrect completions, but will accept something that adds multiple correct, short aliases.

Sherlock-Holo:
thanks for your advice😋

Ville Skyttä <notifications@github.com>于2016年12月12日周一 22:19写道：

PR #101, opened 2016-12-21
new _parse_help_exclude, to not list the options already in use.

jordi-pujol:
Modified: bash_completion to parse GNU help excluding already used options.
Also changes some code for _parse_help and _parse_usage, because some functions are common for all them.                       

Have created a new function _parse_help_exclude that uses four parameters, first two are the same as for _parse_help
param 1: executable name or '-'                                       
param 2: help option (--help by default)                              
param 3: number of positional options already in use, may be omitted
param 4: this regexp lists the options that may be repeated, may be omitted

Has been developed and tested in a Debian Linux system at unstable release.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -761,37 +761,109 @@ _init_completion()
     return 0
 }
 
-# Helper function for _parse_help and _parse_usage.
-__parse_options()
+# Helper function for _parse_help, _parse_help_exclude and _parse_usage.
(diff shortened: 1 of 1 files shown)


jordi-pujol:
Have seen function _parse_help in master has some differences from that in Debian unstable, so I will adapt that part to match the upstream code.

PR #106, opened 2017-01-10
Add sidedoor to _ssh() completion

daradib:
[sidedoor][sidedoor] maintains an SSH connection
and is packaged in Debian. Although its primary use is as a
daemon, it can be invoked like autossh.

[sidedoor]: https://github.com/daradib/sidedoor


scop:
Thanks!

PR #107, opened 2017-01-13
Ifupdown mapping physical interfaces + add ifquery command

jordi-pujol:
ifup: adding ifquery that is also a ifupdown command

List only physical interfaces:

Modfied function _configured_interfaces() in bash-completion:

In a Linux Debian system we can configure mappings that will be managed by ifupdown,
This modification lists only physical interfaces that are attached to the system.

Excerpt & example of a file /etc/network/interfaces:
allow-hotplug wlan0
mapping wlan0
	script /usr/sbin/guessnet-ifupdown
	map autofilter: true
	map verbose: true
	map debug: true
	map iwscan-tries: 3
	map timeout: 10

iface wlan0-anyopen inet dhcp
	# You can also match any open network, if you are desperate :)
	wireless-essid any
	wireless-mode auto
	test wireless open

iface wlan0-casa inet dhcp
	wpa-ssid "myhomewifinet"
	wpa-psk "ABCDEFABCDEFABCDEFABCDEFABCDEF"
	test wireless essid "myhomewifinet"


scop (changes requested):
ifquery part is taken care of (was missing the Makefile* etc additions), will try to find time to take a look at the other stuff later. It seems a bit complex though, and contains at least one regression:
- grep and sed should be invoked with the "command" prefix like they used to

...and switches to extended regexps apparently unnecessarily -- I'm not sure if the -r switch is supported by sed in all Debian and friends versions we would like to support.

scop (changes requested):
I took a closer look, and here are some notes/questions.

First, the code seems to address two different issues: one is to add support for the "mapping" word, and the other to check only infterfaces in /sys/class/net. These should really be two separate commits.

The interfaces man page says that "mapping" is followed by a shell glob pattern (not simply an interface name). So simply treating it as an interface name can produce some incorrect results I presume, when it's a glob.

The code no longer looks into /etc/network/interfaces.d/* for no apparent reason, was that an oversight?

I'm not quite sure what the purpose of looking only into interfaces in /sys/class/net is (if I understand the code correctly). Wouldn't that prevent for example doing an ifup on interface that was just defined and is not yet there in sysfs?

Also, it results in no longer completing lo for ifup/ifdown, for which I don't see a reason for.

All in all, this seems quite complex for my taste in its current form. Perhaps because there is no real specific explanation of what the intent and rationale here is.

jordi-pujol:
On Sun, 27 May 2018 02:53:27 -0700
Ville Skyttä <notifications@github.com> wrote:

> scop requested changes on this pull request.
> 
> I took a closer look, and here are some notes/questions.
> 
> First, the code seems to address two different issues: one is to add
> support for the "mapping" word, and the other to check only
> infterfaces in /sys/class/net. These should really be two separate
> commits.

As a premise, an excerpt from the interfaces man page:
KNOWN BUGS/LIMITATIONS
The ifup and ifdown programs work with so-called "physical" interface names.  These  names
are  assigned  to  hardware  by	the  kernel.

A mapping is followed by a physical ifname and his corresponding iface
sections will use logical names.
This code does access /sys/class/net to check that the names specified in the
interfaces file correspond to physical interface names.

An special case is that this code doesn't complete alternative interface names
composed by a physical name, a colon and a number,
this is the configuration for complementary interfaces of the main physical interface.
Example 
iface eth0 inet dhcp
iface eth0:1 inet static
  address 1.2.3.4 ...
The completion code may be modified to consider this case.

> 
> The interfaces man page says that "mapping" is followed by a shell
> glob pattern (not simply an interface name). So simply treating it as
> an interface name can produce some incorrect results I presume, when
> it's a glob.

Yes, you are right.
This code only considers the case where the mapping word is followed by
a physical interface name, not a glob. Now I see that it's not
difficult to modify this behaviour in some hacking fast manner.
But this is far more complex, when considering glob patterns the list of
interface names is exclusive, and the name's correspondence must be
scanned from top to bottom of the interfaces file, to do it well the code should be
heavily modified.

> 
> The code no longer looks into /etc/network/interfaces.d/* for no
> apparent reason, was that an oversight?

/etc/network/interfaces.d/* must be considered when there is the
corresponding include command inserted in /etc/network/interfaces
But, the problem is that most network utilities do not use the
include command and will not include nothing even when an
include is specified; thus doing so in the completion is out of scope.
For example: guessnet and network-manager will only read plain lines
from /etc/network/interfaces.

> 
> I'm not quite sure what the purpose of looking only into interfaces
> in /sys/class/net is (if I understand the code correctly). Wouldn't
> that prevent for example doing an ifup on interface that was just
> defined and is not yet there in sysfs?
> 
> Also, it results in no longer completing lo for ifup/ifdown, for
> which I don't see a reason for.

May be we should also list interface lo, if anyone wants to disable and re-enable this interface.

> 
> All in all, this seems quite complex for my taste in its current
> form. Perhaps because there is no real specific explanation of what
> the intent and rationale here is.
> 
> 
> 



-- 
Jordi Pujol i Palomer

jordi-pujol:
Hi,
I have modified the code to:
- also consider virtual interfaces like eth0:0
- also consider glob patterns for a mapping specification
Caveats:
the #include directive in /etc/network/interfaces is not widely implemented by the network utilities, so this completion does not treat the #include directives neither.

Observations:
I think that in Debian's Linux, systemd is taking place over network management and /etc/network/interfaces goes towards being obsolete; personally, my systems are using systemd now, thus maybe does not worth too much effort on ifup completion.

PR #108, opened 2017-01-19
Fix bug in 'make' completion when using BSD sed

l3nticular:
On OSX, using BSD sed, this error occurs when trying to complete a space, this error occurs:
'make sed: 46: /dev/fd/63: extra characters at the end of d command'

This does not occur with the above change or with gsed.


scop:
Thanks!

PR #115, opened 2017-03-22
Fix __reassemble_comp_words_by_ref with only space after caret

Vampire:
Without this commit, the cword calculation is done wrongly if there is only whitespace after the caret, but no additional word.

I tried to execute the tests and add a test for this, but unfortunately I was not able to get the test to work properly and didn't really get how to write tests properly in this project.
So it would be nice if the reviewer could execute the tests to see whether my change breaks anything else tested and maybe someone can also add a test for this.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -433,9 +433,6 @@ __reassemble_comp_words_by_ref()
                 ((i == COMP_CWORD)) && printf -v "$3" %s "$j"
                 # Remove optional whitespace + word separator from line copy
                 line=${line#*"${COMP_WORDS[i]}"}
-                # Start new word if word separator in original line is
-                # followed by whitespace.
-                [[ $line == [[:blank:]]* ]] && ((j++))
                 # Indicate next word if available, else end *both* while and
                 # for loop
                 if ((i < ${#COMP_WORDS[@]} - 1)); then
@@ -443,6 +440,9 @@ __reassemble_comp_words_by_ref()
                 else
                     break 2
                 fi
+                # Start new word if word separator in original line is
+                # followed by whitespace.
+                [[ $line == [[:blank:]]* ]] && ((j++))
             done
             # Append word to current word
             ref="$2[$j]"
--- a/test/t/unit/test_unit_get_cword.py
+++ b/test/t/unit/test_unit_get_cword.py
@@ -153,3 +153,12 @@ class TestUnitGetCword(TestUnitBase):
             ]
         )
         assert got == 1
+
+    def test_25(self, bash):
+        """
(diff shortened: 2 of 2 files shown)


Vampire:
2.5 years later and the bug is still present while still my PR fixes it.
Anything blocking this from getting merged?
(The failing CI is the same one as on master)

Vampire:
You can easily see the problem manually if you add between the third `[[ $i == $COMP_CWORD ]] && printf -v "$3" %s "$j"` and the `else` a `echo && declare -p "$3"`.
If you then have `rsync sftp: ` (with a space in the end and the cursor between the colon and the space and then trying to complete.
`cword` will be `2` while it should be `1`.
This is what my patch fixes.

scop:
The reason for the stall is that this touches a pretty intricate part of code, and there are no tests. I tried a bit to add one myself as well, but didn't get it to work.

In case you're interested in trying out again, the test suite has changed a lot meanwhile; it's (mainly) now in python+pexpect instead of tcl+expect, in case you find that combo more welcoming.

Vampire:
Not really.
I'm currently on Windows unfortunately and pexpect is only partly compatible with Windows, e. g. `pexpect.spawn` is not available and some replacement would be needed to be used for Windows.
I also tried to run the tests in a docker container like the CI does, but I get a bunch of failures and errors.

As I said, the test would "simply" be to complete `rsync sftp: ` with the caret between the colon and the space and there the `cword` calculation will result in `2` instead of the expected `1` which then will make the actual completion script return wrong values as it tries to complete the wrong word without my patch.

Vampire:
Ah, some were timeouts, and the other three (`evince`, `kpdf` and `kdvi`) depend on files that are there with upper and lower case, which of course is not the case as I have the project checked out on NTFS which is case-insensitive, so only one is there, only one gets built into the docker image and so the expectation does not match the results as the fixture is not as intended.

Vampire:
Ok, with an additional `git checkout -- .` in the `docker-script.sh` I got it working.
I found the proper place for an easy test that fails without my patch and works properly with my test.
I hope this allows you to finally merge my fix. :-)

Vampire:
@scop another three years later, this error is still present and my PR still provides a failing test and a fix.
I rebased my PR again and adapted it to new rules.
Would you maybe finally consider merging this? :-)

scop:
Thanks for your patience. I took the liberty to tweak the commit comment and commentary in the test file (as well as removed an unnecessary space there to reduce confusion).

Vampire:
Yeah, no problem, I'm happy it finally got merged. :-)

PR #116, opened 2017-04-16
man: use "man -k" (apropos) to find completions

tornaria:
See #73.

The old code attempts to guess manual names from filenames, but this has
a few drawbacks. On one hand the logic to determine the manpath was
incorrect, as explained in #73. On the other hand, not all manual names
correspond to filenames. Both issues are subtle and non-portable as the
exact semantics vary from one man program to another.

The new code uses "man -k" to find potential completions. This method
seems portable and it should be correct -- as long as the man database
is up to date.

The one delicate detail is parsing the output of "man -k". The proposed
version works ok at least with both man-db and mdocml packages.


scop:
An extensive change like this would be much better off with a slew of new test cases for the test suite. Running it is somehow broken in Travis at the moment, so we'll need to resort to local runs for the time being. This patch fails it for me (c`d test; ./runCompletion man.exp`) with:

```
bash: .../bash-completion/completions/man: line 76: syntax error near unexpected token `<'
bash: .../bash-completion/completions/man: line 76: `    done < <(man ${sect:+-s "$sect"} -k "$cur" 2> /dev/null)'
```

More to follow maybe later, but the above (addition of test cases and fixing the current test suite) should be fixed, as welll as "man" invoked as "command man" like it was before.

tornaria:
I don't understand what's wrong with the code. It's process substitution, is that not allowed in your version of bash? Note that this feature is also used in make completion.

The reason to use process substitution instead of a pipe is that a pipe would run in a subshell hiding the array $names. This is a standard workaround (starting with bash 4.2 one can use shopt -s lastpipe instead).

We could instead use a pipe and capture output using command substitution, instead of a variable.

In fact, it seems using a pipe to "sed" may be even faster than the while read loop (it is certainly simpler).

---

Wrt testing, I couldn't make it work. This is what happens:
```
$ ./runCompletion man.exp
[...]
FAIL: ERROR Unexpected output from bash command "source $(cd "$SRCDIR/.."; pwd)/bash_completion"
[...]
```

---

I can think of a couple of tests which show the issues with the current code. First #73:
```
$ export MANPATH=:
$ man man<TAB>
```
should complete at least "man" (assuming if man is present then its manual page is also present).

As a second test, have a man page where the manual name is different than the filename, and check that it still completes by the manual name. This should work fine with mdocml, although I'm not sure this is supported with man-db (they add symlinks for this purpose).

Also note: in order for these tests to work, one needs to build the man database before, this will depend on the man program, the ones I know:

For man-db:
```
$ mandb <path>
```

For mdocml:
```
$ makewhatis <path>
```

scop:
Wrt getting docs to work, see doc/testing.txt. More info for your failing runCompletion should be in test/log/completion.log

tornaria:
Sorry, my bad. I see now it's the same failure you found before, I thought it was a failure in the test framework. In fact if I revert my patch (or if I apply the second patch) the test works.

I think I know what's going on. In ```test/config/bashrc``` it is set ```set -o posix```. Now my commit does the ```set +o posix``` trick I copied from ```completions/make```. Why does it fail for me and not for make? In fact, in ```completions/make``` the use of process substitution is inside a command substitution, so it gets parsed at runtime. But in my code the process substitution is parsed when the source is read, and at that time ```set +o posix``` has not been run.

Conclusion: process substitution cannot be used in completions, except in particular cases.

Fortunately, my new patch doesn't use process substitution, and all the tests pass.

tornaria:
Rebased on top of 2.7.

The first commit 944a507 is only about expanding the test suite to be more comprehensive, including two tests that currently fail (one of them corresponds to #73)

The second commit 2af2986 fixes #73.

PR #117, opened 2017-04-18
test: run bash with --norc to avoid system bashrc

tornaria:
When running bash with `--rcfile` it skips loading the user bashrc, but
it doesn't skip loading the system bashrc.

If bash completions is installed and loaded from system bashrc, this
causes an error:
```
FAIL: ERROR Unexpected output from bash command "BASH_COMPLETION_COMPAT_DIR=$(cd "$SRCDIR/.."; pwd)/completions"
FAIL: ERROR executing bash command "BASH_COMPLETION_COMPAT_DIR=$(cd "$SRCDIR/.."; pwd)/completions"
```
In order to avoid that, run bash with the `--norc` option which skips
loading any bashrc, and immediately source our `config/bashrc`.
It is important to set INPUTRC and PS1 before running bash.

### DIFF
--- a/doc/testing.txt
+++ b/doc/testing.txt
@@ -184,7 +184,6 @@ test-run fails.
 set -e  # Exit if simple command fails
 set -u  # Error if variable is undefined
 
-CRON=running
 LOG=/tmp/bash-completion.log~
 
     # Retrieve latest sources
--- a/test/config/bashrc
+++ b/test/config/bashrc
@@ -1,5 +1,8 @@
 # bashrc file for DejaGnu testsuite
 
+# Note that we do some initialization that would be too late to do here in
+# library.exp's start_bash().
+
 	# Use emacs key bindings
 set -o emacs
 	# Use bash strict mode
@@ -7,17 +10,8 @@ set -o posix
 	# Unset `command_not_found_handle' as defined on Debian/Ubuntu, because this
 	# troubles and slows down testing
 unset -f command_not_found_handle
-	# Set fixed prompt `/@'
 TESTDIR=$(pwd)
-export PS1='/@'
 export PS2='> '
-	# Configure readline
-export INPUTRC=$SRCDIR/config/inputrc
-    # When not running via cron, avoid escape junk at beginning of line from
-    # readline, see e.g.  http://bugs.gentoo.org/246091
-[ "$CRON" ] || export TERM=dumb
-	# Ensure enough columns so expect doesn't have to care about line breaks
-stty columns 150
     # Also test completions of system administrator commands, which are
     # installed via the same PATH expansion in `bash_completion.have()'
 export PATH=$PATH:/sbin:/usr/sbin:/usr/local/sbin
--- a/test/lib/library.exp
+++ b/test/lib/library.exp
@@ -991,8 +991,13 @@ proc start_bash {} {
     if {! [info exists TOOL_EXECUTABLE]} {set TOOL_EXECUTABLE bash}
     set env(SRCDIR) $::srcdir
     set env(SRCDIRABS) $::srcdirabs
-    exp_spawn $TOOL_EXECUTABLE --rcfile $::srcdir/config/bashrc
-    assert_bash_exec {} "$TOOL_EXECUTABLE --rcfile $::srcdir/config/bashrc"
+    set env(PS1) "/@"
+    set env(INPUTRC) "$::srcdir/config/inputrc"
+    set env(TERM) "dumb"
+    set stty_init "columns 150"
+    exp_spawn $TOOL_EXECUTABLE --norc
+    assert_bash_exec {} "$TOOL_EXECUTABLE --norc"
+    assert_bash_exec "source $::srcdir/config/bashrc"
 }
 
 
--- a/test/lib/library.exp
+++ b/test/lib/library.exp
@@ -991,10 +991,20 @@ proc start_bash {} {
     if {! [info exists TOOL_EXECUTABLE]} {set TOOL_EXECUTABLE bash}
     set env(SRCDIR) $::srcdir
     set env(SRCDIRABS) $::srcdirabs
+
+    # PS1, INPUTRC, TERM and stty columns must be initialized
+    # *before* starting bash to take proper effect.
+
+    # Set fixed prompt `/@'
     set env(PS1) "/@"
+    # Configure readline
     set env(INPUTRC) "$::srcdir/config/inputrc"
+    # Avoid escape junk at beginning of line from readline,
+    # see e.g.  http://bugs.gentoo.org/246091
     set env(TERM) "dumb"
+    # Ensure enough columns so expect doesn't have to care about line breaks
     set stty_init "columns 150"
+
     exp_spawn $TOOL_EXECUTABLE --norc
     assert_bash_exec {} "$TOOL_EXECUTABLE --norc"
     assert_bash_exec "source $::srcdir/config/bashrc"


scop:
I cannot reproduce. I do have the (Fedora provided) bash-completion package installed and enabled, and it gets loaded on shell startup. When I cd to my `bash-completion/test` checkout dir and do `./runCompletion id.exp`, it succeeds with no errors.

tornaria:
Maybe you it is loading from your user `~/.bashrc`, instead of the system bashrc. I don't have fedora, but it works that way in debian (and I don't see the issue there).

In void linux the bash-completion is loading from the system bashrc, thus the issue I described.

Note that the system bashrc has a different filename in different distros, in debian it is `/etc/bash.bashrc`, while in void is `/etc/bash/bashrc`.

Try adding
```
source /usr/share/bash-completion/bash_completion
```
to your system bashrc, to be able to reproduce the bug and to test the fix.

scop:
In Fedora (and e.g. CentOS), the system bashrc` /etc/bashrc` gets loaded from users' `~/.bashrc`, so the effect would be the same as loading bash-completion directly from the user one.

tornaria:
Not so in debian or void linux. Indeed, bash has a configuration option `SYS_BASHRC` which is set to `/etc/bash.bashrc` (debian) or `/etc/bash/bashrc` (void), possibly undefined in fedora.

If defined, bash loads this file *before* loading the user `~/.bashrc`.

When you use `bash --rcfile <file>`, this is loaded instead of the user bashrc, but the system bashrc is loaded nonetheless. In a standard debian system, the bash-completion script is loaded from the user bashrc, so this works. But on a standard void system, the bash-completion script is loaded from the system bashrc, thus the issue.

This is what happens when I run bash in interactive mode with `--rcfile`:
```
$ strace -e trace=open bash --rcfile /tmp/empty-bashrc-file -ic exit
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/libdl.so.2", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
open("/dev/tty", O_RDWR|O_NONBLOCK)     = 3
open("/usr/lib/locale/locale-archive", O_RDONLY|O_CLOEXEC) = 3
open("/etc/nsswitch.conf", O_RDONLY|O_CLOEXEC) = 3
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/libnss_files.so.2", O_RDONLY|O_CLOEXEC) = 3
open("/etc/passwd", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/gconv/gconv-modules.cache", O_RDONLY) = -1 ENOENT (No such file or directory)
open("/usr/lib/gconv/gconv-modules", O_RDONLY|O_CLOEXEC) = 3
open("/etc/bash/bashrc", O_RDONLY)      = 3
open("/etc/bash/bashrc.d/", O_RDONLY|O_NONBLOCK|O_DIRECTORY|O_CLOEXEC) = 3
open("/etc/bash/bashrc.d/bash_completion.sh", O_RDONLY) = 3
open("/usr/share/bash-completion/bash_completion", O_RDONLY) = 3
open("/home/tornaria/.bash_completion", O_RDONLY) = 3
open("/tmp/empty-bashrc-file", O_RDONLY) = 3
exit
+++ exited with 0 +++
```

You can see that the system bashrc *is* loaded, and the bash_completion script is loaded through that.

The only way to avoid loading the system bashrc is to pass the `--norc` flag. But then no rc file is loaded at all, even if one is specified by `--rcfile`:
```
$ strace -e trace=open bash --norc --rcfile /tmp/empty-bashrc-file -ic exit
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/libdl.so.2", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
open("/dev/tty", O_RDWR|O_NONBLOCK)     = 3
open("/usr/lib/locale/locale-archive", O_RDONLY|O_CLOEXEC) = 3
open("/etc/nsswitch.conf", O_RDONLY|O_CLOEXEC) = 3
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/libnss_files.so.2", O_RDONLY|O_CLOEXEC) = 3
open("/etc/passwd", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/gconv/gconv-modules.cache", O_RDONLY) = -1 ENOENT (No such file or directory)
open("/usr/lib/gconv/gconv-modules", O_RDONLY|O_CLOEXEC) = 3
exit
+++ exited with 0 +++
```

Hence the proposed patch, which runs `bash --norc` and then interactively sources the custom bashrc. In order for that to work, one needs to set `PS1` and `INPUTRC` before running bash.

Can you check that running the tests with my patch works fine on fedora? I tested on debian and void linux and it works fine on both.

Otherwise, I cannot run the tests without modifying my system bashrc.

scop:
I'll have a closer look a bit later, but before that, it would be interesting to know _why_ the error occurs. Your output in this log only contains the error message, not its cause. The cause can be found in `log/completion.log` (if it was generated by a `runCompletion` call).

FYI I found these two external completion issues related to this in my system:
http://icedtea.classpath.org/bugzilla/show_bug.cgi?id=3365
https://github.com/sheepdog/sheepdog/pull/388

tornaria:
I'm sorry. The error is that `BASH_COMPLETION_COMPAT_DIR` is a readonly variable when bash-completions was already loaded:
```
...
/@BASH_COMPLETION_COMPAT_DIR=$(cd "$SRCDIR/.."; pwd)/completions^M
bash: BASH_COMPLETION_COMPAT_DIR: readonly variable^M
/@FAIL: ERROR Unexpected output from bash command "BASH_COMPLETION_COMPAT_DIR=$(cd "$SRCDIR/.."; pwd)/completions"
echo $?^M
1^M
/@FAIL: ERROR executing bash command "BASH_COMPLETION_COMPAT_DIR=$(cd "$SRCDIR/.."; pwd)/completions"
...
```

In fact, running the system bashrc may do other wrong things, I presume one really wants to run bash without any system configuration for the tests to avoid random issues.

scop:
Could be, but on the other hand if the maintainers have defined a system bashrc that should always be loaded that early on shell init, it could be a sign that there's something essential in it which shouldn't be skipped...

scop:
I'll still get back to this later, but I've removed the read only variable in c41a76237bc9dcbfa326eeddd026b66d7646d91d

tornaria:
That change doesn't fix the issue, because the system provided bash-completion is still declaring the variable to be read only (of course, this will fix itself with time).

I thought the point of running the tests without loading the rc file is to make sure it's running on a well know configuration to avoid any trouble like here.

tornaria:
I have rebased the patch, and added a small fix: one needs to set `TERM=dummy` before invoking `bash --norc`, before by the time `config/bashrc` is sourced, it's to late (it's the same with `INPUTRC`).

scop:
I know it doesn't solve the issue, that's why I'm still going to get back to this one later.

tornaria:
Another small fix: we need to set stty columns to 150 before running bash, because after startup it won't take effect.

Now all the tests pass for me on debian and on void linux (except a very small number which give trouble on master as well: alias, cd, ssh)

tornaria:
Rebased on top of 2.7 in 14b033b

scop:
The fedoradev test setup had started to reproduce this once they added a system default-loaded bashrc. Merged along with some followup cleanup commits. Thanks!

PR #120, opened 2017-04-27
Permit mkdir to optionally complete on files (as well as dirs)

### LINKED ISSUE #119: mkdir should optionally allow completion of files (as well as directories)
A common use case is to want to make a directory with a name that is similar to an existing file.  eg. someone sent you `LongAndComplicated.zip`, which doesn't contain sub-directories, so you want to create a directory before unzipping this file.  It would be useful (and, without bash-completion, it is possible and the default behaviour) to type `mkdir Long<tab>`, then backspace the `.zip`, and press Enter to create the directory.

This makes sense for mkdir to have different behaviour to rmdir, because in the case of mkdir, the directory doesn't yet exist at the time of tab-completion (except possibly in the case of `mkdir -p`) - whereas for rmdir, the argument must exist and be a directory.

This came up in the past in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=376433 , and was fixed at the time in d9988bc84694185f1cc1869afa3370fe6d9bd477.  The fix was then (inadvertently?) reverted in 1b248b55021318e653a0ca81145829c9dc77c97b.

devkev:
Fixes #119.

I _think_ I've adjusted the test case correctly, but I'm not 100% sure.  When I try to run the test case, it comes up as an "untested testcase", and I don't know why.  In any case, feel free to adjust it if necessary.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1882,6 +1882,8 @@ _longopt()
         [[ $COMPREPLY == *= ]] && compopt -o nospace
     elif [[ "$1" == @(rmdir|chroot) ]]; then
         _filedir -d
+    elif [[ "$1" == mkdir && -z "${BASH_COMPLETION_MKDIR_COMPLETES_FILES+_}" ]]; then
+        _filedir -d
     else
         _filedir
     fi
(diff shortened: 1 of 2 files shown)


devkev:
This patch doesn't change the default behaviour.  It should be very simple and straightforward.  It allows people to opt-in (by setting an environment variable) to mkdir completing based on filenames.  Can you please consider merging this?

scop:
Closed/partially applied in master now, without making it optional. I made it so that mkdir completes on files with nospace, so that it's somewhat clearer that the completion needs to be manually adjusted.

PR #123, opened 2017-05-09
Test fixes

tornaria:
Various fixes for tests

### DIFF
--- a/test/lib/completions/alias.exp
+++ b/test/lib/completions/alias.exp
@@ -30,8 +30,8 @@ set cmd "alias foo"
 append cmd \002\002\002; # \002 = ^B = Move cursor left in bash emacs mode
 send "$cmd\t"
 expect {
-    -re "^alias foo\r\nbar +foo *\r\n/@alias foo$" { pass "$test" }
-    -re "^alias foo=$" { fail "$test: Wrong cursor position" }
+    -re "^alias foo(\b\b\b)?\r\nbar +foo *\r\n/@alias foo(\b\b\b)?$" { pass "$test" }
+    -re "^alias (foo\b\b\b)?foo=(foo\b\b\b)?$" { fail "$test: Wrong cursor position" }
     -re /@ { unresolved "$test" }
     default { unresolved "$test" }
 }
--- a/test/lib/completions/finger.exp
+++ b/test/lib/completions/finger.exp
@@ -48,7 +48,7 @@ foreach h [get_known_hosts] {
         }
     }
 }
-assert_complete $hosts "finger test@$char" $test -expect-cmd-minus "test@$char"
+assert_complete $hosts "finger test@$char" $test -nospace -expect-cmd-minus "test@$char"
 
 
 sync_after_int
--- a/test/lib/completions/ssh.exp
+++ b/test/lib/completions/ssh.exp
@@ -76,7 +76,7 @@ foreach h [get_known_hosts] {
         }
     }
 }
-assert_complete $hosts "ssh $char" $test -ltrim-colon-completions -expect-cmd-minus "$char"
+assert_complete $hosts "ssh $char" $test -expect-cmd-minus "$char"
 
 
 sync_after_int
--- a/test/lib/completions/cd.exp
+++ b/test/lib/completions/cd.exp
@@ -25,8 +25,8 @@ append cmd \002\002\002; # \002 = ^B = Move cursor left in bash emacs mode
 #append cmd \033\0133D; # Escape-[-D = Cursor left
 send "$cmd\t"
 expect {
-    -re "cd $::srcdir/fixtures/shared/default/foo\r\n(\.svn/ +|)bar bar.d/ +foo.d/ *(\.svn/ *|)\r\n/@cd $::srcdir/fixtures/shared/default/foo$" { pass "$test" }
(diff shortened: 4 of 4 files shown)


test/lib/completions/alias.exp
```diff
@@ -30,8 +30,8 @@ set cmd "alias foo"
 append cmd \002\002\002; # \002 = ^B = Move cursor left in bash emacs mode
 send "$cmd\t"
 expect {
-    -re "^alias foo\r\nbar +foo *\r\n/@alias foo$" { pass "$test" }
-    -re "^alias foo=$" { fail "$test: Wrong cursor position" }
+    -re "^alias foo(\b\b\b)?\r\nbar +foo *\r\n/@alias foo(\b\b\b)?$" { pass "$test" }
+    -re "^alias (foo\b\b\b)?foo=(foo\b\b\b)?$" { fail "$test: Wrong cursor position" }
```
scop: While this seems to fix all cases I'm aware of, I'd rather like to understand when exactly does the \b get printed, and when it works as the backspace (i.e. not printed). Do you have any further details? The data I have is:

- My Fedora 25 local box, bash 4.3.43: _not_ printed
- bash-completion fedoradev container, bash 4.4.11: _not_ printed
- bash-completion centos6 container, bash 4.1.2: printed
- bash-completion ubuntu14 container, bash 4.3.11: printed

test/lib/completions/finger.exp
```diff
@@ -48,7 +48,7 @@ foreach h [get_known_hosts] {
         }
     }
 }
-assert_complete $hosts "finger test@$char" $test -expect-cmd-minus "test@$char"
+assert_complete $hosts "finger test@$char" $test -nospace -expect-cmd-minus "test@$char"
```
scop: Improved fix for this is in a4c67638eaea89503c1623a78e1680ac1eb2331f

test/lib/completions/ssh.exp
```diff
@@ -76,7 +76,7 @@ foreach h [get_known_hosts] {
         }
     }
 }
-assert_complete $hosts "ssh $char" $test -ltrim-colon-completions -expect-cmd-minus "$char"
+assert_complete $hosts "ssh $char" $test -expect-cmd-minus "$char"
```
scop: The issue needs to be fixed at its source, not just pushed out of sight.

test/lib/completions/alias.exp
tornaria: I think the difference depends on whether bash is compiled with or without curses.
 * If bash is compiled with curses then \b is *not* printed
 * If bash is compiled without curses then \b *is* printed
I've tested this to be true both with bash 4.3.30 and 4.4. Note that using `configure --without-curses` doesn't work (in the sense that the binary produced is still linked against ncurses) so I had to remove `ncurses-devel` to produce the non-curses binaries.

```
$ cat a.exp 
set env(PS1) "/@"
set env(TERM) "dummy"
spawn bash --norc
expect "/@"
send "echo hola\002\002\002\n"
expect "hola\r\n"
$ PATH=~/src/bash/bash-4.4-nocurses/ /usr/bin/expect a.exp | grep "echo hola" | od -c
0000000   /   @   e   c   h   o       h   o   l   a  \b  \b  \b  \r  \n
0000020
$ PATH=~/src/bash/bash-4.4-curses/ /usr/bin/expect a.exp | grep "echo hola" | od -c
0000000   /   @   e   c   h   o       h   o   l   a  \r  \n
0000015
$ PATH=~/src/bash/bash-4.3.30-nocurses/ /usr/bin/expect a.exp | grep "echo hola" | od -c
0000000   /   @   e   c   h   o       h   o   l   a  \b  \b  \b  \r  \n
0000020
$ PATH=~/src/bash/bash-4.3.30-curses/ /usr/bin/expect a.exp | grep "echo hola" | od -c
0000000   /   @   e   c   h   o       h   o   l   a  \r  \n
0000015
```

Note that if one does *not* set `TERM=dummy`, then \b are printed in both cases.

test/lib/completions/ssh.exp
tornaria: I don't have any desire to understand what `-ltrim-colon-completions` option is actually doing, or for what purpose. But this test is the only use of this option, so unless this is really testing something meaningful, I think eliminating the option altogether may be the simplest option.

test/lib/completions/alias.exp
scop: Thanks a lot for your work on figuring this out. I opted to revert my earlier commit related to the `\b`'s and set `TERM=dumb` instead of `dummy`; that seems to be the original intent anyway.

test/lib/completions/ssh.exp
scop: ISTR the colon stuff could be related to IPv6 addresses possibly among the returned completions. Not sure about it though, but I think it's clear that `-ltrim-colon-completions` shouldn't be breaking cases where there are no colons present in the first place.

(Sigh, I'm starting to get really tired of trying to figure out and cope with the test suite code, TCL, and expect. Anyone up to rewriting the test suite in [pexpect](https://pexpect.readthedocs.io/en/stable/)?)

scop:
All issues here should be taken care of now.

PR #124, opened 2017-05-11
mplayer mp4a support

Busindre:
mplayer: Associate .mp4a / MP4A added.


scop:
Different fix in master now, thanks.

PR #126, opened 2017-06-26
Fixed errors when using declare and complete

allenbyerly:
After removing some invalid and misused function options, code now executes without error.  It appears to behave correctly but cannot confirm.
removed an invalid option -A for declare
removed an invalid option -D for complete as well as a misused option -F
Original execution errors from bash: 
-bash: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
-bash: complete: -D: invalid option
complete: usage: complete [-abcdefgjksuv] [-pr] [-o option] [-A action] [-G globpat] [-W wordlist] [-P prefix] [-S suffix] [-X filterpat] [-F function] [-C command] [name ...]


scop:
You are probably using an old version of bash. Those options are quite necessary for bash-completion to function as intended. Please update to bash 4.1 or later.

PR #132, opened 2017-06-30
Makefile: update default compatdir

### LINKED ISSUE #131: If sysconfdir is set, compatdir is not properly updated
In c41a76237bc9dcbfa326eeddd026b66d7646d91d, changes were made to how `compatdir` is set. Whereas before `BASH_COMPLETION_COMPAT_DIR` was set using a default, now `compat_dir` is set using the value of `BASH_COMPLETION_COMPAT_DIR` and a default. Unfortunately, the default [is updated by the Makefile using `sed`][makefile], and the new syntax doesn't match the regular expression that was used. As a result, even if `sysconfdir` is set at compile-time, the default value for `compat_dir` is not updated.

This is the cause of #127.

[makefile]: https://github.com/scop/bash-completion/blob/2.6/Makefile.am#L28

elyscape:
This was broken in c41a76237bc9dcbfa326eeddd026b66d7646d91d.

Fixes #131.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,7 +25,7 @@ EXTRA_DIST = CHANGES $(pkgdata_DATA) bash_completion.sh.in .dir-locals.el \
 
 install-data-hook:
 	tmpfile=`mktemp $${TMPDIR:-/tmp}/bash_completion.XXXXXX` && \
-	sed -e 's|=/etc/bash_completion\.d|=$(compatdir)|' \
+	sed -e 's|-/etc/bash_completion\.d|-$(compatdir)|' \
 	    $(DESTDIR)$(pkgdatadir)/bash_completion > $$tmpfile && \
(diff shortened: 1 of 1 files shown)


elyscape:
As a side note, it would be a good idea to push a new patch version once this fix is merged.

monouser7dig:
> As a side note, it would be a good idea to push a new patch version once this fix is merged.

would be much appreciated, thanks for the super quick response!

scop:
Thanks!

ilovezfs:
@scop do you have the time to create a new release? I'm not sure how to proceed on https://github.com/Homebrew/homebrew-core/issues/15172.

scop:
New release is coming up shortly, I'll get a couple of other bugfixes in first.

ilovezfs:
thanks @scop!

PR #144, opened 2017-07-02
lftp: Support ~/.local/... bookmark location

mgorny:
Support the new ~/.local/share/lftp/bookmarks bookmark location. Both locations
are possible, depending on whether ~/.lftp exists.

Bug: https://bugs.gentoo.org/527450
Original-Author: Andrew Petelin <adrianopol@gmail.com>


mgorny:
Indeed you are correct. Updated now.

PR #150, opened 2017-08-02
ebtables: new completion

audeoudh:
Based on the iptables completion, but slightly adapted as commands differ a little

### DIFF
--- a/completions/ebtables
+++ b/completions/ebtables
@@ -59,13 +59,12 @@ _ebtables()
             --ip6-source-port --ip6-sport --ip6-src --ip6-tclass
             --ip-destination --ip-destination-port --ip-dport --ip-dst
             --ip-proto --ip-protocol --ip-source --ip-source-port --ip-sport
-            --ip-src --ip-tos --jump --Lc --Lc] --limit --limit-burst --list
-            --Lmac2 --Lmac2] --Ln --Ln] --log --log-arp --logical-in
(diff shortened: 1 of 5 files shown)


audeoudh:
I did the two same commits as yours, adapted to `ebtables`.

completions/ebtables
```diff
@@ -0,0 +1,84 @@
...
+            --arpreply-target --atomic-commit --atomic-file --atomic-init
+            --atomic-save --change-counters --concurrent --delete
+            --delete-chain --destination --dnat-target --dst --flush --help
+            --in-if --in-interface --init-table --insert --ip6-destination
+            --ip6-destination-port --ip6-dport --ip6-dst --ip6-icmp-type
+            --ip6-prococol --ip6-proto --ip6-protocol --ip6-source
+            --ip6-source-port --ip6-sport --ip6-src --ip6-tclass
+            --ip-destination --ip-destination-port --ip-dport --ip-dst
+            --ip-proto --ip-protocol --ip-source --ip-source-port --ip-sport
+            --ip-src --ip-tos --jump --Lc --Lc] --limit --limit-burst --list
```
scop: The `--Lc]` here should probably be removed? Ditto other options with `]` in them below.

audeoudh: Indeed… I parsed the output of `man` too fast, without checking enough.

PR #154, opened 2017-08-16
clang: new completion

yamaguchi1024:
A new autocompletion feature for clang, by using its new --autocomplete API to get possible flags dynamically.

This is a feature which will be introduced to clang 5.0 released in August 23. 

Fish are already working on this feature as well.
https://github.com/fish-shell/fish-shell/issues/4174

### DIFF
--- a/completions/clang
+++ b/completions/clang
@@ -1,0 +1,81 @@
+_clang_filedir()
+{
+    # _filedir function provided by recent versions of bash-completion package is
+    # better than "compgen -f" because the former honors spaces in pathnames while
+    # the latter doesn't. So we use compgen only when _filedir is not provided.
+    _filedir 2> /dev/null || COMPREPLY=( $( compgen -f ) )
+}
+
+_clang()
+{
+    local cur prev words cword arg flags w1 w2
+    # If latest bash-completion is not supported just initialize COMPREPLY and
+    # initialize variables by setting manualy.
+    _init_completion -n 2> /dev/null
+    if [[ "$?" != 0 ]]; then
+        COMPREPLY=()
+        cword=$COMP_CWORD
+        cur="${COMP_WORDS[$cword]}"
+    fi
+
+    w1="${COMP_WORDS[$cword - 1]}"
+    if [[ $cword > 1 ]]; then
+        w2="${COMP_WORDS[$cword - 2]}"
+    fi
+
+    # Clang want to know if -cc1 or -Xclang option is specified or not, because we don't want to show
+    # cc1 options otherwise.
+    if [[ "${COMP_WORDS[1]}" == "-cc1" || "$w1" == "-Xclang" ]]; then
+        arg="#"
+    fi
+
+    # bash always separates '=' as a token even if there's no space before/after '='.
+    # On the other hand, '=' is just a regular character for clang options that
+    # contain '='. For example, "-stdlib=" is defined as is, instead of "-stdlib" and "=".
+    # So, we need to partially undo bash tokenization here for integrity.
+    if [[ "$cur" == -* ]]; then
+        # -foo<tab>
+        arg="$arg$cur"
+    elif [[ "$w1" == -*  && "$cur" == '=' ]]; then
+        # -foo=<tab>
+        arg="$arg$w1=,"
+    elif [[ "$cur" == -*= ]]; then
+        # -foo=<tab>
+        arg="$arg$cur,"
+    elif [[ "$w1" == -* ]]; then
+        # -foo <tab> or -foo bar<tab>
+        arg="$arg$w1,$cur"
+    elif [[ "$w2" == -* && "$w1" == '=' ]]; then
+        # -foo=bar<tab>
+        arg="$arg$w2=,$cur"
+    elif [[ ${cur: -1} != '=' && ${cur/=} != $cur ]]; then
+        # -foo=bar<tab>
+        arg="$arg${cur%=*}=,${cur#*=}"
+    fi
+
+    # expand ~ to $HOME
+    eval local path=${COMP_WORDS[0]}
+    flags=$( "$path" --autocomplete="$arg" 2>/dev/null | sed -e 's/\t.*//' )
+    # If clang is old that it does not support --autocomplete,
+    # fall back to the filename completion.
+    if [[ "$?" != 0 ]]; then
+        _clang_filedir
+        return
+    fi
+
+    # When clang does not emit any possible autocompletion, or user pushed tab after " ",
+    # just autocomplete files.
+    if [[ "$flags" == "$(echo -e '\n')" || "$arg" == "" ]]; then
+        # If -foo=<tab> and there was no possible values, autocomplete files.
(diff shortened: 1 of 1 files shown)


completions/clang
```diff
@@ -0,0 +1,81 @@
+_clang_filedir()
+{
```
scop: We don't need this, just use _filedir.

completions/clang
```diff
@@ -0,0 +1,81 @@
...
+    # the latter doesn't. So we use compgen only when _filedir is not provided.
+    _filedir 2> /dev/null || COMPREPLY=( $( compgen -f ) )
+}
+
+_clang()
+{
+    local cur prev words cword arg flags w1 w2
+    # If latest bash-completion is not supported just initialize COMPREPLY and
+    # initialize variables by setting manualy.
+    _init_completion -n 2> /dev/null
```
scop: We know latest bash-completion is installed if this file is, please simplify accordingly.

completions/clang
```diff
@@ -0,0 +1,81 @@
...
+    if [[ "$?" != 0 ]]; then
+        COMPREPLY=()
+        cword=$COMP_CWORD
+        cur="${COMP_WORDS[$cword]}"
+    fi
+
+    w1="${COMP_WORDS[$cword - 1]}"
```
scop: I suppose we already have $prev for this?

completions/clang
```diff
@@ -0,0 +1,81 @@
...
+    # Clang want to know if -cc1 or -Xclang option is specified or not, because we don't want to show
+    # cc1 options otherwise.
+    if [[ "${COMP_WORDS[1]}" == "-cc1" || "$w1" == "-Xclang" ]]; then
+        arg="#"
+    fi
+
+    # bash always separates '=' as a token even if there's no space before/after '='.
+    # On the other hand, '=' is just a regular character for clang options that
+    # contain '='. For example, "-stdlib=" is defined as is, instead of "-stdlib" and "=".
+    # So, we need to partially undo bash tokenization here for integrity.
```
scop: Wouldn't `_init_completion -s` work for this?

completions/clang
```diff
@@ -0,0 +1,81 @@
...
+    # If clang is old that it does not support --autocomplete,
+    # fall back to the filename completion.
+    if [[ "$?" != 0 ]]; then
+        _clang_filedir
+        return
+    fi
+
+    # When clang does not emit any possible autocompletion, or user pushed tab after " ",
+    # just autocomplete files.
+    if [[ "$flags" == "$(echo -e '\n')" || "$arg" == "" ]]; then
```
scop: The echo can probably be simplified as `$'\n'`

completions/clang
```diff
@@ -0,0 +1,81 @@
...
+        _clang_filedir
+    elif [[ "$cur" == '=' ]]; then
+        COMPREPLY=( $( compgen -W "$flags" -- "") )
+    else
+        # Bash automatically appends a space after '=' by default.
+        # Disable it so that it works nicely for options in the form of -foo=bar.
+        [[ "${flags: -1}" == '=' ]] && compopt -o nospace 2> /dev/null
+        COMPREPLY=( $( compgen -W "$flags" -- "$cur" ) )
+    fi
+}
```
scop: `&&` missing here, see existing completions.

completions/clang
```diff
@@ -0,0 +1,81 @@
...
+    elif [[ "$w2" == -* && "$w1" == '=' ]]; then
+        # -foo=bar<tab>
+        arg="$arg$w2=,$cur"
+    elif [[ ${cur: -1} != '=' && ${cur/=} != $cur ]]; then
+        # -foo=bar<tab>
+        arg="$arg${cur%=*}=,${cur#*=}"
+    fi
+
+    # expand ~ to $HOME
+    eval local path=${COMP_WORDS[0]}
```
scop: This looks unnnecessary. Why not just invoke `$words[0]`?

completions/clang
```diff
@@ -0,0 +1,81 @@
...
+    flags=$( "$path" --autocomplete="$arg" 2>/dev/null | sed -e 's/\t.*//' )
+    # If clang is old that it does not support --autocomplete,
+    # fall back to the filename completion.
+    if [[ "$?" != 0 ]]; then
```
scop: `[[ $? -ne 0 ]]`

completions/clang
```diff
@@ -0,0 +1,81 @@
...
+    elif [[ "$cur" == '=' ]]; then
+        COMPREPLY=( $( compgen -W "$flags" -- "") )
+    else
+        # Bash automatically appends a space after '=' by default.
+        # Disable it so that it works nicely for options in the form of -foo=bar.
+        [[ "${flags: -1}" == '=' ]] && compopt -o nospace 2> /dev/null
+        COMPREPLY=( $( compgen -W "$flags" -- "$cur" ) )
+    fi
+}
+complete -F _clang clang
```
scop: `# ex: filetype=sh` missing at EOF

completions/clang
```diff
@@ -0,0 +1,81 @@
+_clang_filedir()
```
scop: First line including `-*- shell-script -*-` comment missing

completions/clang
yamaguchi1024: If we replace it with $words[0], it wouldn't work with something like ~/bin/clang where ~ needs to be expanded.

yamaguchi1024:
@scop 
Since this clang --autocomplete API is implemented in clang 5.0, Ubuntu 14.04 doesn't support this newest clang version. What do you think I can do about this?

Teemperor:
Travis has the LLVM repository setup, so I think we could get the most recent clang from there for completion testing.

yamaguchi1024:
@scop Could you review?

scop:
Have you suggested including this with clang, instead of in bash-completion? That would solve any versioning issues, and the completion could just assume that the corresponding clang is installed etc.

yamaguchi1024:
We already included this script in clang, as below. But we wanted to make this completion default on bash.

https://github.com/llvm-mirror/clang/blob/master/utils/bash-autocomplete.sh

scop:
I'm not sure if I understand correctly what you mean by "make this completion default on bash". But I guess what you're after is that it gets installed properly and auto-enabled as appropriate -- if so, you should be just installing it properly along with the rest of clang. There should be no reason to ship another copy with bash-completion; the duplication just causes problems and increases maintenance.

We have support files for cmake and instructions how to use them in README.md, so proper install of the completion file should be straightforward. Something like this (untested) should work, feel free to submit it to clang upstream for me: 

https://github.com/scop/clang/tree/bash-completion-install
https://github.com/scop/clang/commit/ceb95f6a46f5841bff97b961c54c0730270931e9.patch

Teemperor:
@scop Hmm, I'm not sure how this would for example work on OS X. Because clang there is provided by Apple, but bash-completion scripts by the brew maintainers. Any idea how to solve this? Because I fear that this will end with the brew maintainers patching in this completion on their own.

PR #155, opened 2017-08-18
Add support for .lzo extension (--lzop) to tar

maaku:
The --lzop option can be used for tar to read lzop-compressed archives, often with the .lzo extension. We extend the regular expression to support completion of .lzo tar files as well.


scop:
Thanks!

PR #157, opened 2017-08-19
BTRFS completion

xtremae:
bash completion for BTRFS userspace tools

### DIFF
--- a/completions/btrfs
+++ b/completions/btrfs
@@ -1,0 +1,148 @@
+#BTRFS completion
+
+_btrfs_devs()
+{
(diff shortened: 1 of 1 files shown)


Lekensteyn:
Hi @xtremae, btrfs-progs already includes bash completion, there is probably no reason to maintain one here that is prone to become out-of-date:
https://git.kernel.org/pub/scm/linux/kernel/git/kdave/btrfs-progs.git/log/btrfs-completion

Therefore I'm closing this PR. Thanks though for the proposal!

PR #163, opened 2017-09-04
add completion rules for the tox tool

gaborbernat:
documentation for it at https://tox.readthedocs.io/en/latest/

### DIFF
--- a/completions/tox
+++ b/completions/tox
@@ -1,4 +1,4 @@
-# tox(1) completion                                   -*- shell-script -*-
+# tox completion                                           -*- shell-script -*-
 
 _tox()
 {
@@ -6,21 +6,15 @@ _tox()
     _init_completion || return
 
     case $prev in
-        -h|--help|--version|\
-        -n|--num|\
-        -i|--index-url|\
-        --hashseed|\
-        --force-dep)
+        -h|--help|--version|-n|--num|-i|--index-url|--hashseed|--force-dep)
             return
             ;;
         -c)
-            _filedir ini # we expect an ini file
+            _filedir ini
             return
             ;;
-        --installpkg|\
-        --result-json|\
-        --workdir)
-        	_filedir
+        --installpkg|--result-json|--workdir)
+            _filedir
             return
             ;;
         -e)
--- a/completions/tox
+++ b/completions/tox
@@ -1,0 +1,38 @@
+# tox(1) completion                                   -*- shell-script -*-
+
+_tox()
+{
+    local cur prev words cword
+    _init_completion || return
+
+    case $prev in
+        -h|--help|--version|\
+        -n|--num|\
+        -i|--index-url|\
+        --hashseed|\
+        --force-dep)
+            return
+            ;;
+        -c)
+            _filedir ini # we expect an ini file
+            return
+            ;;
+        --installpkg|\
+        --result-json|\
+        --workdir)
+        	_filedir
+            return
+            ;;
+        -e)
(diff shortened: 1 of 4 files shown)


completions/tox
```diff
@@ -0,0 +1,26 @@
...
+{
+    local cur prev words cword
+    _init_completion || return
+
+    case $prev in
+        -h|--help|--version)
+            return
+            ;;
+        -e)
+            local envs=`tox --listenvs-all`
```
scop: Use `$(...)` instead of backticks. Also, `tox` should be invoked as `$1` here.

completions/tox
```diff
@@ -0,0 +1,26 @@
...
+            COMPREPLY=( $(compgen -W "${envs}" -- ${cur}) )
```
scop: Add spaces around `$( ... )`, and quote $cur, and remove superfluous curly braces from envs and cur.

completions/tox
```diff
@@ -0,0 +1,26 @@
...
+            return 0
```
scop: Just return, no need for 0.

completions/tox
```diff
@@ -0,0 +1,26 @@
...
+            ;;
+    esac
```
scop: Bunch of options are lacking completions, e.g. -c --installpkg, -i, --result-json etc etc. At least, ones that take an argument should have a non-completion installed (like in the --help etc case) if their arguments are not completed.

gaborbernat: @scop I thought the parse help  will take care of that, so do I need to enlist the no completion for all that don't take an argument?

gaborbernat:
@scop is now ok? Thanks!

completions/tox
scop: No, help parsing doesn't do anything about option arguments. At least -c, --installpkg, -i, --result-json, --hashseed, --force-dep, and --workdir require an argument, so the completion should either provide argument completions for those as appropriate (like what is currently done for -e), or provide a non-completion for them, which is usually handled by just adding those options to the same case where --help and friends are.

completions/tox
```diff
@@ -0,0 +1,26 @@
...
+{
+    local cur prev words cword
+    _init_completion || return
+
+    case $prev in
+        -h|--help|--version)
+            return
+            ;;
+        -e)
+            local envs=$( $1 --listenvs-all )
```
scop: $1 should be enclosed in quotes.

obestwalter:
@gaborbernat - would be great if that would be fixed up and @scop can merge - looks like there is only a var enclosure missing ... 🥇

gaborbernat:
@obestwalter will do today

gaborbernat:
@scop done

scop:
Looks like one of my review comments was lost in the "outdated" sections, so reposting:

No, help parsing doesn't do anything about option arguments. At least -c, --installpkg, -i, --result-json, --hashseed, --force-dep, and --workdir require an argument, so the completion should either provide argument completions for those as appropriate (like what is currently done for -e), or provide a non-completion for them, which is usually handled by just adding those options to the same case where --help and friends are.

gaborbernat:
@scop sorry for missing that, fixed now it should be complete; although for some reason locally the ini file completion doesn't seem to work; any ideas? thanks!

scop:
Merged with some code style tweaks, will push a fix for an error message I got for -e separately. ini completion works fine here.

gaborbernat:
@scop thanks a lot!

PR #165, opened 2017-10-03
Added "*.gar" to unzip/zipinfo.

ezarko:
Ref https://docs.oracle.com/middleware/1221/wls/WLCOH/create-application.htm#WLCOH722


scop:
Thanks!

PR #172, opened 2017-10-25
Add ProxyJump completion

jmwallach:
ProxyJump has been in place since 7.3

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -44,8 +44,8 @@ _ssh_options()
         IdentityFile IdentitiesOnly IPQoS KbdInteractiveDevices KexAlgorithms
         LocalCommand LocalForward LogLevel MACs
         NoHostAuthenticationForLocalhost NumberOfPasswordPrompts
-        PasswordAuthentication PermitLocalCommand PKCS11Provider Port
-        PreferredAuthentications Protocol ProxyCommand PubkeyAuthentication
+        PasswordAuthentication PermitLocalCommand PKCS11Provider ProxyJump Port
+        PreferredAuthentications Protocol ProxyCommand ProxyJump PubkeyAuthentication
         RekeyLimit RemoteForward RequestTTY RhostsRSAAuthentication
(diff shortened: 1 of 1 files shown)


completions/ssh
```diff
@@ -44,8 +44,8 @@ _ssh_options()
         IdentityFile IdentitiesOnly IPQoS KbdInteractiveDevices KexAlgorithms
         LocalCommand LocalForward LogLevel MACs
         NoHostAuthenticationForLocalhost NumberOfPasswordPrompts
-        PasswordAuthentication PermitLocalCommand PKCS11Provider Port
-        PreferredAuthentications Protocol ProxyCommand PubkeyAuthentication
+        PasswordAuthentication PermitLocalCommand PKCS11Provider ProxyJump Port
```
scop: Remove this, one addition is enough :)

completions/ssh
```diff
@@ -215,7 +219,6 @@ _ssh()
         local configfile
         _ssh_configfile
         _known_hosts_real -a -F "$configfile" -- "$cur"
-
```
scop: This whitespace change seems unrelated, please remove it.

completions/ssh
```diff
@@ -116,6 +116,10 @@ _ssh_suboption()
         Ciphers)
             _ssh_ciphers "$2"
             ;;
+        ProxyJump)
+            _known_hosts_real -- "$cur"
```
scop: Shouldn't `-a` be used for `_known_hosts_real` here?

completions/ssh
jmwallach: I fixed the issues as mentioned inline and was preparing the -J option when I noticed that ProxyJump accepts comma-separated lists. Since _ssh_suboptions is not supposed to accept these arguments I am going to move the addition to my local config and not push the update.

scop:
Not supporting comma separated values is better than nothing, so I've pushed a change finishing support for that.

jmwallach:
Glad to have inspired someone to do the work :)

PR #178, opened 2018-01-03
VBoxManage completion

guillomovitch:


completions/VBoxManage
```diff
@@ -0,0 +1,1531 @@
...
+        --basefolder)
+            _filedir -d
+            return 0
+            ;;
+        --groups)
+            _vboxmanage_get_group
+            return 0
+            ;;
+        --ostype)
+            _vboxmanage_get_ostypes
```
scop: Should be Should be _vboxmanage_get_ostype (no trailing "s").

completions/VBoxManage
```diff
@@ -0,0 +1,1531 @@
...
+        _vboxmanage_get_uuid_vmname
+    fi
+}
+
+_vboxmanage_createvm()
+{
+    case $prev in
+        --name)
+            _vboxmanage_get_vmname
+            return 0
```
scop: Please replace these with plain returns (no 0) or consistency with rest of the tree.

completions/VBoxManage
```diff
@@ -0,0 +1,1531 @@
...
+            ;;
+        --hostonlyadapter*)
+            _vboxmanage_get_hostonlyif
+            return 0
+            ;;
+    esac
+
+    if [[ "$cur" == -* ]]; then
+        COMPREPLY=( $( compgen -W "--name --groups --description --ostype \
```
scop: Can't these huge hardcoded lists be parsed with e.g. _parse_help?

scop:
Oh, and the commit message needs work, something like "VBoxManage: New completion" would be appropriate.

guillomovitch:
Sorry for the answer delay. Here is a new proposal, addressing most (if not all) comments.
And I'm volonteering to maintain it if needed, of course.

gryf:
There is at least several different bash completions for vboxmanage with different state and quality. I'm wondering if there is a point to keep such complicated completion. Bash-completion itself have separate package at most Linux systems, so the VirtualBox. In perfect world vboxmanage completion should match version of installed VirtualBox, since completion may offer options which doesn't exists in in installed version of VirtualBox, or on the other hand, it may not provide right completion, if completion is too old.
I see the point to have it, but I'd rather have the right version matching installed VirtualBox, than a broken one.
Also, depending o Linux distribution, it can happen, that they main line will contain different version of VirtualBox.

Disclaimer: I'm the author of another [vboxmanage completion](https://github.com/gryf/vboxmanage-bash-completion) script.

PR #179, opened 2018-01-07
java -jar completes .war

martinda:
@scop I would like `java -jar` to complete the `.war` files. For example, it is possible to run `java -jar jenkins.war` from the command line. This pull-request makes this possible.


scop:
Applied a slightly modified version. Thanks!

PR #180, opened 2018-01-09
OpenSSH: Allow completition of smart cards related options

Jakuje:
For some reason, the completion of smartcard-related options were excluded, even though they accept path to PKCS#11 module (`*.so`) and it was very annoying to write all these paths manually. For example the following commands are completed correctly with this change and complete the files with "`*.so`" extension:
```
ssh -I /usr/lib64/pkcs11/opensc-pkcs11.so
ssh-add -s /usr/lib64/pkcs11/opensc-pkcs11.so
ssh-add -e /usr/lib64/pkcs11/opensc-pkcs11.so
ssh-keygen -D /usr/lib64/pkcs11/opensc-pkcs11.so
```

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -172,6 +172,10 @@ _ssh()
             _filedir
             return
             ;;
+        -I)
+            _filedir "so"
+            return
+            ;;
         -c)
             _ssh_ciphers "$1"
             return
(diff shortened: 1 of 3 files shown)


scop:
Applied with minor changes, thanks!

Issue #182, opened 2018-01-17
mplayer - user config interferes with completion


0xf4:
Changing the log level in `~/.mplayer/config` (e.g. `msglevel=all=3`) prevents completions from working.
I'll propose a patch right away.

PR #183, opened 2018-01-17
Prevents changes in mplayer users' config from breaking completion.

0xf4:
Fixes scop/bash-completion#182

### DIFF
--- a/completions/mplayer
+++ b/completions/mplayer
@@ -3,7 +3,7 @@
 _mplayer_options_list()
 {
     cur=${cur%\\}
-    COMPREPLY=( $( compgen -W "$( $1 -nomsgcolor -nomsgmodule $2 help 2>/dev/null | \
+    COMPREPLY=( $( compgen -W "$( $1 -noconfig all $2 help 2>/dev/null | \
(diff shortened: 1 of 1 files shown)


scop:
Hm, I think I'd rather address the known issue instead of disabling configs altogether. So I suppose we could add `-msglevel all=5` to the current list of `-nomsg*` options. Thoughts?

0xf4:
I opted for disabling every possible config altogether to avoid whatever unforeseen similar interferences may arise, and as it looked like a better approach than to add yet another specific `-no<whatever>` switch, as it doesn't seem like the most scalable strategy. Another thing that wouldn't harm (absent from my commit) would be forcing the locale, along the lines of: `LC_ALL=C $cmd -noconfig all -list-options`.
But, OTOH, maybe I'm over-thinking it, and as long as no one reports back any other bugs related to this issue, it'd make sense to go for the more conservative solution of minimal changes.

In the end it's probably you who'll end up maintaining the code in the long term, so I'll go the route you prefer. Tell me your opinion and I'll do the changes myself if you want me to ;-)

scop:
Thanks for the explanation. I opted to commit your original approach (with some commit message modifications) in df36cae3f85fbf2c386560f4e2da62bb7a67196b and extended the test suite to cover it.

PR #191, opened 2018-03-09
Completions for openssl *pkey* commands

gftg85:
This patch set adds completion for openssl commands: genpkey, pkey, pkeyparam, and pkeyutl.

They have been requested by a Debian user [1], and, although upstream support was not available, I thought it could be a good learning exercise, since I have never programmed bash-completion.

I split the changes into 5 different patches, one for each openssl command, and a fifth that suggests arguments to the -kdf option.  I'm OK with squashing the changes...  I just split them because I thought it could ease review, specially for the fifth patch (and also because CONTRIBUTING.md suggests small changes...  Maybe it got too small?).

[1] https://bugs.debian.org/779814#10


scop:
Looks good to me, merged. I opted to take the squash route (mainly because there were a few inconsistencies wrt the command names and the commit messages, keyparam vs pkeyparam etc and I didn't want to ask for fixes to those and was too lazy to fix them myself).

PR #193, opened 2018-03-11
mkdir: complete also on non dirs

g-raud:
mkdir creates directories that do not exist; there is no reason why the
names may not be derived from non directory files (for example creating
~/.bash_completion.d when ~/.bash_completion exists).

This effectively reverts commit 1b248b55 ("Make mkdir complete only on
dirs.", 2010-10-23).


scop:
See #119 and #120

PR #194, opened 2018-03-18
cvs: Add completion for the log command

gftg85:


completions/cvs
```diff
@@ -331,6 +331,13 @@ _cvs()
                 _cvs_command_options "$1" $mode
             fi
             ;;
+        log)
+            if [[ "$cur" != -* ]]; then
+                _filedir
```
scop: I'm pretty sure we want to use `_cvs_entries` here instead of `_filedir`. In fact, I suppose the whole patch could be reduced to simply adding `log` to the already existing `diff)` or `editors|watchers)` cases.

gftg85: Indeed.  Thanks for spotting and for the suggestion...  I submitted it again with the changes.

scop:
Thanks!

PR #195, opened 2018-03-22
ssh: avoid ssh-aliases/functions for remote filename completions

dougpagani:
Justification: As a function (_scp) running in the current
    bash-environment, all aliases & functions will be defined in that
    environment. For ssh-aliases/fxns which act as a "hook", this has
    the potential of cluttering $COMP_WORDS, and ruining completion as a
    consequence.
        Alternatives, (e.g. wrapping the function-definition with 'shopt
    -u/s expand_aliases'), are insufficient, because:
        1) aliases still expand in subshells
        2) This still does not prevent function-resolution of functions
        named "ssh".
----
    I have an ssh-hook which was causing issues with `scp` filename completion. I did figure out a better alternative personally that works in the current state of `master`, (I added a condition for the hook which was printing to stdout/err to check that the sdout `[ -t 1 ]` and stderr `[ -t 2 ]` were attached to a terminal, but I thought submitting a simple patch would give better default behavior for users in general. For background, in my hook, I was checking if `stdin` was attached to a terminal, but the way this completion works, `/dev/stdin` is left still-attached to a terminal, while it is only stdout & stderr which are redirected.

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -345,19 +345,19 @@ _scp_remote_files()
 
     # default to home dir of specified user on remote host
     if [[ -z $path ]]; then
-        path=$(ssh -o 'Batchmode yes' $userhost pwd 2>/dev/null)
+        path=$(command ssh -o 'Batchmode yes' $userhost pwd 2>/dev/null )
     fi
 
     local files
     if [[ $1 == -d ]]; then
         # escape problematic characters; remove non-dirs
-        files=$( ssh -o 'Batchmode yes' $userhost \
+        files=$(command ssh -o 'Batchmode yes' $userhost \
             command ls -aF1dL "$path*" 2>/dev/null | \
             command sed -e 's/'$_scp_path_esc'/\\\\\\&/g' -e '/[^\/]$/d' )
     else
         # escape problematic characters; remove executables, aliases, pipes
         # and sockets; add space at end of file names
-        files=$( ssh -o 'Batchmode yes' $userhost \
+        files=$(command ssh -o 'Batchmode yes' $userhost \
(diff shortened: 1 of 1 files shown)


scop:
Don't know what to do requests like this. Pretty much the same justification could be given for every single command we invoke, they could be defined as aliases or functions as well. I wouldn't want to clutter the codebase with `command` everywhere, and then again I'm sure we'd run into cases where we actually *should* be invoking the alias/function, and command invocation doesn't work for someone.

While I don't think bash-completion has a good general way to deal with these cases, users might be able to adjust their functions so that they don't interfere with completions. For example by checking if various `$COMP_*` variables are set, and doing what's appropriate for completions to not break in those cases.

dougpagani:
That's a good workaround for context-differentiation -- checking for `$COMP_*` vars. Since the `ssh`'s stdout is gathered, I just checked for a tty on filde 1 ( `test -t 1` ). 

I would be inclined to agree with your hesitation to clutter the codebase with `command`. But I think the scp case is different. Though it's obvious to me now, it took a fair bit of digging a couple months ago to figure out what was going on behind the scenes. I think `scp` is unique in this (deserving of a `command` to skip-over UDF aliases & functions), since it employs a different command altogether (`ssh` of course). It's not exactly intuitive that by making an alias for `ssh`, I break my `scp`. I mean, the output can get pretty wacky.

Again, I have my workaround personally, but I do think that this is a scenario which warrants `command` wrapping. When simple typing tab, it has the potential to spit a whole lot of confusing junk onto the screen. For those unaware of the mechanisms of `bash-completion`, relation of `scp` to `ssh`, I think it would be a nice problem preemption.

PR #198, opened 2018-04-11
pkg-config: datadir relative variables

inigomartinez:
Although `datarootdir` is defined as `${prefix}/share`, the user has the possibility to define `datadir` in a different directory.

Due to this `completionsdir` and `helpersdir` that are defined in the generated pkg-config `.pc` file should be relative to `datadir` variable.

This helps any pkg-config user to set these variables to proper directories by using the `define-variable` command line argument.

### DIFF
--- a/bash-completion.pc.in
+++ b/bash-completion.pc.in
@@ -1,7 +1,8 @@
 prefix=@prefix@
+datadir=@datadir@
 compatdir=@compatdir@
-completionsdir=@datarootdir@/@PACKAGE@/completions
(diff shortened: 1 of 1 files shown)


scop:
Applying your change would have resulted in a warning from, `./configure`:
```shell
$ ./configure | grep WARNING
config.status: WARNING:  'bash-completion.pc.in' seems to ignore the --datarootdir setting
```
...and there were similar potential issues in `bash-completion-config.cmake`. I believe 81ba2c7e7dfbaefbafa1e8615727c9612e5fb314 fixes these.

PR #200, opened 2018-04-16
Add basic completion for 'iw'

Synray:
Only completes up to the 4th level of a command, at the moment. And only for commands starting with `iw dev`. The rest will come next, though this completion covers all of the most frequently used, as well as the `iw help` commands.
External: `sed` and `_available_interfaces` (iwconfig / ip).

### DIFF
--- a/completions/iw
+++ b/completions/iw
@@ -1,0 +1,326 @@
+# iw(8) completion                                         -*- shell-script -*-
+
+_iw()
+{
+    local cur prev words cword
+    _init_completion || return
+
+    local subcword obj cmd
+    for (( subcword=1; subcword < cword; subcword++ )); do
+        [[ "${words[subcword]}" == '--version' ]] && return
+        if [[ -n $obj ]]; then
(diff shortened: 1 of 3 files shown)


Abominas:
Completion for all "iw dev" commands.

I'll probably try to refactor the duplicate completions.
It's not wrapped at column 79 yet, but besides that is the code style correct?

Synray:
Sorry it's been a while since I worked on this. I don't really use `iw` anymore, but since I think `iw`'s horrendous syntax (and tendency to barf unsorted help to stdout when it fails) creates a drastic need for bash completion, I'll try to finish this.

I made the style changes you requested, apart from wrapping all the lines to 79 characters (which I'll do once the code is relatively stable). I also added completion for `iw phy`. The 3 commits were then squashed into one.

I'll add a test case next.

Synray:
Added a test, I looked at some other tests, but the length of each varies a lot, so let me know if I should add more to it.

PR #202, opened 2018-04-18
ab completion

ghost:
version 1.0

### DIFF
--- a/completions/ab
+++ b/completions/ab
@@ -1,0 +1,38 @@
+# POSIX ab completion                                   -*- shell-script -*-
+
+_ab()
+{
+    local cur prev words cword
+    _init_completion || return
+
+    case $prev in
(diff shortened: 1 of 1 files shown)


completions/ab
```diff
@@ -0,0 +1,38 @@
+# POSIX ab completion                                   -*- shell-script -*-
```
scop: Hardly POSIX :), "ab(1) completion" would sound more appropriate.

scop (changes requested):
In addition to the inline comments:
* Unit tests missing
* completions/Makefile.am modifications missing
* Should handle {non-,}completion of last bundled option's arg, see for example what was done to completions/bind in commit d664aebf6bd6ef5ca0602f6440887273bbe000e3

PR #210, opened 2018-05-08
Declare _xspecs as global.

jdswinbank:
By default, associative arrays are local. If bash_completion is sourced from
within a funciton, they won't propagate to the caller, and the system is not
initialized properly. By making this explicitly global, it propagates as
expected.


jdswinbank:
This would be more compelling if it worked with bash 4.1. As it is, I leave it here for future reference.

scop:
Merged with a fix to the bash version check, thanks.

PR #211, opened 2018-05-24
mplayer: Add common supported module music formats

theyamo:


scop:
Thanks. Uppercase S[3T]M missing and cloud add *.med while at it, but I'll take care of those.

PR #222, opened 2018-06-29
Support new GCC --completion option for bash completion

marxin:
GCC 9.1 will come up with a new option that will provide option hints. It's more precise that parsing of gcc --help output. First patch does a clean-up where I unified option providing. And I removed legacy compiler binary names.

### DIFF
--- a/completions/gcc
+++ b/completions/gcc
@@ -1,50 +1,58 @@
 # gcc(1) completion                                        -*- shell-script -*-
-#
-# The only unusual feature is that we don't parse "gcc --help -v" output
-# directly, because that would include the options of all the other backend
-# tools (linker, assembler, preprocessor, etc) without any indication that
-# you cannot feed such options to the gcc driver directly.  (For example, the
-# linker takes a -z option, but you must type -Wl,-z for gcc.)  Instead, we
-# ask the driver ("g++") for the name of the compiler ("cc1"), and parse the
-# --help output of the compiler.
 
 _gcc()
 {
-    local cur prev words cword
+    local cur prev prev2 words cword argument prefix prefix_length
     _init_completion || return
 
-    local cc backend
+    # Test that GCC is recent enough and if not fallback to
+    # parsing of --completion option.
+    if ! $1 --completion=" " 2>/dev/null; then
+        if [[ "$cur" == -* ]]; then
+            local cc=$($1 -print-prog-name=cc1 2>/dev/null)
+            [[ $cc ]] || return
+            COMPREPLY=($( compgen -W "$($cc --help 2>/dev/null | tr '\t' ' ' |\
+                command sed -e '/^  *-/!d' -e 's/ *-\([^][ <>]*\).*/-\1/')" -- "$cur" ))
+            [[ $COMPREPLY == *= ]] && compopt -o nospace
+            return
+        else
+            _filedir
+            return
+        fi
+    fi
 
-    case $1 in
-        gcj)
-            backend=jc1
-            ;;
-        gpc)
-            backend=gpc1
-            ;;
-        *77)
-            backend=f771
-            ;;
-        *95)
-            backend=f951
-            ;;
-        *)
-            backend=cc1 # (near-)universal backend
-            ;;
-    esac
+    # extract also for situations like: -fsanitize=add
+    if [[ $cword -gt 2 ]]; then
+        prev2="${COMP_WORDS[$cword - 2]}"
+    fi
 
+    # sample: -fsan
     if [[ "$cur" == -* ]]; then
-        cc=$($1 -print-prog-name=$backend 2>/dev/null)
-        [[ $cc ]] || return
-        # sink stderr:
-        # for C/C++/ObjectiveC it's useless
-        # for FORTRAN/Java it's an error
-        COMPREPLY=( $(compgen -W "$($cc --help 2>/dev/null | tr '\t' ' ' |\
-            command sed -e '/^  *-/!d' -e 's/ *-\([^][ <>]*\).*/-\1/')" \
-            -- "$cur") )
-        [[ $COMPREPLY == *= ]] && compopt -o nospace
-    else
+        argument=$cur
+        prefix=""
+    # sample: -fsanitize=
+    elif [[ "$cur" == "=" && $prev == -* ]]; then
+        argument=$prev$cur
+        prefix=$prev$cur
+    # sample: -fsanitize=add
+    elif [[ "$prev" == "=" && $prev2 == -* ]]; then
+        argument=$prev2$prev$cur
+        prefix=$prev2$prev
+    # sample: --param lto-
+    elif [[ "$prev" == "--param" ]]; then
+        argument="$prev $cur"
+        prefix="$prev "
+    fi
+
+    if [[ "$argument" == "" ]]; then
         _filedir
+    else
+        # In situation like '-fsanitize=add' $cur is equal to last token.
+        # Thus we need to strip the beginning of suggested option.
+        prefix_length=$((${#prefix}+1))
+        local flags=$($1 --completion="$argument" | cut -c $prefix_length-)
+        [[ "${flags}" == "=*" ]] && compopt -o nospace 2>/dev/null
+        COMPREPLY=($(compgen -W "$flags" -- ""))
     fi
 } &&
 complete -F _gcc gcc{,-5,-6,-7,-8} g++{,-5,-6,-7,-8} g77 g95 \
--- a/completions/gcc
+++ b/completions/gcc
@@ -12,13 +12,13 @@ _gcc()
             local cc=$($1 -print-prog-name=cc1 2>/dev/null)
             [[ $cc ]] || return
             COMPREPLY=($( compgen -W "$($cc --help 2>/dev/null | tr '\t' ' ' |\
-                command sed -e '/^  *-/!d' -e 's/ *-\([^][ <>]*\).*/-\1/')" -- "$cur" ))
+                command sed -e '/^  *-/!d' -e 's/ *-\([^][ <>]*\).*/-\1/')" \
+                -- "$cur" ))
             [[ $COMPREPLY == *= ]] && compopt -o nospace
-            return
         else
             _filedir
-            return
         fi
+        return
     fi
 
     # extract also for situations like: -fsanitize=add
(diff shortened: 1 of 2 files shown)


scop:
Eh, by the "I don't think this change is acceptable or necessary" part, I mean the "Rapidly simplify the script" commit.

completions/gcc
```diff
@@ -1,21 +1,67 @@
...
+      argument="$prev $cur"
+      prefix="$prev "
+    fi
+
+    if [[ "$argument" == "" ]]; then
+      _filedir
+    else
+      # In situation like '-fsanitize=add' $cur is equal to last token.
+      # Thus we need to strip the beginning of suggested option.
+      flags=$( gcc --completion="$argument" 2>/dev/null | sed "s/^$prefix//")
```
scop: * `flags` should probably be made `local`
* Should invoke `"$1"` instead of `gcc`
* The sed will fail if there are some problematic regex chars in `$prefix`. Would it be possible to just remove (length of characters in `$prefix`) from the start of these strings?
* If continuing to use sed nevertheless, should invoke `command sed` instead of `sed`

scop:
Yes, feel free to squash, with the indentation changes I'll need to re-review from scratch anyway. And no, I don't think the legacy binary support should be removed without a good, stated reason.

marxin:
> Yes, feel free to squash, with the indentation changes I'll need to re-review from scratch anyway. And no, I don't think the legacy binary support should be removed without a good, stated reason.

Good, I've just squashed the changes and included all your notes. Should be better now.

marxin:
Hello, may I please ping patch review. Thanks.

marxin:
> I'm in the middle of rewriting the test suite to pexpect, so other progress is pretty much stalled right now.

I see.

> 
> Anyway, a very brief look at the revised PR:
> 
> The gccgo addition should be a separate commit. I suggest submitting it as a separate PR altogether, so it can be applied quickly out of the way.

Done in: https://github.com/scop/bash-completion/pull/227

> The change silently remove the detection whether cc, c++ etc are actually gcc, g++ etc. This is not acceptable (especially when done without a notice in the commit message), because we don't want to apply gcc completion to non-gcc cc and friends.

You are right, that should remain there. I'll update that in next version.

> More reviewing sometime later when I find time, but as said, the gccgo change would be something good to have out of the way.

marxin:
@scop: Any progress on this please?

marxin:
Thanks, I updated the patch based on your requests.

marxin:
May I please gently ping the patch review process.

completions/gcc
```diff
@@ -1,49 +1,60 @@
...
-    local cur prev words cword
+    local cur prev prev2 words cword argument prefix prefix_length
     _init_completion || return
+    _expand || return
 
-    local cc backend
+    # Test that GCC is recent enough and if not fallback to
+    # parsing of --completion option.
+    $1 --completion=" " 2> /dev/null
+    if [[ "$?" -ne 0 ]]; then
```
Lekensteyn: Instead of checking `$?`, what about:

    if ! $1 --completion=" " 2>/dev/null; then

marxin: Sure, thanks. I'll rebase the patch.

jakubjelinek:
Any progress with this?  GCC 9 will be released shortly and is already used in some Linux distributions (e.g. Fedora 30).

marxin:
May I please ping the pull request.

marxin:
May I please ping that? note that GCC was already released.

marxin:
I've just rebased the patch. Can it be merged please?

completions/gcc
```diff
@@ -1,50 +1,60 @@
...
-    local cur prev words cword
+    local cur prev prev2 words cword argument prefix prefix_length
     _init_completion || return
+    _expand || return
 
-    local cc backend
+    # Test that GCC is recent enough and if not fallback to
+    # parsing of --completion option.
+    $1 --completion=" " 2> /dev/null
+    if ! $1 --completion=" " 2>/dev/null; then
```
scop: I wonder why two calls here, i.e. what's the first one for?

completions/gcc
```diff
@@ -1,50 +1,60 @@
...
-# linker takes a -z option, but you must type -Wl,-z for gcc.)  Instead, we
-# ask the driver ("g++") for the name of the compiler ("cc1"), and parse the
-# --help output of the compiler.
 
 _gcc()
 {
-    local cur prev words cword
+    local cur prev prev2 words cword argument prefix prefix_length
     _init_completion || return
+    _expand || return
```
scop: Could you clarify how addition of _expand relates to the topic of the change?

completions/gcc
```diff
@@ -1,50 +1,60 @@
...
 
-    local cc backend
+    # Test that GCC is recent enough and if not fallback to
+    # parsing of --completion option.
+    $1 --completion=" " 2> /dev/null
+    if ! $1 --completion=" " 2>/dev/null; then
+        if [[ "$cur" == -* ]]; then
+            local cc=$( $1 -print-prog-name=cc1 2> /dev/null )
```
scop: As noted in earlier review comments, please observe the existing coding style, e.g. we don't add spaces after `>`. As a new stylistic item, we no longer put spaces around stuff in `$( )`. Please check these in the whole patch, not just here.

completions/gcc
```diff
@@ -1,50 +1,60 @@
...
+        argument="$prev $cur"
+        prefix="$prev "
+    fi
+
+    if [[ "$argument" == "" ]]; then
         _filedir
+    else
+        # In situation like '-fsanitize=add' $cur is equal to last token.
+        # Thus we need to strip the beginning of suggested option.
+        prefix_length=`expr length "$prefix" + 1`
```
scop: Something pure bash like `$((${#prefix}+1))` should work here I think.

completions/gcc
```diff
@@ -1,50 +1,60 @@
...
+        prefix=$prev$cur
+    # sample: -fsanitize=add
+    elif [[ "$prev" == "=" && $prev2 == -* ]]; then
+        argument=$prev2$prev$cur
+        prefix=$prev2$prev
+    # sample: --param lto-
+    elif [[ "$prev" == "--param" ]]; then
+        argument="$prev $cur"
+        prefix="$prev "
+    fi
```
scop: The above looks quite complex to me, will need to have another look and try to digest later when I find time unless someone beats me to it.

completions/gcc
marxin: Sure, that's not needed.

completions/gcc
marxin: This one is also not needed, it's some kind of leftover.

marxin:
> Left some comments. As an additional overall thing, we should add a bunch of unit tests that cover both old and new functionality in test/t/test_gcc.py

Sure, I've just added some new GCC tests and will add some for older GCC version.

completions/gcc
marxin: Yes, it's quite complex due to bash argument parsing, where characters like space or `=` are separators. I tried to provide an example for each of the command line snippets.

marxin:
@scop I've just added some tests, but I will probably need help with the Docker. According to failing tests, it seems to me that `gcc` is not installed in the image. Can you please take a look?

scop:
Nope, gcc is in the relevant images, the builds automatically install packages for all our executables. So there's probably something else wrong with the fixtures.

I haven't had a deeper look yet, but note that you don't need to work on the docker tests from the start, you can just run the pytest tests on your local box. In a nutshell, setup a python 3 virtualenv, pip install -r test/requirements.txt, pytest test/t/test_gcc.py. Be sure to try out both new and old gcc cases.

Also, from now on, please submit further changes as separate commits instead of amending and force-pushing. That way I don't have to go through all the changes again each time I look. I can squash and tidy up in the eventual merge.

test/t/test_gcc.py
```diff
@@ -1,7 +1,64 @@
...
 
+from conftest import assert_bash_exec
+
 
 class TestGcc:
+    @pytest.fixture(scope="class")
+    def gcc_with_completion(self, bash):
+        got = assert_bash_exec(bash,
+                               "gcc --help=common | grep completion || :",
+                               want_output=True)
```
scop: I think this will fail always when there's no `completion` in the output, because there's no completion at all then, but `want_output=True` insists there must be.

marxin: You are right, I fixed that and it seems that it runs both of new and an older GCC release.

marxin:
> Also, from now on, please submit further changes as separate commits instead of amending and force-pushing. That way I don't have to go through all the changes again each time I look. I can squash and tidy up in the eventual merge.

Sure, I've just added the tests as one another git commit.

completions/gcc
```diff
@@ -1,50 +1,58 @@
...
+
+    if [[ "$argument" == "" ]]; then
         _filedir
+    else
+        # In situation like '-fsanitize=add' $cur is equal to last token.
+        # Thus we need to strip the beginning of suggested option.
+        prefix_length=$((${#prefix}+1))
+        local flags=$(gcc --completion="$argument" 2>/dev/null | cut -c $prefix_length-)
```
scop: Should invoke `$1`, not `gcc` here.

completions/gcc
```diff
@@ -1,50 +1,58 @@
...
+        [[ "${flags: -1}" == '=' ]] && compopt -o nospace 2>/dev/null
```
scop: `$flags == *=` for consistency with rest of codebase?

completions/gcc
scop: I think bash substring would be preferable over `cut` here, and `${flags:$prefix_length}` could be used in `compgen` below (no need to trim it before that).

scop: What kind of cases will cause stderr output here? Support for `--completion` was already determined above.

marxin: I would like to use it, but output of `--completion` is multiline and I can't use bash substring format.

scop:
To repeat an earlier request I made:

> from now on, please submit further changes as separate commits instead of amending and
> force-pushing. That way I don't have to go through all the changes again each time I look.

marxin:
> To repeat an earlier request I made:
> 

Sorry, I will do it next time.

scop:
Applied with some minor fixups, thanks, and for the patience too.

marxin:
> Applied with some minor fixups, thanks, and for the patience too.

I thank you!

PR #241, opened 2018-10-11
Tshark completion updates and fixes

Lekensteyn:
A couple of updates and fixes:

* tshark: update -T and -t completions
  The new -t values are available since 1.12 while the the most recent -T
  value was added in 2.4. Current as of v2.9.0rc0-2173-g9fcb4af6b6.
  Hardcode options as executing tshark can be slow while the options do
  not change that option.
* tshark: prevent a single-character file from breaking -G completion
* tshark: support .gz and .cap files for -r expansion
  Wireshark performs content sniffing to recognize the file type, so
  perhaps the extension filter should be removed completely. Until that
  happens, let's recognize .gz-compressed captures and .cap files too.

As for further improvements, executing an non-optimized debug build of `tshark` is slow (for `-O`, `-G`, etc.). Are there any standard methods for memoization that can be used?

### DIFF
--- a/completions/tshark
+++ b/completions/tshark
@@ -58,13 +58,15 @@ _tshark()
             return
             ;;
         -*T)
-            # TODO: could be parsed from "-T ." output
-            COMPREPLY=( $( compgen -W 'ps text pdml psml fields' -- "$cur" ) )
+            # Parse from: tshark -T . 2>&1 | awk -F \" '/^\t*"/ { print $2 }'
+            COMPREPLY=( $( compgen -W \
+                'pdml ps psml json jsonraw ek tabs text fields' -- "$cur" ) )
             return
             ;;
         -*t)
-            # TODO: could be parsed from "-t ." output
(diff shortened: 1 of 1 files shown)


scop:
Looks good, applied.

No standards for memoization exist, and I'm not sure we want to go there. But if you want to give it a go anyway, please at least do it without writing to files -- we have a policy for that in place (see CONTRIBUTING.md).

Lekensteyn:
> No standards for memoization exist, and I'm not sure we want to go there. But if you want to give it a go anyway, please at least do it without writing to files -- we have a policy for that in place (see CONTRIBUTING.md).

Yes, I was planning on using a shell variable and no temporary file. There is one generic example specific for git here: https://github.com/git/git/blob/5a0cc8aca797dbd7d2be3b67458ff880ed45cddf/contrib/completion/git-completion.bash#L380

My ASAN+Debug build takes almost a full second to run `tshark -G protocols`, its output is 117 kiB. Before I noticed this project I had an override in place long time ago (`complete -F _minimal tshark`) due to `-r` not completing all kind of files. That should be fixed now though :)

PR #242, opened 2018-10-11
chromium-browser: support more binaries, add more options

Lekensteyn:
Only Ubuntu seems to call the binary chromium-browser, Debian, Arch
Linux and probably all others call it chromium.

The `chromium --help` output only suggests --password-store which is not
very helpful. The full list of options is pretty large and might change
over time, so just add the (extension) developer-oriented options that
are used in my private completion script.
___
Validated all options for existence using:
```sh
links=$(xsel)  # from X selection
curl https://peter.sh/experiments/chromium-command-line-switches/ > petersh.html
text=$(html2text --ignore-links petersh.html)
for i in ${links//=};do echo "$text" | grep -qe "$i" || echo "$i"; done
```


completions/Makefile.am
```diff
@@ -759,7 +759,7 @@ symlinks: $(targetdir) $(DATA)
 		rm -f $(targetdir)/$$file && \
 			$(LN_S) cardctl $(targetdir)/$$file ; \
 	done
-	for file in google-chrome google-chrome-stable ; do \
+	for file in google-chrome google-chrome-stable chromium chrome; do \
```
scop: Need these in .gitignore and Makefile's CLEANFILES too

Lekensteyn:
I've split off the binary changes into #245 (and fixed Makefile.am and .gitignore, thanks for the suggestion). I've also elaborated the commit message of this change with the motivation for hardcoding these options, hopefully it is convincing :-)

PR #244, opened 2018-10-11
conftest: fix RemovedInPytest4Warning due to use of node.get_marker

Lekensteyn:
There should only be one marker, so get_closest_marker (introduced with
pytest 3.6.0 (2018-05-23)) is the most accurate replacement.

Fixes the following warning while running tests:

    test/t/conftest.py:155: RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
    Please use node.get_closest_marker(name) or node.iter_markers(name).
    Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
___
This reduces noice while running tests with newer pytest versions (locally and with Travis). A documentation update in doc/testing.rst is probably necessary though as only Debian buster and Ubuntu cosmic have the required pytest version. Older distributions have to install a more recent pytest version using pip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ standard commands for GNU autotools packages:
 autoreconf -i  # if not installing from prepared release tarball
 ./configure
 make
-make check # optional, requires python3 with pytest and pexpect, dejagnu, and tcllib
+make check # optional, requires python3 with pytest >= 3.6 and pexpect, dejagnu, and tcllib
 make install # as root
 ```
 
--- a/doc/testing.txt
+++ b/doc/testing.txt
@@ -26,7 +26,12 @@ http://wiki.tcl.tk/708[Tcl Style Guide].
 Installing dependencies
 -----------------------
 
-Installing dependencies should be easy using your local package manager.
+Installing dependencies should be easy using your local package manager. Note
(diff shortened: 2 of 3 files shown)


scop:
Maybe we could just try calling `get_current_marker`, catch exception if it doesn't exist and try again with `get_marker`? Requiring >= 3.6.0 would mean trouble/extra work for a bunch of users for pretty much cosmetic reasons at this point.

scop:
Or even monkey patch pytest at runtime to alias `get_current_marker` to `get_marker` if it doesn't exist.

Lekensteyn:
The node class is internal to pytest so monkey patching that seems risky. What do you think about this ugly-but-functional approach?

scop:
I think it would be ok. However, I just did a bit of testing, and apparently all of what we use this stuff for doesn't even work with pytest < 3.6 -- e.g. the class level bashcomp markers are not found at all with those versions with `get_marker`. If that's really so and is not easily fixable, I think your original approach would be fine after all, along with the documentation update about >= 3.6.0 being required.

Lekensteyn:
Reverted to the original version and updated documentation. For full test coverage you probably want to install a lot of dependencies, and at that point you might as well just install an updated pytest (as you probably did in the docker image).

PR #245, opened 2018-10-13
chromium-browser: consider chrome and chromium as aliases

Lekensteyn:
Only Ubuntu calls it "chromium-browser", all others (Debian, Arch Linux,
Fedora) call it "chromium". When built from source, the resulting binary
is called "chrome".
___
This will likely not suggest anything because `chromium --help` by default does not contain any options. It is recommended to apply #242 for more helpful completion.

For reference, this is my `chromium --help` output:
```
Chromium launcher v6 -- for Chromium help, see chromium(1)

Custom flags are read from the following file:

  /home/peter/.config/chromium-flags.conf

Arguments contained in that file are split on whitespace and shell quoting
rules apply but no further parsing is performed. Lines starting with a hash
symbol (#) are skipped. Lines with unbalanced quotes are skipped as well.

Currently detected flags:

  --password-store=basic
```

If I remove that file (or set `HOME=/bla`), then no options will be suggested at all.

### DIFF
--- a/completions/.gitignore
+++ b/completions/.gitignore
@@ -17,6 +17,8 @@ btdownloadgui.py
 c++
 cc
 cdrecord
+chrome
+chromium
 ci
 ciptool
 civclient
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -481,6 +481,8 @@ CLEANFILES = \
 	c++ \
 	cc \
 	cdrecord \
+	chrome \
+	chromium \
 	ci \
 	ciptool \
 	civclient \
@@ -759,7 +761,7 @@ symlinks: $(targetdir) $(DATA)
 		rm -f $(targetdir)/$$file && \
 			$(LN_S) cardctl $(targetdir)/$$file ; \
 	done
-	for file in google-chrome google-chrome-stable ; do \
+	for file in google-chrome google-chrome-stable chromium chrome; do \
 		rm -f $(targetdir)/$$file && \
 			$(LN_S) chromium-browser $(targetdir)/$$file ; \
 	done
--- a/completions/chromium-browser
+++ b/completions/chromium-browser
@@ -30,6 +30,6 @@ _chromium_browser()
     _filedir "@(?([xs])htm?(l)|pdf)"
 } &&
 complete -F _chromium_browser chromium-browser google-chrome \
-    google-chrome-stable
+    google-chrome-stable chromium chrome
 
 # ex: filetype=sh


scop:
Thanks! Here's my `chromium-browser --help` output (Ubuntu 18.04). Not that much stuff there either, but more than in yours.

```
chromium-browser [-h|--help] [-g|--debug] [--temp-profile] [--no-touch-pinch] [options] [URL]

        --verbose               Events logged to stderr.
        -g or --debug           Start within /usr/bin/gdb
        -h or --help            This help screen
        --temp-profile          Start with a new and temporary profile
        --no-touch-pinch        Disable pinch gestures.

 Other supported options are:
       Chromium has hundreds of undocumented command-line flags that are added
       and removed at the whim of the developers.   Here,  we  document  rela‐
       tively stable flags.

       --user-data-dir=DIR
              Specifies  the directory that user data (your "profile") is kept
              in.  Defaults to $HOME/.config/chromium .  Separate instances of
              Chromium must use separate user data directories; repeated invo‐
              cations of chromium-browser will reuse an existing process for a
              given user data directory.

       --app=URL
              Runs URL in "app mode": with no browser toolbars.

       --incognito
              Open in incognito mode.

       --new-window
              If PATH or URL is given, open it in a new window.

       --proxy-server=host:port
              Specify the HTTP/SOCKS4/SOCKS5 proxy server to use for requests.
              This overrides any environment variables or settings picked  via
              the  options  dialog.   An  individual proxy server is specified
              using the format:

                [<proxy-scheme>://]<proxy-host>[:<proxy-port>]

              Where <proxy-scheme> is the protocol of the proxy server, and is
              one of:

                "http", "socks", "socks4", "socks5".

              If  the  <proxy-scheme>  is omitted, it defaults to "http". Also
              note that "socks" is equivalent to "socks5".

              Examples:

                --proxy-server="foopy:99"
                    Use the HTTP proxy "foopy:99" to load all URLs.

                --proxy-server="socks://foobar:1080"
                    Use the SOCKS v5 proxy "foobar:1080" to load all URLs.

                --proxy-server="socks4://foobar:1080"
                    Use the SOCKS v4 proxy "foobar:1080" to load all URLs.

                --proxy-server="socks5://foobar:66"
                    Use the SOCKS v5 proxy "foobar:66" to load all URLs.

              It is also possible to specify a separate proxy server for  dif‐
              ferent URL types, by prefixing the proxy server specifier with a
              URL specifier:

              Example:

                --proxy-server="https=proxy1:80;http=socks4://baz:1080"
                    Load https://* URLs using the HTTP proxy "proxy1:80".  And
              load http://*
                    URLs using the SOCKS v4 proxy "baz:1080".

       --no-proxy-server
              Disables  the proxy server.  Overrides any environment variables
              or settings picked via the options dialog.

       --proxy-auto-detect
              Autodetect proxy configuration.  Overrides any environment vari‐
              ables or settings picked via the options dialog.

       --proxy-pac-url=URL
              Specify  proxy autoconfiguration URL.  Overrides any environment
              variables or settings picked via the options dialog.

       --password-store=<basic|gnome|kwallet>
              Set the password store to use.  The default is to  automatically
              detect  based  on  the  desktop  environment.  basic selects the
              built in,  unencrypted  password  store.   gnome  selects  Gnome
              keyring.  kwallet selects (KDE) KWallet.  (Note that KWallet may
              not work reliably outside KDE.)

       --version
              Show version information.

       As a GTK+ app, Chromium also obeys GTK+  command-line  flags,  such  as
       --display.  See the GTK documentation for more:

              <http://library.gnome.org/devel/gtk/stable/gtk-running.html>
              <http://library.gnome.org/devel/gtk/stable/gtk-x11.html>

 See 'man chromium-browser' for more details
```

PR #246, opened 2018-10-15
test: try to solve sporadic failures for 'no completion' tests

Lekensteyn:
A race condition exists between the program writing data and the moment
where 'expect' reads the output. Travis showed sporadic failures that
indicate reading a truncated version of MAGIC_MARK which was
subsequently interpreted as part of the completion output (instead of
"no completion").

To avoid this wrong match, always check for the magic marker. Tested
with test_look.py and various versions of /usr/share/dict/words (one
empty file, one with 'foo' and one with 'foo' and 'food' as contents).


Lekensteyn:
Yay, the important jobs of travis are passing now (except for some pylint issues).

Lekensteyn:
Rebased and fixed a flake8 complaint about having only one space before the `'#` comment sign. I'll address the remaining (existing) flake8 issues in a different PR.

scop:
Awesome! This has been nagging to me for a long time but I haven't found the cycles to look into it. :100:

PR #248, opened 2018-10-24
7z: add .msi support

perlun:
The 7z command line tool supports unpacking `.msi` files. Reference: https://www.7-zip.org/

(I have also tried it personally and noted that bash-completion didn't include these files in its completion.)


scop:
Thanks!

PR #249, opened 2018-11-01
tshark: Support preferences (-o) completion with memoization

Lekensteyn:
Support both completion of `-o tcp....` and `-otcp....`. Memoize the
tshark output since it can take 1 second with ASAN+Debug. _tshark_prefs
is 28190 bytes for tshark v2.9.0rc0-2375-gc672124881.

### DIFF
--- a/completions/tshark
+++ b/completions/tshark
@@ -2,10 +2,34 @@
 
 _tshark()
 {
-    local cur prev words cword
+    local cur prev words cword prefix
     _init_completion -n : || return
 
-    case $prev in
+    case $cur in
+        -o*)
+            prefix=-o
+            ;;
+    esac
+
+    case ${prefix:-$prev} in
+        --*)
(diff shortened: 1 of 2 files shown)


Lekensteyn:
Thanks for the feedback, I've removed grep and modified the pattern to match the expected conditions (preferences names are expected to contain a-z 0-9 _ and . only while modules seem to additionally permit `-`).

I've kept `tr` since the resulting variable is persisted and 1.2k extra lines makes the output from `set` harder to read. (I think that the original version without `tr` caused tests to fail due to newlines in the value.)

Lekensteyn:
Rebased, if this passes I plan to merge it since you saw no other issues so far :)

Lekensteyn:
Merged as 9b4098f748302568ff8c014f634fa822e962f029

scop:
Cool, thanks.

PR #252, opened 2018-11-03
Fallbacks for man completion

gftg85:
Paul Wise mentioned this problem in a bug report in Debian [1].

[1] https://bugs.debian.org/892307#27

### DIFF
--- a/completions/man
+++ b/completions/man
@@ -51,7 +51,8 @@ _man()
         return
     fi
 
-    local manpath=$( manpath 2>/dev/null || command man -w 2>/dev/null )
+    local manpath=$( manpath 2>/dev/null || command man -w 2>/dev/null || command man --path 2>/dev/null )
+    [[ -z $manpath ]] && manpath=$MANPATH
     [[ -z $manpath ]] && manpath="/usr/share/man:/usr/local/share/man"
 
     # determine manual section to search


completions/man
```diff
@@ -51,7 +51,8 @@ _man()
         return
     fi
 
-    local manpath=$( manpath 2>/dev/null || command man -w 2>/dev/null )
+    local manpath=$( manpath 2>/dev/null || command man -w 2>/dev/null || command man --path 2>/dev/null )
```
scop: Are there systems have support for `--path` but not `-w`? See commit 492773098004700d670b4e1c12eebac506471d90

pabs3:
I don't know of any specific situation that this change fixes, however even just amongst Linux distributions there are several man implementations and if one looks at all the BSD distributions, all the Unix distributions and all the other operating systems that have a man implementation, there are likely quite a few more. I think the right approach would be to look at the available options on each of the OSes that have a man implementation and order them by decreasing availability/portability.

I also wonder if any man implementations actually have a way to do completion without the shell parts having to do the completion, since that would have way more chance of being accurate.

cjwatson:
http://mandoc.bsd.lv/man/man.options.1.html#w is an excellent reference on portability.

pabs3:
So that page and this one indicate that `man -w` does not do the same thing on Illumos. It also sounds like using `man -w` from bash-completion as root on Illumos will have undesired side effects (`Updates the whatis database.`).

https://illumos.org/man/1/man
https://github.com/illumos/illumos-gate/blob/master/usr/src/cmd/man/man.c

pabs3:
It sounds like the `-l` option is what is needed on Illumos (and probably Solaris too) but *not anywhere else*, but the bash-completion implementation would need to be different:

https://github.com/illumos/illumos-gate/blob/master/usr/src/cmd/man/man.c#L1202-L1205

I conclude my addition of --path is worthless at best and harmful at worst.

scop:
With regards to falling back to `$MANPATH`, I'm not sure if that's necessarily the right thing to do either -- consider for example a setup that supports `manpath` or `man -w`, but returns empty from one of those even when `$MANPATH` is set to something. It might not be the right thing to do for us to use `$MANPATH` either in those cases. Perhaps a far fetched example but it _could_ result in a regression, which is why I'm not keen on touching this stuff except to fix demonstrable breakage.

gftg85:
Thanks for all the feedback.  I'm closing the bug, because I understood that the patch is not helpful at this time.  If you believe otherwise, feel free to reopen.

PR #254, opened 2018-11-16
Add completions for OpenRC rc-service

gftg85:
I did some testing, but I wasn't technically able to write a test case. :(


gftg85:
Any comments on this patch?

scop:
Sorry for taking this long. It seems OpenRC already comes with bash completions, so we shouldn't add more here. https://github.com/OpenRC/openrc/tree/master/bash-completion

PR #255, opened 2018-11-17
tar: Support completions for zstd compression extensions

gftg85:
(description in the commit message)

### DIFF
--- a/completions/tar
+++ b/completions/tar
@@ -433,9 +433,9 @@ __tar_cleanup_prev()
 
 __tar_detect_ext()
 {
-    local tars='@(@(tar|gem|spkg)?(.@(Z|[bgx]z|bz2|lz?(ma|o)))|t@([abglx]z|b?(z)2))'
+    local tars='@(@(tar|gem|spkg)?(.@(Z|[bgx]z|bz2|lz?(ma|o)|zst))|t@([abglx]z|b?(z)2|zst))'
     ext="$tars"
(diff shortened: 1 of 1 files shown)


completions/tar
```diff
@@ -433,9 +433,9 @@ __tar_cleanup_prev()
 
 __tar_detect_ext()
 {
-    local tars='@(@(tar|gem|spkg)?(.@(Z|[bgx]z|bz2|lz?(ma|o)))|t@([abglx]z|b?(z)2))'
+    local tars='@(@(tar|gem|spkg)?(.@(Z|[bgx]z|bz2|lz?(ma|o)|zst))|t@([abglx]z|b?(z)2|zst))'
     ext="$tars"
-    regex='\(\(tar\|gem\|spkg\)\(\.\(Z\|[bgx]z\|bz2\|lz\(ma\|o\)\?\)\)\?\|t\([abglx]z\|bz\?2\)\)'
+    regex='\(\(tar\|gem\|spkg\)\(\.\(Z\|[bgx]z\|bz2\|lz\(ma\|o\)\?\|zst\)\)\?\|t\([abglx]z\|bz\?2\)\)'
```
scop: Shouldn't we have tzst for regex too?

gftg85: You're right. I'll submit a new version. I wonder, though, what test I could have done that would have exposed this omission.

scop: See the completions involving listing/completing filenames within archives.

scop:
BTW there was one more affected regex, that's now handled in commit a143560f315ddfbe7c4662cba8a27b93137f425b

PR #256, opened 2018-11-20
Add version warning

mojoTX:
I spent a good bit of time trying to figure out why bash-completion was not working on my Mac, and it took me a while to realize that it wasn't loading because the stock version of Bash on macOS is ancient. For instance, on macOS Sierra (10.12.6) the version of Bash is 3.2.57. 

This change would simply emit a gentle warning to stderr when the version is less than 4.1, instead of silently doing nothing. My hope is that this would make it faster for people to realize that there's an issue. This will likely only affect Mac users.


scop:
This would mean that users on systems with old bash without access to modify the system profile.d scripts or loading of them would be spammed on each shell startup, and we don't want that.

PR #257, opened 2018-11-21
Associate  *.xar (eXist-db application package) to unzip

gmella:
Please find a new file extension we can use with unzip.
Thx for your job


scop:
This has some potential to generate false positives, I would guesstimate that the other xar format is more widespread than the eXist-db one. Well, applied anyway, thanks.
https://en.wikipedia.org/wiki/Xar_(archiver)

PR #258, opened 2018-11-24
Fix perl completions with a space between option and argument

gftg85:
This patch is highly based on the patch provided in a Debian bug report [1]

The diff is a bit hard to read because of the indentation changes.  Using git's `--ignore-space-changes`, I get the following, easier-to-review diff:

```
$ git diff HEAD~1 --ignore-space-change 
diff --git a/completions/perl b/completions/perl
index 7b91d1b4..7752fee8 100644
--- a/completions/perl
+++ b/completions/perl
@@ -27,7 +27,6 @@ _perl()
         optPrefix=-P$prev
         optSuffix=-S/
         prefix=$prev
-    fi
 
         case $prev in
             -*[DeEiFl])
@@ -68,7 +67,7 @@ _perl()
                 ;;
         esac
 
-    if [[ "$cur" == -* ]]; then
+    elif [[ "$cur" == -* ]]; then
         COMPREPLY=( $( compgen -W '-C -s -T -u -U -W -X -h -v -V -c -w -d -D -p
             -n -a -F -l -0 -I -m -M -P -S -x -i -e' -- "$cur" ) )
     else
```
[1] https://bugs.debian.org/614775

### DIFF
--- a/completions/perl
+++ b/completions/perl
@@ -27,7 +27,6 @@ _perl()
         optPrefix=-P$prev
         optSuffix=-S/
         prefix=$prev
-    fi
 
     case $prev in
         -*[DeEiFl])
@@ -68,7 +67,23 @@ _perl()
             ;;
     esac
 
-    if [[ "$cur" == -* ]]; then
+    # Unlike other perl options, having a space between the `-e' and
+    # `-E' options and their arguments, e.g. `perl -e "exit 2"', is
+    # valid syntax.  However, the argument is neither a filename nor a
+    # directory, but one line of perl program, thus do not suggest
+    # _filedir completion.
+    elif [[ "$prev" == -e ]] || [[ "$prev" == -E ]]; then
+        return
+
+    # Likewise, `-I' also accepts a space between option and argument
+    # and it takes a directory as value.
+    elif [[ "$prev" == -I ]]; then
+        local IFS=$'\n'
+        compopt -o filenames
+        COMPREPLY=( $(compgen -d $optPrefix $optSuffix -- "$cur") )
+        return
+
+    elif [[ "$cur" == -* ]]; then
         COMPREPLY=( $(compgen -W '-C -s -T -u -U -W -X -h -v -V -c -w -d -D -p
             -n -a -F -l -0 -I -m -M -P -S -x -i -e' -- "$cur") )
     else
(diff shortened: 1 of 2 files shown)


gftg85:
I'm working on a fix for the test case failures.

gftg85:
OK, now I think that the patch is correct and that the tests will pass.

The previous submission unintentionally modified the behaviour of the `-e` and `-E` completions, by suggesting `_filedir` completion with a space.  The new submission reverts this behavioural change.

On the other hand, the new submission intentionally modifies the behaviour of the `-x` completion, because it is not correct to suggest directory completion for it with a space (see commit message for the detailed explanation).  It updates the expected result of test case, accordingly.

Thank you.

gftg85:
Hmm, one of the tests (DIST=ubuntu14) failed, but it seems like the failure is not related to my change.

gftg85:
Is there a way to request a rebuild of the failing test?

gftg85:
I updated against current master just to make the test suite run again, because I had the impression that the test suite failed due to reasons unrelated to the patch.  I'm glad that that was actually true and that the tests are now green.

Thanks for your patience, @scop! :)

gftg85:
Now I stalled for a long time. :)

Yes, I'm still interested in this pull request and I agree with your comment about the `-I` option (it does work with or without the space between switch and argument).  Nice catch!  I fixed the code and updated the commit message accordingly (and I removed the indentation "fixes"). Thank you!

By the way, I very much like what you're doing with the test suite.  Thanks for that, too! <3

gftg85:
Hi, have you had a chance to look at the updated version?

gftg85:
I rebased this patch against current master to ping its review. :)

Ville, is there anything preventing you from accepting it?  I know I wrote a very long commit message, and there it takes time to read it, but my intention was to make it faster to review, as it details the rationale for the patch (nobody needs to guess what the patch is doing, nor why).

Thanks!

gftg85:
> is there anything preventing you from accepting it?

This message sounds harsh after re-reading it, but that was not my intention.  What I wish I had asked is...  Is there anything I can do better to make this pull request easier to review and eventually merged?

scop:
Nah, applied in 8eba6d0c6b88dacd80762ada3a3bc5fe14673fb2 (GH doesn't seem to be able to track it back here and mark as merged). Also tweaked the indentation and added a pair of trivial test cases.

PR #260, opened 2018-11-25
Fallback to suggesting all files in _filedir_xspec

gftg85:


bash_completion
```diff
@@ -1938,6 +1938,13 @@ _filedir_xspec()
         }
         ))
 
+    # Try without filter if it failed to produce anything and configured to
+    [[ -n ${COMP_FILEDIR_FALLBACK:-} && ${#toks[@]} -lt 1 ]] && {
+        reset=$(shopt -po noglob); set -o noglob
```
scop: Need local for reset.

bash_completion
```diff
@@ -1938,6 +1938,13 @@ _filedir_xspec()
...
+        toks+=( $( eval compgen -f -- "\$(quote_readline "\$cur")" ) )
```
scop: Can we get away without the eval here? I.e. maybe just:
```
        toks+=( $( compgen -f -- "$(quote_readline "$cur")" ) )
```

gftg85:
I applied your suggestions and tested that they still produce the output describe in the commit message.

Thanks!

bash_completion
```diff
@@ -1904,7 +1904,7 @@ else
 fi
 _filedir_xspec()
 {
-    local cur prev words cword
+    local cur prev words cword local
```
scop: Um no, this is not correct :) Please remove this change.

bash_completion
scop: Just add `local` before `reset` right here on this line and it's all good.

gftg85:
Thanks, I have submitted an updated version.

PR #261, opened 2018-11-26
mutt: Expand sourced filenames, which may contain quotes

gftg85:
Description in the commit message


gftg85:
Now with a test case (I think I finally understood how the test suite works)

gftg85:
I have updated this branch against current master and fixed the conflicts. I think the test case failures are unrelated to this patch.

gftg85:
Hi, @scop,

Is there anything with this pull request that you believe should be changed? I'll happily write a new version, if so.

PS: I have just rebased the patches against current master.

scop:
Sorry about taking this long. The reason for that is that every time I eyeball this I see the `eval`, and that doesn't give me a warm fuzzy feeling at all. I wonder if there's a safer yet similarly concise way to achieve this?

Issue #235, opened 2018-09-02
Update test docs for pytest+pexpect


scop:
doc/testing.txt

scop:
In a nutshell it is:
1) Make sure Python 3.x (where x is >= 4 I believe) and pytest and pexpect for it are installed. Not sure about the minimum needed pytest/pexpect versions, that would have to be checked.
2) `cd test ; pytest t  # be sure to invoke the correct pytest`
For more details see test/docker/docker-script.sh

PR #262, opened 2018-11-28
Update usage message to match reality

0mp:
In order to change the tool one has to use `--tool` instead of `--tool_exec`.

### DIFF
--- a/test/run
+++ b/test/run
@@ -9,7 +9,7 @@ usage() {
     echo "Unrecognized options are passed through to dejagnu by default."
     echo
     echo "Interesting options:"
-    echo "  --tool=    Test against a different bash executable."
+    echo "  --tool=         Test against a different bash executable."
     echo "  --buffer-size   Change expect match buffer size from our default of 20000 bytes."
     echo "  --debug         Create a dbg.log in the test directory with detailed expect match information."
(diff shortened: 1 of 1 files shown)


test/run
```diff
@@ -9,7 +9,7 @@ usage() {
     echo "Unrecognized options are passed through to dejagnu by default."
     echo
     echo "Interesting options:"
-    echo "  --tool_exec=    Test against a different bash executable."
+    echo "  --tool=    Test against a different bash executable."
```
scop: Please add some space to align the description with the rest (can't see it, but I suppose we're not using tabs there).

0mp:
@scop Oh, alright!

It's just that I wanted to run those tests on FreeBSD and add them to the official port (see [a pending revision on FreeBSD Phabricator](https://reviews.freebsd.org/D18370) if you are interested). This way it will be easier to improve Bash completions on FreeBSD (and write some additional tests).

I guess I'll subscribe to #235 and wait for more updates on how to run the new tests suite.

scop:
Actually, this change is incorrect. `--tool_exec` is the right option for changing the tested bash executable, it gets passed all the way through to `runtest`.  `--tool` does different things.

Added some comments to #235.

PR #272, opened 2019-01-04
Distribute test files with make dist

gftg85:
Downstream maintainers might want to run the full test suite in their
continuous integration.  For instance, in Debian, it should be possible
to use autopkgtest to run all the test cases under test/t/.  This patch
adds the test cases to the tarball generated with make dist.


scop:
Thanks for the heads up, however we already enumerate all desired dist files in t/Makefile.am and t/unit/Makefile.am, so a better fix was to hook them up with the build appropriately; done in master now.

PR #276, opened 2019-01-22
docker: New completion

WqyJh:
Add new completion for docker cli. The completion file is copied from [docker's official repository](https://github.com/docker/docker-ce) under the path of [components/cli/contrib/completion/bash](https://github.com/docker/docker-ce/tree/master/components/cli/contrib/completion/bash).


scop:
No thanks, it's actually preferred to ship completions with their corresponding upstream projects rather than here.

PR #284, opened 2019-02-23
xm: Remove completion for unavailable command

gftg85:
I ran the test case locally and I didn't see any regressions.

### DIFF
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -439,7 +439,7 @@ bashcomp_DATA = 2to3 \
 		xfreerdp \
 		xgamma \
 		xhost \
-		xm \
+		_xm \
 		xmllint \
 		xmlwf \
 		xmms \
--- a/completions/_xm
+++ b/completions/_xm
@@ -1,5 +1,9 @@
 # bash completion for xm                                   -*- shell-script -*-
 
+# Use of this file is deprecated.  The 'xm' command itself is no longer
+# provided by upstream.  It has been replaced with the 'xl' command, for
+# which upstream provides completion, use that instead.
+
 _xen_domain_names()
 {
     COMPREPLY=( $( compgen -W "$( xm list 2>/dev/null | \


scop:
Some distros which are still on extended support still ship it. For example RHEL/CentOS 5 which are on extended life cycle support even have xen 3.x, and some RHEL/CentOS 6.x have 4.4.x versions.

If there are no actual known problems with a completion, I don't think we should be removing them at least without a deprecation period. I'm fine with deprecating this though if you think it would be a good idea; just modify this PR to do that (rename it to `_xm`, add comment at top explaining deprecation reason, see others starting with an underscore for examples).

Just to verify that I didn't miss anything: not sure what you mean by having ran the test case locally -- this PR removes the command and the test case, so... what test case was there left for you to run?

gftg85:
> Some distros which are still on extended support still ship it. For example RHEL/CentOS 5 which are on extended life cycle support even have xen 3.x, and some RHEL/CentOS 6.x have 4.4.x versions.

I see, but I also think they are not shipping updated version of bash-completion. As an example, RHEL 7.x ships bash-completion 2.1.

> If there are no actual known problems with a completion, I don't think we should be removing them at least without a deprecation period. I'm fine with deprecating this though if you think it would be a good idea; just modify this PR to do that (rename it to `_xm`, add comment at top explaining deprecation reason, see others starting with an underscore for examples).

I saw your recent commit to deprecate adb completions, and I though about doing the same for xm, but then I thought it would be OK to delete it at once, because the case for adb deprecation is different, in the sense that it's been replaced with a completion from upstream, as opposed to the case of xm, where the command itself does not exist anymore.

Anyhow, I understand your concern about a deprecation period and I'll write a new patch.  :)

> Just to verify that I didn't miss anything: not sure what you mean by having ran the test case locally -- this PR removes the command and the test case, so... what test case was there left for you to run?

I chose my words poorly...  What I meant to say is that I ran the test *suite* and I didn't see any regressions on *any* of the tests cases, even though the test suite does not succeed completely.  I always run the full test suite.  Sorry for the misunderstanding.

gftg85:
Updated version that deprecates xm completion, instead of removing it.  Thanks.

scop:
On Mon, Feb 25, 2019 at 2:13 PM Gabriel F. T. Gomes <
notifications@github.com> wrote:

> Some distros which are still on extended support still ship it. For
> example RHEL/CentOS 5 which are on extended life cycle support even have
> xen 3.x, and some RHEL/CentOS 6.x have 4.4.x versions.
>
> I see, but I also think they are not shipping updated version of
> bash-completion. As an example, RHEL 7.x ships bash-completion 2.1.
>
Yes. That's so old that I'm sure quite a few people may be tempted to
upgrade to a new upstream version, so...

>

PR #286, opened 2019-03-07
Don't override any pattern by its repetition in _longopt

przepompownia:
The option --recursive for `grep` command did not show up
because it was overriden by --directories.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1878,7 +1878,7 @@ _longopt()
 
     if [[ "$cur" == -* ]]; then
         COMPREPLY=( $( compgen -W "$( LC_ALL=C $1 --help 2>&1 | \
-            command sed -ne 's/.*\(--[-A-Za-z0-9]\{1,\}=\{0,1\}\).*/\1/p' | sort -u )" \
+            command grep --basic-regexp --only-matching -- '--[-A-Za-z0-9]\{1,\}=\{0,1\}' | sort -u )" \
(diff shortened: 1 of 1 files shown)


bash_completion
```diff
@@ -1878,7 +1878,9 @@ _longopt()
 
     if [[ "$cur" == -* ]]; then
         COMPREPLY=( $( compgen -W "$( LC_ALL=C $1 --help 2>&1 | \
-            command sed -ne 's/.*\(--[-A-Za-z0-9]\{1,\}=\{0,1\}\).*/\1/p' | sort -u )" \
+            command grep --basic-regexp --line-number --only-matching -- '--[-A-Za-z0-9]\{1,\}=\{0,1\}' | \
```
scop: These options to `grep` are not portable, see CONTRIBUTING.md

bash_completion
```diff
@@ -1878,7 +1878,9 @@ _longopt()
...
+            mawk -F: '{ if (!options[$1]) options[$1] = $2 } END { for (line in options) printf("%s\n", options[line]);  }' | \
```
scop: `mawk` isn't portable either, plain awk would be acceptable but I hope these commands can be simplified overall.

przepompownia: Right. Clear bash is sufficient.

scop:
Applied with a clarified commit message and a few followup commits, thanks!

PR #289, opened 2019-03-07
ethtool: New completion

kevinoid:
This commit adds completion for [ethtool](https://kernel.org/pub/software/network/ethtool/), "the standard Linux utility for controlling network drivers and hardware".

Some tests require the fix from #288 due to line wrapping for some completions.

Thanks for considering,
Kevin

### DIFF
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -94,6 +94,7 @@ bashcomp_DATA = 2to3 \
 		_eject \
 		eog \
 		ether-wake \
+		ethtool \
 		evince \
 		explodepkg \
 		export \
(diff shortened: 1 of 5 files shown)


kevinoid:
Rebased onto `master` now that #288 has been merged.

kevinoid:
Thanks for the feedback!

I will ask the ethtool maintainers if they would be interested in shipping the bash completion script with ethtool and follow-up with their response.

kevinoid:
This bash-completion script [will be included in the next ethtool release](https://lore.kernel.org/netdev/20190424195428.GA16350@tuxdriver.com/).  No need to ship it in bash-completion.  Thanks again for reviewing it and the suggestion to submit it upstream.

scop:
Awesome, thanks!

PR #291, opened 2019-03-10
_longopt: Skip _filedir completion with --no-filename

gftg85:
The '--no-filename' switch from 'grep' does not expect file and
directory argument.


scop:
Thanks, I opted to make it a bit more generic rather than exclude specific ones.

gftg85:
Thank you!

PR #293, opened 2019-03-13
_known_hosts_real: "Include" directive support in ssh_config

gvych:
Added support for expanding ssh hosts from configs included in default SSH configs. Useful for ssh configs shared between teammates
You can add next directive into main config:
```bash
user@note ~ $ head ~/.ssh/config 
Include /home/user/git/myrepo/ssh_additional_hosts
```
After that hosts from ssh_additional_hosts also in completion


gvych:
Oops! Sorry about that :)
Canceling than.

PR #295, opened 2019-03-28
Add a postfix opt to _known_hosts_real

lonyai:
I create an opt like -c because I like develop an ansible bash_completion file. And the colon suffix is required to ansible's inventory opts.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1488,6 +1488,7 @@ _included_ssh_config_files()
 #           -c             Use `:' suffix
 #           -F configfile  Use `configfile' for configuration settings
 #           -p PREFIX      Use PREFIX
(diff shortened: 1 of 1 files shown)


scop:
Could you elaborate on your use case more? You mention needing a colon suffix, but we already have the `-c` option which adds a colon suffix. Note also that you can add some other suffix to completions using for example something like this (adds a comma, add `compopt -o nospace` if you want it):

```
    _known_hosts_real -- "$cur"
    COMPREPLY=( $( compgen -S , -W '${COMPREPLY[@]}' -- "$cur" ) )
```

lonyai:
It's may be beautiful in one-liner.

scop:
What about the colon, we already have `-c` which adds it, and you mentioned that's what your completion requires?

PR #299, opened 2019-04-16
test for _xspecs declaration

bid0uille:
Test if \_xspecs is declared before to use xspec completion to avoid syntax error (invalid arithmetic operand) when the command contains non word character ([^a-zA-Z0-9_]). Typically, $cmd with a dot or an hyphen or a non ascii char triggers this syntax error when completing args because if _xspecs is not explicitely declared, ${_xspecs[$cmd]} is interpreted as a simple rather than associative array, and so $cmd in this context has to be a valid variable name rather than a key.
 Test only for declaration, but considered to test for associative array declaration: [[ $(declare -p _xspecs) =~ "declare -A _xspecs"]]

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2058,7 +2058,7 @@ __load_completion()
     done
 
     # Look up simple "xspec" completions
-    [[ "${_xspecs[$cmd]}" ]] && complete -F _filedir_xspec "$cmd" && return 0
+    [[ -n ${_xspecs+true} ]] && [[ "${_xspecs[$cmd]}" ]] && complete -F _filedir_xspec "$cmd" && return 0
 
     return 1
 }


bid0uille:
I'm on bash 5.0.2 but forgot to precise bash-completion version: 2.8 (debian testing/unstable)

1. I triggered the error by typing:
$ myscript.sh <tab>
which give me
$ myscript.sh bash: myscript.sh: syntax error: invalid arithmetic operator (error token is ".sh")

But thanks to your point 2., I notice that in current version of bash-completion, _xspecs is declared with "-Ag" option if bash > 4.2. In mine it's "-A" anyway, making _xspecs, I guess, not global and so indexed rather than associative in the __load_completion function.
Putting "-Ag" is solving the problem for me.
So sorry about the noise, you may forget about this bug as it's already fixed :)
Thanks for attention, and prettier solution, anyway!

PR #311, opened 2019-04-29
Added "luksClose" and "luksOpen" to cryptsetup

Wurzelmann:
Added "luksClose" and "luksOpen" to cryptsetup


scop:
We already suggest `open` and `close` there, the luks* ones added here are old backwards compatibility aliases which are intentionally omitted.

PR #314, opened 2019-04-30
test_arp: Skip if ARP tables are empty

mgorny:
Skip test_arp if 'arp' yields no output.  This e.g. happens when
the host is not networking-enabled.

### DIFF
--- a/test/t/test_arp.py
+++ b/test/t/test_arp.py
@@ -1,11 +1,8 @@
 import pytest
 
-from conftest import in_docker
-
 
 class TestArp:
-    @pytest.mark.xfail(in_docker(), reason="Probably fails in docker")
-    @pytest.mark.complete("arp ")
+    @pytest.mark.complete("arp ", skipif='test -z "$(arp 2>/dev/null)"')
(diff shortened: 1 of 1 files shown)


test/t/test_arp.py
```diff
@@ -5,7 +5,7 @@
 
 class TestArp:
     @pytest.mark.xfail(in_docker(), reason="Probably fails in docker")
-    @pytest.mark.complete("arp ")
+    @pytest.mark.complete("arp ", skipif='test -z "$(arp 2>/dev/null)"')
```
scop: I think the docker xfail and related in_docker import can be removed, this seems a better and more general fix?

mgorny: Sure. I wasn't sure what was the problem with docker.

scop:
Thanks!

test/t/test_arp.py
scop: I'm fairly confident that it was this one. We probably have some other ones that could be skipped more cleanly than doing a container check, too (nowadays grep for `in_container` if interested).

PR #315, opened 2019-04-30
test/unit: Fix expected values when ${HOME} is overriden

mgorny:
Tilde substitution '~' respects the value of ${HOME}, while '~user'
always uses home directory from passwd.  Adjust the tests using
the latter to do the same.  This fixes a number of test failures when
${HOME} is set to a temporary directory by the Gentoo build environment.

### DIFF
--- a/test/unit/__expand_tilde_by_ref.exp
+++ b/test/unit/__expand_tilde_by_ref.exp
@@ -1,13 +1,16 @@
 # @param string $out  Reference to variable to hold value of bash environment
 #                     variable $HOME.
-proc setup {home user} {
+proc setup {home user userhome} {
     upvar $home _home
     upvar $user _user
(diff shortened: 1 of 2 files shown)


scop:
Thanks, but I'm sorry, the old dejagnu/tcl/expect based test suite is on its way out, I don't want to land any new work or improvements on it. It's just pending getting the remaining few unit and completion tests ported over to pytest/pexpect.

mgorny:
Can I expect those tests to be fixed before the next release? I'd really love to finally have tests passing on Gentoo. Thanks for all the work on them, by the way!

scop:
I would very much like to see that happen. In fact I would have liked that to happen already for 2.9, but the remaining ones are somewhat hairy and I didn't want to delay the release any more because of them. Help is welcome!

PR #320, opened 2019-05-21
Allow completing '?' to the latest file in a directory

compholio:
This completion change allows retrieving the latest file in any
_filedir completion (whether it is the current directory or not).

Examples:
mv ?<TAB> -> mv latest-file.pdf
mv ~/Downloads/?<TAB> -> mv ~/Downloads/latest-file.pdf

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -557,6 +557,7 @@ _filedir()
 
     local -a toks
     local x reset
+    local newest=0
 
     reset=$(shopt -po noglob); set -o noglob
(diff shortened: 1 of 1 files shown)


scop:
Let's discuss at https://github.com/scop/bash-completion/discussions/637

This implementation has a few issues that would need fixing before accepting anyway, but I think the time is not right to go that deep into details as there are some larger concerns to resolve first.

PR #326, opened 2019-06-06
Improve completion of ip route subcommands

felixonmars:
Implemented frequently used completions for via, dev, and existing
routes.

### DIFF
--- a/completions/ip
+++ b/completions/ip
@@ -198,8 +198,7 @@ _ip()
                         COMPREPLY=( $(compgen -W "table default \
                         $($1 r|command cut -d ' ' -f 1)" -- "$cur") )
                     elif [[ $prev == dev ]]; then
-                        COMPREPLY=( $(compgen -W "$($1 r|command sed -ne \
-                        's/.*dev \([a-z0-9-]*\).*/\1/p')" -- "$cur") )
+                        _available_interfaces
                     elif [[ $prev == table ]]; then
                         : # TODO
                     else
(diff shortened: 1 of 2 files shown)


felixonmars:
The test failure looks irrelevant.

felixonmars:
Sorry for the delay. I should have adjusted accordingly :)

completions/ip
```diff
@@ -192,14 +192,14 @@ _ip()
...
+                        's/.*via \([0-9.]*\).*/\1/p')" -- "$cur") )
                     elif [[ $prev == $subcmd ]]; then
                         COMPREPLY=( $(compgen -W "table default \
-                        $(ip r|cut -d ' ' -f 1|sort -u)" -- "$cur") )
+                        $($1 r|command cut -d ' ' -f 1)" -- "$cur") )
                     elif [[ $prev == dev ]]; then
-                        COMPREPLY=( $(compgen -W "$(ip r|grep -oP \
-                        '(?<=dev )[a-z0-9-]*'|sort -u)" -- "$cur") )
+                        COMPREPLY=( $(compgen -W "$($1 r|command sed -ne \
+                        's/.*dev \([a-z0-9-]*\).*/\1/p')" -- "$cur") )
```
scop: Shouldn't we complete `dev` with `_available_interfaces` (maybe with `-a`) instead?

scop:
Also, some additions to the unit test would be nice.

completions/ip
felixonmars: It could be more correct but too verbose for me (as it includes interfaces that are hardly used for routing like `lo`, and also interfaces that are not even up currently (even with the "-a" switch). I have changed to use `_available_interfaces -a` though.

felixonmars:
The test failure looks irrelevant.

felixonmars:
A friendly ping to revive this pr :)

scop:
Yeah, guess I sat on this long enough. Thanks for the contribution and patience!

PR #329, opened 2019-06-29
chromium-browser: Add support for .mhtml files

ezr:
Chromium can open MHTML documents (single-page html files). This adds support for opening files with the `.mhtm?(l)` file extension.


scop:
Thanks, applied with commit message word-wrapped.

PR #332, opened 2019-07-08
Update SSH completions to current version

Jakuje:


Jakuje:
Thank you for the comment. I modified the check to take also older openssh versions into account.

Jakuje:
I am sorry, I wrote the last commit late and probably did not read it correctly. It works like a charm. Thanks for patience :)

completions/ssh
```diff
@@ -108,15 +108,21 @@ _ssh_suboption()
         canonicalizehostname)
             COMPREPLY=( $(compgen -W 'yes no always' -- "$cur") )
             ;;
-        *file|identityagent|include)
+        *file|identityagent|include|controlpath|proxycommand|revokedhostkeys|xauthlocation)
```
scop: Wouldn't proxycommand be better completed with `compgen -c -- "$cur"` than `_filedir`?

Jakuje: You are right. In this context, I noticed that I missed also the `LocalCommand` and `RemoteCommand` options, so I added them to the same branch as `ProxyCommand` in the last commit

scop:
Thanks!

Issue #333, opened 2019-07-12
Test suite requires non-standard, non-portable `getconf -a` parameter.


awilfox:
The standard for [`getconf(1)`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/getconf.html) does not specify the glibc extension `-a` (to print all).  However, the bash-completion test suite fails on POSIX conformant systems that have `getconf(1)` that does not support `-a`:

```
____________________________ TestGetconf.test_1 ____________________________
self = <test_getconf.TestGetconf object at 0x3fff8c1ab240>, completion = <CompletionResult []>

    @pytest.mark.complete("getconf P")
    def test_1(self, completion):
>       assert completion
E       assert <CompletionResult []>

../t/test_getconf.py:7: AssertionError
____________________________ TestGetconf.test_4 ____________________________

self = <test_getconf.TestGetconf object at 0x3fff8c1b51d0>, completion = <CompletionResult []>

    @pytest.mark.complete("getconf -v ")
    def test_4(self, completion):
>       assert completion
E       assert <CompletionResult []>

../t/test_getconf.py:19: AssertionError
```

awilfox:
Sorry for the noise, this appears already fixed in 70afc1ed3697c3171a004b7db2f19220117d2862.

Issue #312, opened 2019-04-30
Alpine Linux: Tests failing on latest release (v2.9)


oxr463:
I'm the maintainer of bash-completion for Alpine Linux,
and I have a few tests that are failing on the latest release:

- arping - using version from iputils.
- getconf - using version from musl libc.
- iconv - same as above.
- ifup/ifdown - using only version (as far as I know), from debian.

NOTE: I'm developing in an LXC container, so I'm assuming that's why ifup/ifdown are failing, the log mentions something about failing to run due to being in a Docker container.

See: https://gitlab.com/oxr463/aports/issues/31

[log.txt](https://github.com/scop/bash-completion/files/3129713/log.txt)

scop:
I believe container stuff (thus arping and ifup/ifdown) and getconf are now taken care of in master per the above two commits, but I'm not sure what to do about iconv. It doesn't seem to recognize any options whatsoever, just responds with "unrecognized option" with all `-*` I've tried.

oxr463:
I did some further digging, and apparently, there are functional differences between glibc and musl versions of iconv, (See: https://wiki.musl-libc.org/functional-differences-from-glibc.html#iconv).

Is there a way to force-skip a test?

scop:
Sure, there may be that kind of functional differences but that doesn't mean it'd have to affect bash-completion in any way. Adding a skip for option completion because the musl one doesn't output an usage message we could use to parse options from: https://git.alpinelinux.org/aports/tree/main/musl/iconv.c

oxr463:
Oh that's true. Sounds good to me!

scop:
Pretty much all of this should be taken care of in master, apart from killall/pidof failures which are tracked in #318.

oxr463:
Thanks!

nh2:
Could you make a release? NixOS with musl also suffers from this.

PR #335, opened 2019-07-12
test_getconf: finish fix to #312 / #333

awilfox:
The first test for getconf also relies on the non-portable `-a` option.
Expect failure on this test as well if it isn't available.

Fixes: 70afc1ed ("test_getconf: skip if -a doesn't output any POSIX_V*")

### DIFF
--- a/test/t/test_getconf.py
+++ b/test/t/test_getconf.py
@@ -2,7 +2,9 @@ import pytest
 
 
 class TestGetconf:
-    @pytest.mark.complete("getconf P")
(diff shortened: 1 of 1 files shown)


scop (changes requested):
The idea is correct, but the implementation doesn't quite match it. We don't want to check if POSIX_V* are returned here, but rather exactly what your description says -- i.e. to check if the `-a` option is available in the first place. Perhaps just a `! getconf -a &>/dev/null` would do the trick.

PR #337, opened 2019-08-01
Add exe extension to support some common rar self-extracting (SFX) archives

andreabravetti:
Add exe extension to support some common rar self-extracting (SFX) archives


scop:
Thanks!

PR #339, opened 2019-08-05
Adds `cd-shortcuts` a Bash completion function to improve the "cd"

plainspooky:
`_cd_shortcuts()` is a Bash completion function to improve "cd" command adding two new shortcuts:

#### up : `^{number}`
To go {number} directories up from current path, for example:
```
cd /usr/src/linux-headers-5.0.0-20-generic/arch/x86/kernel
cd ^4<Tab> → cd ../../../../
```
#### at : `@<name>`
To go to first {name} directory from the right to the left starting to current path.
```
cd /usr/src/linux-headers-5.0.0-20-generic/arch/x86/kernel
cd @src<Tab> → cd /usr/src
```
If `^` or `@` are not detected it calls the original completion for "cd".

### DIFF
--- a/completions/cd-shortcuts
+++ b/completions/cd-shortcuts
@@ -24,6 +24,7 @@
 #   You should have received a copy of the GNU General Public License
 #   along with this program.  If not, see http://www.gnu.org/licenses/
 #
+
 _cd_shortcuts(){
 
     local let valid_words=0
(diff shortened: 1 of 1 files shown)


plainspooky:
I've checked travis-ci logs and looks like that some error happened during tests on Debian 10 but not related with my PR.

PR #340, opened 2019-08-09
Add .aab to unzip extension list.

pcc:
.aab is the extension used for Android App Bundles, which are zip files.
https://developer.android.com/guide/app-bundle


scop:
Thanks!

PR #345, opened 2019-09-05
Backport several upstream fixes

timvisher:


scop:
I don't use 1.x anymore myself and have too little time for 2.x already, so we need someone else to look after that. Personally, I think it would be best to leave 1.x to rest.

timvisher:
That's a shame because for older macOS users 1.x is what they're stuck with. :(

PR #351, opened 2019-09-24
apt-get: fix if the version number contains a colon

przepompownia:
The output from `apt-cache madison` consist of two lines only and does not require using regular expressions.

### DIFF
--- a/completions/apt-get
+++ b/completions/apt-get
@@ -35,11 +35,12 @@ _apt_get()
                     return
                 elif [[ $cur == *=* ]]; then
                     IFS='=' read -r package cur <<< "$cur"
-                    mapfile -t COMPREPLY < <(compgen -W "$( \
+                    COMPREPLY=($(IFS=$'\n' compgen -W "$( \
+                        apt-cache madison "$package" 2>/dev/null | \
                         while IFS=' |' read -r _ version _; do
                             echo "$version"
-                        done < <(apt-cache madison "$package"))" \
-                        -- "$cur")
+                        done )" \
+                            -- "$cur"))
                     __ltrim_colon_completions "$cur"
                     return
                 fi
--- a/completions/apt-get
+++ b/completions/apt-get
@@ -2,7 +2,7 @@
 
 _apt_get()
 {
-    local cur prev words cword
+    local cur prev words cword package
     _init_completion -n ':=' || return
 
     local special i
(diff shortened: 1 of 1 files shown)


przepompownia:
Currently, for example, package `thunderbird` has versions `1:60.9.0-1` `1:68.0~b1-1` on Debian sid/experimental.

completions/apt-get
```diff
@@ -34,11 +34,13 @@ _apt_get()
                     _filedir deb
                     return
                 elif [[ $cur == *=* ]]; then
-                    COMPREPLY=( $(compgen -W "$(\
-                        apt-cache --no-generate show "${cur%%=*}" 2>/dev/null |
-                        command sed -ne \
-                            's/^Version:[[:space:]]*\([^[:space:]]\)/\1/p')" \
-                        -- "${cur#*=}") )
+                    IFS='=' read -r package cur <<< "$cur"
+                    mapfile -t COMPREPLY < <(compgen -W "$( \
```
scop: Process substitution will break under posix mode, and turning it off reportedly breaks other things, see 153d6d4fab1faafbd33fccb0ba8a99964f05a95b. So this should be implemented some other way.

completions/apt-get
```diff
@@ -34,11 +34,13 @@ _apt_get()
...
+                        while IFS=' |' read -r _ version _; do
+                            echo "$version"
+                        done < <(apt-cache madison "$package"))" \
```
scop: `--no-generate` and `2>/dev/null` were dropped, any particular reason? They seem as good ideas as before.

przepompownia: The output from `apt-cache madison` does not consist of unneeded lines (not only of two as I mentioned above by mistake ;)) and is simpler to parse (it does not require any external tool like `sed`).

Right - the redirection of the error output was missing.

scop: But we still want `--no-generate` with it, no?

przepompownia:
Right: --no-generate is an option of the whole `apt-cache` command, not only `apt-cache show`. Forgive careless reading ;)

przepompownia:
It is possible that package cache regeneration caused exceeded timeout for tests execution. The last commit is the first without it.

completions/apt-get
```diff
@@ -34,11 +34,14 @@ _apt_get()
                     _filedir deb
                     return
                 elif [[ $cur == *=* ]]; then
-                    COMPREPLY=( $(compgen -W "$(\
-                        apt-cache --no-generate show "${cur%%=*}" 2>/dev/null |
-                        command sed -ne \
-                            's/^Version:[[:space:]]*\([^[:space:]]\)/\1/p')" \
-                        -- "${cur#*=}") )
+                    IFS='=' read -r package cur <<< "$cur"
```
scop: Two remaining notes on this line:
- `package` should quite likely be made `local`
- herestrings i.e. `<<<` use temporary files, would be better to find a way to implement this without them

scop: For example something like (untested):
```
local package=${cur%%=*}
cur=${cur#*=}
```

scop:
Thanks!

Issue #190, opened 2018-03-03
"make" completion is overwriting readline keybindings


dsifford:
For whatever reason, the `make` completion is overwriting the globally set tab keybinding (which I have set to `menu-complete`).

This is easily reproducible. Check the example below.

[![asciicast](https://asciinema.org/a/aNeXIX8iOyrYZ2CYMLRDkD8CZ.png)](https://asciinema.org/a/aNeXIX8iOyrYZ2CYMLRDkD8CZ)

Why could this be happening?

scop:
No idea offhand. Is make completion the only thing this happens with?

dsifford:
To the best of my knowledge, yes.

I haven't systematically tested this with any of the others, but I can say with some level of certainty that this is the first time that I've ever ran into this. I use a pretty wide array of tools that have associated completions on a frequent basis, so I'm somewhat certain that this issue stems from the make completion itself.

bewuethr:
As far as I can tell, the completion for `make` is the only one that checks the `COMP_TYPE` variable, i.e., checks what type of completion is being attempted &ndash; maybe related? I don't see where the binding could be affected, though.

dsifford:
Are either of you able to reproduce this on your end @scop @bewuethr ?

bewuethr:
Yes, I can reproduce it; haven't tried to narrow it down, though.

bewuethr:
I tried isolating the cause, but now (on a different machine) I am unable to reproduce. The binding isn't removed any longer.

dsifford:
Nope, nothing special about the file.

I'm able to reproduce with the following Makefile on my machine:

```make
.PHONY: foo
foo:
        @echo foo

.PHONY: bar
bar:
        @echo bar
```

```
$ uname -a
Linux archlinux 4.15.5-1-ARCH #1 SMP PREEMPT Thu Feb 22 22:15:20 UTC 2018 x86_64 GNU/Linux
$ bash --version
GNU bash, version 4.4.19(1)-release (x86_64-unknown-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

scop:
Can't reproduce using the given Makefile, current bash-completion master, and

```
$ uname -a
[...]

Issue #143, opened 2017-07-02
make: 'bash: set +o posix: command not found' when IFS=$'\n'


mgorny:
To reproduce:

```
$ IFS=$'\n'
$ make <tab>
```

It causes the output to look like:
```
$ make bash: set +o posix: command not found
bash: set +o posix: command not found

Display all 124 possibilities? (y or n)
```

etc., i.e. it works but prints two `set +o posix` errors every time it completes.

Original report: https://bugs.gentoo.org/530246

scop:
This in completions/make appears to be one candidate fix, not sure if it's the best one:

```diff
-        $reset
+        local IFS+=' '; $reset
```

And we probably have a bunch of other cases (at least all similar `shopt` constructs) lurking here and there, possibly others as well. Not sure if we should even try to support arbitrary `IFS` environments...

mgorny:
Well, to be honest I don't know whether it's worth supporting either. This has been reported by an user who actually did hit this, so might be worth fixing it on per-case basis. However, if you feel like it's too much of an hassle, feel free to reject it and I'll forward the answer appropriately.

As for the technical details, I'm afraid IFS is too much of a brainf* for me. I always get lost on where it exactly gets used and where not. I'm not sure if the best solution wouldn't be to simply override IFS for the whole function scope, unless we actually need to respect it somewhere.

scop:
There's no central place we could use for setting an expected IFS, so I opted to tweak the cases I could find. I kept the IFS setting close to where it's needed in order to give a hint/reminder why it is there.

PR #352, opened 2019-10-07
bind: add variable and fix function completion

algorythmic:
This adds completion to the common invocation forms of:

    bind 'set variable-name value'
and

    bind '"keyseq":function-name'

A little tricky because the whole argument is likely to be quoted and/or has colons in it.

Supported completions (works w/ or w/o `:` in `$COMP_WORDBREAKS`):

	bind '"\C-r":ab    ==>    '"\C-r":abort'
    bind "\C-r":ab     ==>    "\C-r":abort
	bind "Con          ==>    "Control-          "Control:
    bind 'set ke       ==>    'set keymap        'set keyseq-timeout
    bind 'set keymap'  ==>    'set keymap emacs  'set keymap vi       ...

The variable names and values are completed in a case-insensitive and case-preserving manner.

I added some test but they seem to all be getting skipped, unless I put the `_bind()` function definition first in the file -- not sure what's up with that.

### DIFF
--- a/completions/bind
+++ b/completions/bind
@@ -1,16 +1,73 @@
 # bash bind completion                                     -*- shell-script -*-
 
+_bind_var_vals()
+{
+    local pre=$1 var=$2 val=$3 vals
+    case $var in
+        bind-tty-special-chars|blink-matching-paren|byte-oriented|\
+        colored-completion-prefix|colored-stats|completion-ignore-case|\
+        completion-map-case|convert-meta|disable-completion|\
+        echo-control-characters|enable-bracketed-paste|enable-keypad|\
+        enable-meta-key|expand-tilde|history-preserve-point|\
+        horizontal-scroll-mode|input-meta|mark-directories|\
+        mark-modified-lines|mark-symlinked-directories|match-hidden-files|\
+        menu-complete-display-prefix|meta-flag|output-meta|page-completions|\
+        prefer-visible-bell|print-completions-horizontally|\
+        revert-all-at-newline|show-all-if-ambiguous|show-all-if-unmodified|\
+        show-mode-in-prompt|skip-completed-text|visible-stats)
+            vals=(on off)
+            ;;
+        bell-style)
+            vals=(audible none visible)
+            ;;
+        editing-mode)
+            vals=(emacs vi)
+            ;;
+        keymap)
+            vals=(emacs emacs-standard emacs-meta emacs-ctlx vi vi-command
+                vi-insert)
+            ;;
+        *)
+            return 1
+            ;;
+    esac
+    local v reset=$(shopt -p nocasematch); shopt -s nocasematch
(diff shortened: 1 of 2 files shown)


completions/bind
```diff
@@ -1,16 +1,73 @@
...
+            ;;
+        *)
+            return 1
+            ;;
+    esac
+    local v reset=$(shopt -p nocasematch); shopt -s nocasematch
+    for v in "${vals[@]}"; do
+        [[ $v == "$val"* ]] && COMPREPLY+=("${pre}${val}${v/#$val}")
+    done
+    eval "$reset"
```
scop: I think a plain `$reset` should work here, no need to eval. And the `shopt` probably needs protection from nondefault `$IFS`. See existing similar constructs, commit 5240618033fecc3032a421d39c2eaeb2e4c20d57 and #143

completions/bind
```diff
@@ -1,16 +1,73 @@
...
+}
+
+_bind_vars()
+{
+    local pre=$1 var=$2
+    local v reset=$(shopt -p nocasematch); shopt -s nocasematch
```
scop: Same as above wrt shopt and eval

completions/bind
```diff
@@ -1,16 +1,73 @@
...
+    while IFS=' ' read _ v _; do
+        [[ $v == "$var"* ]] && COMPREPLY+=("${pre}${var}${v/#$var}")
+    done < <(bind -v)
```
scop: Process substitution doesn't work under posix mode, see commit 153d6d4fab1faafbd33fccb0ba8a99964f05a95b and #190 for why simply unsetting it is not a good solution, need to come up with some other way.

completions/bind
```diff
@@ -1,16 +1,73 @@
...
+            COMPREPLY=( "${pre}${var} " "${pre}${var}  " )
+        fi
+    fi
+}
+
 _bind()
 {
-    local cur prev words cword
-    _init_completion || return
+    local cur prev words cword IFS=$'\n'
```
scop: Hm, this might actually provide for the IFS protection I mentioned above. Right?

completions/bind
algorythmic: I was thinking `eval "$reset"` might be a better/cleaner construct to use in these cases, rather than doing the `$IFS` switcheroo, since `eval "$reset"` works the same way regardless of `$IFS` value and so doesn't require protection from non-default values.  Curious what you think.

completions/bind
algorythmic: I think `IFS` would have to contain a space in order for the expansion of `$reset` to be evaluated as separate words

completions/bind
algorythmic: Got it. Replaced with a command substitution approach.

completions/bind
scop: Consistency and eval avoidance are good things to pursue.

PR #359, opened 2019-11-03
Let find complete with directories after -path option

gftg85:
The argument to find's -path option is supposed to restrict the search
results to items that belong to the specified directory, so let
bash-completion suggest directories to it.

Reported by 積丹尼 Dan Jacobson [1].

[1] https://bugs.debian.org/918430

### DIFF
--- a/completions/find
+++ b/completions/find
@@ -17,6 +17,10 @@ _find()
             _filedir
             return
             ;;
+        -path|-ipath)
+            _filedir -d
+            return
+            ;;
         -fstype)
             _fstypes
             [[ $OSTYPE == *bsd* ]] && \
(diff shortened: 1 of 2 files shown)


gftg85:
Maybe that would be more inline with other completions currently do, for instance, suggestion of perl modules after `perl -M<TAB>`.

On the other hand, I fell like that is re-implementing find itself.  I do not feel like working on it, so I'll probably close this pull request.

Thanks for the feedback, Ville.

scop:
If you don't mind, I'd appreciate leaving it open even though you may not be working on it yourself, I or someone will pick it up eventually and the issue won't be forgotten.

I don't think improving this is a matter of re-implementing find at all. Just examine the arguments given to see what is the `starting-point` arg, and filter all dir completions that don't start with it. Defaulting to the starting point would probably be just a matter of `${cur:=$starting_point}`, grepping `:=` gives some ideas.

PR #365, opened 2019-11-12
cppcheck: Remove deprecated option 'posix'  for '--std='

versat:
'--std=posix' is deprecated and will be removed.
See this Cppcheck commit:
https://github.com/danmar/cppcheck/commit/cb06aebdab894c2ca767869889df8e1175aeb379

### DIFF
--- a/completions/cppcheck
+++ b/completions/cppcheck
@@ -52,7 +52,7 @@ _cppcheck()
             ;;
         --std)
             COMPREPLY=( $(compgen -W 'c89 c99 c11 c++03 c++11 c++14 c++17
-                c++20 posix' -- "$cur") )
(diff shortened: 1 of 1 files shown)


scop:
Applied, thanks. It would be nice if they'd output available standards in with for example --std=help and similarly --platform=help and --enable=help in some form we could parse available values from instead of having to hardcode them. The current --help output is pretty clearly not intended to be parsed.

versat:
> It would be nice if they'd output available standards in with for example --std=help and similarly --platform=help and --enable=help in some form we could parse available values from instead of having to hardcode them. The current --help output is pretty clearly not intended to be parsed.

Yes, that sounds like a good idea. I will create a ticket with your suggestion.

PR #366, opened 2019-11-13
allow autocomplete with serial ports

jerkey:
an important feature of GNU screen is its usefulness as a serial terminal.  Autocomplete normally does not work with serial devices because they are not real files, so this change autocompletes serial ports after $ screen /dev<TAB>

### DIFF
--- a/completions/screen
+++ b/completions/screen
@@ -25,6 +25,13 @@ _screen()
     local cur prev words cword
     _init_completion || return
 
+    if [[ $cword -eq 1 && $cur == /dev* ]]; then
+        COMPREPLY=( $(compgen -W "$(shopt -s nullglob; printf '%s\n' \
+            /dev/serial/by-id/* /dev/ttyUSB* /dev/ttyACM* 2>/dev/null)" \
+            -- "$cur") )
+        return
+    fi
+
     if ((cword > 2)); then
         case ${words[cword-2]} in
             -*[dD])


completions/screen
```diff
@@ -86,6 +89,10 @@ _screen()
     if [[ "$cur" == -* ]]; then
         COMPREPLY=( $(compgen -W '$(_parse_help "$1")' -- "$cur") )
     fi
+
+    if [[ "$cur" == /dev* ]]; then
+        COMPREPLY=( $( compgen -W "$(ls -1 /dev/serial/by-id/ /dev/ttyUSB* /dev/ttyACM* 2>/dev/null)" -- "$cur" ) )
```
scop: We should invoke `ls` as `command ls`.

scop: Stylistic nit: no space after `$(` or before the corresponding `)`.

completions/screen
```diff
@@ -44,6 +44,9 @@ _screen()
             -*)
                 continue
                 ;;
+            /dev*)
+                continue
+                ;;
```
scop: The screen man page documents that the serial device must be the first argument, so I think it would be better to move the whole serial case to a block of its own, immediately after `_init_completion`, something like
```bash
if [[ $cword -eq 1 && $cur == /dev* ]]; then
```

jerkey:
oops there is a problem with stuff like /dev/serial/by-id/usb-Prolific_Technology_Inc._USB-Serial_Controller_D-if00-port0
for some reason it autocompletes to
"$ screen /dev/serial/by-id/: " and that space is causing it to go on to a new word of the commandline.  I guess this is because of the dash in "by-id" and I am not sure how to fix it

completions/screen
jerkey: this is my first work with autocomplete!  whatever you think is best should be done, go ahead.  Please see my comment below about a bug with /dev/serial/by-id/... where the hyphen ("-") is causing interpretation to change to options.  Maybe your change would fix that?

scop: Nope, that didn't have anything to do with it, that's just what `ls` does:
```
/dev/ttyACM0

/dev/serial/by-id/:
usb-FIBOCOM_...
```

scop:
Modifed version applied, anyway congrats on your first contribution here! For future reference, the usual workflow here is that I (or someone else) reviews PR's and requests changes, and the submitter (in this case you) is expected to make the changes. This may go back and forth a couple of times, but eventually most PR's end up merged.

I'll push a couple of related screen improvements shortly.

jerkey:
@scop thank you for explaining that, I guess I didn't fully understand what changes you wanted me to make.  When I figure out the weird behavior with /dev/serial/by-id/* i'll try to do a better job with a PR then.  Thanks for enabling my first PR to this project!

PR #372, opened 2019-12-08
Use kaffeine completions for totem

gaul:
Extensions sourced from mplayer.


scop:
Also, in any case, a test suite item should be added for all new completions.

gaul:
> If we want to do just filename extensions, I think it would be a better to just add totem e.g. to the kaffeine xspec completion (or add a separate line for totem if the supported file types differ "enough") in bash_completion than to ship a separate file for it.

@scop Done.

scop:
Requested test case added in 9c33f68a04e0ec6766cb224ab025efbd42ca54e7

PR #373, opened 2019-12-08
Fix/#368

### LINKED ISSUE #368: Adding \ to a command to bypass alias use wrong completion
Hello,

I just discovered that using \ to bypass an alias cause bash-completion to use the wrong complete function.

For example :
```
$ cd <TAB>
$ complete | grep 'cd$'
complete -o nospace -F _cd cd
$ \cd <TAB>
$ complete | grep 'cd$'
complete -F _minimal \cd
complete -o nospace -F _cd cd
```

Anyway to fix it and considering \command equivalent to command ?

Regards,

rockandska:
Fix #368 

@scop , not sure my tests are good since i don't see executed the second test on `test/lib/completions/cd.exp`

Comments / corrections welcome

Edit: ok, i didn't get that we need to rebuild the docker test image each time we made modifications and see now the same errors at travis

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2116,16 +2116,27 @@ __load_completion()
     return 1
 }
 
+__load_completion_alias()
+{
+    local cmd="$1"
+    if [[ "$cmd" =~ ^\\ ]];then
+        cmd=$(complete -p "${cmd##*\\}" 2>/dev/null | tr -d $'\n' | sed -e 's/ [^ ]*$//'; printf " '%s'\n" "$cmd")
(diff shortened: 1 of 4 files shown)


rockandska:
CI tests seem not stable and failed randomly even on local
https://travis-ci.com/rockandska/bash-completion/builds/140140742

scop:
Thanks, I think I'd prefer something like #375 though. Lacks tests, but I believe it does the right thing in a simpler way than this PR. What do you think?

rockandska:
As said in #375 , look far better than mine ^^

PR #375, opened 2019-12-10
*: complete commands when prefixed with a backslash

scop:


rockandska:
Seems cleaner than #373 indeed.
Just one though, you assume there will be only one `\`.
Even if `\` is the correct way to unalias, the unalias seems to work even with more `\`.
I would better to assume this.

Thanks for the fix

scop:
I tested multiple backslashes already when creating the patch, but they don't work for me.
```
$ \\cd /tmp
bash: \cd: command not found
```

Issue #358, opened 2019-10-29
What does this do?


mariusa:
The README starts with installation, troubleshooting, FAQ...
Would you please add a few paragraphs and 1-2 examples with what does this do, expanding on 'Programmable completion functions for bash' ? 

Thanks

scop:
Pull requests welcome :) Preferably concise ones that point to bash's docs on this stuff.

PR #376, opened 2019-12-23
README: add introduction

cjmayo:
My attempt at describing the project (#358).
Do edit!


scop:
Thanks, added a different version, more geared towards general completion uses rather than project internals.

PR #389, opened 2020-02-23
python: support executables with minor version

kevinoid:
Debian, Ubuntu, and likely other distributions install the Python executable with a minor version (e.g. `python2.7`, `python3.8`) in addition to the major version (`python2`, `python3`) and unversioned (`python`) aliases.  Apply the completion to the minor versioned executables for versions supported in the last 3 years.

Thanks for considering,
Kevin

### DIFF
--- a/completions/.gitignore
+++ b/completions/.gitignore
@@ -163,7 +163,9 @@ pydoc3
 pylint-[23]
 pytest-[23]
 python2
+python2.7
 python3
+python3.[345678]
 pypy
 pypy3
 pyvenv-3.[45678]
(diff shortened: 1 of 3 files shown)


kevinoid:
> Thanks, could you add the symlinked ones to completions/.gitignore

Thanks @scop, good catch!  Should be fixed now.

scop:
Merged, thanks. There are/have been also ones with `m` and `u` suffix after the version, but we can add them in a separate change in case someone wants.

kevinoid:
My pleasure.  Thanks for the very fast review and merge!

Interesting.  I hadn't run across the `m` and `u` versions.  Sounds like a good plan to me.

scop:
IIRC the u ones were some unicode ones a long long time ago. Probably no longer relevant.

PR #390, opened 2020-02-23
Check that ruptime is present before calling

Self-Perfection:
Avoids calling command_not_found_handle which might take considerable
time.

---

For instance on my system command_not_found_handle takes more than 200ms. I don't want this delay every time I complete hostname!

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1685,9 +1685,11 @@ _known_hosts_real()
     fi
 
     # Add hosts reported by ruptime.
-    COMPREPLY+=( $(compgen -W \
-        "$(ruptime 2>/dev/null | awk '!/^ruptime:/ { print $1 }')" \
-        -- "$cur") )
+    if type ruptime &>/dev/null; then
+        COMPREPLY+=( $(compgen -W \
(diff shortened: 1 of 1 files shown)


scop:
We do this a lot, and I'd rather not start changing all of them or bits here and there. I wonder if there's a clean and robust way we could disable command_not_found_handle for the duration of running completions instead, for example in _init_completion.

Actually this might be something to suggest upstream bash do by default.

Self-Perfection:
When I was preparing change I noticed 10 lines above `avahi-browse` is tested for presence so I just went with the flow.

It might be possible to save command_not_found_handle state in the beginning, unset this function and restore after completion just like described here: https://mharrison.org/post/bashfunctionoverride/

Don't sure how safe and robust it is.

scop:
Ok, fair enough re avahi-browse. A centralized approach would still be preferred, as just calling stuff is what we do pretty much everywhere else in the codebase, and FWIW I'd rather keep it that way. The linked solution irks me though, and I'm not sure if we have a place to cleanly restore the original function.

PR #391, opened 2020-03-01
ip: add support for netns (based on #21)

felixonmars:
The original PR looks like abandoned. I added some improvements and hopefully addressed the issues mentioned there.

- Embed the available netns part inside completion
- Add 2>/dev/null for ls command
- Add missing subcommands
- Only complete the netns name once after subcmd

### DIFF
--- a/completions/ip
+++ b/completions/ip
@@ -327,8 +327,7 @@ _ip()
                     ;;
                 delete|exec|pids|set)
                     [[ $prev == $subcmd ]] && \
-                        COMPREPLY=( $(compgen -W "$( command \
-                            $1 netns list )" -- "$cur") )
+                        COMPREPLY=( $(compgen -W "$($1 netns list)" -- "$cur") )
                     ;;
                 *)
                     [[ $cword -eq $subcword ]] && \
--- a/completions/ip
+++ b/completions/ip
@@ -320,10 +320,10 @@ _ip()
 
         netns)
             case $subcmd in
-                list|add|identify|pids|monitor)
+                list|list-id|add|identify|monitor)
                     # TODO
                     ;;
-                add|delete|exec|pids)
+                delete|exec|pids|set)
                      _available_netns
                     ;;
                 *)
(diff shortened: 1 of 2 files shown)


completions/ip
```diff
@@ -318,6 +318,24 @@ _ip()
             esac
             ;;
 
+        netns)
+            case $subcmd in
+                add|identify|list|list-id|monitor)
```
scop: list, list-id, and monitor don't seem to be actual TODOs, because they take no args. So for clarity, I think they should be in a separate case.

completions/ip
```diff
@@ -318,6 +318,24 @@ _ip()
...
+                    # TODO
+                    ;;
+                delete|exec|pids|set)
+                    [[ $prev == $subcmd ]] && \
+                        COMPREPLY=( $( compgen -W '$( command \
```
scop: Cosmetic: no space after the paren in `$(` for consistency, ditto for the closing paren and similarly below.

completions/ip
```diff
@@ -318,6 +318,24 @@ _ip()
...
+                        ls /var/run/netns 2>/dev/null )' -- "$cur" ) )
```
scop: One indent level missing here.

completions/ip
```diff
@@ -318,6 +318,24 @@ _ip()
...
+                    ;;
+                *)
+                    [[ $cword -eq $subcword ]] && \
+                        COMPREPLY=( $( compgen -W 'help add delete exec
+                        identify list list-id monitor pids set' -- "$cur" ) )
```
scop: Spaces within inner parens, indent level

completions/ip
scop: Instead of listing /var/run/netns, it would seem better to me to use the output from `"$1" netns list`

completions/ip
felixonmars: list-id seems to take optional arguments, so it should remain in TODO. I'll split off the other two.

completions/ip
felixonmars: You are right, using `$1 netns list` is actually better since it doesn't return error when the directory is not present.

felixonmars:
Hopefully addresses all issues :)

scop:
Sure does, thanks! I took the liberty to remove a `command` "prefix" as we don't use it within this file to invoke the `ip` to be completed elsewhere either.

felixonmars:
Thanks :)

PR #396, opened 2020-03-11
Speed up Python completion

jakseb:
* Make completion of dotted names (e.g. `python -m foo.b`) faster.
* (minor one) Ignore errors in `walk_packages()` so that a package raising an exception does not prevent other modules from being listed.

### DIFF
--- a/helpers/python
+++ b/helpers/python
@@ -3,12 +3,20 @@
 import pkgutil
 import sys
 
-# walk_packages() is much slower than iter_modules(), use it only when
-# completing something with a dot in it.
-if len(sys.argv) > 1 and "." in sys.argv[1]:
-    walker = pkgutil.walk_packages
-else:
-    walker = pkgutil.iter_modules
 
-for mod in walker():
+def iter_submodules(pkgname):
+    __import__(pkgname)
+    pkg = sys.modules[pkgname]
+    return pkgutil.walk_packages(
+        pkg.__path__, pkgname + ".", onerror=lambda _: None
+    )
+
+
+if len(sys.argv) > 1 and "." in sys.argv[1]:
+    pkgname = sys.argv[1].rpartition(".")[0]
+    modules = iter_submodules(pkgname)
+else:
+    modules = pkgutil.iter_modules()
+
+for mod in modules:
     print (mod[1])  # noqa: E211
--- a/test/t/test_python.py
+++ b/test/t/test_python.py
@@ -33,3 +33,7 @@ class TestPython:
     @pytest.mark.complete("python -m sy", require_cmd=True)
     def test_8(self, completion):
         assert completion
+
+    @pytest.mark.complete("python -m json.", require_cmd=True)
+    def test_9(self, completion):
+        assert "json.tool" in completion
--- a/test/t/test_python3.py
+++ b/test/t/test_python3.py
@@ -33,3 +33,7 @@ class TestPython3:
     @pytest.mark.complete("python3 -m sy", require_cmd=True)
     def test_8(self, completion):
         assert completion
+
+    @pytest.mark.complete("python3 -m json.", require_cmd=True)
+    def test_9(self, completion):
+        assert "json.tool" in completion


jakseb:
Side note: `helpers/python` is skipped by pre-commit hooks (black, flake8). Is this intentional?

helpers/python
```diff
@@ -3,12 +3,20 @@
 import pkgutil
 import sys
 
-# walk_packages() is much slower than iter_modules(), use it only when
-# completing something with a dot in it.
+
+def iter_submodules(pkgname):
+    __import__(pkgname)
```
scop: Some concerns here.

First, the docs of `__import__` say
> Direct use of `__import__()` is also discouraged in favor of `importlib.import_module()`.
...and in pydoc
> ... meant for use by the Python interpreter and not for general use ...

Second, I'm not that happy with importing in the first place, due to possible side effects in doing so, also possible security issues arising from that.

WDYT?

jakseb: > First, the docs of `__import__` say
> 
> > Direct use of `__import__()` is also discouraged in favor of `importlib.import_module()`.

I didn't know what versions of Python `bash-completion` is supposed to support. I decided to stick to the lowest version able to run the current helper. For CPython it seems to be 2.5 and 3.0 (not tested, I only inspected source code of `pkgutil`: [v2.5a2][src-pkgutil-py2], [v3.0a1][src-pkgutil-py3]). On the other hand, `importlib` was introduced in Python 2.7 and 3.1 (docs: [Python 2][doc-importlib-py2], [Python 3][doc-importlib-py3]). So I chose to use `__import__()` instead. Moreover, this is what the implementation of `walk_packages()` uses.

[src-pkgutil-py2]: https://github.com/python/cpython/blob/v2.5a2/Lib/pkgutil.py#L70-L108
[src-pkgutil-py3]: https://github.com/python/cpython/blob/v3.0a1/Lib/pkgutil.py#L71-L150
[doc-importlib-py2]: https://docs.python.org/2/library/importlib.html
[doc-importlib-py3]: https://docs.python.org/3/library/importlib.html

I can switch to using `importlib` if compatibility with old Pythons is not an issue. Just let me know.

> Second, I'm not that happy with importing in the first place, due to possible side effects in doing so, also possible security issues arising from that.

The current script is already doing imports. Quoting [the docs of `walk_packages()`][doc-walk_packages]:

[doc-walk_packages]: https://docs.python.org/3/library/pkgutil.html#pkgutil.walk_packages

> Note that this function must import all packages (not all modules!) on the given path, in order to access the `__path__` attribute to find submodules.

I must admit, though, that my code may in some cases import something that the original version wouldn't. I will modify it to check whether the module I am going to import is a package, in order to match the behavior of the current version. (Although this is non-trivial).

scop: Hm.

What i'd ideally like to see is:
1) old python versions not dropped
2) no use of discouraged functions
3) no arbitrary code execution issues out of the box in expected default configs

For 1) and 2) I suppose using importlib if available and __import__ only if not could be the way to go here, unless it gets "too ugly".

For 3) I think it's not actually as bad as I initially thought it'd be, as I don't think we're actually importing anything from the current dir unless the user has explicitly wanted to do that through use of `PYTHONPATH`. So from that POV we're fine with importing stuff in the first place.

Wrt old python versions, I think our CI setup goes down to 2.6.6 on CentOS 6.

jakseb:
@scop I have pushed a new version (sorry that nearly a month later). It is rather awful. Could you take a look anyway and express your opinion? :wink: (Not meant to be merged yet. The code could get some minor refinements and the commit history should be squashed).

You were actually right when you were concerned about potential side effects of imports. If the helper blindly imports arbitrary modules, then for example trying to complete `pydoc antigravity.` results in launching a web browser. Another problematic module is `idlelib.idle` as well as some modules whose names end with `.__main__` (although this case is totally understandable and could be easily filtered out). For this reason, I decided to restrict imports to only packages.

In addition, I implemented filtering of module names in recursive walk in order to further speed up the traversal (in some cases).

Unfortunately, those two things caused the code to grow considerably in size and complexity.

There is one more issue I should mention. I did some tests, comparing the outputs of this new helper and of the old one (the one present in master). There are some items that the old version prints but my version does not. However, this is not a problem IMO. The unlisted modules aren't directly importable, so they are useless as arguments to `pydoc` or `python -m`. The helper is also used by `pylint` and `mypy` completions. The missing modules are (on my machine) `samba.subunit.run` and things like `pip._vendor.{something}`. I doubt anybody needs to run `pylint` or `mypy` on those.

(The script I used to compare helpers' outputs is available [here][regrpy-branch].)

[regrpy-branch]: https://github.com/jakseb/bash-completion/blob/pyhelper/regression/test/test_pyhelper.py "test/test_pyhelper.py on branch pyhelper/regression"

helpers/python
```diff
@@ -3,12 +3,135 @@
...
+    parentname = pkgname.rpartition(".")[0]
+    if parentname:
+        parent = import_package(parentname)
+    if find_spec is not None:
+        spec = find_spec(pkgname)
+        if spec is not None and spec.submodule_search_locations is not None:
+            return import_module(pkgname)
+    # find_spec() unavailable, fall back to searching iter_modules(...).
+    if parentname:
+        path = parent.__path__ or []
```
scop, line 52: `__path__` is documented as
> A list of strings specifying the search path within a package. This attribute is not set on modules.

Can we take advantage of that when deciding if something is (not) a package?

PR #398, opened 2020-03-17
New completion: apt-mark

JCGoran:
Mostly borrowed code from `apt-get` completion. I've also added some basic tests for edge case options.
Note that I've implemented the completion to display short options as well, which I could of course remove if they are deemed unnecessary.

### DIFF
--- a/completions/apt-mark
+++ b/completions/apt-mark
@@ -6,8 +6,8 @@ _apt_mark()
     _init_completion -s || return
 
     local special i
-    for (( i=0; i < ${#words[@]}-1; i++ )); do
-        if [[ ${words[i]} == @(hold|unhold|auto|manual|install|remove|purge|showhold|showauto|showmanual|showinstall|showremove|showpurge|minimize-manual) ]]; then
+    for (( i=1; i < ${#words[@]}-1; i++ )); do
+        if [[ ${words[i]} == @(auto|manual|minimize-manual|showauto|showmanual|hold|unhold|showhold|install|remove|purge|showinstall|showremove|showpurge) ]]; then
             special=${words[i]}
         fi
     done
@@ -18,8 +18,7 @@ _apt_mark()
                 return
                 ;;
             *)
-                COMPREPLY=( $(apt-cache --no-generate pkgnames "$cur" \
-                    2>/dev/null) )
+                COMPREPLY=( $(_xfunc apt-cache _apt_cache_packages) )
                 ;;
         esac
         return
@@ -29,7 +28,11 @@ _apt_mark()
         --help|--version|--option|-!(-*)[hvo])
             return
             ;;
-        --config-file|--file|-!(-*)[cf])
+        --config-file|-!(-*)c)
+            _filedir conf
+            return
+            ;;
+        --file|-!(-*)f)
             _filedir
             return
             ;;
@@ -38,12 +41,10 @@ _apt_mark()
     $split && return
 
     if [[ "$cur" == -* ]]; then
-        COMPREPLY=( $(compgen -W '-f --file= --help --version -c --config-file -o --option' -- "$cur") )
+        COMPREPLY=( $(compgen -W '--file= --help --version --config-file --option' -- "$cur") )
         [[ $COMPREPLY == *= ]] && compopt -o nospace
     else
-        COMPREPLY=( $(compgen -W 'hold unhold install remove purge
-            showhold showinstall showremove showpurge showauto showmanual
-            auto manual minimize-manual' -- "$cur") )
+        COMPREPLY=( $(compgen -W 'auto manual minimize-manual showauto showmanual hold unhold showhold install remove purge showinstall showremove showpurge' -- "$cur") )
     fi
 
 } &&
--- a/completions/apt-mark
+++ b/completions/apt-mark
@@ -7,7 +7,7 @@ _apt_mark()
 
     local special i
     for ((i = 1; i < ${#words[@]} - 1; i++)); do
-        if [[ ${words[i]} == @(auto|manual|minimize-manual|showauto|showmanual|hold|unhold|showhold|install|remove|purge|showinstall|showremove|showpurge) ]]; then
+        if [[ ${words[i]} == @(auto|manual|minimize-manual|showauto|showmanual|hold|unhold|showhold|install|remove|deinstall|purge|showinstall|showremove|showpurge) ]]; then
             special=${words[i]}
         fi
     done
--- a/completions/apt-mark
+++ b/completions/apt-mark
@@ -15,6 +15,14 @@ _apt_mark()
 
     if [[ -v special ]]; then
         case $special in
+            auto | manual | unhold)
+                local -A showcmds=([auto]=manual [manual]=auto [unhold]=hold)
+                local showcmd=${showcmds[$special]}
+                COMPREPLY=($(
+                    compgen -W "$("$1" show$showcmd 2>/dev/null)" -- "$cur"
+                ))
+                return
+                ;;
             minimize-manual)
                 return
                 ;;
--- a/completions/apt-mark
+++ b/completions/apt-mark
@@ -1,0 +1,52 @@
+# Debian apt-mark(8) completion                             -*- shell-script -*-
+
+_apt_mark()
+{
+    local cur prev words cword split
+    _init_completion -s || return
+
+    local special i
+    for (( i=0; i < ${#words[@]}-1; i++ )); do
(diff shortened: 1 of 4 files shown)


scop:
I'd rather see all `apt-*` completions being shipped with apt itself, as the `apt` completion is already there and IMO in the big picture it's better to have completions shipped along with the commands. I'm not "ruling out" inclusion of this in bash-completion in the meantime, as there is no active effort in getting them moved, and adding bits like this don't make it much harder when it actually starts to take place.

Will have a look at the actual PR content later unless someone else beats me to it.

JCGoran:
Understood, though I don't see the completions being shipped with apt anytime soon, see [this bug on Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=799545), which was reported 5 years ago, and last activity was almost 3 years ago. Additionally, from what I can tell, `apt` itself ships only `apt` completion, while `bash-completion` offers the rest (`apt-get`, `apt-cache`, `apt-build`).

scop:
Yeah, I've been involved with that bug. But I think the lack of response there should not be counted as a rejection. Someone just needs to do some work and then we'll see for real.

Anyway, addition of apt-mark here isn't a biggie, will have a quick look now.

test/t/test_apt_mark.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+
+
+@pytest.mark.bashcomp(cmd="apt-mark")
+class TestAptMark:
+    @pytest.mark.complete("apt-mark ")
+    def test_1(self, completion):
+        assert all(x in completion for x in "auto manual remove showinstall showremove \
+        hold minimize-manual showauto showmanual unhold install \
```
scop: Not sure if it's just the patch viewer, or if we have a bit of an indentation issue here.

Anyway no need to use backslashes for line continuation, just use multiple string constants, they'll auto concatenate like
```python
... for x in (
    "auto manual remove showinstall showremove "
    "hold minimize-manual showauto showmanual unhold install "
    # [...]
)

File: test/t/test_apt_mark.py [hunk 2]
```diff
@@ -0,0 +1,27 @@
...
+        purge showhold showpurge".split())
+
+    @pytest.mark.complete("apt-mark minimize-manual ")
+    def test_2(self, completion):
+        assert completion == []
```
scop: `assert not completion`

test/t/test_apt_mark.py
```diff
@@ -0,0 +1,27 @@
...
+    def test_3(self, completion):
+        assert completion == "--file=bash-completion-test-subject.deb"
+
+    @pytest.mark.complete("apt-mark --config-file ", cwd="dpkg")
+    def test_4(self, completion):
+        assert completion == "bash-completion-test-subject.deb"
+
+    @pytest.mark.complete("apt-mark --option ")
+    def test_5(self, completion):
+        assert completion == []
```
scop: `assert not completion`

completions/apt-mark
```diff
@@ -0,0 +1,52 @@
...
+        esac
+        return
+    fi
+
+    case $prev in
+        --help|--version|--option|-!(-*)[hvo])
+            return
+            ;;
+        --config-file|--file|-!(-*)[cf])
+            _filedir
```
scop: Do you thing we could add some file extensions to narrow down the completions here? `*.conf` would probably work well enough for configs (I see we don't do that for apt-get either for reasons unknown to me though). Do the stats files accepted by `--file` follow a naming pattern we could use?

completions/apt-mark
```diff
@@ -0,0 +1,52 @@
...
+            return
+            ;;
+    esac
+
+    $split && return
+
+    if [[ "$cur" == -* ]]; then
+        COMPREPLY=( $(compgen -W '-f --file= --help --version -c --config-file -o --option' -- "$cur") )
```
scop: Include only the long options here, search for "available options" in CONTRIBUTING.md

completions/apt-mark
```diff
@@ -0,0 +1,52 @@
+# Debian apt-mark(8) completion                             -*- shell-script -*-
+
+_apt_mark()
+{
+    local cur prev words cword split
+    _init_completion -s || return
+
+    local special i
+    for (( i=0; i < ${#words[@]}-1; i++ )); do
+        if [[ ${words[i]} == @(hold|unhold|auto|manual|install|remove|purge|showhold|showauto|showmanual|showinstall|showremove|showpurge|minimize-manual) ]]; then
```
scop: This list would be better sorted correspondingly with the source for easier mainenance/crosscheck. The only source for the full list I'm aware of is the manpage, and at least in the version I have (1.6.12 on Ubuntu 18.04) they're not in this order.

My version doesn't have `minimize-manual`, can you confirm that its inclusion is intentional?

completions/apt-mark
```diff
@@ -0,0 +1,52 @@
...
+            ;;
+    esac
+
+    $split && return
+
+    if [[ "$cur" == -* ]]; then
+        COMPREPLY=( $(compgen -W '-f --file= --help --version -c --config-file -o --option' -- "$cur") )
+        [[ $COMPREPLY == *= ]] && compopt -o nospace
+    else
+        COMPREPLY=( $(compgen -W 'hold unhold install remove purge
```
scop: Options list order and minimize-manual note from above applies here too.

completions/apt-mark
```diff
@@ -0,0 +1,52 @@
...
+
+_apt_mark()
+{
+    local cur prev words cword split
+    _init_completion -s || return
+
+    local special i
+    for (( i=0; i < ${#words[@]}-1; i++ )); do
+        if [[ ${words[i]} == @(hold|unhold|auto|manual|install|remove|purge|showhold|showauto|showmanual|showinstall|showremove|showpurge|minimize-manual) ]]; then
+            special=${words[i]}
```
scop: Should we `break` after this?

completions/apt-mark
```diff
@@ -0,0 +1,52 @@
...
+    done
+
+    if [[ -n $special ]]; then
+        case $special in
+            minimize-manual)
+                return
+                ;;
+            *)
+                COMPREPLY=( $(apt-cache --no-generate pkgnames "$cur" \
+                    2>/dev/null) )
```
scop: Should reuse apt-cache's `_apt_cache_packages`, see examples in d90fbe3bfd7e53c365897fa584edc9d2d0a2e9bf which I just pushed.

completions/apt-mark
```diff
@@ -0,0 +1,52 @@
+# Debian apt-mark(8) completion                             -*- shell-script -*-
+
+_apt_mark()
+{
+    local cur prev words cword split
+    _init_completion -s || return
+
+    local special i
+    for (( i=0; i < ${#words[@]}-1; i++ )); do
```
scop: Could start from `i=1` (I'll fix this in a bunch of other completions we have)

completions/apt-mark
JCGoran: I've added `*.conf` in the latest version for `--config-file`, but for `--file=` I'm not 100% sure if there are any extensions, since by default it seems to read the file `/var/lib/apt/extended_states` which doesn't have one.

completions/apt-mark
JCGoran: Actually, since using more than one special option results in an error, here's a potential solution using an array, effectively disabling the completion if the user inputs more than one option:
```diff
diff --git a/completions/apt-mark b/completions/apt-mark
index a88aeacd..213ab6a4 100644
--- a/completions/apt-mark
+++ b/completions/apt-mark
@@ -5,15 +5,19 @@ _apt_mark()
     local cur prev words cword split
     _init_completion -s || return
 
-    local special i
+    local i
+    local special=()
     for (( i=1; i < ${#words[@]}-1; i++ )); do
         if [[ ${words[i]} == @(auto|manual|minimize-manual|showauto|showmanual|hold|unhold|showhold|install|remove|purge|showinstall|showremove|showpurge) ]]; then
-            special=${words[i]}
+            special+=(${words[i]})
         fi
     done
+    if [[ ${#special[@]} -gt 1 ]]; then
+        return
+    fi
 
-    if [[ -n $special ]]; then
-        case $special in
+    if [[ -n ${special[0]} ]]; then
+        case ${special[0]} in
             minimize-manual)
                 return
                 ;;
```

scop: > Actually, since using more than one special option results in an error,

True, but:
```
# apt-mark unhold hold
E: Unable to locate package hold
E: No packages found
```
So the two specials are not actually treated as specials, the latter is treated as a package name. It just happens I don't have a package named "hold" installed, that's why the error. I don't think we should prevent cases like that from happening, so I opted for the simple `break`.

scop:
Thanks, and apologies for lettings this stall for so long.

PR #400, opened 2020-03-30
Add new ssh -Q completions in OpenSSH 8.2p1

Jakuje:
Mostly based on the upstream code:

https://github.com/openssh/openssh-portable/blob/master/ssh.c#L738

One place for improvement would be accepting case insensitive aliases, but I was not sure how to do that easily on fast review of the manual pages.

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -4,10 +4,8 @@ _ssh_queries()
 {
     COMPREPLY+=( $(compgen -W \
         "cipher cipher-auth help mac kex key key-cert key-plain key-sig
-         protocol-version compression sig" \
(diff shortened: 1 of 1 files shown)


Jakuje:
Thanks for the pointer. It did a trick to work good enough.

I separated the case insensitive options to separate `compgen` (at least for clarity that there is something else), while left the case sensitive ones in the original. Hope it is fine. If not, I can merge them as in the end, functionality should be the same as everything is converted to lowecase.

scop:
I tweaked the change a bit -- left the grouping in, but made it so that we invoke `compgen` only once. Thanks!

PR #401, opened 2020-03-30
Support xz compressed kernel modules.

hramrach:


hramrach:
Yes, you need to enable support in more places than just bash-completion.
I am running kernel with xz compressed modules right now so I suppose they do work.

PR #403, opened 2020-04-18
openssl: update -starttls completions

jakseb:


jakseb:
It is worth noting that for sufficiently new versions of `openssl` the completions could be easily [extracted from program's output][scraping-commit]. I don't know if that idea is good (e.g. [backward compatibility][fallback-commit] concerns) and worth introducing, so this PR contains only an uncontroversial change of extending the list of predefined completions.

[scraping-commit]: https://github.com/jakseb/bash-completion/commit/9dd2d8b5ecf78a905d600c66f55a6c92fc386f10
[fallback-commit]: https://github.com/jakseb/bash-completion/commit/82e2bc21009b5e5928dbae3e7a2002eab9c4acde

scop:
Thanks! I do think that scraping would be preferable to hardcoding where sanely doable, but such a change should come with an addition to our test suite.

PR #404, opened 2020-04-18
Change dead Gmane link to lists.gnu.org

jakseb:


scop:
Thanks, the PR number was a very nice fit, BTW ;)

PR #409, opened 2020-05-04
fix for "bash_completion is a directory" error message

beantaxi:
I was getting `~/.bash_completion is a directory` error messages, which I did not like seeing on every ssh login. So I added a `-f` to make sure it's a file, in addition to the other checks.

Perhaps this isn't quite correct; among other things, installing bash_completion and then having a directory called ~/.bash_completion might be a problem. But I thought I'd submit my fix anyway.

Thanks!


scop:
Yeah, guess there's no harm in doing that check. Thanks!

beantaxi:
Happy to help! Cheers

PR #422, opened 2020-06-08
Allow all file types for tshark -r

pdnm:
The current autocomplete only accepts pcap/pcapng(.gz), but Tshark allows many other [file types](https://www.wireshark.org/docs/wsug_html_chunked/ChIOOpenSection.html#ChIOInputFormatsSection).

I think instead of adding all those types to the pattern, we should just list all files for these reasons:

- Tshark doesn't care about the file extension
- The user typically types a prefix before hitting Tab
- Capture files are typically stored in a directory without many other files

### DIFF
--- a/completions/tshark
+++ b/completions/tshark
@@ -58,11 +58,8 @@ _tshark()
             [[ ${COMPREPLY-} == *: ]] && compopt -o nospace
             return
             ;;
-        -*r)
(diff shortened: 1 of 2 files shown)


scop:
The three bulleted reasons listed are not usually something for which we enable completing all files, but in this case the sheer number of different file types is. Applied a slightly modified version. Thanks!

PR #428, opened 2020-07-07
support Android AAR libraries as 'unzip' file type

eighthave:
Android AAR is a ZIP container with required files:
https://developer.android.com/studio/projects/android-library.html#aar-contents


scop:
Thanks!

PR #433, opened 2020-07-31
Allow completion of farbfeld images in feh

mlugg:
Add common file extensions for farbfeld images (*.ff / *.ff.gz / *.ff.bz2) to feh completions


scop:
Thanks!

Issue #434, opened 2020-08-06
Missing ps completion


relrelb:
`ps` has no completion at all.
If there is no special reason for such a basic command to be missing, I would like to contribute it.

PR #435, opened 2020-08-06
Add ps completion

relrelb:
For #434.


scop:
Awesome, thanks; applied with some changes.

Issue #436, opened 2020-08-08
Color-Bug if ip is an alias


johannes-schmatz:
If ip is an alias to 'ip -c' (colored output), then the completion for ip addr show dev outputs some worse escape sequences.
The completed device names are not correct.

FIX:
put in command for absolute path to command, do not use alias.
```
$ PAGER=cat git diff
diff --git a/bash_completion b/bash_completion
index 1a7f5634..16cd47d8 100644
--- a/bash_completion
+++ b/bash_completion
@@ -1006,11 +1006,11 @@ _available_interfaces()

    COMPREPLY=($({
        if [[ ${1:-} == -w ]]; then
-            iwconfig
+            command iwconfig
        elif [[ ${1:-} == -a ]]; then
-            ifconfig || ip link show up
+            command ifconfig || command ip link show up
        else
-            ifconfig -a || ip link show
+            command ifconfig -a || command ip link show
        fi
    } 2>/dev/null | awk \
        '/^[^ \t]/ { if ($1 ~ /^[0-9]+:/) { print $2 } else { print $1 } }'))
```

scop:
A generic problem with bypassing aliases is that doing so will break setups that have the commands aliased for a real reason which makes them actually work (BTW the proposed fix touches not only ip, but iwconfig and ifconfig as well without a mention). If we accept this, by the same reasoning we should do this for every single command we invoke, which is not pretty. We do it for a select few currently only.

Offhand, I suppose the best thing would be if `ip` had an auto-color mode or something that would not try to colorize if stdout is not a pty.

johannes-schmatz:
ip has a no-color mode with `-c=never`.
use `ip -c=never` to don't get colored output.

scop:
Can't use it, it's a too new addition. For example on Ubuntu 18 (and we support much older distros too):
```
$ ip -V
ip utility, iproute2-ss180813
$ ip -c=never a
Option "-c=never" is unknown, try "ip -help".
```

scop:
I guess we could look into ugliness like `ip -c=never foo || ip foo` but we need to address at least in 5 places in bash_completion, and completions/ip. Would be better if the tools that colorize would just automatically DTRT. But meh, could you submit a real PR for those mentioned places along with some test cases that set the `ip -c` alias to verify that the fixes work and don't break older versions?

johannes-schmatz:
pull request is out. done.

PR #439, opened 2020-08-10
fix issue #436 Color-Bug if ip is an alias

johannes-schmatz:
change all
```
ip link show
```
to something like this
```
(ip -c=never link show || ip link show)
```

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -917,7 +917,7 @@ _mac_addresses()
     # - ip link: link/ether
     COMPREPLY+=($(
         {
-            LC_ALL=C ifconfig -a || (ip -c=never link show || ip link show)
+            LC_ALL=C ifconfig -a || ip -c=never link show || ip link show
         } 2>/dev/null | command sed -ne \
             "s/.*[[:space:]]HWaddr[[:space:]]\{1,\}\($re\)[[:space:]].*/\1/p" -ne \
             "s/.*[[:space:]]HWaddr[[:space:]]\{1,\}\($re\)[[:space:]]*$/\1/p" -ne \
@@ -927,7 +927,7 @@ _mac_addresses()
 
     # ARP cache
     COMPREPLY+=($({
-        arp -an || (ip -c=never neigh show || ip neigh show)
+        arp -an || ip -c=never neigh show || ip neigh show
     } 2>/dev/null | command sed -ne \
         "s/.*[[:space:]]\($re\)[[:space:]].*/\1/p" -ne \
         "s/.*[[:space:]]\($re\)[[:space:]]*$/\1/p"))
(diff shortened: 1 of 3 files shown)


bash_completion
```diff
@@ -917,7 +917,7 @@ _mac_addresses()
     # - ip link: link/ether
     COMPREPLY+=($(
         {
-            LC_ALL=C ifconfig -a || ip link show
+            LC_ALL=C ifconfig -a || (ip -c=never link show || ip link show)
```
scop, line 920: Please remove the parenthesis from everywhere, they're not needed, and actually start an unnecessary subshell.

scop:
completions/ip still needs to be addressed...

johannes-schmatz:
AFAIK completions/ip does call _available_interfaces, so there's no ip invocation

scop:
It calls at least `$1 help`, `$1 r`, and `$1 netns` -- `$1` there is the `ip` command being invoked.

johannes-schmatz:
Thank you! I didn't find them by searching 'ip'

completions/ip
```diff
@@ -48,7 +48,7 @@ _ip()
                 COMPREPLY=($(compgen -W "$c" -- "$cur"))
                 ;;
             *)
-                COMPREPLY=($(compgen -W "help $($1 help 2>&1 | command sed -e \
+                COMPREPLY=($(compgen -W "help $($1 -c=never help 2>&1 || $1 help 2>&1 | command sed -e \
```
scop, line 51: This is not correct; both `$1` calls need to be wrapped in `{ ... }` like `{ $1 -c=never help || $1 help; } 2>&1 || command ...`

completions/ip
```diff
@@ -191,11 +191,11 @@ _ip()
                     ;;
                 a | add | d | del | change | append | r | replace | monitor)
                     if [[ $prev == via ]]; then
-                        COMPREPLY=($(compgen -W "$($1 r | command sed -ne \
+                        COMPREPLY=($(compgen -W "$($1 -c=never r || $1 r | command sed -ne \
```
scop, line 194: Same as above.

completions/ip
```diff
@@ -191,11 +191,11 @@ _ip()
...
                             's/.*via \([0-9.]*\).*/\1/p')" -- "$cur"))
                     elif [[ $prev == "$subcmd" ]]; then
                         COMPREPLY=($(compgen -W "table default \
-                            $($1 r | cut -d ' ' -f 1)" -- "$cur"))
+                            $($1 -c=never r || $1 r | cut -d ' ' -f 1)" -- "$cur"))
```
scop, line 198: Same as above.

completions/ip
```diff
@@ -349,7 +349,7 @@ _ip()
                     ;;
                 delete | exec | pids | set)
                     [[ $prev == "$subcmd" ]] &&
-                        COMPREPLY=($(compgen -W "$($1 netns list)" -- "$cur"))
+                        COMPREPLY=($(compgen -W "$($1 -c=never netns list || $1 netns list)" -- "$cur"))
```
scop, line 352: The first one needs a `2>&1` here.

scop (changes requested):

scop:
Applied with some output redirection fixes to completions/ip.

PR #451, opened 2020-10-08
xrandr: improve completion

uno20001:
* complete both arguments of --setprovider{offloadsink,outputsource}
 * complete both arguments of --{add,del}mode
 * complete second and (partially) third argument of --setmonitor
 * complete single argument of --delmonitor, --dpi, --filter
 * only complete per-output options when an output is specified

---

I'm not sure how tests should be modified. I'd appreciate some help.


completions/xrandr
```diff
@@ -1,62 +1,165 @@
 # bash completion for xrandr                               -*- shell-script -*-
 
+_xrandr_outputs()
+{
+    "$1" 2>/dev/null | awk '/connected/ {print $1}'
+}
+
+_xrandr_monitors()
+{
+    "$1" --listmonitors 2>/dev/null | sed -ne 's/.* [0-9]*: [+\*]*\([^ ]*\).*/\1/p'
```
scop, line 10: `command sed` instead of plain `sed` everywhere, see examples in existing completions.

scop: Also, `test/runLint` might catch some non-portable sed etc uses, be sure to run it in case you didn't yet.

completions/xrandr
```diff
@@ -1,62 +1,165 @@
...
+        --output | --addmode | --delmode)
+            local outputs=$(_xrandr_outputs "$1")
+            COMPREPLY=($(compgen -W "$outputs" -- "$cur"))
             return
             ;;
-        --output | --left-of | --right-of | --above | --below | --same-as)
-            local outputs=$("$1" | awk '/connected/ {print $1}')
+        --dpi)
+            local outputs=$(_xrandr_outputs "$1")
             COMPREPLY=($(compgen -W "$outputs" -- "$cur"))
```
scop, line 58: Hmm, this doesn't look quite right. Completing `--dpi` with outputs?

completions/xrandr
```diff
@@ -1,62 +1,165 @@
...
-                    -e "s/\([^[:space:]]\)[[:space:]].*/\1/")
+                local modes=$(_xrandr_modes "$1" "$output")
                 COMPREPLY=($(compgen -W "$modes" -- "$cur"))
             fi
             return
             ;;
         -o | --orientation)
-            COMPREPLY=($(compgen -W 'normal inverted left right 0 1 2 3' -- \
-                "$cur"))
+            COMPREPLY=($(compgen -W 'normal inverted left right 0 1 2 3' -- "$cur"))
```
scop, line 76: This seems like an unrelated whitespace change, in the future please keep them as separate commits.

scop: In general it's a good thing to do one thing in one commit, here everything's lumped in one. Just something to remember for future changes.

completions/xrandr
```diff
@@ -1,62 +1,165 @@
...
+            COMPREPLY=($(compgen -W "$monitors" -- "$cur"))
             return
             ;;
     esac
 
-    COMPREPLY=($(compgen -W '$("$1" -help 2>&1 |
-        command sed -e "s/ or / /g" -e "s/<[^>]*>]//g" | _parse_help -)' \
-        -- "$cur"))
+    # second arguments
+    if [[ $cword -ge 2 ]]; then
```
scop, line 115: `(( arithmetic evaluation ))`

completions/xrandr
```diff
@@ -1,62 +1,165 @@
...
+           --setprovideroffloadsink)
+               local providers=$(_xrandr_providers "$1" "Sink Offload")
+               COMPREPLY=($(compgen -W "0x0 $providers" -- "$cur"))
+               return
+               ;;
+       esac
+    fi
+
+    # third arguments
+    if [[ $cword -ge 3 ]]; then
```
scop, line 143: `(( arithmetic evaluation ))`

completions/xrandr
```diff
@@ -1,62 +1,165 @@
...
+                COMPREPLY=($(compgen -W "none $outputs" -- "$cur"))
+                # TODO: the third argument is actually a comma-separated list
+                return
+                ;;
+        esac
+    fi
+
+    local options
+    if [[ -v output ]]; then
+        options=`_xrandr_help "$1" | sed -e 's/ or /\n  /g' -e 's/<[^>]*>]//g' | _parse_help -`
```
scop, line 156: `$(...)` instead of backticks here and below

scop (changes requested):
Thanks! Wrt additional tests, I think we should add a mock executable/output. See e.g. `test/t/test_dnssec_keygen.py` and `test/fixtures/dnssec-keygen/dnssec-keygen` for an example, and there's some similar for at least `arp` and `ifconfig`.

completions/xrandr
uno20001: Sorry, I don't know why I removed that newline. I can put it back if necessary.

completions/xrandr
uno20001: Yes, this is strange, but both the man page, the help text, and the [source code](https://gitlab.freedesktop.org/xorg/app/xrandr/-/blob/master/xrandr.c#L3125) of `xrandr` indicate that the name of an output is an acceptable argument.

completions/xrandr
uno20001: > `command sed` instead of plain `sed` everywhere, see examples in existing completions.

Should `awk` be prefixed by `command` as well?

completions/xrandr
uno20001: Done.

completions/xrandr
uno20001: Done.

completions/xrandr
uno20001: Done.

completions/xrandr
scop: No need, as said, just for the future.

completions/xrandr
scop: Ok :P In that case, I suggest combining it with the identical `--output | --addmode | --delmode` switch case or adding a comment why it's a separate one.

completions/xrandr
scop: No, we currently do it generally for grep, ls, and sed only. (Ref test/runLint)

scop:
BTW it's easier for me to review changes if they're separate, new commits -- if you force-push everything I practically need to re-review everything too. I can tidy up the commit/history at merge time, no need to worry about that too much.

completions/xrandr
uno20001: Done.

completions/xrandr
uno20001: Done.

uno20001:
I created some tests, but I think something's still incorrect since it seems that on some of the targets it runs the installed `xrandr` not the mock script I created. Could you please take a look? I'm also not sure when `env`, `require_cmd` arguments are needed, or what `pre_cmds` is.

scop:
Will try to find time to look in a few days. `require_cmd=True` causes tests to be skipped if the command they test is not available. The idea of `pre_cmds`  here (and e.g. in dnssec-keygen) is that we should prepend our mock binary to the path and have it there for all tests in the file. However for some reason just like here, the dnssec-keygen tests additionally set the path using an `env=...` for some cases. That doesn't look right, either something's not working as intended or there's something redundant.

scop:
> However for some reason just like here, the dnssec-keygen tests additionally set the path using an `env=...` for some cases. That doesn't look right, either something's not working as intended or there's something redundant.

dnssec-keygen's tests use pre_cmds to _append_ the mocked binary dir to `$PATH` in order to be sure to have some version to test against (preferring the system one). Per-test `env=` settings on the other hand _prepend_ it, that's used in cases where we're using a hardcoded completion list against which we do an exact match test.

scop (approved):

scop (approved):

uno20001:
Thanks. Is there anything I need to do?

scop:
Nope, I was just waiting for the CI to finish running. Thanks!

PR #458, opened 2020-11-06
kaffeeine, dragon, totem: Add opus file extension

laodzu:
I noticed this ommission while using totem but expect it to work also for the other commands sharing the completion.


scop:
Thanks!

laodzu:
Cool, thanks for the quick merge!

PR #460, opened 2020-11-10
Make lua completion work for different versions

JakobDev:
If a system has different versions of lua installed, the completion only works for the normal lua command, not for a specific version. This PR fixes that.


scop (changes requested):
Thanks! In addition to the suggestions, please add symlink creation and cleanups to completions/Makefile.am (symlinks, CLEANFILES), as well as completions/.gitignore

JakobDev:
@scop Should be done now

scop:
Thanks!

PR #465, opened 2020-11-11
mutt: do not lose the + in mutt -f +foo<tab>

### LINKED ISSUE #464: mutt -f +file<tab> loses the + upon expansion
With #463 applied I can expand mail folders with `mutt -f =nam<tab>`, but `mutt -f +nam<tab>` eats the `+` upon expansion, producing `mutt -f name`, which doesn't work.

I fear that preparing a PR for this might be slightly beyond my capabilities.

mgedmin:
Closes #464.

### DIFF
--- a/completions/mutt
+++ b/completions/mutt
@@ -114,6 +114,7 @@ _muttfiledir()
         compopt -o filenames
         COMPREPLY=($(compgen -f -- "$folder/${cur:1}"))
         COMPREPLY=(${COMPREPLY[@]#$folder/})
+        [[ $cur == [+]* ]] && COMPREPLY=(${COMPREPLY[@]/#/+})
         return
     elif [[ $cur == !* ]]; then
         spoolfile="$($muttcmd -F "$muttrc" -Q spoolfile 2>/dev/null |


mgedmin:
I am baffled: I have a machine with Ubuntu 18.04 LTS (bash 4.4.20(1)-release, bash-completion 1:2.8-1ubuntu1) where `mutt -f =<tab>` eats the `=` and where a similar fix would be needed.  On my other machine with Ubuntu 20.10 (bash 5.0.17(1)-release, bash-completion 1:2.11-2ubuntu1) `mutt -f =<tab>` works fine and only `mutt -f +<tab>` needs to restore the `+`.

mgedmin:
I've found a solution that works on both Ubuntu 18.04 and 20.10, but I don't entirely understand the situation: https://github.com/mgedmin/dotfiles/commit/4758e78e2ea8e57636873d79e796691eff237ed2

mgedmin:
Welp this doesn't actually work on Ubuntu 20.10.  I thought I'd tested it!  Back to the drawing board...

mgedmin:
> Welp this doesn't actually work on Ubuntu 20.10. I thought I'd tested it!

I did test it, and `mutt -f =fold<tab>` is correctly expanded to `mutt -f =folder`.

What breaks is `mutt -f =<tab>` instead of showing all possible mailboxes it expands to `mutt -f =\=`.

Fixed now.

mgedmin:
I now have tests for this in #467.  Without this fix they fail as follows:

```
____________________________________ TestMutt.test_plus_is_not_eaten ____________________________________

self = <test_mutt.TestMutt object at 0x7fa06943ff10>, completion = <CompletionResult ['\x08\x08foo.mbx']>

    @pytest.mark.complete("mutt -F muttrc -f +f", require_cmd=True, cwd="mutt")
    def test_plus_is_not_eaten(self, completion):
>       assert completion == "oo.mbx".split()
E       AssertionError: assert <CompletionRe...\x08foo.mbx']> == ['oo.mbx']
E         Use -v to get the full diff

/home/mg/src/bash-completion/test/t/test_mutt.py:30: AssertionError
```

PR #488, opened 2021-01-04
7z, tar, unace, unrar, unzip, zipinfo: recognize comic book archives

fbriere:
Comic book archives¹ are simply archive files containing images.  While
they are typically treated as e-books and fed to programs that can
handle those, they are also plain archive files and can therefore be
manipulated by the corresponding program.

 ¹ https://en.wikipedia.org/wiki/Comic_book_archive

The following file extension mappings are added:

    .cb7 → 7z
    .cba → unace
    .cbr → unrar
    .cbt → tar
    .cbz → unzip, zipinfo

(Note that .cbt files, unlike .tar files, are never compressed.)


scop:
Thanks!

PR #496, opened 2021-02-11
screen: Match all files in directories under /dev/serial

tms13:
The original rule didn't complete from `/dev/serial/by-id/`.

It also makes sense to complete from any other directory under `/dev/serial`, as that's a logical place for custom udev rules to add symlinks for meaningful device names.

### DIFF
--- a/completions/screen
+++ b/completions/screen
@@ -30,7 +30,7 @@ _screen_sessions()
                 COMPREPLY=($(compgen -W "$(
                     shopt -s nullglob
                     printf '%s\n' \
-                        /dev/serial/by-id/* /dev/ttyUSB* /dev/ttyACM* 2>/dev/null
+                        /dev/serial/*/* /dev/ttyUSB* /dev/ttyACM* 2>/dev/null
                 )" \
                     -- "$cur"))
                 return


scop:
> The original rule didn't complete from `/dev/serial/by-id/`.

Hm, it did, and only that subdir -- do you mean `by-path`? Waiting for your clarification on that, other than that, LGTM. (No need to push changes just for that, I can change them and the commit message on the fly.)

tms13:
>> The original rule didn't complete from `/dev/serial/by-id/`.

0> In article <scop/bash-completion/pull/496/c779626094@github.com>,
0> Ville Skyttä <URL:mailto:notifications@github.com> ("Ville") wrote:

Ville> Hm, it did, and only that subdir -- do you mean `by-path`?  Waiting
Ville> for your clarification on that, other than that, LGTM. (No need to
Ville> push changes just for that, I can change them and the commit message
Ville> on the fly.)

Oops, yes I wrote the wrong thing in the commit.

I'm working on a bunch of embedded boards and have udev rules to put
their consoles in /dev/serial/board/<name>, and it was a little
frustrating being unable to complete on those.  That was my main
motivation, but I thought people might also value /d/s/by-path, too.

Thanks for reviewing - I was concerned because the auto-tests seem to be
failing; worried it might "taint" this change (my first ever Github PR).

scop:
Cool, thanks for the confirmation. Nevermind the test suite, it's broken due to unrelated reasons at the moment. Welcome!

Issue #497, opened 2021-02-13
Completion for ssh not finding mdns hosts with avahi


sirnewton01:
When I run "ssh <TAB>" I get a list of hosts and some special ipv6 addresses, but only one of the machines on my local network, but not the one that actually has ssh running on it. I've enabled COMP_KNOWN_HOSTS_WITH_AVAHI=true to get the avahi behaviour.

It turns out that the completion for avahi in this case is using _workstation._tcp as the service type, which doesn't appear to be used very often across Linux and MacOS at least. Perhaps _ssh._tcp would be a good addition here since there's already alot of auto-detection of ssh hostnames using the .known_hosts as an example.

PR #499, opened 2021-02-14
_known_hosts_real: look up _ssh._tcp from avahi, too

scop:
As a side effect, we no longer support avahi-browse versions without -k.

Closes https://github.com/scop/bash-completion/issues/497


sirnewton01:
Thanks so much for this. I'll update my machines with the patch when it becomes available.

PR #503, opened 2021-02-25
Make configure fall back to bashdefault and default

villevoutilainen:
This is necessary for some configure scripts where the --help
of the configure doesn't describe option arguments accurately.
Qt, for instance, has a configure script where -android-sdk is followed
by a path. The path is not completed because this script doesn't know
how to do that, and this fallback makes it work. It seems reasonable
in general to fall back on default completions if option completions
didn't work.

### DIFF
--- a/completions/configure
+++ b/completions/configure
@@ -37,6 +37,6 @@ _configure()
         [[ ${COMPREPLY-} == *= ]] && compopt -o nospace
     fi
 } &&
-    complete -F _configure configure
+    complete -o bashdefault -o default -F _configure configure
 
 # ex: filetype=sh


villevoutilainen:
Those check failures seem completely borked, is there a way to perhaps re-run them once the test environment is not failing catastrophically?

completions/configure
```diff
@@ -37,6 +37,6 @@ _configure()
         [[ ${COMPREPLY-} == *= ]] && compopt -o nospace
     fi
 } &&
-    complete -F _configure configure
+    complete -o bashdefault -o default -F _configure configure
```
scop, line 40: While I agree with:
> It seems reasonable in general to fall back on default completions if option completions didn't work.

...for this particular completion, the problem with the above implementation is that we don't have a way to say what worked or not. For example, we explicitly want to prevent all completions following --help, --version, --program-suffix and friends, but if we slap the defaults in the mix, our prevention will no longer work. I wish there was a way to signal success/not someway to prevent the defaults from kicking in (e.g. via some return value), but I'm not aware of a way.

scop (changes requested):

scop:
> Those check failures seem completely borked

Yeah, I've been kind of hoping for them to fix themselves, because I believe they also appeared out of the blue without us doing anything to provoke them. Help welcome if there are ideas though. (It's not limited to Alpine, nor caused by an upgrade of pytest to a fresh version.)

completions/configure
villevoutilainen: I would rather have possible superfluous completions after --help than have the completion of a directory not actually complete after an option that the completion doesn't recognize, but wants a directory anyway. The former is just a nuisance, the latter leads to copypaste typos that instead of failing the configure itself will fail the subsequent build.

scop: Mileages vary. Anyway I think it's possible to accomplish the original intent without sacrificing non-completions by not touching `complete` options at all, but doing this:
```diff
--- a/completions/configure
+++ b/completions/configure
@@ -25 +25,4 @@ _configure()
-    [[ $cur != -* ]] && return
+    if [[ $cur != -* ]]; then
+        _filedir
+        return
+    fi
```

villevoutilainen: I verified that that works; plz patch that in. :)

scop:
Done, slightly revised.

PR #504, opened 2021-02-26
Fix install of declare completion

algorythmic:
After https://github.com/scop/bash-completion/commit/3671cee9e0f9981b3452f1d7b459a6f3b0b2f6a2 the `declare` completion and its links were not getting installed.


scop:
Thanks!

PR #514, opened 2021-03-30
Update issue templates

scop:


akinomyoga:
It seems the new files `bug_report.md` and `feature_request.md` were already pushed to the master branch, so this PR appears to be ready to be closed. Is there remaining issues that needs to be solved or discussed in this PR?

scop:
I think this is a result of my fooling around with/testing the web UI editor, probably didn't realize I had submitted a PR or just forgot about it. But right, nothing to see here, thanks!

PR #523, opened 2021-05-13
cvs: added support for 'cvs status' completions

kcaran:
I noticed 'cvs status' commands were not completing correctly. This change adds the 'status' command to the 'diff' and 'log' command cases. It is very similar to this pull:

https://github.com/scop/bash-completion/pull/194

I also added a few test cases. Thank you!


scop:
Thanks!

PR #525, opened 2021-05-21
conftest: set HISTFILE=/dev/null in order not to affect user's history

### LINKED ISSUE #321: Tests clutter user's bash history
After running `make check`, some of the test commands end up in my `~/.bash_history` which clutters history completion. I think this could be solved via setting `HISTFILE=/dev/null` early on.

akinomyoga:
After running `make check` or `cd test/t; pytest-3 ...`, my command histories in `~/.bash_history` were washed out. I believe we should set `HISTFILE=/dev/null` in `conftest.py` not to affect the users's history.

This time I didn't add the test for this behavior. If you think we need to add tests for tests, please let me know.

**Edit**: Resolve #321

### DIFF
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -191,6 +191,7 @@ def bash(request) -> pexpect.spawn:
             INPUTRC="%s/config/inputrc" % testdir,
             TERM="dumb",
             LC_COLLATE="C",  # to match Python's default locale unaware sort
+            HISTFILE="/dev/null", # not to break the user's history file
         )
     )
 
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -192,7 +192,7 @@ def bash(request) -> pexpect.spawn:
             TERM="dumb",
             LC_COLLATE="C",  # to match Python's default locale unaware sort
             HISTFILE="/dev/null",  # to leave user's history file alone
-         )
+        )
     )
 
     fixturesdir = os.path.join(testdir, "fixtures")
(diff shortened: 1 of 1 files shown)


akinomyoga:
Now I found an existing issue #321 

> I vaguely remember that others preferred that a long time ago. But I've always found it very annoying, just haven't got around to it especially now that probably most of my tests are run in dedicated containers. But a clean fix would be very welcome nevertheless. [(comment by @scop #321)](https://github.com/scop/bash-completion/issues/321#issuecomment-498015578)

I'm not sure what was the purpose of the "others", but if one needs some information on what commands are executed while the tests, how about instead set `HISTFILE=./conftest.log` so that the commands will be appended to `test/fixtures/conftest.log`?

scop:
Thanks!

akinomyoga:
Thank you for taking your time to review the PR and fix my English!

scop:
Forgot to note that I opted to not do anything about a history file of our own for now, let's revisit if there's a case where we'd need that. Loads of output (maybe even too much) can already be obtained by using `$BASHCOMP_TEST_LOGFILE`.

akinomyoga:
OK, thank you for the information.

By the way, as for #492, I'm now trying to understand what `@pytest.mark.complete` does. It seems `@pytest.mark.complete` cannot be directly used for the present purpose, but maybe I should add another decorator that allows changing the directory and shell settings. I think it takes time.

PR #526, opened 2021-05-22
Fix the printf unescaping in "_quote_readline_by_ref"

akinomyoga:
This is the PR for the fix discussed in the conversation https://github.com/scop/bash-completion/pull/492#discussion_r637213822. We want to revert the following change:

```diff
-          printf -v value "$value" # Decode escape sequences of \....
+            printf -v value %s "$value" # Decode escape sequences of \....
```

I have included the tests along with the fix.

### conftest.py (52814dd4): add an optional argument `rendered_cmd`

Because I need to make a test that involves a TAB character in the filenames, I would like to add this new argument to support command lines in which the actual input sequence and the result rendered to the terminal are different. For example, suppose we think about a command-line string `"echo alpha\\\tb"`:

- A. Inputs from a keyboard would be `"echo alpha\\\026\tb"` i.e., <kbd>e c h o SP a l p h a \ C-v TAB b</kbd>
- B. The internal command-line string becomes `"echo alpha\\\tb"` i.e., <kbd>e c h o SP a l p h a \ TAB b</kbd>
- C. Bash renders the command in the terminal as `"/@echo alpha\\   b"` i.e. <kbd>/ @ e c h o SP a l p h a \ SP SP SP b</kbd> where `/@` is the prompt.

The current problem of the testing framework is that it assumes that `A == C` and just performs `bash.expect_exact` for the input string A. In this commit, I allow users to optionally specify C through the named argument `rendered_cmd="..."`. As I'm not a native English speaker, I don't have a good sense of choosing the argument name. If you have a better idea for the name of the new argument, could you please specify it?

### unit/test_quote_readline.py (608badfa): add a failing test

This tests the completion results for `"echo alpha\\\tb"`.

### bash_completion (ab583900): fix `_quote_readline_by_ref`

This is the suggested change of removing `%s` from `printf`.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -554,7 +554,8 @@ _quote_readline_by_ref()
         if [[ ${!2} == \$\'*\' ]]; then
             local value=${!2:2:-1}      # Strip beginning $' and ending '.
             value=${value//'%'/%%}      # Escape % for printf format.
-            printf -v value %s "$value" # Decode escape sequences of \....
+            # shellcheck disable=SC2059
+            printf -v value "$value"    # Decode escape sequences of \....
             local "$2" && _upvars -v "$2" "$value"
         fi
     fi
--- a/bash_completion
+++ b/bash_completion
@@ -552,10 +552,10 @@ _quote_readline_by_ref()
         # to drop the additional quoting.  See also:
         # https://www.mail-archive.com/bash-completion-devel@lists.alioth.debian.org/msg01942.html
         if [[ ${!2} == \$\'*\' ]]; then
-            local value=${!2:2:-1}      # Strip beginning $' and ending '.
-            value=${value//'%'/%%}      # Escape % for printf format.
+            local value=${!2:2:-1} # Strip beginning $' and ending '.
+            value=${value//'%'/%%} # Escape % for printf format.
             # shellcheck disable=SC2059
-            printf -v value "$value"    # Decode escape sequences of \....
+            printf -v value "$value" # Decode escape sequences of \....
             local "$2" && _upvars -v "$2" "$value"
         fi
     fi
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -511,7 +511,7 @@ def assert_complete(
         bash.send(cmd + "\t")
         # Sleep a bit if requested, to avoid `.*` matching too early
         time.sleep(kwargs.get("sleep_after_tab", 0))
-        bash.expect_exact(cmd)
+        bash.expect_exact(kwargs.get("rendered_cmd", cmd))
         bash.send(MAGIC_MARK)
         got = bash.expect(
             [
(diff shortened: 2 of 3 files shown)


scop:
Cool, thanks. This is pretty intricate stuff, frankly much more low level than I'm comfortable with, but not at all your fault and seems necessary here. Added some lint fixes (highly recommended to install pre-commit, added a blurb about that in e1c400a4a325685db87c8a5f901e9d1e0a7c4590).

I guess the only bit of an issue I'd like to see addressed is that the rendered cmd appears to assume PS1 length of 2, seems kind of magical. Maybe add some comment clarifying that, or even define it in terms of PS1 length, e.g.
```python
rendered_cmd="echo alpha\\"
rendered_cmd += " " * (8 - (len(conftest.PS1 + rendered_cmd) % 8)) + "b"
```
...but that's not too clear either, maybe a comment would do better.

akinomyoga:
> (highly recommended to install pre-commit, added a blurb about that in [e1c400a](https://github.com/scop/bash-completion/commit/e1c400a4a325685db87c8a5f901e9d1e0a7c4590)).

Thank you for the instruction! I have installed `pre-commit` (and also `perl-CPAN` required by `pre-commit` and `mypy`). But I'm not sure I understand the usage. If I just run the command `pre-commit` everything is skipped. I don't know how to use it but ended up with the following command:

```bash
$ pre-commit run --from-ref @~ --to-ref @ -v --show-diff-on-failure --all-files
```

Is there any other correct (shorter) command to perform the lint check on the previous commit? Or do I need to run this long command every time?

Anyway, some lint errors have been automatically corrected and reflected in the actual files. But there are still errors by `mypy`, and I don't know how to check the details of the error:

```
mypy.....................................................................Failed
- hook id: mypy
- duration: 0.03s
- exit code: 1
```

If I directly run `mypy` (though I'm not sure if I correctly run `mypy`), I got the following errors:

```
$ mypy --config-file=test/setup.cfg test/t/unit/test_unit_quote_readline.py                   bash-com
test/t/unit/test_unit_quote_readline.py:3: error: Skipping analyzing 'pytest': found module but no type hints or library stubs  [import]
test/t/unit/test_unit_quote_readline.py:3: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
test/t/unit/test_unit_quote_readline.py:5: error: Cannot find implementation or library stub for module named 'conftest'  [import]
Found 2 errors in 1 file (checked 1 source file)
```

Anyway, I push the commit of the current status e4ef29e6.

----

> I guess the only bit of an issue I'd like to see addressed is that the rendered cmd appears to assume PS1 length of 2, seems kind of magical.

Yes.

> Maybe add some comment clarifying that, or even define it in terms of PS1 length, e.g.
> 
> [...]
> 
> ...but that's not too clear either, maybe a comment would do better.

I added a comment in `unit/test_unit_quote_readline.py.

scop:
`pre-commit install` will install it so that it'll run automatically on commit, in git's pre-commit hook. I've encountered the mypy failure myself too, will have a look sometime.

But very nice, thanks a bunch again!

akinomyoga:
> pre-commit install will install it so that it'll run automatically on commit, in git's pre-commit hook

Ah, OK. I see. Thank you!

PR #528, opened 2021-05-23
completions: invoke "cd" as "builtin cd"

akinomyoga:
In searching the uses of `cd` command in completions, I found that `cd` is directly called instead of through `builtin cd`. The builtin `cd` is one of the commands that users define an alias or a function (to save the visited directory, etc.). I suggest prefixing them with `builtin` keyword.

I searched for the policy of `bash-completion` on this kind of treatment and found in `CONTRIBUTING.md` the following item:

> - Use printf(1) instead of echo(1) for portability reasons, and be sure to invoke commands that are often found aliased (such as ls or grep etc) using the `command` (or `builtin`) command as appropriate.

So I think we should also apply this rule to `cd` builtin. (By the way, I found that so far no builtin commands have been prefixed by `builtin` in any completions, so `cd` is the first one).

I have also added a rule to `test/runLint`. There are a few false positives, and many `cd` in test codes. If we should also prefix `cd` in test codes, please let me know. Then, I will add them.

### DIFF
--- a/completions/removepkg
+++ b/completions/removepkg
@@ -16,7 +16,7 @@ _removepkg()
 
     local root=${ROOT:-/}
     COMPREPLY=($(
-        cd "$root/var/log/packages" 2>/dev/null || return 1
+        command cd "$root/var/log/packages" 2>/dev/null || return 1
         compgen -f -- "$cur"
     ))
 } &&
--- a/completions/removepkg
+++ b/completions/removepkg
@@ -16,7 +16,7 @@ _removepkg()
 
     local root=${ROOT:-/}
     COMPREPLY=($(
-        command cd "$root/var/log/packages" 2>/dev/null || return 1
+        command cd -- "$root/var/log/packages" 2>/dev/null || return 1
         compgen -f -- "$cur"
     ))
 } &&
--- a/completions/_mount.linux
+++ b/completions/_mount.linux
@@ -38,14 +38,14 @@ _mount()
             ;;
         -L)
             COMPREPLY=($(
-                cd "/dev/disk/by-label/" 2>/dev/null || return
+                command cd "/dev/disk/by-label/" 2>/dev/null || return
                 compgen -f -- "$cur"
             ))
             return
             ;;
         -U)
             COMPREPLY=($(
-                cd "/dev/disk/by-uuid/" 2>/dev/null || return
+                command cd "/dev/disk/by-uuid/" 2>/dev/null || return
                 compgen -f -- "$cur"
             ))
             return
--- a/completions/_umount.linux
+++ b/completions/_umount.linux
@@ -86,7 +86,7 @@ _linux_fstab()
             local i
             for i in ${!COMPREPLY[*]}; do
                 [[ ${COMPREPLY[i]} == "$realcur"* ]] &&
-                    COMPREPLY+=($(cd "$dircur" 2>/dev/null &&
+                    COMPREPLY+=($(command cd "$dircur" 2>/dev/null &&
                         compgen -f -d -P "$dircur" \
                             -X "!${COMPREPLY[i]##"$dirrealcur"}" -- "$basecur"))
             done
--- a/completions/_umount.linux
+++ b/completions/_umount.linux
@@ -86,7 +86,7 @@ _linux_fstab()
             local i
             for i in ${!COMPREPLY[*]}; do
                 [[ ${COMPREPLY[i]} == "$realcur"* ]] &&
-                    COMPREPLY+=($(command cd "$dircur" 2>/dev/null &&
+                    COMPREPLY+=($(command cd -- "$dircur" 2>/dev/null &&
                         compgen -f -d -P "$dircur" \
                             -X "!${COMPREPLY[i]##"$dirrealcur"}" -- "$basecur"))
             done
--- a/completions/feh
+++ b/completions/feh
@@ -27,13 +27,13 @@ _feh()
             fi
             local font_path
             # font_path="$(imlib2-config --prefix 2>/dev/null)/share/imlib2/data/fonts"
-            # COMPREPLY=( $(command cd "$font_path" 2>/dev/null; compgen -f \
+            # COMPREPLY=( $(command cd -- "$font_path" 2>/dev/null; compgen -f \
             #     -X "!*.@([tT][tT][fF])" -S / -- "$cur") )
             for ((i = ${#words[@]} - 1; i > 0; i--)); do
                 if [[ ${words[i]} == -@(C|-fontpath) ]]; then
                     font_path="${words[i + 1]}"
                     COMPREPLY+=($(
-                        command cd "$font_path" 2>/dev/null
+                        command cd -- "$font_path" 2>/dev/null
                         compgen -f \
                             -X "!*.@([tT][tT][fF])" -S / -- "$cur"
                     ))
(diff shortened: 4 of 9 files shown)


scop:
Hm, I don't think I've heard of `cd` being aliased to something, and nobody has to my knowledge reported a problem with this in nearly two decades. (BTW cd does have some/previous visited dir support out of the box.)

I've resisted changes like this, because one could make the same argument for pretty much any command out there leading to prefixing everything which would be annoying. In some cases would not be the right thing to do either, because there might be very real reasons for why something is aliased and having bash-completion use it is desirable.

If there is a case why we'd like to do this for `cd` anyway despite the above, why wouldn't we do it via `command` like we do for others, but use `builtin` for it instead? (Sure, `builtin` is not applicable to sed, grep and the like, but on the other hand `command` would seem to work for `cd`.)

akinomyoga:
> Hm, I don't think I've heard of cd being aliased to something,

For some simple aliases, maybe you can search with `alias cd='cd -P'` or `alias cd=pushd` (though I don't think this kind of alias is useful). More typical cases define a function named `cd` (or define a function `mycd` and alias it as `cd`). Usually, people define these tiny functions by themselves, but there is also a popular plugin:

- [b4b4r07/enhancd: A next-generation cd command with your interactive filter](https://github.com/b4b4r07/enhancd)

I also wrote my own, being inspired by some blog article (it is about ten years ago, so I don't remember the page. Anyway it was written in Japanese). Also, I have received in my project a report on the problem caused by alias/function `cd` before. I can also find many examples by searching in the internet:

- [all-bash-history/acd_func.sh at master · djoot/all-bash-history](https://github.com/djoot/all-bash-history/blob/42fa6ddd9c840611ce5ef8890eae2cd6aa9f1927/acd_func.sh#L57)
- [cdhist/cdhist.rc at b6ddb70938cfcbd0ffd93ca6db119f0b54df7c2d · bulletmark/cdhist](https://github.com/bulletmark/cdhist/blob/b6ddb70938cfcbd0ffd93ca6db119f0b54df7c2d/cdhist.rc#L28)
- [cdext/cdext-setup.bash at master · ndragon70/cdext](https://github.com/ndragon70/cdext/blob/master/cdext-setup.bash#L15)
- [mshex/cdhist.sh at a017140c9d2b09327215a844e6da38521c54d34c · akinomyoga/mshex](https://github.com/akinomyoga/mshex/blob/a017140c9d2b09327215a844e6da38521c54d34c/shrc/cdhist.sh#L173) This is mine
- [bash - Do we have more history for cd? - Unix &amp; Linux Stack Exchange](https://unix.stackexchange.com/questions/157763/do-we-have-more-history-for-cd)
- [replace cd in bash to (silent) pushd](https://gist.github.com/mbadran/130469)

There seem to be also discussions on `alias cd='cd -P'` and `alias cd=pushd`

- [bash - Always use `-P` flag with `cd` command - Ask Ubuntu](https://askubuntu.com/questions/320127/always-use-p-flag-with-cd-command)
- [command line - Using `alias' to add a default option - e.g. `alias cd="cd -P" ' - Stack Overflow](https://stackoverflow.com/questions/11697478/using-alias-to-add-a-default-option-e-g-alias-cd-cd-p)
- [bash - aliasing cd to pushd - is it a good idea? - Unix &amp; Linux Stack Exchange](https://unix.stackexchange.com/questions/4290/aliasing-cd-to-pushd-is-it-a-good-idea)
- [Alias cd to record your directory travelling Using alias](https://www.commandlinefu.com/commands/view/7196/alias-cd-to-record-your-directory-travelling)

> and nobody has to my knowledge reported a problem with this in nearly two decades.

I think that is because these aliases and functions usually preserve the visible behavior of `cd` so do not cause problems immediately, but later, for example, one would notice that their `cd` history is contaminated by some paths the user didn't visit. This is not a serious problem, and it is not clear at a glance when the contamination has occurred and what is the culprit, but it is simply annoying.

> (BTW cd does have some/previous visited dir support out of the box.)

builtin `cd` only remembers one previous place. `pushd` may be used to remember all the visited directory in the session, but it is not shared with other sessions and will disappear once the session ends.
A useful example that cannot be achieved by the built-in bash features is above-mentioned [b4b4r07/enhancd](https://github.com/b4b4r07/enhancd). Another plugin [wting/autojump](https://github.com/wting/autojump) doesn't override `cd` but define a different command `j`, but it shows how the behavior of `cd` can be usefully modified.

> I've resisted changes like this, because one could make the same argument for pretty much any command out there leading to prefixing everything which would be annoying.

I think `cd` is one of the most frequently used commands in the interactive shells, so it is not surprising to me that people modify the behavior by alias/functions.

> In some cases would not be the right thing to do either, because there might be very real reasons for why something is aliased and having bash-completion use it is desirable.

I don't think these interactive features (history recording, automatic corrections/expansions, etc.) are wanted for internal completions by users.

> If there is a case why we'd like to do this for `cd` anyway despite the above, why wouldn't we do it via `command` like we do for others, but use `builtin` for it instead? (Sure, `builtin` is not applicable to sed, grep and the like, but on the other hand `command` would seem to work for `cd`.)

OK, I don't have particular preferences on them, so make it the same as `sed`, `grep`, etc.

akinomyoga:
I think I should move the change `cd => builtin cd` in #527 to here as you have requested to separate the PR. It will causes conflicts with #527, so I'll update this PR after #527 has been merged.

scop:
> I don't think these interactive features (history recording, automatic corrections/expansions, etc.) are wanted for internal completions by users.

That's right. But with this I was trying to refer to blanket `builtin`/`command` prefixing _everything_, not `cd` in particular which I agree would unlikely be adversely affected.

scop:
Anyway, thank you for the references, I'm sold on it now :) Let's have a look after #527 as you said.

akinomyoga:
I have updated to use `command cd` (instead of `builtin cd`). I also included the changes on `cd` which was originally in #527.

completions/removepkg
```diff
@@ -16,7 +16,7 @@ _removepkg()
 
     local root=${ROOT:-/}
     COMPREPLY=($(
-        cd "$root/var/log/packages" 2>/dev/null || return 1
+        command cd "$root/var/log/packages" 2>/dev/null || return 1
```
scop, line 19:
```suggestion
        command cd -- "$root/var/log/packages" 2>/dev/null || return 1
```

I suppose `$ROOT` can contain something that would make this fail.

scop (changes requested):

completions/removepkg
akinomyoga: Thank you for pointing this out! I somehow wrongly assumed that it will be treated as an ordinary argument when it contains / (just like `echo -e'a/b/c'`) but it was not the case. I also modified `completions/slackpkg`.

https://github.com/scop/bash-completion/blob/d8eab4776d1368e5a400cd244f543eaa437186d7/completions/slackpkg#L66

In `completions/slackpkg`, `$confdir` is currently a fixed string `/etc/slackpkg`

https://github.com/scop/bash-completion/blob/d8eab4776d1368e5a400cd244f543eaa437186d7/completions/slackpkg#L44

but it is not clear at a glance. Also, maybe some changes can be made in future, so I added `--` to the `cd` in `completions/slackpkg` as well. I pushed a new commit 1df7fde6e6e4eabe5a18e7f8b35dcd366173611f.

scop:
Thanks!

Issue #223, opened 2018-07-02
rsync completions stalled by --exclude


khanrahan:
Hello,

Completion on rsync works for me when I begin with `rsync -Pahvi`  but not if I use `rsync -Pahvi --exclude=".*"`

I have been working around this by setting up my source and destination paths and then going back and inserting the part to exclude hidden files.

Below is my `set -v` output.  Thanks in advance.

```
MacBook-Air:~ kieran$ rsync -Pahvin --exclude=".*" /Vwords=("${@:3:5}")
cword="$3"
cur="$3"
cur="$3"
cword="$3"
prev="$3"
words=("${@:3:5}")
```

```
Kierans-MacBook-Air:~ kieran$ echo $BASH_VERSION
4.4.19(1)-release
```

scop:
Reproduced. Seems this is not limited to rsync but is probably a deeper problem, it's the quoting in `".*"` that breaks things somehow. We end up splitting words on the `.` within the quotes, no idea why.

PR #529, opened 2021-05-29
test(man): fix skip failure of `test_10`

akinomyoga:
This fixes the test failure of `test_man.py`. `@pytest.mark.complete(require_cmd=True)` is referenced only when the fixture `completion` is requested. Since `completion` is not requested in the new version of `test_10`, `require_cmd=True` is inactive now. This PR explicitly checks the command (non-)existence to properly skip the test.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2369,8 +2369,12 @@ fi
 unset compat_dir i _blacklist_glob
 
 # source user completion file
+#
+# Remark: We explicitly check that $user_completion is not '/dev/null' since
+#   /dev/null may be a regular file in broken systems and can contain arbitrary
+#   garbages of suppressed command outputs.
 user_completion=${BASH_COMPLETION_USER_FILE:-~/.bash_completion}
-[[ ${BASH_SOURCE[0]} != "$user_completion" && -r $user_completion && -f $user_completion ]] &&
+[[ $user_completion != "${BASH_SOURCE[0]}" && $user_completion != /dev/null && -r $user_completion && -f $user_completion ]] &&
     . $user_completion
 unset user_completion
 
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -399,7 +399,12 @@ def assert_bash_exec(
 
 
 class bash_env_saved:
+    counter: int = 0
+
     def __init__(self, bash: pexpect.spawn, sendintr: bool = False):
+        bash_env_saved.counter += 1
+        self.prefix: str = "_BASHCOMP_TEST%d" % bash_env_saved.counter
+
         self.bash = bash
         self.cwd_changed: bool = False
         self.saved_shopt: Dict[str, int] = {}
@@ -465,53 +470,58 @@ class bash_env_saved:
     def _save_cwd(self):
         if not self.cwd_changed:
             self.cwd_changed = True
-            self._copy_variable("PWD", "_BASHCOMP_TEST_OLDPWD")
+            self._copy_variable("PWD", "%s_OLDPWD" % self.prefix)
 
     def _check_shopt(self, name: str):
         self._safe_assert(
-            '[[ $(shopt -p %s) == "${_BASHCOMP_TEST_NEWSHOPT_%s}" ]]'
-            % (name, name),
+            '[[ $(shopt -p %s) == "${%s_NEWSHOPT_%s}" ]]'
+            % (name, self.prefix, name),
         )
 
     def _unprotect_shopt(self, name: str):
         if name not in self.saved_shopt:
             self.saved_shopt[name] = 1
             self._safe_exec(
-                "_BASHCOMP_TEST_OLDSHOPT_%s=$(shopt -p %s || true)"
-                % (name, name),
+                "%s_OLDSHOPT_%s=$(shopt -p %s || true)"
+                % (self.prefix, name, name),
             )
         else:
             self._check_shopt(name)
 
     def _protect_shopt(self, name: str):
         self._safe_exec(
-            "_BASHCOMP_TEST_NEWSHOPT_%s=$(shopt -p %s || true)" % (name, name),
+            "%s_NEWSHOPT_%s=$(shopt -p %s || true)"
+            % (self.prefix, name, name),
         )
 
     def _check_variable(self, varname: str):
         try:
             self._safe_assert(
-                '[[ ${%s-%s} == "${_BASHCOMP_TEST_NEWVAR_%s-%s}" ]]'
-                % (varname, MAGIC_MARK2, varname, MAGIC_MARK2),
+                '[[ ${%s-%s} == "${%s_NEWVAR_%s-%s}" ]]'
+                % (varname, MAGIC_MARK2, self.prefix, varname, MAGIC_MARK2),
             )
         except Exception:
-            self._copy_variable("_BASHCOMP_TEST_NEWVAR_" + varname, varname)
+            self._copy_variable(
+                "%s_NEWVAR_%s" % (self.prefix, varname), varname
+            )
             raise
         else:
             if self.noexcept and self.captured_error:
                 self._copy_variable(
-                    "_BASHCOMP_TEST_NEWVAR_" + varname, varname
+                    "%s_NEWVAR_%s" % (self.prefix, varname), varname
                 )
 
     def _unprotect_variable(self, varname: str):
         if varname not in self.saved_variables:
             self.saved_variables[varname] = 1
-            self._copy_variable(varname, "_BASHCOMP_TEST_OLDVAR_" + varname)
+            self._copy_variable(
+                varname, "%s_OLDVAR_%s" % (self.prefix, varname)
+            )
         else:
             self._check_variable(varname)
 
     def _protect_variable(self, varname: str):
-        self._copy_variable(varname, "_BASHCOMP_TEST_NEWVAR_" + varname)
+        self._copy_variable(varname, "%s_NEWVAR_%s" % (self.prefix, varname))
 
     def _restore_env(self):
         self.noexcept = True
@@ -523,23 +533,25 @@ class bash_env_saved:
         # variables because "cd" affects "OLDPWD".
         if self.cwd_changed:
             self._unprotect_variable("OLDPWD")
-            self._safe_exec('command cd -- "$_BASHCOMP_TEST_OLDPWD"')
+            self._safe_exec('command cd -- "$%s_OLDPWD"' % self.prefix)
             self._protect_variable("OLDPWD")
-            self._unset_variable("_BASHCOMP_TEST_OLDPWD")
+            self._unset_variable("%s_OLDPWD" % self.prefix)
             self.cwd_changed = False
 
         for name in self.saved_shopt:
             self._check_shopt(name)
-            self._safe_exec('eval "$_BASHCOMP_TEST_OLDSHOPT_%s"' % name)
-            self._unset_variable("_BASHCOMP_TEST_OLDSHOPT_" + name)
-            self._unset_variable("_BASHCOMP_TEST_NEWSHOPT_" + name)
+            self._safe_exec('eval "$%s_OLDSHOPT_%s"' % (self.prefix, name))
+            self._unset_variable("%s_OLDSHOPT_%s" % (self.prefix, name))
+            self._unset_variable("%s_NEWSHOPT_%s" % (self.prefix, name))
         self.saved_shopt = {}
 
         for varname in self.saved_variables:
             self._check_variable(varname)
-            self._copy_variable("_BASHCOMP_TEST_OLDVAR_" + varname, varname)
-            self._unset_variable("_BASHCOMP_TEST_OLDVAR_" + varname)
-            self._unset_variable("_BASHCOMP_TEST_NEWVAR_" + varname)
+            self._copy_variable(
+                "%s_OLDVAR_%s" % (self.prefix, varname), varname
+            )
+            self._unset_variable("%s_OLDVAR_%s" % (self.prefix, varname))
+            self._unset_variable("%s_NEWVAR_%s" % (self.prefix, varname))
         self.saved_variables = {}
 
         self.noexcept = False
@@ -602,7 +614,7 @@ def diff_env(before: List[str], after: List[str], ignore: str):
         if not re.search(r"^(---|\+\+\+|@@ )", x)
         # Ignore variables expected to change:
         and not re.search(
-            r"^[-+](_|PPID|BASH_REMATCH|_BASHCOMP_TEST_\w+)=",
+            r"^[-+](_|PPID|BASH_REMATCH|_BASHCOMP_TEST\w+)=",
             x,
             re.ASCII,
         )
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -186,29 +186,55 @@ def partialize(
 def bash(request) -> pexpect.spawn:
 
     logfile = None
+    histfile = None
+    tmpdir = None
+    bash = None
+
     if os.environ.get("BASHCOMP_TEST_LOGFILE"):
         logfile = open(os.environ["BASHCOMP_TEST_LOGFILE"], "w")
     elif os.environ.get("CI"):
         logfile = sys.stdout
+
     testdir = os.path.abspath(
         os.path.join(os.path.dirname(__file__), os.pardir)
     )
-    env = os.environ.copy()
-    env.update(
-        dict(
-            SRCDIR=testdir,  # TODO needed at least by bashrc
-            SRCDIRABS=testdir,
-            PS1=PS1,
-            INPUTRC="%s/config/inputrc" % testdir,
-            TERM="dumb",
-            LC_COLLATE="C",  # to match Python's default locale unaware sort
-            HISTFILE="/dev/null",  # to leave user's history file alone
-        )
+
+    # Create an empty temporary file for HISTFILE.
+    #
+    # To prevent the tested Bash processes from writing to the user's
+    # history file or any other files, we prepare an empty temporary
+    # file for each test.
+    #
+    # - Note that HISTFILE=/dev/null may not work.  It results in the
+    #   removal of the device /dev/null and the creation of a regular
+    #   file at /dev/null when the number of commands reach
+    #   HISTFILESIZE by Bash-4.3's bug.  This causes the execution of
+    #   garbages through BASH_COMPLETION_USER_FILE=/dev/null.  - Note
+    #   also that "unset -v HISTFILE" in "test/config/bashrc" was not
+    #   adopted because "test/config/bashrc" is loaded after the
+    #   history is read from the history file.
+    #
+    histfile = tempfile.NamedTemporaryFile(
+        prefix="bash-completion-test_", delete=False
     )
 
-    tmpdir = None
-    bash = None
     try:
+        # release the file handle so that Bash can open the file.
+        histfile.close()
+
+        env = os.environ.copy()
+        env.update(
+            dict(
+                SRCDIR=testdir,  # TODO needed at least by bashrc
+                SRCDIRABS=testdir,
+                PS1=PS1,
+                INPUTRC="%s/config/inputrc" % testdir,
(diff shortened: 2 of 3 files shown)


akinomyoga:
The failure of `test_man.py` is solved by commit 489fc72, but this revealed another problem in `distcheck (ubuntu14)`.  It is only reproduced in `distcheck (ubuntu14)`.  There is no problem in the other environments including my local one. I tried to fix the problem, but I'm still not sure what is happening there.

The previous code using `bash_{save,resore}_variable` passes the test, but the new code `bash_env_saved` fails if the detection of unwanted `OLDPWD` changes is turned on, which means that something is wrong in testing. Even though I can make it pass by turning off the detection, it is not the true solution. Anyway, I add changes to make `bash_env_saved` more robust though they don't seem to solve the problem.

### 7f1d8cbf [Bug fix] Correctly restore `cwd`

In the previous version of `bash_env_saved`, I used `bash.cwd` to get the current working directory of the Bash process, but it turned out that `bash.cwd` just retains the value of the startup time of the process but not the current one. I instead change to store the real current working directory in a shell variable inside the Bash process.

### fb41aec7 Try to continue the remaining restoring process even when some restoration operations fail

Although the entire test anyway fails If one of the restoration operations fails, `bash_env_saved` now tries to perform the remaining restoration process to reduce unnecessary errors caused by the restoration failure.

### ec91250b Support nested call of `bash_env_saved`

In the previous version of `bash_env_saved`, when the same variables are saved/restored in nested `with bash_env_saved()`, the variables are not correctly restored.  I'm not sure whether there are real instances of such cases in the testing, but `bash_env_saved` now supports the nested call of `with bash_env_saved()` in this commit.

akinomyoga:
## I found that `/dev/null` is a regular file in distcheck (ubuntu14)!

Since we have set `HISTFILE=/dev/null`,

https://github.com/scop/bash-completion/blob/7e0803f7c92252b436bc05d9bc6c5b818450a2ee/test/t/conftest.py#L205

the command history which contains `source .../bash_completion` are saved in the file `/dev/null`. Then, `bash_completion` tries to source `BASH_COMPLETION_USER_FILE=/dev/null` (set up in `test/config/bashrc`).

https://github.com/scop/bash-completion/blob/7e0803f7c92252b436bc05d9bc6c5b818450a2ee/test/config/bashrc#L31

https://github.com/scop/bash-completion/blob/7e0803f7c92252b436bc05d9bc6c5b818450a2ee/bash_completion#L2371-L2374

This causes an infinite source chain of `bash_completion` -> File `/dev/null` -> `bash_completion` -> File `/dev/null` -> ...

----

I have added fixes to this problem 492150b6. I have also rebased the commits and added the corresponding description to each commit log. Now it's ready to be reviewed and merged.

bash_completion
```diff
@@ -2369,8 +2369,12 @@ fi
 unset compat_dir i _blacklist_glob
 
 # source user completion file
+#
+# Remark: We explicitly check that $user_completion is not '/dev/null' since
+#   /dev/null may be a regular file in broken systems and can contain arbitrary
+#   garbages of suppressed command outputs.
```
scop, line 2375: I'm not sure what to think about this. On the other hand it seems quite benign, but on the other a broken system is a broken system, and if we start to add workarounds like this for specific cases, where does it end? What if the user completion points to some other file with random junk in it?

akinomyoga: Yeah, that is a valid concern. Broken systems are broken systems, so we don't have to care about them too much in order not to mess up the code. But at the same time, considering the facts that `/dev/null` as `BASH_COMPLETION_USER_FILE` may be frequently used to suppress user configurations, and also that it can possibly cause disasters in such a broken system though it is a rare case, I thought `/dev/null` might be treated as a special path.

In another way of thinking, `/dev/null` can be regarded as a special *keyword* to mean "*do nothing*". When users specify `/dev/null` to some variables, the users actually think about the behavior of this bit bucket but not the actual device linked to this particular location in the filesystem. In this sense, I don't think we should add more special rules to deal with other variations to represent the same location, such as `//dev/null`, `/tmp/../dev/null`, `/./././dev/null`, etc. (by using e.g. `readlink -f`) and also rules for other possible junk files `/dev/random`, `/dev/zero`, or whatever. The `/dev/null` is just special. I don't think there are so many other cases equally frequent and special as `/dev/null` for `BASH_COMPLETION_USER_FILE`. I suspect there are no such cases, so we don't have to worry about the endless workarounds.  We may add special treatment for this particular string just like `/dev/{tcp,udp}/$hostname/$port` in Bash or `/dev/{stdin,stdout,stderr,fd/$n}` in GNU awk.

I think that is also a valid perspective. I think it's not a matter that which one is correct or incorrect, so I'm not reluctant to drop this fix if you prefer not to have it in `bash_completion`! I'd like to hear your preference.

akinomyoga:
I had thought that the Docker image was already broken, but looking at the test log, I guess that `/dev/null` is broken in the middle of the testing:

```
2021-05-31T02:41:03.5296710Z ../../../test/t/unit/test_unit_filedir.py::TestUnitFiledir::test_25[f] PASSED [ 89%]
2021-05-31T02:41:04.3482533Z ../../../test/t/unit/test_unit_filedir.py::TestUnitFiledir::test_25[f2] PASSED [ 89%]
2021-05-31T02:41:05.1673845Z ../../../test/t/unit/test_unit_filedir.py::TestUnitFiledir::test_26[f] PASSED [ 89%]
2021-05-31T02:41:05.9883034Z ../../../test/t/unit/test_unit_filedir.py::TestUnitFiledir::test_26[f2] PASSED [ 89%]
2021-05-31T02:41:05.9901600Z ../../../test/t/unit/test_unit_filedir.py::TestUnitFiledir::test_27[f] SKIPPED [ 89%]
2021-05-31T02:41:06.2023157Z ../../../test/t/unit/test_unit_filedir.py::TestUnitFiledir::test_27[f2] SKIPPED [ 89%]
2021-05-31T02:41:06.2043100Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_1 PASSED [ 89%]
2021-05-31T02:41:06.2062051Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_2 PASSED [ 89%]
2021-05-31T02:41:06.2079446Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_3 PASSED [ 89%]
2021-05-31T02:41:06.2097433Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_4 PASSED [ 89%]
2021-05-31T02:41:06.2115222Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_5 PASSED [ 89%]
2021-05-31T02:41:06.2133984Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_6 PASSED [ 89%]
2021-05-31T02:41:06.2153268Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_7 PASSED [ 89%]
2021-05-31T02:41:06.2171814Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_8 PASSED [ 89%]
2021-05-31T02:41:06.2195394Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_9 PASSED [ 89%]
2021-05-31T02:41:06.2230009Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_10 PASSED [ 89%]
2021-05-31T02:41:06.2242000Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_11 PASSED [ 90%]
2021-05-31T02:41:06.2258583Z ../../../test/t/unit/test_unit_find_unique_completion_pair.py::TestUnitFindUniqueCompletionPair::test_12 PASSED [ 90%]
2021-05-31T02:41:07.2114565Z ../../../test/t/unit/test_unit_get_comp_words_by_ref.py::TestUnitGetCompWordsByRef::test_1 ERROR [ 90%]
2021-05-31T02:41:07.2443361Z ../../../test/t/unit/test_unit_get_comp_words_by_ref.py::TestUnitGetCompWordsByRef::test_2 ERROR [ 90%]
2021-05-31T02:41:07.2767841Z ../../../test/t/unit/test_unit_get_comp_words_by_ref.py::TestUnitGetCompWordsByRef::test_3 ERROR [ 90%]
2021-05-31T02:41:07.3092224Z ../../../test/t/unit/test_unit_get_comp_words_by_ref.py::TestUnitGetCompWordsByRef::test_4 ERROR [ 90%]
2021-05-31T02:41:07.3411014Z ../../../test/t/unit/test_unit_get_comp_words_by_ref.py::TestUnitGetCompWordsByRef::test_5 ERROR [ 90%]
2021-05-31T02:41:07.3729791Z ../../../test/t/unit/test_unit_get_comp_words_by_ref.py::TestUnitGetCompWordsByRef::test_6 ERROR [ 90%]
```

The error starts to happen from `unit/test_unit_get_comp_words_by_ref.py` and all the later tests fail. The test immediately preceding it is `unit/test_unit_find_unique_completion_pair.py`, but this is unrelated because it only tests `conftest`'s utility `find_unique_completion_pair`. The culprit seems to be the second previous one `unit/test_unit_filedir.py`. In fact, the log includes error messages like `bash: cd: /tmp/bash-completion_filedirx_kom9mv: No such file or directory` which is the directory created by `unit/test_unit_filedir.py`. Probably, some test in `unit/test_unit_filedir.py` removes the device `/dev/null` and then a new regular file of the same name is created on the termination of the Bash process. I'm not sure what causes the removal of `/dev/null`.

scop:
> I'm not sure what causes the removal of `/dev/null`.

Huh, if it's something triggered by us, then that's something we need to fix. Maybe that old bash removes `$HISTFILE` in some circumstances.

It would seem safer to `unset HISTFILE` after shell startup instead of setting it to `/dev/null`, e.g. `test/config/bashrc` seems to be an appropriate place.

akinomyoga:
I found that it was actually a bug (or a quirk) of Bash 4.3 and lower. I found the following report in bug-bash:

- [If $HISTFILE is set to /dev/null and you execute more commands than $HISTFILESIZE, /dev/null is deleted.](https://lists.gnu.org/archive/html/bug-bash/2015-01/msg00138.html)

This means that the number of commands in the history hits the limit `$HISTFILESIZE` in `unit/test_unit_filedir.py`. I previously mentioned that dropping the detection of the `OLDPWD` change made the test pass, but now I understand that it was just because the number of executed commands in the test was reduced. So there is nothing wrong in our test code.

> It would seem safer to `unset HISTFILE` after shell startup instead of setting it to `/dev/null`, e.g. `test/config/bashrc` seems to be an appropriate place.

I have rebased the commits on the latest `master` and added the mentioned workaround 1dbe515. In case HISTFILE is accessed before `unset -v HISTFILE` in `test/config/bashrc` is executed, I set `HISTFILE=/tmp/bash_completion.bash_history` as a trasient value.

test/t/conftest.py
```diff
@@ -202,7 +202,7 @@ def bash(request) -> pexpect.spawn:
             INPUTRC="%s/config/inputrc" % testdir,
             TERM="dumb",
             LC_COLLATE="C",  # to match Python's default locale unaware sort
-            HISTFILE="/dev/null",  # to leave user's history file alone
+            HISTFILE="/tmp/bash_completion.bash_history",  # to leave user's history file alone
```
scop, line 205: If we end up writing to this file, I believe it will also be a file overwrite security issue through a symlink attack.

If we are to use a temporary file for this, we want to use a proper one created through either Python's `tempfile` facilities, or `mktemp` in shell (but that's not portable), and cleaned up appropriately. Unless we know of a real world situation in which we will end up writing there, I wouldn't personally bother but would just remove this and rely on the `unset`.

bash_completion
scop: I'm still kind of torn. It seems out of place in this PR now that we're no longer using `/dev/null` for `HISTFILE`. But nevermind, let's just keep it, it's a clean separate commit anyway.

scop (changes requested):

test/t/conftest.py
akinomyoga: > If we end up writing to this file, I believe it will also be a file overwrite security issue through a symlink attack.

Oh, you are right. Thank you.

Maybe I'm too nervous because I just lost my history of the recent few months, but I modified it to use `tempfile` 60c55437. Since `test/config/bashrc` is loaded after the first prompt is shown, the command history is already read when `unset -v HISTFILE` is executed. Maybe this doesn't affect the default user history file, but it is good to exclude dependencies of tests on the user's history file.

----

It seems that [the test for the new commit](https://github.com/scop/bash-completion/actions/runs/897929639) failed by timeout errors, but exactly the same commit has passed [the tests in Actions of my fork repository](https://github.com/akinomyoga/bash-completion/actions/runs/897929610).

test/t/conftest.py
```diff
@@ -186,29 +186,39 @@ def partialize(
...
+            logfile = sys.stdout
+
+        testdir = os.path.abspath(
+            os.path.join(os.path.dirname(__file__), os.pardir)
+        )
+
+        # create temporary file for HISTFILE
+        histfile = tempfile.NamedTemporaryFile(
+            prefix="bash-completion-test_", delete=False
+        )
```
scop, line 206: We could move the `try:` here before `histfile.close()`, as the above lines do not create any resources we could/should clean up if they fail -- better to keep `try` blocks as focused as possible. But no biggie, leaving up to you to decide.

test/t/conftest.py
```diff
@@ -186,29 +186,39 @@ def partialize(
...
+        if os.environ.get("BASHCOMP_TEST_LOGFILE"):
+            logfile = open(os.environ["BASHCOMP_TEST_LOGFILE"], "w")
+        elif os.environ.get("CI"):
+            logfile = sys.stdout
+
+        testdir = os.path.abspath(
+            os.path.join(os.path.dirname(__file__), os.pardir)
+        )
+
+        # create temporary file for HISTFILE
```
scop, line 203: Let's add a comment here and modify the commit message to elaborate why we're doing the tempfile here -- I think it is no longer obvious now that we're also unsetting `HISTFILE`. I know that at least I would have likely forgotten about the details in a year or two and would think of it as an oversight ripe fore cleanup :)

scop (changes requested):

test/t/conftest.py
scop: (...or maybe move it back to where it was, I'm not sure we can expect to be able to remove a tempfile that has failed to `close()`, but still not a biggie, feel free to leave as is if you feel it's better this way.)

test/t/conftest.py
akinomyoga: OK, thank you for your comments. I fixed up the last commit with the following change. I also added a description in the commit log. ce3583f6

<details><summary><code>git diff -b 60c5543 ce3583f</code></summary>

```diff
$ git diff -b 60c5543 ce3583f
diff --git a/test/t/conftest.py b/test/t/conftest.py
index fab52d6b..29d7418b 100644
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -189,7 +189,6 @@ def bash(request) -> pexpect.spawn:
     histfile = None
     tmpdir = None
     bash = None
-    try:

     if os.environ.get("BASHCOMP_TEST_LOGFILE"):
         logfile = open(os.environ["BASHCOMP_TEST_LOGFILE"], "w")
@@ -200,10 +199,20 @@ def bash(request) -> pexpect.spawn:
         os.path.join(os.path.dirname(__file__), os.pardir)
     )

-        # create temporary file for HISTFILE
+    # Create an empty temporary file for HISTFILE.
+    #
+    # To prevent the tested Bash processes from writing to the user's
+    # history file or any other files, we will unset the variable
+    # "HISTFILE" in "test/config/bashrc".  However, the history is
+    # loaded before the first prompt is shown and "test/config/bashrc"
+    # is loaded.  We create an empty history file using tempfile for a
+    # transient value of HISTFILE.
     histfile = tempfile.NamedTemporaryFile(
         prefix="bash-completion-test_", delete=False
     )
+
+    try:
+        # release the file handle so that Bash can open the file.
         histfile.close()

         env = os.environ.copy()
@@ -317,7 +326,10 @@ def bash(request) -> pexpect.spawn:
         if tmpdir:
             tmpdir.cleanup()
         if histfile:
+            try:
                 os.remove(histfile.name)
+            except OSError:
+                pass
         if logfile and logfile != sys.stdout:
             logfile.close()

```
</details>

1. I moved the position of `try:` to immediately before `histfile.close()`.
2. I added `try:`...`except OSError: pass` in case we failed to remove `histfile` (possibly after failing `histfile.close()` as you commented) in the cleanup code.
3. I added comment for the reasoning for the temporary file.

But maybe we don't have to unset `HISTFILE` anymore because we now have a temporary file which can be used as the output target of the command history. Just using the temporary file as `HISTFILE` is clearer than combining it with `unset -v HISTFILE` and thus easy to explain. Nevertheless we may explain why we didn't choose to use neither of `HISTFILE=/dev/null` or `unset -v HISTILFE`

test/t/conftest.py
```diff
@@ -186,29 +186,48 @@ def partialize(
...
+    # "HISTFILE" in "test/config/bashrc".  However, the history is
+    # loaded before the first prompt is shown and "test/config/bashrc"
+    # is loaded.  We create an empty history file using tempfile for a
+    # transient value of HISTFILE.
+    histfile = tempfile.NamedTemporaryFile(
+        prefix="bash-completion-test_", delete=False
     )
 
-    tmpdir = None
-    bash = None
```
scop, line 210: I don't think we should remove these -- if an error is raised before their definitions within the `try` block, the `finally` block will fail for the undefined one.

scop (changes requested):

test/t/conftest.py
scop: Agreed, clearer to remove the `unset` now that we have the tempfile.

test/t/conftest.py
akinomyoga: Thank you. These lines are moved to the beginning of the function. I just thought it is more symmetric to put together these four resources (`logfile`, `histfile`, `tmpdir`, and `bash`) that will be cleaned up in the `finally` block.

test/t/conftest.py
akinomyoga: May I confirm what is that "👍" for? Can I remove `unset -v HISTFILE` from `test/config/bashrc`?

I actually noticed that `test/config/bashrc` is linked from a symbolic link `docs/bashrc`, so I feel we shouldn't include test-specific codes in `test/config/bashrc`. Or, if we keep the `unset -v HISTFILE` line in `test/config/bashrc`, we may create a real copy at `docs/bashrc` that doesn't have `unset -v HISTFILE`.

akinomyoga: I have updated the commit. 0ee13b2

test/t/conftest.py
scop: Ah, missed that if that's how it was done already at the time I took a look at this.

scop (approved):

akinomyoga:
@scop Thank you for the patient review on my PR and the related discussions!

Can I delete these two GitHub Workflow actions, `#223` and `#225`? They have gigantic sizes of output log caused by the infinite source chain of `bash_completion` -> `/dev/null` -> `bash_completion` -> `/dev/null` -> ..., so I'd like to delete them. They have raw sizes of about 4.0GB and 3.4GB (or compressed sizes 500MB and 400MB), respectively. The other actions have "finite" sizes of logs because of `test_man` failure fortunately or unfortunately.

image

scop:
No problem, thank you!

By all means, go and delete them, nice catch.

akinomyoga:
Thanks! I've deleted them.

PR #532, opened 2021-06-04
feat(ssh-keyscan): new completion

scop:


completions/ssh-keyscan
```diff
@@ -0,0 +1,39 @@
...
+            ;;
+        -*f)
+            _filedir
+            return
+            ;;
+        -*p | -*T)
+            return
+            ;;
+        -*t)
+            COMPREPLY=($(compgen -W 'dsa ecdsa ed25519 rsa' -- "$cur"))
```
akinomyoga, line 25: I'm not familiar with `ssh-keyscan`, but `man ssh-keyscan` says (boldfaced by me)

> - -t type
>   Specify the type of the key to fetch from the scanned hosts.  The possible values are “dsa”, “ecdsa”, “ed25519”, or “rsa”.  **Multiple values may be specified by separating them with commas.**  The default is to fetch “rsa”, “ecdsa”, and “ed25519” keys.

Maybe something like the following?

```suggestion
            local prefix=
            [[ $cur == *,* ]] &&
                local prefix=${cur%,*}, cur=${cur%%*,}
            COMPREPLY=($(compgen ${prefix:+-P "$prefix"} -W 'dsa ecdsa ed25519 rsa' -- "$cur"))
```

scop: Ah, good catch! Implemented in 1893481a482c6d6088149d55a1930b4df8252e8b using `_comp_delimited`, it's nicer because it shows only the currently being completed value in its suggestions instead of the whole chain, and removes values already present.

akinomyoga: OK! Thank you! I didn't know this useful utility `_comp_delimited`!

akinomyoga:
Thank you for updating the PR. I have tested the new commit and found that the fields before the comma is removed when there are multiple completions for the string after the comma. For example, in the following cases, it works as expected:

<pre>
$ ssh-keyscan -t rsa,<kbd>TAB TAB</kbd>
dsa      ecdsa    ed25519     # <- the list is shown
$ ssh-keyscan -t rsa,d<kbd>TAB</kbd>
$ ssh-keyscan -t rsa,dsa     # <- correctly completed
</pre>

But in the following ambiguous case, the fields before comma is deleted by <kbd>TAB</kbd>.
<pre>
$ ssh-keyscan -t rsa,e<kbd>TAB</kbd>
$ ssh-keyscan -t e     # <- "rsa," is lost
</pre>

scop:
Seems what you describe happens with `set show-all-if-ambiguous off` only, I have it set to `on` and it works as expected for me (Ubuntu 20.04, bash 5.0.17(1)-release). Opened #552 for that, I suggest we don't wait for that here.

akinomyoga:
OK, fine! I now approve this PR for merging.

akinomyoga (approved):

PR #540, opened 2021-06-06
test(tar,known_hosts,p{gids,pids,names}): use "bash_env_saved"

akinomyoga:
Now that the test problems have been solved, I'd like to finish the fixes that had been pending.


akinomyoga:
Thanks

PR #541, opened 2021-06-06
Document commit message convention, add commitlint

scop:


akinomyoga (approved):
I'm actually not familiar with Conventional Commits, but I'm fine with adopting it since it seems to be used widely.

Also, I'm not familiar with `commitlintrc.yaml` and `.pre-commit-config.yaml`, so I'm not able to review the details of the change in those files but instead tested it by creating several dummy commits on top of the branch `docs/commit-messages`. As far as I have tested, there is no problem.

Currently, I don't see any problems, so I agree with merging this PR. I think we can handle any problems when they come out.

PR #543, opened 2021-06-06
test(conftest): prevent MAGIC_MARK from history expansions

akinomyoga:
[The failure of Workflow Action `#246`](https://github.com/scop/bash-completion/actions/runs/911061422) is caused by history expansions induced by `MAGIC_MARK = "__MaGiC-maRKz!__"`. In this pull request, we remove `!` from `MAGIC_MARK` to avoid unexpected history expansions. Instead, I appended 10 characters (coming from my machine's entropy, `head -c  10 /dev/urandom | base64`).

### DIFF
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -24,7 +24,7 @@ import pexpect  # type: ignore[import]
 import pytest
 
 PS1 = "/@"
-MAGIC_MARK = "__MaGiC-maRKz!__"
+MAGIC_MARK = "__MaGiC-maRKz-NEtXZVZfKC__"
 MAGIC_MARK2 = "Re8SCgEdfN"
 
 


scop:
Nice catch. However as a first impression (not having had a look at any code) I can't help thinking if we have a general failure to prevent things like this happening (missing escaping or something) that might bite us with something else if we "fix" this particular case by removing the offending `!`. Thoughts?

akinomyoga:
> I can't help thinking if we have a general failure to prevent things like this happening (missing escaping or something) that might bite us with something else if we "fix" this particular case by removing the offending `!`.

Sorry for the blank, but probably I fail to understand your concern correctly. Is `!` inside `MAGIC_MARK` used to prevent some other possible problems such as missing escaping?

I actually don't exactly understand why this problem happened in the action `#246`. After putting `MAGIC_MARK` in `stdin` of Bash, we send `SIGINT` to the Bash process and wait for the new prompt. But, in `#246`, somehow SIGINT has been ignored, and also the new prompt checking is skipped. As a result, the next command is directly appended in the line ending with `MAGIC_MARK` and then attempted for the execution. Hmm, one possibility is that `SIGINT` has been delayed and the prompt checking captured the prompt of the previous commands. Maybe I'll later look at the code again.

For the case that the completion has been tested against a command with incomplete quoting, such as `echo "xxx[TAB]`, it may also cause some unexpected results, but the problems caused by such cases are limited. For the history expansion, we cannot really predict what happens as we cannot predict to what command in the history `!...` is expanded. If there is no proper purpose for using `!` in `MAGIC_MARK`, I feel the behavior is more well-defined without `!` in some test failures.

scop:
There's no real significance with the `!` in the value contents, this PR is just fine from that point of view. Still can't shake the feeling what we're seeing could be symptom of a generic failure to quote bug, and just changing the mark value to something that doesn't expand to anything ever would be brushing the real problem under the carpet so to speak -- we could trigger the problem with something else besides the mark. But maybe not, meh, I'll just merge this.

There are a some things/occurrences I don't quite understand with the test suite. There are fewer now that it's in Python than there used to be when it was Tcl, but it's weird still. And flaky, especially if running in parallel with pytest-xdist (haven't dug deeply in that, hence just disabled for now).

akinomyoga:
> Still can't shake the feeling what we're seeing could be symptom of a generic failure to quote bug

OK, now I see what you described in the first reply. And yes, I agree that there is some generic problem in the testing suite. My guess is that this is related to a race condition of `SIGINT` and `stdin` of the tested Bash process rather than the quoting problem.
If the unexpected test failure persists in future workflows, I'll try to take a look at them.

PR #546, opened 2021-06-20
fix(make): fix stripped directory names with `bind 'set show-all-if-ambiguous on'`

### LINKED ISSUE #544: Make target completion doesn't work with subdirectories
## Describe the bug

Make target completion doesn't work when target files are in subdirectories.

## To reproduce

There are two behaviours that happen here. Sample makefile:
```make
.PHONY: abc/xyz

abc/xyz:
	mkdir -p abc
	date > $@
```
For clarity, i've marked spaces with `·`
1. When typing `make·TAB` the result is `make·abc/·`
2. When typing `make·abc/TAB` the result is `make·xyz·`

## Expected behavior

Both described cases should auto complete to `make·abc/xyz·` because it is unambiguous.

- [ ] Operating system name/distribution and version: Artix Linux on WSL
- [ ] bash version: `5.1.8(1)-release`
- [ ] bash-completion version: `2.11` (md5sum completions/make = `84866af8ff59c9e1698f90a8f3a67f22`)

## Debug trace

```shell
+++ _on_command_start
[?2004h[38;5;214mᛋ [97;1mmake abc/+ local cur prev words cword split
+ _init_completion -s
+ local exclude= flag outx errx inx OPTIND=1
+ getopts n:e:o:i:s flag -s
+ case $flag in
+ split=false
+ exclude+==
+ getopts n:e:o:i:s flag -s
+ COMPREPLY=()
+ local 'redir=@(?([0-9])<|?([0-9&])>?(>)|>&)'
+ _get_comp_words_by_ref -n '=<>&' cur prev words cword
+ local exclude flag i OPTIND=1
+ words=()
+ local cur cword words
+ upargs=()
+ upvars=()
+ local upargs upvars vcur vcword vprev vwords
+ getopts c:i:n:p:w: flag -n '=<>&' cur prev words cword
+ case $flag in
+ exclude='=<>&'
+ getopts c:i:n:p:w: flag -n '=<>&' cur prev words cword
+ [[ 6 -ge 3 ]]
+ case ${!OPTIND} in
+ vcur=cur
+ (( OPTIND += 1 ))
+ [[ 6 -ge 4 ]]
+ case ${!OPTIND} in
+ vprev=prev
+ (( OPTIND += 1 ))
+ [[ 6 -ge 5 ]]
+ case ${!OPTIND} in
+ vwords=words
+ (( OPTIND += 1 ))
+ [[ 6 -ge 6 ]]
+ case ${!OPTIND} in
+ vcword=cword
+ (( OPTIND += 1 ))
+ [[ 6 -ge 7 ]]
+ __get_cword_at_cursor_by_ref '=<>&' words cword cur
+ words=()
+ local cword words
+ __reassemble_comp_words_by_ref '=<>&' words cword
+ local exclude i j line ref
+ [[ -n =<>& ]]
+ exclude='[=<>&]'
+ printf -v cword %s 1
+ [[ -v exclude ]]
+ line='make abc/'
+ (( i = 0, j = 0 ))
+ (( i < 2 ))
+ [[ 0 -gt 0 ]]
+ ref='words[0]'
+ printf -v 'words[0]' %s make
+ line=' abc/'
+ (( i == COMP_CWORD ))
+ (( i++, j++ ))
+ (( i < 2 ))
+ [[ 1 -gt 0 ]]
+ [[ abc/ == +([=<>&]) ]]
+ ref='words[1]'
+ printf -v 'words[1]' %s abc/
+ line=
+ (( i == COMP_CWORD ))
+ printf -v cword %s 1
+ (( i++, j++ ))
+ (( i < 2 ))
+ (( i == COMP_CWORD ))
+ local i cur= index=9 'lead=make abc/'
+ [[ 9 -gt 0 ]]
+ [[ -n make abc/ ]]
+ [[ -n makeabc/ ]]
+ cur='make abc/'
+ (( i = 0 ))
+ (( i <= cword ))
+ [[ 9 -ge 4 ]]
+ [[ make != \m\a\k\e ]]
+ (( i < cword ))
+ local old_size=9
+ cur=' abc/'
+ local new_size=5
+ (( index -= old_size - new_size ))
+ (( ++i ))
+ (( i <= cword ))
+ [[ 5 -ge 4 ]]
+ [[  abc != \a\b\c\/ ]]
+ cur=abc/
+ (( index > 0 ))
+ (( index-- ))
+ [[ 4 -ge 4 ]]
+ [[ abc/ != \a\b\c\/ ]]
+ (( i < cword ))
+ (( ++i ))
+ (( i <= cword ))
+ [[ -n abc/ ]]
+ [[ ! -n abc/ ]]
+ (( index < 0 ))
+ local words cword cur
+ _upvars -a2 words make abc/ -v cword 1 -v cur abc/
+ (( 10 ))
+ (( 10 ))
+ case $1 in
+ [[ -n 2 ]]
+ printf %d 2
+ [[ -n words ]]
+ unset -v words
+ eval 'words=("${@:3:2}")'
++ words=("${@:3:2}")
+ shift 4
+ (( 6 ))
+ case $1 in
+ [[ -n cword ]]
+ unset -v cword
+ eval 'cword="$3"'
++ cword=1
+ shift 3
+ (( 3 ))
+ case $1 in
+ [[ -n cur ]]
+ unset -v cur
+ eval 'cur="$3"'
++ cur=abc/
+ shift 3
+ (( 0 ))
+ [[ -v vcur ]]
+ upvars+=("$vcur")
+ upargs+=(-v $vcur "$cur")
+ [[ -v vcword ]]
+ upvars+=("$vcword")
+ upargs+=(-v $vcword "$cword")
+ [[ -v vprev ]]
+ [[ 1 -ge 1 ]]
+ upvars+=("$vprev")
+ upargs+=(-v $vprev "${words[cword - 1]}")
+ [[ -v vwords ]]
+ upvars+=("$vwords")
+ upargs+=(-a${#words[@]} $vwords ${words+"${words[@]}"})
+ (( 4 ))
+ local cur cword prev words
+ _upvars -v cur abc/ -v cword 1 -v prev make -a2 words make abc/
+ (( 13 ))
+ (( 13 ))
+ case $1 in
+ [[ -n cur ]]
+ unset -v cur
+ eval 'cur="$3"'
++ cur=abc/
+ shift 3
+ (( 10 ))
+ case $1 in
+ [[ -n cword ]]
+ unset -v cword
+ eval 'cword="$3"'
++ cword=1
+ shift 3
+ (( 7 ))
+ case $1 in
+ [[ -n prev ]]
+ unset -v prev
+ eval 'prev="$3"'
++ prev=make
+ shift 3
+ (( 4 ))
+ case $1 in
+ [[ -n 2 ]]
+ printf %d 2
+ [[ -n words ]]
+ unset -v words
+ eval 'words=("${@:3:2}")'
++ words=("${@:3:2}")
+ shift 4
+ (( 0 ))
+ _variables
+ [[ abc/ =~ ^(\$(\{[!#]?)?)([A-Za-z0-9_]*)$ ]]
+ [[ abc/ =~ ^(\$\{[#!]?)([A-Za-z0-9_]*)\[([^]]*)$ ]]
+ [[ abc/ =~ ^\$\{[#!]?[A-Za-z0-9_]*\[.*]$ ]]
+ case ${prev-} in
+ return 1
+ [[ abc/ == @(?([0-9])<|?([0-9&])>?(>)|>&)* ]]
+ [[ make == @(?([0-9])<|?([0-9&])>?(>)|>&) ]]
+ local i skip
+ (( i = 1 ))
+ (( i < 2 ))
+ [[ abc/ == @(?([0-9])<|?([0-9&])>?(>)|>&)* ]]
+ (( i++ ))
+ (( 1 ))
+ (( i < 2 ))
+ (( cword <= 0 ))
+ prev=make
+ [[ -n false ]]
+ _split_longopt
+ [[ abc/ == --?*=* ]]
+ return 1
+ return 0
+ makef_dir=('-C' '.')
+ local makef makef_dir i
+ case $prev in
+ false
+ [[ abc/ == -* ]]
+ [[ abc/ == *=* ]]
+ (( i = 1 ))
+ (( i < 2 ))
+ [[ abc/ == -@(C|-directory) ]]
+ (( i++ ))
+ (( i < 2 ))
+ (( i = 1 ))
+ (( i < 2 ))
+ [[ abc/ == -@(f|-?(make)file) ]]
+ (( i++ ))
+ (( i < 2 ))
+ local mode=--
+ (( COMP_TYPE != 9 ))
+ mode=-d
++ _make_target_extract_script -d abc/
++ local mode=-d
++ shift
++ local prefix=abc/
+++ command sed 's/[][\,.*^$(){}?+|/]/\\&/g'
++ local 'prefix_pat=abc\/'
++ local basename=
++ local dirname_len=4
++ [[ -d == -d ]]
++ local 'output=\2'
++ cat
++ [[ -z abc\/ ]]
++ [[ abc\/ == */ ]]
++ cat
++ cat
+ local 'IFS= 	
' 'script=    1,/^# * Make data base/           d;        # skip any makefile output
    /^# * Finished Make data base/,/^# * Make data base/{
                                      d;        # skip any makefile output
    }
    /^# * Variables/,/^# * Files/     d;        # skip until files section
    /^# * Not a target/,/^$/          d;        # skip not target blocks
    /^abc\//,/^$/!            d;        # skip anything user dont want

    # The stuff above here describes lines that are not
    #  explicit targets or not targets other than special ones
    # The stuff below here decides whether an explicit target
    #  should be output.

    /^# * File is an intermediate prerequisite/ {
      s/^.*$//;x;                               # unhold target
      d;                                        # delete line
    }

    /^$/ {                                      # end of target block
      x;                                        # unhold target
      /^$/d;                                    # dont print blanks
      s|^\(.\{4\}\)\(.\{0\}[^:/]*/\{0,1\}\)[^:]*:.*$|\2|p;
      d;                                        # hide any bugs
    }

    # This pattern includes a literal tab character as \t is not a portable
    # representation and fails with BSD sed
    /^[^#	:%]\{1,\}:/ {         # found target block
      /^\.PHONY:/                 d;            # special target
      /^\.SUFFIXES:/              d;            # special target
      /^\.DEFAULT:/               d;            # special target
      /^\.PRECIOUS:/              d;            # special target
      /^\.INTERMEDIATE:/          d;            # special target
      /^\.SECONDARY:/             d;            # special target
      /^\.SECONDEXPANSION:/       d;            # special target
      /^\.DELETE_ON_ERROR:/       d;            # special target
      /^\.IGNORE:/                d;            # special target
      /^\.LOW_RESOLUTION_TIME:/   d;            # special target
      /^\.SILENT:/                d;            # special target
      /^\.EXPORT_ALL_VARIABLES:/  d;            # special target
      /^\.NOTPARALLEL:/           d;            # special target
      /^\.ONESHELL:/              d;            # special target
      /^\.POSIX:/                 d;            # special target
      /^\.NOEXPORT:/              d;            # special target
      /^\.MAKE:/                  d;            # special target
      /^abc\/[^a-zA-Z0-9]/d;            # convention for hidden tgt
      h;                                        # hold target
      d;                                        # delete line
    }'
+ COMPREPLY=($(LC_ALL=C             $1 -npq __BASH_MAKE_COMPLETION__=1             ${makef+"${makef[@]}"} "${makef_dir[@]}" .DEFAULT 2>/dev/null |
            command sed -ne "$script"))
++ LC_ALL=C
++ make -npq __BASH_MAKE_COMPLETION__=1 -C . .DEFAULT
++ command sed -ne '    1,/^# * Make data base/           d;        # skip any makefile output
    /^# * Finished Make data base/,/^# * Make data base/{
                                      d;        # skip any makefile output
    }
    /^# * Variables/,/^# * Files/     d;        # skip until files section
    /^# * Not a target/,/^$/          d;        # skip not target blocks
    /^abc\//,/^$/!            d;        # skip anything user dont want

    # The stuff above here describes lines that are not
    #  explicit targets or not targets other than special ones
    # The stuff below here decides whether an explicit target
    #  should be output.

    /^# * File is an intermediate prerequisite/ {
      s/^.*$//;x;                               # unhold target
      d;                                        # delete line
    }

    /^$/ {                                      # end of target block
      x;                                        # unhold target
      /^$/d;                                    # dont print blanks
      s|^\(.\{4\}\)\(.\{0\}[^:/]*/\{0,1\}\)[^:]*:.*$|\2|p;
      d;                                        # hide any bugs
    }

    # This pattern includes a literal tab character as \t is not a portable
    # representation and fails with BSD sed
    /^[^#	:%]\{1,\}:/ {         # found target block
      /^\.PHONY:/                 d;            # special target
      /^\.SUFFIXES:/              d;            # special target
      /^\.DEFAULT:/               d;            # special target
      /^\.PRECIOUS:/              d;            # special target
      /^\.INTERMEDIATE:/          d;            # special target
      /^\.SECONDARY:/             d;            # special target
      /^\.SECONDEXPANSION:/       d;            # special target
      /^\.DELETE_ON_ERROR:/       d;            # special target
      /^\.IGNORE:/                d;            # special target
      /^\.LOW_RESOLUTION_TIME:/   d;            # special target
      /^\.SILENT:/                d;            # special target
      /^\.EXPORT_ALL_VARIABLES:/  d;            # special target
      /^\.NOTPARALLEL:/           d;            # special target
      /^\.ONESHELL:/              d;            # special target
      /^\.POSIX:/                 d;            # special target
      /^\.NOEXPORT:/              d;            # special target
      /^\.MAKE:/                  d;            # special target
      /^abc\/[^a-zA-Z0-9]/d;            # convention for hidden tgt
      h;                                        # hold target
      d;                                        # delete line
    }'
+ [[ -d != -d ]]
xyz 
[C[C[Kset +x && exec 2>/dev/tty
[?2004l
++ _on_command_start
+ set +x
```

akinomyoga:
Resolve #544 

- Commit 53060e3b: support `@pytest.mark.bashcomp(required_cmd=True)` to skip all the tests in the class.
- Commit eb12832c: add failing test cases to illustrate the problems
- Commit 0ddae4c1: do not strip directory names when `COMP_TYPE=37`. This happens when the user presses <kbd>TAB</kbd> under the readline setting `bind 'set show-all-if-ambiguous on'` is set.
- Commit d5181c55: complete subdirectories as far as the path is unique

## Test suite problem on COMP_TYPE

The test `test_github_issue_544_4` (`test_make.py`) currently fails because of the design of the test suite. The test suite extracts the generated completions by changing the readline settings as

```diff
@@ -22,20 +22,20 @@
 set menu-complete-display-prefix off
 set meta-flag on
 set output-meta on
-set page-completions on
+set page-completions off
 set prefer-visible-bell on
 set print-completions-horizontally off
 set revert-all-at-newline off
-set show-all-if-ambiguous off
+set show-all-if-ambiguous on
 set show-all-if-unmodified off
 set show-mode-in-prompt off
 set skip-completed-text off
 set visible-stats off
-set bell-style audible
+set bell-style none
 set comment-begin #
-set completion-display-width -1
+set completion-display-width 0
 set completion-prefix-display-length 0
-set completion-query-items 100
+set completion-query-items 0
 set editing-mode emacs
 set emacs-mode-string @
 set history-size 100000
```

Under these settings, the completion functions are called with `COMP_TYPE=33` (`!`, show list), which is different from the usual `COMP_TYPE=9` (TAB) or `COMP_TYPE=37` (`%`, menu-complete). The completion for `make` changes the behavior depending on `COMP_TYPE` so that it doesn't produce the expected results in the test suite. Note that `make` completion is the only one that changes its behavior depending on the value of `COMP_TYPE` among the completions in `bash-completion`.

@scop Do you have any thoughts on how we should solve this problem?
- Option 1: Use another approach to get the generated completions if any? I'm not sure if there is any.
- Option 2: Make the completion results independent of `COMP_TYPE`?
- Option 3: Add a special code in `completions/make` that detects the test suite and changes the behavior only in the test suite?
- Option X: any other idea?

## Possibility of using awk

`sed` has been used to extract the target names by `make`. It is difficult to use `sed` to test whether the filenames in subdirectories are unique and to generate the full path only when it is unambiguous. In commit d5181c5, `sed` scripts has been changed so as to always produce the full paths, and instead, the generated results are afterward filtered using Bash code. But, maybe another approach is to rewrite everything using `awk`. It should be more efficient than processing in Bash.

### DIFF
--- a/completions/make
+++ b/completions/make
@@ -46,7 +46,7 @@ _make_target_extract_script()
     /^$/ {                                      # end of target block
       x;                                        # unhold target
       /^$/d;                                    # dont print blanks
-      s|^${dirname_re-}\(.\{${#basename}\}[^:/]*/\{0,1\}\)[^:]*:.*$|${output}|p;
+      s|^${dirname_re-}\(.\{${#basename}\}[^:]*\):.*$|${output}|p;
       d;                                        # hide any bugs
     }
 
@@ -85,6 +85,56 @@ EOF
     }
 
 EOF
+}
+
+# Truncate the non-unique filepaths in COMPRELY to only generate unique
+# directories or files.  This function discards the files under subdirectories
+# unless the path is unique under each subdirectory and instead generate the
+# subdirectory path.  For example, when there are two candidates, "abc/def" and
+# "abc/xyz", we generate "abc/" instead of generating both candidates directly.
+# When there is only one candidate "abc/def", we generate the full path
+# "abc/def".
+#
+# @var[in] cur
+# @var[in] mode
+# @var[in,out] COMPREPLY
+_comp_make__truncate_non_unique_paths()
+{
+    local prefix=$cur
+    [[ $mode == -d ]] && prefix=
+    if ((${#COMPREPLY[@]} > 0)); then
+        # collect the possible completions including the directory names in
+        # `paths' and count the number of children of each subdirectory in
+        # `nchild'.
+        local -A paths nchild
+        local target
+        for target in "${COMPREPLY[@]}"; do
+            local path=${target%/}
+            while [[ ! ${paths[$path]+set} ]] &&
+                paths[$path]=1 &&
+                [[ $path == "$prefix"*/* ]]; do
+                path=${path%/*}
+                nchild[$path]=$((${nchild[$path]-0} + 1))
+            done
+        done
+
+        COMPREPLY=()
+        local nreply=0
+        for target in "${!paths[@]}"; do
+            # generate only the paths that do not have a unique child and whose
+            # all parent and ancestor directories have a unique child.
+            ((${nchild[$target]-0} == 1)) && continue
+            local path=$target
+            while [[ $path == "$prefix"*/* ]]; do
+                path=${path%/*}
+                ((${nchild[$path]-0} == 1)) || continue 2
+            done
+
+            # suffix `/' when the target path is a subdiretory, which has
+            # at least one child.
+            COMPREPLY[nreply++]=$target${nchild[$target]+/}
+        done
+    fi
 }
 
 _make()
@@ -171,6 +221,8 @@ _make()
             $1 -npq __BASH_MAKE_COMPLETION__=1 \
             ${makef+"${makef[@]}"} "${makef_dir[@]}" .DEFAULT 2>/dev/null |
             command sed -ne "$script"))
+
+        _comp_make__truncate_non_unique_paths
 
         if [[ $mode != -d ]]; then
             # Completion will occur if there is only one suggestion
--- a/completions/make
+++ b/completions/make
@@ -162,7 +162,7 @@ _make()
         # recognise that possible completions are only going to be displayed
         # so only the base name is shown
         local mode=--
-        if ((COMP_TYPE != 9)); then
+        if ((COMP_TYPE != 9 && COMP_TYPE != 37 && COMP_TYPE != 42)); then
             mode=-d # display-only mode
         fi
 
--- a/completions/make
+++ b/completions/make
@@ -209,12 +209,19 @@ _make()
             fi
         done
 
-        # recognise that possible completions are only going to be displayed
-        # so only the base name is shown
+        # recognise that possible completions are only going to be displayed so
+        # only the base name is shown.
+        #
+        # Note: This is currently turned off because the test suite of
+        # bash-completion conflicts with it; it uses "set show-all-if-ambiguous
+        # on" (causing COMP_TYPE == 37) to retrieve the action completion
+        # results, and also the compact form with only the basenames is not
+        # essentially needed.  To re-enable it, please uncomment the following
+        # if-statement.
         local mode=--
-        if ((COMP_TYPE != 9 && COMP_TYPE != 37 && COMP_TYPE != 42)); then
-            mode=-d # display-only mode
-        fi
+        # if ((COMP_TYPE != 9 && COMP_TYPE != 37 && COMP_TYPE != 42)); then
+        #     mode=-d # display-only mode
+        # fi
 
         local IFS=$' \t\n' script=$(_make_target_extract_script $mode "$cur")
         COMPREPLY=($(LC_ALL=C \
--- a/test/t/test_make.py
+++ b/test/t/test_make.py
@@ -2,6 +2,8 @@ import os
 
 import pytest
 
+from conftest import assert_complete
+
 
 class TestMake:
     @pytest.mark.complete("make -f Ma", cwd="make")
@@ -45,3 +47,38 @@ class TestMake:
     @pytest.mark.complete("make -", require_cmd=True)
     def test_9(self, completion):
         assert completion
+
+
+@pytest.mark.bashcomp(require_cmd=True, cwd="make/test2")
+class TestMake2:
+    def test_github_issue_544_1(self, bash):
+        completion = assert_complete(bash, "make ab")
+        assert completion == "c/xyz"
+
+    def test_github_issue_544_2(self, bash):
+        completion = assert_complete(bash, "make 1")
+        assert completion == "23/"
+
+    def test_github_issue_544_3(self, bash):
+        completion = assert_complete(bash, "make 123/")
+        assert completion == ["xaa", "xbb"]
+
+    def test_github_issue_544_4(self, bash):
+        completion = assert_complete(bash, "make 123/xa")
+        assert completion == "a"
+
+    def test_subdir_1(self, bash):
+        completion = assert_complete(bash, "make sub1")
+        assert completion == "test/bar/"
+
+    def test_subdir_2(self, bash):
+        completion = assert_complete(bash, "make sub2")
+        assert completion == "test/bar/alpha"
+
+    def test_subdir_3(self, bash):
+        completion = assert_complete(bash, "make sub3")
+        assert completion == "test/"
+
+    def test_subdir_4(self, bash):
+        completion = assert_complete(bash, "make sub4")
+        assert completion == "sub4test/bar/ sub4test2/foo/gamma".split()
--- a/test/t/test_make.py
+++ b/test/t/test_make.py
@@ -18,7 +18,7 @@ class TestMake:
 
     @pytest.mark.complete("make .cache/", cwd="make", require_cmd=True)
     def test_3(self, bash, completion):
-        assert completion == "1 2".split()
+        assert completion == ".cache/1 .cache/2".split()
         os.remove(f"{bash.cwd}/make/extra_makefile")
 
     @pytest.mark.complete("make ", cwd="shared/empty_dir")
@@ -36,7 +36,7 @@ class TestMake:
 
     @pytest.mark.complete("make .cache/.", cwd="make", require_cmd=True)
     def test_7(self, bash, completion):
-        assert completion == ".1 .2".split()
+        assert completion == ".cache/.1 .cache/.2".split()
         os.remove(f"{bash.cwd}/make/extra_makefile")
 
     @pytest.mark.complete("make -C make ", require_cmd=True)
@@ -61,7 +61,7 @@ class TestMake2:
 
     def test_github_issue_544_3(self, bash):
         completion = assert_complete(bash, "make 123/")
-        assert completion == ["xaa", "xbb"]
+        assert completion == ["123/xaa", "123/xbb"]
 
     def test_github_issue_544_4(self, bash):
         completion = assert_complete(bash, "make 123/xa")
--- a/test/fixtures/make/test2/Makefile
+++ b/test/fixtures/make/test2/Makefile
@@ -1,0 +1,23 @@
+# makefile
+
+all: abc/xyz
+.PHONY: abc/xyz
+abc/xyz 123/xaa 123/xbb:
+	mkdir -p $(@:/%=)
+	date > $@
+
+sub1test/bar/alpha sub1test/bar/beta:
+	mkdir -p $(@:/%=)
+	date > $@
+
+sub2test/bar/alpha:
+	mkdir -p $(@:/%=)
+	date > $@
+
+sub3test/bar/alpha sub3test/foo/alpha:
+	mkdir -p $(@:/%=)
+	date > $@
+
+sub4test/bar/alpha sub4test/bar/beta sub4test2/foo/gamma:
+	mkdir -p $(@:/%=)
+	date > $@


test/t/conftest.py
```diff
@@ -305,6 +305,9 @@ def bash(request) -> pexpect.spawn:
             )
             if match:
                 cmd = match.group(1)
+        if marker and marker.kwargs and marker.kwargs.get("require_cmd", False):
+            if not is_bash_type(bash, cmd):
+                pytest.skip("Command not found")
```
scop, line 310: This looks like it's useful no matter what happens to the rest of this PR, so it would have been better to place it in a PR of its own, for cleanliness and if we happen to need to revert the rest for some reason later. But I don't mind keeping it in here this time.

test/t/test_make.py
```diff
@@ -45,3 +47,32 @@ def test_8(self, bash, completion):
     @pytest.mark.complete("make -", require_cmd=True)
     def test_9(self, completion):
         assert completion
+
+
+@pytest.mark.bashcomp(require_cmd=True, temp_cwd=True)
+class TestMake2:
+    def test_github_issue_544_1(self, bash):
+        print(os.getcwd())
+        print(bash.cwd)
```
scop, line 56: Debug remainders or something that should be removed, no?

test/t/test_make.py
```diff
@@ -45,3 +47,32 @@ def test_8(self, bash, completion):
...
+        with open("Makefile", "w") as f:
```
scop, line 57: It seems this could be placed in a static fixture dir of its own instead of generating on the fly in a temp cwd.

scop (changes requested):

scop:
Added a few rather superficial comments. The `make` completion continues to terrify me -- the code is quite complex, and my `make`-fu isn't really up to speed to anywhere near always tell how the completion should behave. So I'm not sure how much I can be of assistance here. Let's just keep adding test cases for all changes we make, like you've done here.

>     * Option 2: Make the completion results independent of `COMP_TYPE`?

This sounds good to me in principle, but I can't tell offhand what it would mean to user experience.

> But, maybe another approach is to rewrite everything using `awk`. It should be more efficient than processing in Bash.

Do we think that Makefiles can get sufficiently large to make efficiency here a issue we need to care much about? To me, code understandability and portability are more important unless there is an actual performance issue we do want to tackle, and in general bash wins over sed or awk on the understandabilty/portability front. But things can get hairy and overly verbose with bash, too, so this is something to look into case by case.

test/t/conftest.py
akinomyoga: OK. I'll separate the PR.

test/t/test_make.py
akinomyoga: Yes, these are debug remnants.

test/t/test_make.py
akinomyoga: Do you mean we add a Makefile at `/test/fixture/make/somewhere` in the tree?

scop: Yep, that's what I was thinking of. I think it's better to have static files there as is, and generate only when that's not feasible (e.g. dynamic content, chars that are problematic in some environments in file names).

akinomyoga:
> > ```
> > * Option 2: Make the completion results independent of `COMP_TYPE`?
> > ```
> 
> 
> This sounds good to me in principle, but I can't tell offhand what it would mean to user experience.

Removing the dependence on COMP_TYPE means that we effectively revert the feature introduced by 39f00f9. This feature strips the directory name in the list of completions displayed in the terminal. For example, when we try to complete `subdir/x` and there are matching make targets `subdir/x1`, `subdir/x2`, and `subdir/x3`, the displayed list becomes `x1 x2 x3` with this feature instead of `subdir/x1 subdir/x2 subdir/x3`. 

> > But, maybe another approach is to rewrite everything using `awk`. It should be more efficient than processing in Bash.
> 
> Do we think that Makefiles can get sufficiently large to make efficiency here a issue we need to care much about?

I think basically a handwritten Makefile shouldn't be so large (unless the user uses fancy GNU make macros to automatically generate a bunch of rules). However, the result of Makefile generators such as autotools and cmake can be extremely large.

> To me, code understandability and portability are more important unless there is an actual performance issue we do want to tackle, and in general bash wins over sed or awk on the understandabilty/portability front.

For understandability, I believe `awk` is better than the current `sed` or a Bash implementation. For the portability, `awk` may be harder than `sed` but is still manageable if we carefully write the awk script. I think the most challenging implementations for `awk`/`sed` among contemporary systems are Solaris `awk` and Solaris `sed`. At least, Solaris `awk` lacks many important features of POSIX `awk`. I actually have no experience with Solaris `sed` so am not even sure whether the POSIX `sed` scripts work as expected.

> But things can get hairy and overly verbose with bash, too, so this is something to look into case by case.

Maybe I should make a sample implementation by `awk` and we compare it with the current implementation?

test/t/test_make.py
akinomyoga: OK

scop:
> Maybe I should make a sample implementation by `awk` and we compare it with the current implementation?

Sure, if you're interested.

While at it, no matter the language, I think it would be cool if we could extract the whole implementation to a separately invoked static script in `helpers/` instead of generating it from bash code on the fly. That'd help with editor syntax highlighting and formatting, avoid some potentially hairy escaping issues, and likely make it easier to test.

akinomyoga:
I'm going to update this PR after another PR #548, which the new commits for this PR would depend on, is processed.

> > > * Option 2: Make the completion results independent of `COMP_TYPE`? (comment by @akinomyoga)
> > 
> > This sounds good to me in principle, but I can't tell offhand what it would mean to user experience. (comment by @scop)
> 
> Removing the dependence on COMP_TYPE means that we effectively revert the feature introduced by [39f00f9](https://github.com/scop/bash-completion/commit/39f00f92e52b783e7e9e43aac4b4274cc9dee152). (comment by @akinomyoga)

If you are fine with removing the COMP_TYPE-dependent feature 39f00f9, I can remove it. What do you think?

> > Maybe I should make a sample implementation by `awk` and we compare it with the current implementation?
> 
> Sure, if you're interested.

I would do it in another PR if I have time after this PR is settled.

test/t/conftest.py
akinomyoga: This was merged in PR #547.

test/t/test_make.py
akinomyoga: I have added a static file `/test/fixtures/make/test2/Makefile` in the tree.

test/t/test_make.py
akinomyoga: I have removed them, thank you.

akinomyoga:
> > Under these settings, the completion functions are called with `COMP_TYPE=33` (`!`, show list), which is different from the usual `COMP_TYPE=9` (TAB) or `COMP_TYPE=37` (`%`, menu-complete). The completion for `make` changes the behavior depending on `COMP_TYPE` so that it doesn't produce the expected results in the test suite. Note that `make` completion is the only one that changes its behavior depending on the value of `COMP_TYPE` among the completions in `bash-completion`.
> >
> > @scop Do you have any thoughts on how we should solve this problem?
> >
> > * Option 1: Use another approach to get the generated completions if any? I'm not sure if there is any.
> > * Option 2: Make the completion results independent of `COMP_TYPE`?
> > * Option 3: Add a special code in `completions/make` that detects the test suite and changes the behavior only in the test suite?
> > * Option X: any other idea?
>
> > ```
> > * Option 2: Make the completion results independent of `COMP_TYPE`?
> > ```
> 
> This sounds good to me in principle, but I can't tell offhand what it would mean to user experience.

I still hesitate to choose the solution for this issue. In the current test framework, the completions are extracted with `COMP_TYPE=33` (display candidates) which is different from the usual `COMP_REPLY=9` (tab-completion) and `COMP_REPLY=37` (menu-completion). Considering the reason that Bash provides the variable `COMP_REPLY`, I feel it is a too strong restriction to require not using `COMP_REPLY` in completion functions.

scop:
This is an area in which I have little experience, so unsure how much I can be of assistance. A couple of (likely stupid) comments/questions though:

> In the current test framework, the completions are extracted with COMP_TYPE=33 

Not sure I follow -- the only `git grep` hit for that I find is the one in the current `make` completion that test it's not equal to 9...?

> Isn't there any way to update the test framework so that it can handle different cases of COMP_REPLY?

Doesn't strike me as impossible off the bat, but nothing concrete to offer right now either.

akinomyoga:
The issue is reported again in #858. I think there is no reason to block the fixes just for the test suite problem. Here, I have rebased it on top of the latest master and, in commit d21ebba8, deactivated the *display-only mode* (39f00f92e52b783e7e9e43aac4b4274cc9dee152) causing the test-suite issue. If we would like to still use the "display-only mode", we can discuss how to make it work with the test framework in a separate issue, but I'm currently not sure if the "display-only mode" is worth re-enabling.

completions/make
```diff
@@ -157,19 +157,65 @@ _make()
...
+                        ((paths[\$path]++))
+                    fi
+                    [[ $path == "$prefix"*/* ]]
+                do
+                    path=${path%/*}
+                done
+            done
+
+            COMPREPLY=()
+            local nreply=0 target
```
scop, line 203: `target` was already defined above. 

```suggestion
            local nreply=0
```

completions/make
```diff
@@ -157,19 +157,65 @@ _make()
...
+        #     mode=-d # display-only mode
+        # fi
 
         local IFS=$' \t\n' script=$(_make_target_extract_script $mode "$cur")
         COMPREPLY=($(LC_ALL=C \
             $1 -npq __BASH_MAKE_COMPLETION__=1 \
             ${makef+"${makef[@]}"} "${makef_dir[@]}" .DEFAULT 2>/dev/null |
             command sed -ne "$script"))
 
+        local prefix=$cur
```
scop, line 180: Could we add some commentary here what the added hunk of code does, and why? I may be even slower today than usual, but it's not clear to me after eyeballing it for a while, let alone if revisiting afterwards without the context in the PR available to crosscheck.

scop (dismissed):

completions/make
akinomyoga: Thank you for the suggestion. I have added necessary code comments in commit 0ef11ca15b98abb47394f7c39208812239a6c379 (which was later squashed). I also noticed that this implementation does not work expected, so I modified the code too. The corresponding tests are added in commit 5df3eef90babd6b4553fa5c011bd1c3cd074d4dd (which was also squashed).

completions/make
akinomyoga: Thank you! I have removed it as a part of commit 0ef11ca15b98abb47394f7c39208812239a6c379.

ncfavier:
Can confirm this works fine. Not sure why we need all that code though, bash should be able to do the right thing with paths (`complete -o filenames`?)

ncfavier:
I am talking about processing, not generation. I guess what we're doing isn't even specific to paths. `complete -W 'test/abc test/xyz' foo` already seems to do the right thing, why do we need to reinvent anything?

akinomyoga:
OK, I see. Thank you for your clarification!

What we are doing here is not equivalent to `compgen -W 'test/abc test/xyz' -o filenames`. We reduce the list `test/abc test/xyz` to `test/`, which is not available by `-o filenames`. This is the behavior in the current `master`, though I'm not sure for the background of this behavior. This PR partially suppresses this reduction only in the unambiguous cases as requested in #544. If there is anything built in Bash that can be used to implement this conditional reduction in a simple way, I'm not reluctant to switch to it. Do you have any idea?

akinomyoga:
> OK, I see. Thank you for your clarification!
> 
> What we are doing here is not equivalent to `compgen -W 'test/abc test/xyz' -o filenames`. We reduce the list `test/abc test/xyz` to `test/`, which is not available by `-o filenames`. This is the behavior in the current `master`, though I'm not sure for the background of this behavior. This PR partially suppresses this reduction only in the unambiguous cases as requested in #544. If there is anything built in Bash that can be used to implement this conditional reduction in a simple way, I'm not reluctant to switch to it. Do you have any idea?
> 
> **P.S.**
> 
> > though I'm not sure for the background of this behavior.
> 
> It seems the reduction is introduced in commit [b28d710](https://github.com/scop/bash-completion/commit/b28d7108d3677c61bd01c51ccee8bb1cf9e3bfba). This seems to predate the time we moved to GitHub, so I don't know where to find the corresponding discussion.

ncfavier:
> If there is anything built in Bash that can be used to implement this conditional reduction in a simple way, I'm not reluctant to switch to it.

I mean, leaving aside `-o filenames`, if I just run `complete -W 'test/abc test/xyz' foo` then `foo <Tab>` completes to `foo test/`, and if I remove `test/abc` then it completes to `test/xyz`. Why do we need to do anything?

akinomyoga:
Ah, I see what you mean now. I have been considering the case where there are multiple directories and multiple files therein and we do not want to list the entire file tree just by `show-all-if-ambiguous` or by two consecutive TABs. For the normal completion with `complete -f`, we just obtain the list of filenames in the current directory or in the subdirectory already input, and the current implementation in the master and also this PR implements it.

But maybe the behavior you suggest makes sense.

ncfavier:
Ah, I see what you mean, we're reducing the list of options shown. That makes sense.

akinomyoga:
It's hard to decide for me. This is related to the background that introduced [this line](https://github.com/scop/bash-completion/commit/b28d7108d3677c61bd01c51ccee8bb1cf9e3bfba#diff-1776d90047a6c8d16281efc474f61be0173178c5564bbeddf154f7c4ceb3b07aR24) ten years ago. This is just my guess, but people might have wanted to mimic the behavior of Bash for normal paths.

ncfavier (approved):
Let's not delay this any longer

scop:
As noted, the make completion is already too complex for my taste and this makes it even more complex. Then again, we have the test cases here which appear to make sense as well as good amount of commentary, and no alternative implementations for taking care of the issue, so leaving it up to @akinomyoga to merge if he's comfortable with what we have here.

akinomyoga:
@ncfavier @scop Thank you for your comments. After another look, I decided to separate the corresponding code in a new function (1ea3326e is the updated patch). I think I will merge this version after waiting for a week.

PR #547, opened 2021-06-30
test(conftest): support `@pytest.mark.bashcomp(required_cmd=True)`

akinomyoga:
Separate commit 53060e3 in #546

### DIFF
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -305,6 +305,13 @@ def bash(request) -> pexpect.spawn:
             )
             if match:
                 cmd = match.group(1)
+        if (
+            marker
+            and marker.kwargs
+            and marker.kwargs.get("require_cmd", False)
+        ):
+            if not is_bash_type(bash, cmd):
+                pytest.skip("Command not found")
 
         request.cls.cmd = cmd
 


scop:
Needs to be run through `black`, feel free to merge after fixing. Highly recommended to use `pre-commit` locally to avoid roundtrips like this.

akinomyoga:
Ah, OK. I've fixed it. I thought I have run `pre-commit` previously for #544, but it seems I have skipped it because #544 is still a draft PR.

akinomyoga:
The test has passed, so I merge this PR.

scop:
> I thought I have run `pre-commit` previously for #544, but it seems I have skipped it because #544 is still a draft PR.

I suggest installing pre-commit locally with `pre-commit install` so there is never need to remember to run it, it will just happen automatically when one tries to commit. `git commit -n` can be used to bypass it if it causes problems e.g. with one's workflow (intermediate WIP commits that should not be linted yet or so).

akinomyoga:
Yeah, I once installed the pre-commit hook after you have suggested it to me before. But after that, pre-commit interfered with `git rebase -i` and caused troubles.  Since I frequently do `git rebase`, I have uninstalled it and decided to run it manually.

scop:
FWIW I use pre-commit and rebase with and without `-i` a lot, and I don't think I've ever run into problems with that combo.

akinomyoga:
Oh, yeah. I didn't remember the exact situation that I met with `git rebase`, so I had actually re-installed the pre-commit hook again after my previous comment. Then, I remembered that it was related to misconfigured `mypy` tests. In my environment, `mypy` produces the following error, which prevent me of making commits touching any `*.py` files:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

No virtual environments found

```

When I first installed the hook, I first faced this issue in `git rebase` where I tried to modify an older commit and there are some newer commits touching `*.py` after the modified commit. Since you have mentioned that there is the same issue in your environment with `mypy` in https://github.com/scop/bash-completion/pull/526#issuecomment-846460351, I just gave up to try to fix this `mypy` issue. Do you have any idea on how to fix this problem?

scop:
Our dev env setup instructions are inadequate, indeed a python virtualenv must be set up in the bash-completion working dir, then the above should work. Basically the commands in [this `run` section](https://github.com/scop/bash-completion/blob/a00b719edbc500b7bec08a3f4b3edea7f6b89cad/.github/workflows/check.yaml#L16) need to be run locally, or replacing the first two with a functionally equivalent [pyenv](https://github.com/pyenv/pyenv/) and [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv/) setup (that's what I'm using), `venv-run` then picks the env up and things should work in principle. The other mypy related issue you mentioned is something different, this doesn't fix that one, but I don't think I'm hitting it every time.

akinomyoga:
OK thank you for your instruction! It now works correctly.

PR #548, opened 2021-06-30
test(conftest): make `@pytest.mark.bashcomp(cwd=path)` relative from `test/fixtures`

akinomyoga:
In commit 2ed0e858a, I have added `@pytest.mark.bashcomp(cwd=path)` along with `@pytest.mark.bashcomp(temp_cwd=True)`, but this has never been used. Now I'd like to use `@pytest.mark.bashcomp(cwd=path)` in `test_make.py` in the updated #546 to specify the directory that contains a static `Makefile` for testing #544. I noticed that it is unuseful that `cwd` is the relative path to the current working directory of pytest so would like to change to interpret it as the relative path to the `test/fixtures`.


akinomyoga:
This is blocking #546, so I'd like to have this PR reviewed first.

akinomyoga:
I have updated the commit message. @scop Could you check this PR? I need to have this PR finished before finalizing #546.

PR #555, opened 2021-07-05
test(conftest): check error messages in `assert_complete()`

akinomyoga:
- 2548107a I'd like to let `assert_complete()` check that no error messages are output while the completion. This would be used in the next PR for the fix of the `scp` completion.
- 1c354b3a It detected an error message in `test_2` (`test_convert.py`) that has been missed before.

### DIFF
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -765,6 +765,8 @@ def assert_complete(
                 r"^" + re.escape(MAGIC_MARK),
                 # 2: on same line, result in .match
                 r"^([^\r]+)%s$" % re.escape(MAGIC_MARK),
+                # 3: error messages
+                r"^([^\r].*)%s$" % re.escape(MAGIC_MARK),
                 pexpect.EOF,
                 pexpect.TIMEOUT,
             ]
@@ -777,6 +779,9 @@ def assert_complete(
         elif got == 2:
             output = bash.match.group(1)
             return CompletionResult(output)
+        elif got == 3:
+            output = bash.match.group(1)
+            raise AssertionError("Unexpected output: [%s]" % output)
         else:
             # TODO: warn about EOF/TIMEOUT?
             return CompletionResult()
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -755,12 +755,16 @@ def assert_complete(
         bash.send(cmd + "\t")
         # Sleep a bit if requested, to avoid `.*` matching too early
         time.sleep(kwargs.get("sleep_after_tab", 0))
-        bash.expect_exact(kwargs.get("rendered_cmd", cmd))
+        rendered_cmd = kwargs.get("rendered_cmd", cmd)
+        bash.expect_exact(rendered_cmd)
         bash.send(MAGIC_MARK)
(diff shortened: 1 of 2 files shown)


test/t/conftest.py
```diff
@@ -755,16 +755,22 @@ def assert_complete(
...
-        bash.expect_exact(kwargs.get("rendered_cmd", cmd))
+        rendered_cmd = kwargs.get("rendered_cmd", cmd)
+        bash.expect_exact(rendered_cmd)
         bash.send(MAGIC_MARK)
         got = bash.expect(
             [
                 # 0: multiple lines, result in .before
-                r"\r\n" + re.escape(PS1 + cmd) + ".*" + re.escape(MAGIC_MARK),
+                r"\r\n"
+                + re.escape(PS1 + rendered_cmd)
```
scop, line 765: This switches from expecting `cmd` to expecting `rendered_cmd` here, but the commit message doesn't mention it. It may very well be the right thing to do, but I think we should mention if/that the change is deliberate as now it's in a commit that only mentions checking error messages which doesn't explain it to the casual reader.

Could you add the elaboration on this? Also the explanations you're adding to PR descriptions for each commit are great, but they would be better placed in the actual commit message bodies.

akinomyoga: OK. I think I'll separate the commit and also add commit messages. Also, it seems this change found a few existing errors of `nounset`, so I'll add their fixes.

scop:
#558 should fix the unrelated xmlwf test suite failure.

completions/ssh
```diff
@@ -246,7 +246,7 @@ _ssh_identityfile()
     [[ -z $cur && -d ~/.ssh ]] && cur=~/.ssh/id
     _filedir
     if ((${#COMPREPLY[@]} > 0)); then
-        COMPREPLY=($(compgen -W '${COMPREPLY[@]}' \
+        COMPREPLY=($(compgen -W '${COMPREPLY[@]+"{COMPREPLY[@]}"}' \
```
scop, line 249: IIRC I've guarded against this in existing code by an explicit check on the number of elements in the accessed array, such as is visible in this particular change here, a few lines above. I don't have that strong opinions how to check it, but I think it would be good to be consistent.

But re this particular change, I think it's unnecessary, as the array access is within the said guarding `if` block already.

akinomyoga: You are right. I actually replaced some of these `-W '${COMPREPLY[@]}'` just by the regular expression I will again go through the changes to see if they are needed one by one.

Anyway, these changes are yet not the complete list for the `compgen -W '${arr[@]}'` pattern, and I locally have additional changes and still looking for remaining. It seems like this will be another big change, now I'm thinking of separating the PR.

akinomyoga: Oh, I have been misreading your comment. Now I'm going to rewrite them to use explicit checks.

akinomyoga: There are so many related fixes, so I decided to separate the PR for `-W '${COMPREPLY[@]}'` at #559. Now in this PR, I don't include the fixes for `-W '${COMPREPLY[@]}'`

PR #559, opened 2021-07-08
fix(misc): fix bugs and errors found by tests

akinomyoga:
Details are described in the commit messages.

- e7b81ae3c288814861109be16096024f21dc7b9a fix(_xinetd_services,aspell,dpkg,etc.): avoid unnecessary word splitting
- 9938f7e50cc7845dc0e3d337f70be727b81fa781 fix(_ltrim_colon_completions): do compaction of sparse `COMPREPLY`
- 231a29b09cf8ac19fbd6846d441583f28c364d7d fix(secret-tool): fix empty candidates for `secret-tool search ` options
- 149a13ae118ac8afb4c7331543d995d843025c2b fix(secret-tool): protect unset opts[...] from pathname expansions
- c6574f3dbf6508ce0339028472636ded8f94488f fix(nounset): fix nounset errors of `-W '${COMPRELY[@]}'`
- b56a094ee8cb8bedd1670df41926122ec4886bf1 fix(nounset): fix nounset errors of `-W '${arr[@]}'`
- 8302aa3dc381e081162be8ea149fa17c236b1c29 fix(tree): work around bash-4.2 bug of `[[ ${arr[*]} == *text* ]]`
- 93dfc1fdf7375d965f191eb24ec38d42737b78a6 fix(man): work around the case both `manpath` and `man -w` are missing

### DIFF
--- a/completions/man
+++ b/completions/man
@@ -58,7 +58,17 @@ _man()
     fi
 
     local manpath=$(manpath 2>/dev/null || command man -w 2>/dev/null)
-    [[ -z $manpath ]] && manpath="/usr/share/man:/usr/local/share/man"
+    if [[ -z $manpath ]]; then
+        # Note: Both "manpath" and "man -w" may be unavailable, in
+        # which case we determine the man paths based on the
+        # environment variable MANPATH.
+        manpath=:${MANPATH-}:
+        # Note: An empty path (represented by two consecutive colons
+        # or a preceding/trailing colon) represents the system man
+        # paths.
+        manpath=${manpath//::/':/usr/share/man:/usr/local/share/man:'}
+        manpath=${manpath:1:-1}
+    fi
 
     # determine manual section to search
     local sect
--- a/completions/man
+++ b/completions/man
@@ -84,8 +84,10 @@ _man()
         # weed out directory path names and paths to man pages
         COMPREPLY=(${COMPREPLY[@]##*/?(:)})
         # strip suffix from man pages
-        COMPREPLY=(${COMPREPLY[@]%$comprsuffix})
-        COMPREPLY=($(compgen -W '${COMPREPLY[@]%.*}' -- "${cur//\\\\/}"))
+        ((${#COMPREPLY[@]})) &&
+            COMPREPLY=(${COMPREPLY[@]%$comprsuffix})
+        ((${#COMPREPLY[@]})) &&
+            COMPREPLY=($(compgen -W '"${COMPREPLY[@]%.*}"' -- "${cur//\\\\/}"))
     fi
 
     # shellcheck disable=SC2053
--- a/bash_completion
+++ b/bash_completion
@@ -1233,7 +1233,7 @@ else
     # @param $1 if -s, don't try to avoid truncated command names
     _pnames()
     {
-        local -a procs
+        local -a procs=()
         if [[ ${1-} == -s ]]; then
             procs=($(command ps ax -o comm | command sed -e 1d))
         else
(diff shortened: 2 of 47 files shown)


akinomyoga:
@scop Thank you for reviewing and merging the PR #555! I have completed this PR, which solves the problems found by the new error detections in PR #555. The details of the fixes are explained in the commit message of the corresponding commits (I have updated the list in the top comment).

As this PR contains different kinds of fixes, I thought I could separate the PR, but I decided to keep it as a single PR for now. The reason is that I believe all the tests should be kept *green* not to miss the possible new bugs, but I'm afraid that the reviewing and merging of all the related commits are delayed if I separate the PR. In fact, 8302aa3dc381e081162be8ea149fa17c236b1c29 is the fix of the problem of the `tree` completion introduced in #566 which was overlooked because of the failing tests.

Nevertheless, I will separate the PR if necessary (for example, when we need further discussion on a particular commit, or when we think it is better to make it easy to later revert them separately). Please feel free to tell me that.

completions/man
```diff
@@ -58,7 +58,16 @@ _man()
...
 
     local manpath=$(manpath 2>/dev/null || command man -w 2>/dev/null)
-    [[ -z $manpath ]] && manpath="/usr/share/man:/usr/local/share/man"
+    if [[ -z $manpath ]]; then
+        # Note: Both "manpath" and "man -w" may be unavailable, in
+        # which case we determine the man paths based on the
+        # environment variable MANPATH.
+        manpath=:${MANPATH-}:
+        # Note: An empty path (represented by two consecutive colons or
+        # precedign/trailing colon) represents the system man paths.
```
scop, line 67: Typo: precedign -> preceding

scop (approved):
Very cool stuff. TIL a lot: didn't know `foo=(/some/path/*)` would Just Work with filenames containing spaces, didn't know/remember the behavior of `${#foo[*]}` for sparse arrays, didn't know `unset -v 'foo[$bar]'` would actuall expand `$bar`, didn't obviously know about the bash 4.2 bug affecting `tree`.

I pushed one cosmetic commit, added one trivial spelling suggestion, the only thing I'd like to see in addition to these is fixing up the commit messages of the two "fix nounset errors of" commits: they mention things like
> In this commit, we replace ${COMPRELY[@]} with
${arr[@]+"${arr[@]}"}.

...but we no longer do that, we do different kinds of empty/missing protections.

Please feel free to push after addressing these, thanks a lot!

completions/man
akinomyoga: Thank you!. I have fixed it.

akinomyoga:
> didn't obviously know about the bash 4.2 bug affecting `tree`.

Neither did I. I have just realized the bash-4.2 bug by investigating the cause of the test failure in the `centos7` environment.

> I pushed one cosmetic commit,

Thanks!

> the only thing I'd like to see in addition to these is fixing up the commit messages of the two "fix nounset errors of" commits: they mention things like
> 
> > In this commit, we replace `${COMPRELY[@]}` with `${arr[@]+"${arr[@]}"}`.

Oh, I forgot to update it. Thank you.

```diff
-we replace ${COMPRELY[@]} with ${arr[@]+"${arr[@]}"}.
+we check that COMPREPLY is non-empty before trying to call "compgen -W".
```

I also modified other commit messages.

```diff
-"arr8~8~=($(printf '%s ' <glob-pattern>))" are synonym of
+"arr=($(printf '%s ' <glob-pattern>))" are synonym of
```
They (`8~8~`) were stray characters produced by terminal escape sequence of some special keys.

```diff
-Even if there is a matching filename,
+Even if there is no matching filenames,
```
It was opposite.

I have applied suggested changes, rebased and force-pushed the new commits to the PR branch. 

> Please feel free to push after addressing these, thanks a lot!

Do you mean merging in the master by "push"?

scop:
> Do you mean merging in the master by "push"?

Oops, yes, exactly that. Thanks again :tada:

akinomyoga:
Thanks. I have merged it.

PR #560, opened 2021-07-22
curl: fix option listing in newer versions

algorythmic:
`_parse_help` only takes a single parameter (which is then splits) for the options that it invokes the tool with, so `curl` was being invoked with just `--help`, not with `--help all`.

### DIFF
--- a/completions/curl
+++ b/completions/curl
@@ -159,7 +159,7 @@ _comp_cmd_curl()
     esac
 
     if [[ $cur == -* ]]; then
-        COMPREPLY=($(compgen -W '$(_parse_help "$1" --help all)' -- "$cur"))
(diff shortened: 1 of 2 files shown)


algorythmic:
The test failure seems unrelated, is there something else I should do here?

danyspin97:
Hi! Any update on this PR?

akinomyoga:
I need this change before performing the interface change of `_parse_help`.

scop (approved):
Cheers, added the requested test case, I'm good with merging this if/when it passes CI.

akinomyoga:
Thank you for adding the test!

akinomyoga (approved):

PR #561, opened 2021-07-23
Add markdown completion for okular

SuibianP:


dreua:
Came here to add this just to see that it's already done and merged, great! :+1: 

All we need now to get this (and probably many other) improvement to the users would be a release. So please please, make a release and thanks you for all the work that goes into this :pray:

scop:
https://github.com/scop/bash-completion/issues/785

PR #562, opened 2021-07-31
feat(pylint): message ids completion

scop:
Complete message ids for `--fail-on`, `--help-msg`, `--enable`, and
`--disable`. We scrape the output of `--list-msgs-enabled` instead of
`--list-msgs`, because it outputs both enabled and disable message ids
and thus allows for one implementation for all cases, even though the
output is somewhat less parseable than that of the latter.

Curiously, `--list-msgs-enabled` also outputs some message ids that
`--list-msgs` does not.

### DIFF
--- a/completions/pylint
+++ b/completions/pylint
@@ -1,5 +1,17 @@
 # pylint(1) completion                                     -*- shell-script -*-
 
+_comp_cmd_pylint_message_ids()
+{
+    local filter=p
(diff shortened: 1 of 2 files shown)


scop:
Background about the pylint status on this at https://github.com/PyCQA/pylint/issues/4778, merged PR at https://github.com/PyCQA/pylint/pull/4793

PR #565, opened 2021-07-31
WIP feat(sqlite3): sqlar support

scop:
Support for https://www.sqlite.org/sqlar.html


scop:
WIP in particular is the options handling when working on an `.sqlar` file.

PR #566, opened 2021-07-31
feat(tree): new completion

scop:
Upstream doesn't seem that active, so I thought I'd just add it here, can maintain it.

### DIFF
--- a/completions/tree
+++ b/completions/tree
@@ -1,0 +1,39 @@
+# tree(1) completion                                       -*- shell-script -*-
+
+_comp_cmd_tree()
+{
+    local cur prev words cword
+    _init_completion || return
+
+    case $prev in
+        -!(-*)[LPIHT] | --filelimit | --timefmt | --help | --version)
+            return
+            ;;
+        --charset)
+            _xfunc iconv _iconv_charsets
+            return
+            ;;
+        -!(-*)o)
+            _filedir
+            return
+            ;;
+        --sort)
+            COMPREPLY=($(compgen -W "name version size mtime ctime" -- "$cur"))
+            return
+            ;;
+    esac
+
+    if [[ $cur == -* ]]; then
+        COMPREPLY=($(compgen -W '$(_parse_help "$1")' -- "$cur"))
+        return
+    fi
+
+    if [[ ${words[*]} == *\ --fromfile\ * ]]; then
+        _filedir
+    else
+        _filedir -d
+    fi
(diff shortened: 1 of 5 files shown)


completions/tree
```diff
@@ -0,0 +1,39 @@
...
+        --charset)
+            _xfunc iconv _iconv_charsets
+            return
+            ;;
+        -!(-*)o)
+            _filedir
+            return
+            ;;
+        --sort)
+            COMPREPLY=($(compgen -W "name version size mtime ctime" -- "$cur"))
```
akinomyoga, line 21: `man tree` says `--sort[=]type`. As far as I tried, `--sort=name`, etc. work as expected. The other long-name options also seem to accept `--xxxx=yyyy` (though `--timefmt=fmt` is silently ignored).

Another thing that I noticed while the trial is that If one specifies multiple options that take an option argument in one word,  the same number of following arguments are consumed as the option argument of the multiple options. For example, `tree -Lo 1 output.txt` is equivalent to `tree -L 1 -o output.txt`. But I'm not sure if it's worth supporting.

scop: Good catch, long option split thing addressed in f7d5e3fd49f1b71cdc5154c0db76cf55acb51576.

I think we do not need to take approval that seriously; for me it would work nicely if we just took at it as "I checked this out, and don't see anything wrong with it on surface/no blockers, and I'm finished taking that look". No need for it to be a stringent thorough test by someone intimately familiar with the command or everything in bash-completion.

Just having a pair of eyes have a glance on stuff is useful to spread knowledge what we're doing, that already has nice value. If there are no comments, it's not clear if someone has looked into it at all, or if a full review by them is in progress. As noted, explicit approval is a fine way to me to signify that someone did have a look, found no blockers, and is finished having that look.

(I'll try to get to at least some the open PR's and discussions later today. Thanks!)

scop: Re `tree -Lo 1 output.txt`, I think that's a generic feature we lack at the moment, and would be pretty laborous to tackle. I think the way we do handle the arg of the last of bundled short options is good enough, and if we want to improve on that, we should come up with a generic way to handle it.

scop: Still re approval, a thumbs up emoji could serve as a "light" approval, but the problem with that is that no notifications are sent about them, and one might want to express happiness in seeing something implemented with an emoji and not signify an approval about the implementation by it.

PR #567, opened 2021-08-04
Create pyenv

ajshastri:
Added bash completion for pyenv


scop:
Thanks, but pyenv already ships its completions, search for "completion" at https://github.com/pyenv/pyenv

Be also sure to read and follow the contributing guidelines here at https://github.com/scop/bash-completion/blob/master/CONTRIBUTING.md for any future contributions you might be considering.

Issue #569, opened 2021-08-09
Missing double quotes on command variables


cykerway:
## Describe the bug

Command name "${COMP_WORDS[0]}" can actually contain spaces. But some code is using this variable and/or its equivalents without double quotes and triggers word splitting.

## To reproduce

    $ _mycomp() { _command_offset 0; }
    $ complete -F _mycomp '"/tmp/aaa bbb"'
    $ "/tmp/aaa bbb" <Tab><Tab><Tab>
    $ complete -p aaa

## Expected behavior

Expects nothing but the last command shows:

    complete -F _minimal aaa

## Versions (please complete the following information)

- [ ] Operating system name/distribution and version:
- [ ] bash version, `echo "$BASH_VERSION"`:
- [2.11] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`:

PR #570, opened 2021-08-09
Add missing double quotes for command variables.

cykerway:
Some command variables may contain spaces but were used without double
quotes. This commit fixes this error by double quoting these variables.

Issue https://github.com/scop/bash-completion/issues/569.

### DIFF
--- a/test/t/test_sudo.py
+++ b/test/t/test_sudo.py
@@ -1,5 +1,6 @@
 import pytest
 
+from conftest import assert_bash_exec
 from conftest import assert_complete
 
 
@@ -81,3 +82,13 @@ class TestSudo:
         part, _ = part_full_group
         completion = assert_complete(bash, "sudo chown foo:bar:%s" % part)
         assert not completion
+
+    def test_12(self, bash):
+        completion = assert_complete(bash, 'sudo "/tmp/aaa bbb" ')
+        output = assert_bash_exec(
+            bash, "complete -p aaa || true", want_output=True
+        )
+        assert output.endswith(
+            "bash: complete: aaa: no completion specification"
+        )
+
--- a/test/t/test_sudo.py
+++ b/test/t/test_sudo.py
@@ -1,7 +1,6 @@
 import pytest
 
-from conftest import assert_bash_exec
-from conftest import assert_complete
+from conftest import assert_complete, assert_bash_exec
 
 
 class TestSudo:
@@ -84,11 +83,6 @@ class TestSudo:
         assert not completion
 
     def test_12(self, bash):
-        completion = assert_complete(bash, 'sudo "/tmp/aaa bbb" ')
-        output = assert_bash_exec(
-            bash, "complete -p aaa || true", want_output=True
-        )
-        assert output.endswith(
-            "bash: complete: aaa: no completion specification"
-        )
+        assert_complete(bash, 'sudo "/tmp/aaa bbb" ')
+        assert_bash_exec(bash, "! complete -p aaa", want_output=None)
 
--- a/bash_completion
+++ b/bash_completion
@@ -2069,7 +2069,7 @@ _command_offset()
                 done
             else
                 cspec=${cspec#complete}
-                cspec=${cspec%%$compcmd}
+                cspec=${cspec%%"$compcmd"}
                 COMPREPLY=($(eval compgen "$cspec" -- '$cur'))
             fi
         elif ((${#COMPREPLY[@]} == 0)); then
--- a/bash_completion
+++ b/bash_completion
@@ -2030,18 +2030,18 @@ _command_offset()
         COMPREPLY=($(compgen -d -c -- "$cur"))
     else
         local cmd=${COMP_WORDS[0]} compcmd=${COMP_WORDS[0]}
-        local cspec=$(complete -p $cmd 2>/dev/null)
+        local cspec=$(complete -p "$cmd" 2>/dev/null)
 
         # If we have no completion for $cmd yet, see if we have for basename
         if [[ ! $cspec && $cmd == */* ]]; then
-            cspec=$(complete -p ${cmd##*/} 2>/dev/null)
+            cspec=$(complete -p "${cmd##*/}" 2>/dev/null)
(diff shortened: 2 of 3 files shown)


akinomyoga (approved):

scop (changes requested):
Thanks! I took a brief look into this earlier and tried to get something reproduced in an even simpler scenario than #569, but failed if memory serves. Anyway I think we want some test cases for these in the test suite.

Also we need to remember to fix the commit message to conform to conventional commits (see CONTRIBUTING.md) so they make it into the automated release notes, but that can be done if we eventually squash commits on merge unless taken care of before that.

cykerway:
OK, I gave a simpler example in the added test case. You are free to squash the commit and edit its message. I don't have write access anyway. Maybe you can put a commit message like:

    fix(_command_offset): add missing double quotes

test/t/test_sudo.py
```diff
@@ -81,3 +82,13 @@ def test_11(self, bash, part_full_group):
         part, _ = part_full_group
         completion = assert_complete(bash, "sudo chown foo:bar:%s" % part)
         assert not completion
+
+    def test_12(self, bash):
+        completion = assert_complete(bash, 'sudo "/tmp/aaa bbb" ')
```
akinomyoga, line 87: The variable `completion` is unused.
```suggestion
        assert_complete(bash, 'sudo "/tmp/aaa bbb" ')
```

test/t/test_sudo.py
```diff
@@ -81,3 +82,13 @@ def test_11(self, bash, part_full_group):
...
+        output = assert_bash_exec(
+            bash, "complete -p aaa || true", want_output=True
+        )
+        assert output.endswith(
+            "bash: complete: aaa: no completion specification"
+        )
```
akinomyoga, line 93: This test depends on the details of the error messages that might change in future Bash. I feel that we can just test that `complete -p aaa` fails. We don't set `errexit` (`set -e`) while testing. I suggest something like the following simpler test though I didn't test it:

```suggestion
        assert_bash_exec(bash, "! complete -p aaa", want_output=None)
```

What do you think?

test/t/test_sudo.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 
+from conftest import assert_bash_exec
 from conftest import assert_complete
```
akinomyoga, line 4:
```suggestion
from conftest import assert_complete, assert_bash_exec
```

We have never put them in separate lines in other tests. It is good to keep it consistent.

akinomyoga (changes requested):

cykerway:
OK those changes look better and have been taken. Sorry I don't know how to sync github changes to git. I just squashed them then forced push. You can probably do another squash of these 4 commits.

akinomyoga (commented):
@cykerway Thank you for the quick updates! (@cdleonard sorry I have mistakenly made the mention on you. You can just ignore it. Sorry for the noise)

Another thing is that we may add direct tests for `_command_offset` just like the other utility functions defined in `bash_completion`. In the current PR, you have added the test for `sudo` which serves as an indirect test for `_command_offset`. In addition to this, I think we may also test the function directly in a new file e.g. `/test/t/unit/test_unit_command_offset.py`.

akinomyoga:
> OK those changes look better and have been taken. Sorry I don't know how to sync github changes to git. I just squashed them then forced push.

Didn't `git pull` (or `git pull origin gh-issue-569` if you haven't set the upstream branch) work? The changes applied on the GitHub page are applied to the PR branch as new commits. Then I think you can just pull them into your local branch.

cykerway:
> Another thing is that we may add direct tests for `_command_offset` just like the other utility functions defined in `bash_completion`. In the current PR, you have added the test for `sudo` which serves as an indirect test for `_command_offset`. In addition to this, I think we may also test the function directly in a new file e.g. `/test/t/unit/test_unit_command_offset.py`.

How can you run `_command_offset` outside a real bash completion? It gives me an error like:

    compopt: not currently executing completion function

akinomyoga:
Ah, OK. I think we should support something in the test framework. For the time being, for example, I think the tests may be surrounded by the following pair of commands (which replaces `compopt`):

```bash
assert_bash_exec(bash, "compopt() { true; }")
...
assert_bash_exec(bash, "unset -f compopt")
```

@scop What are your thoughts?

cykerway:
I prefer a simple wrapper than a mock. I added a unit test for `_command_offset` which employs a light wrapper function `_co1`, which basically just calls `_command_offset 1`. You are free to go the "mock" way. I didn't read every piece of code and don't know if mocking `compopt` would change other aspects of the function being tested.

akinomyoga (approved):

akinomyoga:
OK! Your solution looks cleaner!

scop:
LGTM too. I'll fix up the black, flake8, and isort failures before merging, but please set up pre-commit (see CONTRIBUTING.md) to flag these already before submitting for future contributions, TIA!

Re squashing, I meant the maintainers here would do the squash here eventually before the merge, which I'll take care of, too.

scop:
I got a permission denied when trying to push to your fork, so did the fixups locally and merged as ae1a164f5a7611b55a1395fafc06d319b4104761. Thanks!

cykerway:
Ah, I don't really know why or what permission is denied on the github web interface, but good to know this has be done from the command line. Wait, I learned only a few days ago that the right thing is [*doing merges from the command line, not using the entirely broken github web interface*](https://lore.kernel.org/lkml/CAHk-=wjbtip559HcMG9VQLGPmkurh5Kc50y5BceL8Q8=aL0H3Q@mail.gmail.com/) blabla

scop:
Yeah, I happened to read that too. Kind of agree, but then again doing it via the web interface is just so convenient that it's hard to beat. GitLab does it better than GH, FWIW.

Re the permission denied, I think there's a checkbox option when opening a PR that says something like "allow repository collaborators to push to this fork", dunno if that might have been off in this case. Or maybe something's just changed elsewhere. Anyway for some other PR's here and elsewhere I have been able to push to the submitter's repo earlier, this was the first time it didn't work.

PR #572, opened 2021-08-18
Minor cleanups

a1346054:


scop:
All changes look good to me on cursory review, but I think the commit messages could use improvements:

- "fix: use upstream license file": we should use "docs:", not "fix:" for this
- "fix: use consistent code-style throughout the whole file": we should use "style" instead of "fix" for this, mention _which_ file, and it'd be better to be more specific, e.g. "style(bash_completion): remove redundant quotes"
- "fix: use same codestyle everywhere": this should also be "style", talk about redundant quotes instead of vague "same codestyle", and it bundles a python3 change which isn't really a code style but a functional change, so it should be another commit altogether
- "fix whitespace" is much too generic, e.g. "style(*.md): trim excess whitespace" would be better. I also think this message should have not passed commitlint, be sure to set up pre-commit for your working tree, see end of CONTRIBUTING.md

I'll slice and dice these into order for you, please take them into account for future constributions you may make. Thanks!

a1346054:
Thanks for the review and explanation, I'll keep those points in mind for all my future contributions. Indeed those commit messages are much better than what I did.

PR #600, opened 2021-08-30
fix(_parse_help): protect word splitting from pathname expansions

akinomyoga:
This solves error messages on the standard option completions under `shopt -s failglob`. For example, with `shopt -s failglob`, we see the following error message:

```bash
$ shopt -failglob 
$ tar -[TAB]bash: no match: --occurrence[=NUMBER]
```

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -888,7 +888,10 @@ __parse_options()
 
     # Take first found long option, or first one (short) if not found.
     option=
+    local reset=$(shopt -po noglob)
+    set -o noglob
     local -a array=($1)
+    $reset
     for i in "${array[@]}"; do
         case "$i" in
             ---*) break ;;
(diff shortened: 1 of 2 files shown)


scop:
Hmm, I cannot reproduce.
```
$ shopt -s failglob
$ tar -[Tab]
-A  -c  -d  -r  -t  -u  -x  
```

bash 5.0 on Ubuntu 20.04.

akinomyoga:
Ah, I found out I had [a mistake in my `BASH_COMPLETION_USER_DIR`](https://github.com/akinomyoga/akinomyoga.dotfiles/commit/04225a223b1f46e2ebd2d31312fe35930c241b17) so that the older `tar` completion in `/usr/share/bash-completion/completions` provided by the distribution has been mistakenlly loaded. The error messages that I observed with `tar` were actually a different issue from the older `tar` completion. 

I actually showed the error message of `tar` as just an example, and there are other (really) affected completions. For another example, you can try the following completion:

```bash
$ bash --norc # inside bash-completion repository
$ shopt -s failglob
$ source bash_completion
$ source completions/xz
$ xz -[TAB]bash: no match: --files[=FILE]
```

scop:
The current way as done here seems fine to me. Thanks again!

akinomyoga:
OK, thanks!

PR #605, opened 2021-09-07
test(mypy): set mypy_path, to find conftest

scop:
Hopefully fixes problems running pre-commit.

### DIFF
--- a/test/setup.cfg
+++ b/test/setup.cfg
@@ -6,6 +6,7 @@ markers =
 
 [mypy]
 python_version = 3.6
+mypy_path = $MYPY_CONFIG_FILE_DIR/t
 show_error_codes = true
 
 [isort]


akinomyoga:
Thank you for the fix! Locally I have confirmed that this fix indeed solves the error! (Personally I also searched for how to fix it and found that `mypy_path` might be related but couldn't find where to put the setting.)

akinomyoga (approved):

scop:
Cool. This was quite confusing as I had various `mypy` executables around, tested with many, sometimes found that `mypy_path` worked (which was my initial suspect and I've done it in other projects, too), but sometimes it *didn't* work while but `MYPYPATH` in env did work (which I can't really explain), sometimes within venv, sometimes not, sometimes with incorrect configs, sometimes ran into some issues that lead me to believe that sometimes manual purging of `.mypy_cache` dirs is in order...

But I hope this is sorted now, also got to clean up my setup quite a bit :)

PR #614, opened 2021-09-30
insmod, modinfo, modprobe: support zstd compressed modules

hramrach:


scop:
Thanks! For future reference, be sure to check out CONTRIBUTING.md on how to set up pre-commit and have it check commit messages for conformance with conventional commits which we've started using here.

bitstreamout:
Is there any reasons why .xz and .bz2 are ignored?  Why is there only the half of the supported compression methods supported?

hramrach:
xz was added by a previous commit. Nobody cared about bz2 so it's not added.

scop:
Ref https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/tree/shared/util.c?id=3821e1971ec53fabbbb9f2679ea988ee12db61c8#n38

PR #628, opened 2021-10-13
feat: Per-user opt-out when installed system-wide

### LINKED ISSUE #627: Per-user opt-out when installed system-wide

Assuming the necessary version requirement checks succeed, the `bash_completion.sh.in` file (installed system-wide as `/etc/profile.d/bash_completion.sh`) unilaterally sources `/usr/share/bash-completion/bash_completion`.

However, some users on the system might not want to use bash-completion's completions (while others do want them, ie. uninstalling the bash-completion package is not a workaround).  The sourcing of `/etc/profile.d/*` happens from `/etc/profile`, which is the first file read by bash at startup, which means that users have no way to intercept, interrupt, or otherwise prevent the sourcing of `bash_completion`.  Additionally, the `bash-completion` script itself has no way for users to avoid obtaining the completions, it just starts setting up the completions straight away.

At least Ubuntu (but presumably most of distros) install `/etc/profile.d/bash_completion.sh` as-is from the upstream bash-completion project.  This means that this problem exists "out of the box" on any distro which doesn't take steps (ie. patch `bash_completion.sh` or `bash_completion`) to avoid it.

Based on [other comments](https://github.com/scop/bash-completion/issues/232#issuecomment-416032155), it seems that bash-completion's stance is that if the completions aren't wanted, then the user simply shouldn't source `bash_completion`.  However, this is at odds with the system-wide profile.d file that always sources `bash_completion`.

Therefore, one of `bash_completion.sh.in` or `bash_completion` should provide a mechanism for users to avoid obtaining bash-completion's completions.  This should preferably be done in `bash_completion.sh.in`, so that users can still obtain the completions later in a session by sourcing `bash_completion`.  For example, it could check for the existence of a file called `disable-systemwide-autoload` in the bash-completion user dir, and avoid sourcing `bash_completion` if it exists.  Another possibility would be to allow users to define `$BASH_COMPLETION_DISABLE_SYSTEMWIDE_AUTOLOAD` in their `~/.config/bash_completion` config file, and then add a check of this variable on line 10 of `bash_completion.sh.in`.  My guess is that the latter of these approaches is slightly preferable, so I'll submit a PR for it.

- [ ] This is a request for new completion
- [ ] Link to upstream project query about shipping the completion:

devkev:
If the `BASH_COMPLETION_DISABLE_SYSTEMWIDE_AUTOLOAD` env var is defined
and not null (eg. inside the `~/.config/bash_completion` config file),
then `/etc/profile.d/bash_completion.sh` will not source
`bash_completion`.

Fixes #627.

### DIFF
--- a/bash_completion.sh.in
+++ b/bash_completion.sh.in
@@ -7,7 +7,8 @@ if [ "x${BASH_VERSION-}" != x -a "x${PS1-}" != x -a "x${BASH_COMPLETION_VERSINFO
         [ "${BASH_VERSINFO[0]}" -eq 4 -a "${BASH_VERSINFO[1]}" -ge 2 ]; then
         [ -r "${XDG_CONFIG_HOME:-$HOME/.config}/bash_completion" ] &&
             . "${XDG_CONFIG_HOME:-$HOME/.config}/bash_completion"
-        if shopt -q progcomp && [ -r @datadir@/@PACKAGE@/bash_completion ]; then
+        if shopt -q progcomp && [ -r @datadir@/@PACKAGE@/bash_completion -a \
+            "x${BASH_COMPLETION_DISABLE_SYSTEMWIDE_AUTOLOAD-}" = x ]; then
             # Source completion code.
             . @datadir@/@PACKAGE@/bash_completion
         fi
--- a/doc/bash_completion.txt
+++ b/doc/bash_completion.txt
@@ -25,6 +25,10 @@ Environment variables
     loaded. If unset or null, the default compatibility directory to use is
     `/etc/bash_completion.d`.
 
+*BASH_COMPLETION_DISABLE_SYSTEMWIDE_AUTOLOAD*::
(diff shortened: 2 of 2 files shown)


devkev:
I tried to install and run the pre-commit hooks, but they failed to run because application-properties (required by pymarkdownlnt) requires python >= 3.8, but I'm on Ubuntu 18.04 which has a default python of 3.6.  I installed python 3.8, but it wasn't used by pre-commit and I couldn't figure out how to make it be used, and I'm not going to change my default python version in case it breaks other aspects of my system.  So, I disabled the hooks.  If there is a problem with the formatting (or otherwise), please feel free to adjust it as much as you'd like.

scop:
Closing due to explanation in #627.

But requiring Python 3.8 for development is not intentional ()requiring 3.6 is), thanks for bringing it up. I'll see if something could be done about that (documenting, if all else fails).

PR #634, opened 2021-10-19
Avoid nounset mode error with various completions

scop:


scop:
Noticed that too with a grep, but was too lazy to extend the scope of this PR, thanks for the nudge ;). Anyway apt-mark, avahi-browse, and curl addressed additionally now; feh, slapt-*, and xgamma have explicit guards one line above (although some of those are a bit dubious implementations). The _expand unit test seems different enough and would fail anyway if we didn't have the expected completion, so I'll leave that one as is.

PR #639, opened 2021-11-01
Support for setfacl

Roy-Orbison:
Could not find an existing completion script for filesystem acl modification and I don't see any completion code in [the upstream acl project](http://git.savannah.nongnu.org/cgit/acl.git/tree/).
Adapted from chown. Understands but does not suggest short options.

### DIFF
--- a/completions/setfacl
+++ b/completions/setfacl
@@ -1,0 +1,92 @@
+# setfacl(1) completion                                    -*- shell-script -*-
+
+_setfacl()
+{
+    local cur prev words cword split
+    # Don't treat user:group as separate words.
+    _init_completion -s -n : || return
+
+    if [[ $cword -ge 2 ]]; then
+        local word
+        for word in "${words[@]:1:$(( cword - 1 ))}"; do
+            if [[ "$word" == '--' ]]; then
+                _filedir
+                return 0
+            fi
+        done
+    fi
+
+    case "$prev" in
+        --restore)
+            _filedir
+            return 0
+            ;;
+
+        -m|--modify|-x|--remove|--set)
+            local other_rules prefix_default mycur
+            other_rules="${cur%"${cur##*,}"}"
+            cur="${cur##*,}"
+            if [[ $cur = d:* || $cur = default:* ]]; then
+                prefix_default=${cur%%"${cur#*:}"}
+                cur="${cur#*:}"
+            else
+                prefix_default=
+            fi
+            mycur="${cur##*:}"
+            if [[ $cur == *:*:*:* ]]; then
+                return
+            elif [[ $cur == *:*:* || $cur == [mo]*:* ]]; then
+                local opts
+                opts='- r w x X rw rx rX wx wX rwx rwX'
+                COMPREPLY=( $( compgen -S ',' -W "$opts" -- "$mycur" ) )
+            elif [[ $cur == u*:* ]]; then
+                local IFS
+                IFS=$'\n'
+                COMPREPLY=( $( compgen -u -S ':' -- "$mycur" ) )
+            elif [[ $cur == g*:* ]]; then
+                local IFS
+                IFS=$'\n'
+                COMPREPLY=( $( compgen -g -S ':' -- "$mycur" ) )
+            else
+                local prefixes
+                prefixes='user: group: mask: other:'
+                if [[ -z $prefix_default ]]; then
+                    prefixes+=' default:'
+                fi
+                COMPREPLY=( $( compgen -W "$prefixes" -- "$mycur" ) )
+            fi
+            if [[ $cur != *:* && -z $prefix_default && "$other_rules" && ${#COMPREPLY[@]} -eq 1 ]]; then
+                COMPREPLY=("${other_rules##*:}${COMPREPLY[0]}")
+            fi
+            [[ $COMPREPLY == *[:,] ]] && compopt -o nospace
+            return 0
+            ;;
+
+        -M|--modify-file|-X|--remove-file|--set-file)
+            _filedir
+            return 0
+            ;;
+
+    esac
+
+    $split && return 0
+
+    if [[ "$cur" == -* || $cword -eq 1 ]]; then
+        local opts
+        opts='--modify-file= --modify --remove-file= --remove --set-file= --set
+            --remove-all --remove-default --no-mask --mask --default --test
+            --recursive --logical --physical --'
+        if [[ ${#words[@]} -eq 2 ]]; then
+            opts+=' --version --help --restore='
+        elif [[ ${#words[@]} -eq 3 && "$prev" == --test ]]; then
+            opts+=' --restore='
+        fi
+        COMPREPLY=( $( compgen -W "$opts" -- "$cur" ) )
+        [[ $COMPREPLY == *= ]] && compopt -o nospace
+    else
+        _filedir
+    fi
+} &&
+complete -F _setfacl setfacl
+
+# ex: ts=4 sw=4 et filetype=sh


Roy-Orbison:
I've marked this as draft for now because I wasn't sure how to best account for groups that contain colons (as `_usergroup()` does), or whether that could be a TODO given completion covering common use cases is better than none.

scop:
Hm, we should add a template for pull requests so that the request to ask upstream to include the completion first is more prominent. Anyway, for time being, see https://raw.githubusercontent.com/scop/bash-completion/master/.github/ISSUE_TEMPLATE/feature_request.md and the start of CONTRIBUTING.md.

Roy-Orbison:
[I asked.](https://lists.nongnu.org/archive/html/acl-devel/2022-01/msg00000.html)

Roy-Orbison:
No response from them. Maybe you'd have more luck.

Roy-Orbison:
Still nothing. Did you have any luck?

Roy-Orbison:
I did hear back, but they never confirmed whether they even tried it, and said it had to be sent as a patch. I kinda gave up.

akinomyoga:
[Re: BASH completion (written but not reviewed)](https://lists.nongnu.org/archive/html/acl-devel/2022-12/msg00026.html)
> i don't have a prob including it with the project if you want to send it as
> a proper patch to the list
> -mike

This means that they *will review* if you submit it in a proper form.

Roy-Orbison:
@akinomyoga and I don't know what that proper form is, because I don't know where they would want that file in the tree, how it should be included/referenced, whether it needs a `make` action, etc. I don't write low-level C programs, so I'd probably just get it wrong.

It's not complicated to test; no compilation, no install, download and source a single script in bash. I took the response as ‘we'll look at this when you've done it the right way, but you haven't, and we won't tell you exactly what that is, so we won't look at it.’

akinomyoga:
The proper form is just the form of posting an email. There is a specific way of posting a patch in the mailing lists of traditional open-source projects, and there are reasons. However, you didn't follow it, which would cause problems. See other posts on the mailing list.

It's not related to how much the actual changes in the patch are correct or not. You should submit a patch with *the temporary location of the file* and wait for their comments. If they don't like the file location, they should tell you the preferred path.

akinomyoga:
> whether it needs a `make` action,

I'd say the `make` action should be included, or otherwise the file is likely to be used by no one. Anyway, you may ask it on the mailing list.

completions/setfacl
```diff
@@ -0,0 +1,92 @@
...
+                local prefixes
+                prefixes='user: group: mask: other:'
+                if [[ -z $prefix_default ]]; then
+                    prefixes+=' default:'
+                fi
+                COMPREPLY=( $( compgen -W "$prefixes" -- "$mycur" ) )
+            fi
+            if [[ $cur != *:* && -z $prefix_default && "$other_rules" && ${#COMPREPLY[@]} -eq 1 ]]; then
+                COMPREPLY=("${other_rules##*:}${COMPREPLY[0]}")
+            fi
```
akinomyoga, line 60: I haven't checked the actual behavior, but I guess this has the same problem as https://github.com/scop/bash-completion/pull/913#issuecomment-1490140309 when `bind 'set show-all-if-ambiguous on'` is set.

completions/setfacl
```diff
@@ -0,0 +1,92 @@
+# setfacl(1) completion                                    -*- shell-script -*-
+
+_setfacl()
```
akinomyoga, line 3:
```suggestion
_comp_cmd_setfacl()
```

completions/setfacl
```diff
@@ -0,0 +1,92 @@
...
+        elif [[ ${#words[@]} -eq 3 && "$prev" == --test ]]; then
+            opts+=' --restore='
+        fi
+        COMPREPLY=( $( compgen -W "$opts" -- "$cur" ) )
+        [[ $COMPREPLY == *= ]] && compopt -o nospace
+    else
+        _filedir
+    fi
+} &&
+complete -F _setfacl setfacl
```
akinomyoga, line 90:
```suggestion
complete -F _comp_cmd_setfacl setfacl
```

Roy-Orbison:
@akinomyoga Exactly my point. I don't know what `make` action to include for installation across different distros, so any patch I put forth would likely be unacceptable. Last time, it took _a year_ before I got any response. You're welcome to ask them if you like.

completions/setfacl
```diff
@@ -0,0 +1,92 @@
+# setfacl(1) completion                                    -*- shell-script -*-
+
+_setfacl()
+{
+    local cur prev words cword split
+    # Don't treat user:group as separate words.
+    _init_completion -s -n : || return
+
+    if [[ $cword -ge 2 ]]; then
```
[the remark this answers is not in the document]
Roy-Orbison:
```suggestion
    if (( cword >= 2 )); then
```

completions/setfacl
Roy-Orbison: This probably wouldn't have been picked up by the `setfacl` team, and you'd know better than I how to fix it.

akinomyoga:
> @akinomyoga Exactly my point. I don't know what `make` action to include for installation across different distros, so any patch I put forth would likely be unacceptable.

That's not the problem. The first patch doesn't need to be at an acceptable level. It is normal to revise the patch multiple times after discussing each version of the submitted patch.

It just needs to be *submitted* in a proper form. I repeat that the proper form is not about the content of the patch. You provided the change with *a link*. This is one of the most problematic ways of posting anything on the mailing list. Links are in general considered to become unavailable in the future. On the other hand, we often need to look back on the mailing list log to investigate the background of the codebase, so all the discussions need to be trackable, i.e., all the necessary information should be included in the emails themselves. Otherwise, they can't even start a discussion. You are blocking the discussion there.

akinomyoga:
Thanks for updates!

akinomyoga:
> Last time, it took _a year_ before I got any response. You're welcome to ask them if you like.

Honestly, this is the problem of the upstream acl project. It's not your fault. If you would like to communicate with them, you need to be patient. I'm not so enthusiastic to make it include the upstream.

completions/setfacl
akinomyoga: There is no robust way to make the displayed list compact. You'll need to always prefix the `other_rules` in the same way as [PR \#913](https://github.com/scop/bash-completion/pull/913/files#diff-a4757074ff650000804fd3eaabe9b0a9e02e33040ca5b8afd4c0275fc5f3e136R905-R908). For example,

```suggestion
            if [[ $cur != *:* && -z $prefix_default && "$other_rules" ]]; then
                local i
                for i in "${!COMPREPLY[@]}"; do
                    COMPREPLY[i]=${other_rules##*:}${COMPREPLY[i]}
                done
            fi
```

But I haven't tried to understand the part `[[ $cur != *:* && -z $prefix_default && "$other_rules" ]]`, so it might not result in an expected result. Could you explain the condition `[[ $cur != *:* && -z $prefix_default && "$other_rules" ]]`?

Roy-Orbison: Re #913:
> I'm not sure why the original only prepended the prefix back when there was only one completion, so I'm worried I'm missing something.

My understanding was that omitting the prefix when there are multiple candidates makes the list more concise, thus easier to visually scan & less likely to cause pagination. It also provides the ability to create richer results, as demonstrated in [this article](https://iridakos.com/programming/2018/03/01/bash-programmable-completion-tutorial#displaying-each-entry-in-a-new-line).

Without your suggestion the behaviour is like this:
```sh
roy@RoyBook:~$ setfacl --set user:w
whoopsie:  www-data:
roy@RoyBook:~$ setfacl --set user:whoopsie:rw
rw,   rwx,  rwX,
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:a
adm:            audio:          avahi:          avahi-autoipd:
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:adm:
-,    r,    rw,   rwx,  rwX,  rx,   rX,   w,    wx,   wX,   x,    X,
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:adm:rw,
default:  group:    mask:     other:    user:
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:adm:rw,other:
-,    r,    rw,   rwx,  rwX,  rx,   rX,   w,    wx,   wX,   x,    X,
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:adm:rw,other:r
```
With it, the previous definition's mode is prepended to the next definition, which is a drop in readability:
```sh
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,
rwX,default:  rwX,group:    rwX,mask:     rwX,other:    rwX,user:
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:a
adm:            audio:          avahi:          avahi-autoipd:
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:adm:
-,    r,    rw,   rwx,  rwX,  rx,   rX,   w,    wx,   wX,   x,    X,
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:adm:rw,
rw,default:  rw,group:    rw,mask:     rw,other:    rw,user:
```

I tried to replicate the prefix loss per the last snippet of [your comment](https://github.com/scop/bash-completion/pull/913#issuecomment-1490140309), but it did not seem to occur. Perhaps this function is not affected due to the slightly different behviour around colons?
```sh
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:avahi
avahi:          avahi-autoipd:
roy@RoyBook:~$ setfacl --set user:whoopsie:rwX,group:av<TAB>
```

PR #642, opened 2021-11-01
ulimit option arg improvements

scop:


completions/ulimit
```diff
@@ -32,10 +32,7 @@ _ulimit()
         fi
     fi
 
-    local args
-    _count_args
-    ((args == 1)) &&
-        COMPREPLY=($(compgen -W "soft hard unlimited" -- "$cur"))
+    [[ ${mode-} ]] && COMPREPLY=($(compgen -W "soft hard unlimited" -- "$cur"))
```
akinomyoga, line 35: When `shopt -s localvar_inherit` (bash 5.0) is enabled and there is a global variable `mode`, `local mode` initializes the variable `mode` with the value of the global `mode`, and thus `[[ ${mode-} ]]` will always success unexpectedly. I think `local mode` should be explicitly initialized with some value. the condition `[[ ! -v mode ]]` also needs to be updated.

scop: Ouch, TIL, thanks! Sounds like we're going to have a _lot_ of things in need of changing if we want to support `localvar_inherit` mode, basically pretty much all `local`s need to be reviewed, and on the other hand we no longer need to be defensive with `${foo-}` for our locals, then.

Would be great if shellcheck could flag these cases.

akinomyoga: Actually, I doubt if anyone uses the option `shopt -s localvar_inherit`. So, I believe it is not so urgent that we need to soon scan all the occurrence of `local`s. Nevertheless, we should gradually change it when noticed.

> Would be great if shellcheck could flag possibly affected cases.

I found an issue that mentions the related possible warning: https://github.com/koalaman/shellcheck/issues/1297, which implies that it is not yet implemented?

scop: Yeah, bash 5 has been out for a while and this is the first time I've heard of it, so in that sense it's not that urgent.

Gradual changing does work, but has the drawback that one needs to be more alert so that we aren't introducing more than if we'd have the entire codebase fixed in one fell swoop. For example copy-pasting some existing completion as a starting point of a new one can easily introduce more of this.

akinomyoga: > For example copy-pasting some existing completion as a starting point of a new one can easily introduce more of this.

I see.

scop: Copied the more general issue to #659. Because the existing `[[ ! -v mode ]]` check would already not do the right thing if `mode` was initialized from a set global, the introduced use of `${mode-}` is not really a complete new regression, and as such would IMHO be better fixed in a separate PR targeted to fix exactly that.

akinomyoga (approved):

PR #645, opened 2021-11-03
ci: test on fedoradev again

scop:
Now that we have the image again.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -84,9 +84,7 @@ jobs:
           - dist: debian10
             bsd: true
             network: none
-          # fedoradev disabled until we have the image,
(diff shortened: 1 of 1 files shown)


akinomyoga:
I actually don't know the backgrounds of disabling the Fedora image, but I believe it's good to merge it if it works correctly.

scop:
Ah yes, I've forgotten about this, thanks for reminding. It did work here, so let's just enable it and see if it needs further attention.

If my memory serves, the reason for disabling it was that it couldn't be built due to https://github.com/moby/moby/issues/42680 when we moved the images from Docker Hub to ghcr. But that's sorted out a while ago already.

akinomyoga:
I see! Thank you for your explanation.

Issue #647, opened 2021-11-07
Using -o option with ssh command seems to break known hosts completion


egamby:
## Describe the bug

When I want to auto-complete a ssh command with -o option, for instance `ssh -o ForwardX11=yes <tab>`, the list of known hosts is not provided anymore.

## To reproduce

1. Type `ssh -o ForwardX11=yes <tab>`
2. Auto-completion proposes all the possibilities: `Display all 2497 possibilities? (y or n)`

## Expected behavior

Completion should propose the list of known hosts.

## Versions

- [ ] Operating system name/distribution and version: Debian 11
- [ ] bash version, `echo "$BASH_VERSION"`: 5.1.4(1)-release
- [ ] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.11

## Debug trace
[debug_trace.txt](https://github.com/scop/bash-completion/files/7492460/debug_trace.txt)

scop:
Oh, good catch! Makes me wonder how we've missed this for this long :thinking:. Anyway #648 contains a candidate fix, but seems there are a few remaining things to iron out, see the PR.

egamby:
Thanks a lot for the quick fix. I confirm this solves the issue :-)

scop:
Let's not close this yet, #648 is not merged yet. It'll autoclose this when it is.

PR #648, opened 2021-11-07
fix(ssh): don't confuse arg counting with option args

scop:
Options that take an arg must not be treated as arg the _count_args way.

Closes https://github.com/scop/bash-completion/issues/647

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -266,6 +266,10 @@ _ssh()
     # Keep cases sorted the same they're in ssh's usage message
     # (but do group ones with same arg completion)
     case $prev in
+        -*B)
+            _available_interfaces -a
+            return
+            ;;
         -*b)
             _ip_addresses
             return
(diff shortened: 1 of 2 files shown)


scop:
Note that this change breaks `TestSsh::test_2`. But the behavior tested by that test seems broken to me: I don't know why we'd ever complete both known hosts and commands. If we already have a host, I don't think we should be completing another but just a command from there on...?

akinomyoga:
> If we already have a host, I don't think we should be completing another but just a command from there on...?

The test case `TestSsh::test_2` seems to test the completion for `ssh -F config ls`, in which a host is not specified. I think we should rather complete a hostname instead of commands.

scop:
Right, just hosts would be the right thing to do for that test case. The general fix would be to split [this else block](https://github.com/scop/bash-completion/blob/c12e4da3e0b4c0535746c0d2eeade26d8b44e679/completions/ssh#L334) to only ever complete hostnames _or_ commands, I think there is no case in which we should complete both.

akinomyoga:
Oh, I see the exact problem now. Thank you for your clarification.

scop:
6c564f4ac8e6c89f67596cc66a68710bd8e6982f is only semi related, but I opted to bundle it here in this PR because it does play a role now with regards to host vs commands completion, ditto options taking args in general. And 79cea404346f104476ec031997d67253d33ae3d4 is related to keeping those in sync, so in that sense also semi related but opted to bundle it here too.

akinomyoga (approved):
looks good

PR #651, opened 2021-11-12
fix ip netns list parsing

remijouannet:
…t/commit/?id=f19966efeedf5292f003aaccecc3ac3a8a069b28

    ip netns list output has change, it's now look like this
    ip netns list
    netns12 (id: 1)
    netns45 (id: 0)

### DIFF
--- a/completions/ip
+++ b/completions/ip
@@ -360,7 +360,9 @@ _ip()
                 delete | exec | pids | set)
                     [[ $prev == "$subcmd" ]] &&
                         COMPREPLY=($(compgen -W "$(
-                            $1 -c=never netns list 2>/dev/null || $1 netns list
+                            {
+                                $1 -c=never netns list 2>/dev/null || $1 netns list
+                            } | awk '{print $1}'
                         )" -- "$cur"))
                     ;;
                 *)


completions/ip
```diff
@@ -360,7 +360,7 @@ _ip()
                 delete | exec | pids | set)
                     [[ $prev == "$subcmd" ]] &&
                         COMPREPLY=($(compgen -W "$(
-                            $1 -c=never netns list 2>/dev/null || $1 netns list
+                            $1 -c=never netns list 2>/dev/null || $1 netns list | awk '{print $1}'
```
scop, line 363:
```suggestion
                            { $1 -c=never netns list 2>/dev/null || $1 netns list; } | awk '{print $1}'
```
We need the awk to affect output of both branches of the statement.

scop (changes requested):
Thanks for the contribution! One comment re the change below.

Other than that, the referenced iproute2 commit does not appear to be where the change happened, it just changed from unsigned to signed integer, the `(id: ...)` part was already there. If we want to refer to a commit here, we should look up the correct one.

Regarding the commit message here, please also observe our commit message formatting and be sure to use pre-commit locally to validate its format, search for gitlint and conventional commits in CONTRIBUTING.md.

remijouannet:
Hi, indeed the correct commit is [d182ee1307c7a83b581d8f6d473bbce2004420c0](https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/commit/?id=d182ee1307c7a83b581d8f6d473bbce2004420c0), i will rebase everything and change the commit message

remijouannet:
@scop does it look better ?

scop:
It does, however I think this commit still didn't go through pre-commit because it looks like its message would fail gitlint validation due to long lines in the message body part, as well as shfmt formatting.

"include netns" in the commit message could be replaced with "includes netnsid if set" to make it more accurate.

If your branch allows me to push, I could fix these up for you this time in case you don't feel like setting up pre-commit or fixing the expectations manually, let me know.

remijouannet:
i'm not gonna say no, i tried to install your pre-commit configuration on my laptop but that was a lot of tool and it fail for me, i sended you have an invite to push on my branch, thx for your patience

scop:
Done. It would be interesting to hear how installing the tools failed for you -- even though there is a bunch, all that should be needed is install of pre-commit (which requires python3), and docker. Everything else should be taken care of pre-commit behind the scenes.

PR #653, opened 2021-11-18
Handle compspec changes during meta-command completion.

potamides:
This pull request addresses a problem that occurs when `_command_offset` is being executed (e.g. when using sudo), and a completion function exists already. When this function signals that completion should be retried (changes compspec and exits with status 124), completion is not attempted again.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2096,6 +2096,10 @@ _command_offset()
 
                 # restart completion (once) if function exited with 124
                 if (($? == 124 && retry_count++ == 0)); then
+                    # Note: When the completion function returns 124, the state
+                    # of COMPREPLY is discarded.
+                    COMPREPLY=()
+
                     cspec=$(complete -p "$compcmd" 2>/dev/null)
 
                     # Note: When completion spec is removed after 124, we do
--- a/bash_completion
+++ b/bash_completion
@@ -2073,48 +2073,55 @@ _command_offset()
             cspec=$(complete -p "$compcmd" 2>/dev/null)
         fi
 
-        if [[ -n $cspec ]]; then
-            for ((i = 0; i <= 1; i++)); do
-                if [[ ${cspec#* -F } != "$cspec" ]]; then
-                    # complete -F <function>
-
-                    # get function name
-                    local func=${cspec#*-F }
-                    func=${func%% *}
-
-                    if ((${#COMP_WORDS[@]} >= 2)); then
-                        $func "$cmd" "${COMP_WORDS[-1]}" "${COMP_WORDS[-2]}"
-                    else
-                        $func "$cmd" "${COMP_WORDS[-1]}"
-                    fi
-
-                    # restart completion (once) if function exited with 124
-                    if (($? == 124 && i == 0)); then
-                        cspec=$(complete -p "$compcmd" 2>/dev/null)
-                        [[ -n $cspec ]] && continue
-                    fi
-
-                    # restore initial compopts
-                    local opt
-                    while [[ $cspec == *" -o "* ]]; do
-                        # FIXME: should we take "+o opt" into account?
-                        cspec=${cspec#*-o }
-                        opt=${cspec%% *}
-                        compopt -o $opt
-                        cspec=${cspec#"$opt"}
-                    done
-                else
-                    cspec=${cspec#complete}
-                    cspec=${cspec%%"$compcmd"}
-                    COMPREPLY=($(eval compgen "$cspec" -- '$cur'))
+        local retry_count=0
+        while true; do # loop for the retry request by status 124
+            if [[ ! $cspec ]]; then
+                if ((${#COMPREPLY[@]} == 0)); then
+                    # XXX will probably never happen as long as completion loader loads
+                    #     *something* for every command thrown at it ($cspec != empty)
+                    _minimal
                 fi
-                break
-            done
-        elif ((${#COMPREPLY[@]} == 0)); then
-            # XXX will probably never happen as long as completion loader loads
-            #     *something* for every command thrown at it ($cspec != empty)
-            _minimal
-        fi
+            elif [[ $cspec == *' -F '* ]]; then
+                # complete -F <function>
+
+                # get function name
+                local func=${cspec#* -F }
+                func=${func%% *}
+
+                if ((${#COMP_WORDS[@]} >= 2)); then
+                    $func "$cmd" "${COMP_WORDS[-1]}" "${COMP_WORDS[-2]}"
+                else
+                    $func "$cmd" "${COMP_WORDS[-1]}"
+                fi
+
+                # restart completion (once) if function exited with 124
+                if (($? == 124 && retry_count++ == 0)); then
+                    cspec=$(complete -p "$compcmd" 2>/dev/null)
+
+                    # Note: When completion spec is removed after 124, we do
+                    # not generate any completions including the default ones.
+                    # This is the behavior of the original Bash progcomp.
+                    [[ -n $cspec ]] || break
+
+                    continue
+                fi
+
+                # restore initial compopts
+                local opt
+                while [[ $cspec == *" -o "* ]]; do
+                    # FIXME: should we take "+o opt" into account?
+                    cspec=${cspec#*-o }
+                    opt=${cspec%% *}
+                    compopt -o $opt
+                    cspec=${cspec#"$opt"}
+                done
+            else
+                cspec=${cspec#complete}
+                cspec=${cspec%%"$compcmd"}
+                COMPREPLY=($(eval compgen "$cspec" -- '$cur'))
+            fi
+            break
+        done
     fi
 }
 complete -F _command aoss command "do" else eval exec ltrace nice nohup padsp \
--- a/bash_completion
+++ b/bash_completion
@@ -2074,33 +2074,42 @@ _command_offset()
         fi
 
         if [[ -n $cspec ]]; then
-            if [[ ${cspec#* -F } != "$cspec" ]]; then
-                # complete -F <function>
+            for ((i = 0; i <= 1; i++)); do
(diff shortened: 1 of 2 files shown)


akinomyoga:
> This pull request addresses a problem that occurs when `_command_offset` is being executed (e.g. when using sudo), and a completion function exists already.

Could you describe what kind of problem occurs?

> When this function signals that completion should be retried (changes compspec and exits with status 124),

When the completion function requests a retry by the exit status 124, comspec is not necessarily changed. Instead, the function definition might be overwritten. For example, the following case is possible:

```bash
complete -F _test1 test1

# loader
_test1() { source somewhere/_test1_definition.sh && return 124; }
```

```bash
# somewhere/_test1_definition.sh
# real definition of the completion function
_test1() {
  COMPREPLY=(....)
}
```

Actually, in this case, I think one could call `_test1` at the end of the loader instead of returning 124, but returning 124 should also work anyway.

potamides:
>Could you describe what kind of problem occurs?

Sure.

Currently, when a completion function which modifies the compsec is executed during meta-command completion in `_command_offset`, the new compspec is not executed. To my understanding this doesn't follow the behavior of the shell.

As an example, this is a problem when using the completion functions of [fzf](https://github.com/junegunn/fzf). When fzf doesn't find its trigger string, it attempts to load the corresponding bash-completion function as a fallback and returns 124 to signal the shell to to retry completion. However, this fails when the command is completed from e.g. `sudo` or when using [complete-alias](https://github.com/cykerway/complete-alias/), since `_command_offset` doesn't have the same behavior as the shell for this and doesn't retry completion.

This pull request is my attempt to extend `_command_offset` for this use case. Though it seems you found it still differs from the shell behavior, unfortunately.

akinomyoga:
OK, thank you for your explanation. Anyway, it is true that this PR makes more cases work as expected.

> When fzf doesn't find its trigger string, it attempts to load the corresponding bash-completion function as a fallback and returns 124 to signal the shell to to retry completion.

I couldn't understand what `fzf` tries to do from this description because it shouldn't work. But after checking [the code](https://github.com/junegunn/fzf/blob/master/shell/completion.bash#L136-L160), I now find that `fzf` actually does something non-trivial than that. It first attempts to run the *original* completion function saved in the variable `__fzf_original_completion_CMDNAME`. If the *original* completion function has not yet saved, it lets `__load_completion` (of bash-completion) load the *original* completion, save the setting to the variable `__fzf_original_completion_CMDNAME`, and again overwrites the completion with the version of `fzf`. Finally it returns 124.

Then, actually your fix does not change any behavior specifically for the `fzf` case because `fzf` restores the comspec after it saves the original completion, i.e., *it doesn't change the comspec* finally.

> This pull request is my attempt to extend _command_offset for this use case. Though it seems you found it still differs from the shell behavior, unfortunately.

Why don't you remove the condition that the comspec is changed? If the completion function returns 124, I think it's sufficient condition that we attempt another completion.

akinomyoga (changes requested):
Another thing is that the original Bash attempts the additional completion by the exit status 124 *at most once*. This PR tries to loop forever as far as the completion function returns 124. It may cause problems with badly designed user completion settings. Could you update the PR?

potamides:
>Then, actually your fix does not change any behavior specifically for the fzf case because fzf restores the comspec after it saves the original completion, i.e., it doesn't change the comspec finally. I suggest testing your fix in the actual cases that you have raised.

Coincidentally, my fix does work for fzf since fzf adds a `-o nospace` to the compspec which changes things. But I agree with you that it would make more sense to remove this check.

akinomyoga:
> Coincidentally, my fix does work for fzf since fzf adds a `-o nospace` to the compspec which changes things.

Oh, I see. You are right.

bash_completion
```diff
@@ -2078,14 +2078,25 @@ _command_offset()
...
+                    func=${func%% *}
+
+                    if ((${#COMP_WORDS[@]} >= 2)); then
+                        $func "$cmd" "${COMP_WORDS[-1]}" "${COMP_WORDS[-2]}"
+                    else
+                        $func "$cmd" "${COMP_WORDS[-1]}"
+                    fi
+
+                    # restart completion (once) if function exited with 124
+                    if [[ $? == 124 ]]; then
```
akinomyoga, line 2093:
```suggestion
                    if (($? == 124)); then
```

bash_completion
```diff
@@ -2078,14 +2078,25 @@ _command_offset()
...
+                        newcspec=$(complete -p "$compcmd" 2>/dev/null)
+                        func=${newcspec#*-F }
```
akinomyoga, line 2095: - What happens when `-F func` is not included in the new completion spec? I think `complete --p ...` will be executed in the second loop.
- What happens when the completion spec is removed, i.e., `newcspec` becomes empty? `$cmd` will be executed unintendedly in the second loop.

I think you may also include the checks `[[ -n $cspec ]]` and `[[ ${cspec#* -F } != "$cspec" ]]` inside the loop.

bash_completion
```diff
@@ -2078,14 +2078,25 @@ _command_offset()
                 # complete -F <function>
 
                 # get function name
-                local func=${cspec#*-F }
-                func=${func%% *}
+                local func=${cspec#*-F } newcspec
```
akinomyoga, line 2081: In this PR, the new completion spec is assigned to `newcspec` and not reflected in `cspec`. However, it seems `cspec` is later tested for whether it contains `-o xxxx`. Do you think we should still perform this check on the older `cspec` and skip the check for the new completion spec? Maybe we can directly update `cspec`?

akinomyoga (commented):
Thank you for updating!

bash_completion
potamides: This required a little bit of digging to find out what this is required for. I found some info on [this](https://alioth-lists-archive.debian.net/pipermail/bash-completion-devel/2011-September/003906.html) mailing list. Seems like the old mount compspec relied on `-o dirname` if the completion function didn't generate matches which thus must also be set during meta-completion to work. I think running this for the new compspec only should be sufficient.

bash_completion
potamides: I think we could just run the existing code which processes the original compspec again to handle these edge cases. In my latest commit I do this by increasing the scope of the for-loop.

akinomyoga: Thanks! I have pushed an additional modification b2409562 in which I moved the fallback for the empty `cspec` also into the loop.

bash_completion
akinomyoga: OK!

akinomyoga:
I have also slightly modified the loop conditions 5d9fad4d. @potamides Could you check these changes?

akinomyoga (approved):

akinomyoga:
@potamides Can you add test cases for the problem that this PR tries to solve? You can add tests in [`test/t/unit/test_unit_command_offset.py`](https://github.com/scop/bash-completion/blob/master/test/t/unit/test_unit_command_offset.py).

potamides:
>Could you check these changes?

Looks fine to me. I also gave it a run and it's working as expected.

>Can you add test cases for the problem that this PR tries to solve?

Sure!

akinomyoga:
FYI: Test can be run by e.g. the following command:

```bash
$ cd test/t/unit
$ pytest -v test_unit_command_offset.py
```

You may also check the module [`test/t/conftest.py`](https://github.com/scop/bash-completion/blob/master/test/t/conftest.py) for the details of the test utilities, and the other test files for useful examples.

potamides:
With your recent change in commit 5c86ce5 `_command_offset` now attempts minimal completion when the new cspec is empty. I think the earlier behavior of not attempting any completion in that case was closer to the behavior of the shell.

Consider this example:

```bash
complete -F _test1 test1

_test1() { complete -r test1 && return 124; }
```

This won't trigger any completion.

akinomyoga:
> With your recent change in commit [5c86ce5](https://github.com/scop/bash-completion/commit/5c86ce5b1eede9a1d684ef989fefd76a57bd5b05) `_command_offset` now attempts minimal completion when the new cspec is empty. I think the earlier behavior of not attempting any completion in that case was closer to the behavior of the shell.

Hmm, OK. I thought we should attempt the minimal completion in this case. I'm not sure if Bash's behavior is the intended one or there is any rationale for that behavior, but I may drop the change.

akinomyoga:
@potamides Sorry for the delay. I added a test case in which the completion spec is not changed (2a1a82a9). Also, I added a line to clear `COMPREPLY` when the completion function returns 124 to make the behavior the same as Bash's treatment of 124 (5508eb9d). Could you check them?

potamides:
No worries! I checked your changes, they make sense to me and work as expected. Thanks!

akinomyoga (approved):

akinomyoga:
I've squashed small changes.

@scop Do you have any comments? If not, I will merge the PR. If you think descriptions are needed in the commit messages, we can add them.

scop:
Just a small one: `fix(_command_offset): refactor the loop` -- if it's a `refactor` commit then the type should be `refactor`, not `fix` (if it doesn't actually fix anything by itself). But if it _does_ fix something, then the message should be clarified what that is.

Other than that, no other comments besides looks like nice collaboration there, I'll trust your judgement on the commit message contents including my note above :+1:

akinomyoga:
@scop Thank you!

> But if it _does_ fix something, then the message should be clarified what that is.

Actually, it includes a tiny fix. I have decided to split the commit into two, 3d21604 and 2596e25. After the CI has passed, I'll merge.

akinomyoga:
@potamides We have merged the PR. Thank you again for making the PR, for your patience with requests, and for updating the PR!

potamides:
@akinomyoga @scop Very nice. Finally I can complete my aliases without any unexpected surprises! Thank you for your help and for merging this PR!

PR #670, opened 2021-12-23
Fix `_expand` for failglob

### LINKED ISSUE #668: ~ complete is broken
## Describe the bug

When I type `code ~<TAB>` I obtain the `bash: no match: COMPREPLY[0]=~root` error:

image

## To reproduce

1. `code ~<TAB>`
2. See problem

## Expected behavior

No error expected even with `failglob` enabled.

- [x] Operating system name/distribution and version: Linux Mint 20.2 Cinnamon
- [x] bash version, `echo "$BASH_VERSION"`: 5.1.0(1)-release
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.10

akinomyoga:
Resolve #668

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1209,7 +1209,7 @@ _expand()
             ;;
         ~*)
             _tilde "$cur" ||
-                eval COMPREPLY[0]="$(printf ~%q "${COMPREPLY[0]#\~}")"
+                eval "COMPREPLY[0]=$(printf ~%q "${COMPREPLY[0]#\~}")"
             return ${#COMPREPLY[@]}
             ;;
     esac
--- a/test/t/unit/test_unit_expand.py
+++ b/test/t/unit/test_unit_expand.py
@@ -1,6 +1,6 @@
 import pytest
 
-from conftest import assert_bash_exec
+from conftest import assert_bash_exec, bash_env_saved
 
 
 @pytest.mark.bashcomp(cmd=None, ignore_env=r"^[+-](cur|COMPREPLY)=")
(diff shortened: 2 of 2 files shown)


akinomyoga:
This isn't directly related to this fix, but I have a question.

### Usage of _expand in completions/{info,man}

When I was checking the use cases of `_expand` in `bash-completion`, I found the following codes in `completions/{info,man}` (see [completions/info\#L9](https://github.com/scop/bash-completion/blob/b0afc79c12f3d9fe7d5cc2e58a373eba0edc0d09/completions/info#L9-L50) and [completions/man#L55](https://github.com/scop/bash-completion/blob/b0afc79c12f3d9fe7d5cc2e58a373eba0edc0d09/completions/man#L55-L78)):

```bash
    if [[ $cur == @(*/|[.~])* ]]; then
        _filedir "$manext"
        return
    fi

    [...]

    _expand || return
```

Since `cur` doesn't seem to be modified between these lines, `cur` will not start with a tilde at the line `_expand` is called. However, `_expand` does something only when `cur` stars with a tilde. These invocations of `_expand` don't seem to do anything. Are they needed?

scop:
Thanks!

And good catch re `info` and `man`, I believe those `_expand` calls can be removed. Related commits are at least 1b85a1bf582fe000c812b2dc025971b1fa1f78c7 and add4e3cd443da011e8302b5d1fcfb5673d850f7a -- unfortunately in the latter I failed to elaborate why exactly were those calls deferred, but I guess it was because they didn't need to be run at that point. And I'm not sure why they'd need to be run at all, at least they have been no-ops since.

PR #673, opened 2021-12-27
doc(styleguide): fix asciidoc format

akinomyoga:
I have noticed that `doc/*.txt` are actually AsciiDoc, and some of the formatting in the HTML file `doc/html~/main.html` generated by `./makeHtml.sh` is broken. This PR fixes the formatting.


scop:
Cool, thanks. Depending on what we want for the doc/ things in the future, some pre-commit syntax checks would be in order.

akinomyoga:
Thank you!

PR #674, opened 2021-12-27
feat(ssh-keygen): sync -O options with openssh 8.4p1, FIDO arg fixes

scop:
The man page is missing bunch of equal signs for FIDO options that take an argument.

### DIFF
--- a/completions/ssh-keygen
+++ b/completions/ssh-keygen
@@ -85,8 +85,8 @@ _ssh_keygen()
                         cur=${cur#*=}
                         _filedir
                         ;;
-                    application=)
-                        COMPREPLY=("ssh:")
+                    application=*([^:=]))
+                        COMPREPLY=($(compgen -W "ssh:" -- "${cur#*=}"))
                         compopt -o nospace
                         ;;
                     user=*)
--- a/completions/ssh-keygen
+++ b/completions/ssh-keygen
@@ -66,11 +66,12 @@ _ssh_keygen()
                     no-x11-forwarding permit-agent-forwarding
                     permit-port-forwarding permit-pty permit-user-rc
                     permit-X11-forwarding no-touch-required source-address=
+                    verify-required
 
                     lines= start-line= checkpoint= memory= start= generator=
 
-                    application challenge= device resident user
-                    write-attestation-path
+                    application= challenge= device= no-touch-required resident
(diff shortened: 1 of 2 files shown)


completions/ssh-keygen
```diff
@@ -80,10 +81,17 @@ _ssh_keygen()
                         compopt -o filenames
                         COMPREPLY=($(compgen -c -- "${cur#*=}"))
                         ;;
-                    checkpoint=* | challenge=*)
+                    checkpoint=* | challenge=* | write-attestation=*)
                         cur=${cur#*=}
                         _filedir
                         ;;
+                    application=)
```
akinomyoga, line 88: `ssh-keygen -O application=s[TAB]` didn't work.

```suggestion
                    application=*([!:=]))
```

If we simply add `*`, it seems `ssh:` will be repeated as many times as the TABs, so `:` and `=` (which breaks the word through `COMP_WORDBREAKS`) are excluded.

completions/ssh-keygen
```diff
@@ -80,10 +81,17 @@ _ssh_keygen()
...
+                        COMPREPLY=("ssh:")
```
akinomyoga, line 89: Currently, any string already input by the user other than `ssh:` is removed. Maybe this is more natural:
```suggestion
                        COMPREPLY=($(compgen -W "ssh:" -- "${cur#*=}"))
                        ((${#COMPREPLY[@]} == 0)) && COMPREPLY=("ssh:${cur#*=}")
```

completions/ssh-keygen
scop: Thanks, added with `[^:=]` in b252360 as using `^` to negate is more common for these cases than `!` in the codebase.

completions/ssh-keygen
scop: Thanks, added the first in b252360. Left out the second because it seems a bit too magical to me and would give incorrect results if new application prefixes are ever introduced.

completions/ssh-keygen
akinomyoga: Thanks!

completions/ssh-keygen
akinomyoga: OK!

akinomyoga (approved):

PR #675, opened 2021-12-28
feat(_modules): support zstd compressed modules

richardschuetz:


scop:
Thanks! Looks good otherwise, but the source is not shfmt'd. FFR, search for pre-commit in CONTRIBUTING.md (or do the equivalent formats and checks manually, see .pre-commit-config.yaml). I'll take care of it for you here.

PR #678, opened 2022-01-02
Namespacing/publicity variable naming changes

scop:
Implementation of variable naming changes discussed in https://github.com/scop/bash-completion/discussions/537 and https://github.com/scop/bash-completion/discussions/539

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1334,7 +1334,7 @@ _backup_glob='@(#*#|*@(~|.@(bak|orig|rej|swp|dpkg*|rpm@(orig|new|save))))'
 #
 _xinetd_services()
 {
-    local xinetddir=${BASHCOMP_XINETDDIR:-/etc/xinetd.d}
+    local xinetddir=${_comp_test_xinetd_dir:-/etc/xinetd.d}
     if [[ -d $xinetddir ]]; then
         local IFS=$' \t\n' reset=$(shopt -p nullglob)
         shopt -s nullglob
--- a/completions/tar
+++ b/completions/tar
@@ -481,7 +481,7 @@ _gtar()
     # FIXME: handle long options
     local tar_mode_arg=
 
-    if [[ -v BASHCOMP_TAR_OPT_DEBUG ]]; then
+    if [[ -v _comp_cmd_tar_debug ]]; then
         set -x
         local PS4='$BASH_SOURCE:$LINENO: '
     fi
@@ -634,7 +634,7 @@ _gtar()
         break
     done # just-for-easy-break while
 
-    if [[ -v BASHCOMP_TAR_OPT_DEBUG ]]; then
+    if [[ -v _comp_cmd_tar_debug ]]; then
         set +x
     fi
 }
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -190,8 +190,8 @@ def bash(request) -> pexpect.spawn:
     tmpdir = None
     bash = None
 
-    if os.environ.get("BASHCOMP_TEST_LOGFILE"):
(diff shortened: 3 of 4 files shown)


scop:
Here's a few initial bits to start the work and to see how it'd look like.

This does not address the publicity part yet; arguably at least the `_comp_test_*` env state saving variables should be marked as private the way we end up wanting private ones to be marked.

This also does not mark any of the changes implemented yet as breaking even though strictly speaking, the xinetd and tar ones kind of would be. But those were never intended to be user settable under normal operation, and I doubt renaming them would cause any real world problems. Then again if we end up making some "really" breaking changes elsewhere which would warrant major version bump per semver, we could mark these as such as well.

scop:
Should be rebased on top of #728 when that's in.

scop:
This is now in sync with the current proposal at https://github.com/scop/bash-completion/discussions/539#discussioncomment-864857. Target branch is docs/markdown, with the intent that this will be merged to master after the markdown branch is in, not to the markdown branch before that.

Therefore draft status for now, although I consider this complete for changes touching public configuration/control variables, internal ones should be done in another PR to keep size at bay.

completions/cvs
```diff
@@ -248,10 +248,10 @@ _cvs()
             esac
 
             if [[ $cur != -* ]]; then
-                # if $COMP_CVS_REMOTE is not null, 'cvs commit' will
-                # complete on remotely checked-out files (requires
+                # if $BASH_COMPLETION_CMD_CVS_REMOTE is not null, 'cvs commit'
+                # will complete on remotely checked-out files (requires
                 # passwordless access to the remote repository
-                if [[ -n ${COMP_CVS_REMOTE:-} ]]; then
+                if [[ -n ${BASH_COMPLETION_CMD_CVS_REMOTE-${COMP_CVS_REMOTE-}} ]]; then
```
akinomyoga, line 254:
```suggestion
                if [[ ${BASH_COMPLETION_CMD_CVS_REMOTE-${COMP_CVS_REMOTE-}} ]]; then
```

bash_completion
```diff
@@ -634,7 +635,8 @@ _filedir()
         IFS=$'\n'
 
         # Try without filter if it failed to produce anything and configured to
-        [[ -n ${COMP_FILEDIR_FALLBACK-} && -n $arg && ${#toks[@]} -lt 1 ]] && {
+        [[ ${BASH_COMPLETION_FILEDIR_FALLBACK-${COMP_FILEDIR_FALLBACK-}} &&
+            -n $arg && ${#toks[@]} -lt 1 ]] && {
```
akinomyoga, line 639:
```suggestion
            $arg && ${#toks[@]} -lt 1 ]] && {
```

bash_completion
```diff
@@ -623,7 +623,8 @@ _filedir()
         # the fallback condition with the "plus" dirs.
         local opts=(-f -X "$xspec")
         [[ $xspec ]] && plusdirs=(-o plusdirs)
-        [[ ${COMP_FILEDIR_FALLBACK-} || -z ${plusdirs-} ]] ||
+        [[ ${BASH_COMPLETION_FILEDIR_FALLBACK-${COMP_FILEDIR_FALLBACK-}} ||
+            -z ${plusdirs-} ]] ||
```
akinomyoga, line 627:
```suggestion
            ! ${plusdirs-} ]] ||
```

It seems there are more cases `[[ ! $word ]]` (5 cases in `bash_completion`) than `[[ -z $word ]]` (2 cases). Another instance of `-z` is in line `bash_completion:1990` (`_cd`).

akinomyoga (approved):

akinomyoga:
Thanks!

Issue #641, opened 2021-11-01
Add GitHub pull request template


scop:
Hm, we should add a template for pull requests so that the request to ask upstream to include the completion first is more prominent. Anyway, for time being, see https://raw.githubusercontent.com/scop/bash-completion/master/.github/ISSUE_TEMPLATE/feature_request.md and the start of CONTRIBUTING.md.

_Originally posted by @scop in https://github.com/scop/bash-completion/issues/639#issuecomment-956342757_

PR #679, opened 2022-01-04
New completion: stunnel

olszomal:


akinomyoga:
@olszomal Thank you for the new completions. Have you read [CONTRIBUTING.md](https://github.com/scop/bash-completion/blob/master/CONTRIBUTING.md) and already asked the upstream project about including it there?

mtrojnar:
https://github.com/scop/bash-completion/blob/master/CONTRIBUTING.md#contributing-to-bash-completion currently says:

> It's nowhere near clear cut always what is the best place for the completion, upstream project or us. Even if it would seem to be upstream, not all upstreams are interested in shipping completions, or their install systems might not easily support installing completion files properly. Or the projects might be stagnant. But give it some thought, and ask if unsure.

Wouldn't it be a good idea to directly clarify that the upstream project should be asked **first**?

akinomyoga:
@mtrojnar Thank you for your feedback! Yes, and that is actually #641.

akinomyoga:
@mtrojnar Oh, so you are the upstream maintainer of `stunnel` and `osslsigncode`. What do you think of including and maintaining these completions in the upstream project?

mtrojnar:
> What do you think of including and maintaining these completions in the upstream projects?

Gladly.  It's indeed easier this way.

completions/stunnel
```diff
@@ -0,0 +1,25 @@
+# bash completion for stunnel                              -*- shell-script -*-
+
+_stunnel()
```
akinomyoga, line 3: We are currently gradually shifting the names of completion functions to `_comp_cmd_CommandName` as described in [`doc/styleguide.txt`](https://github.com/scop/bash-completion/blob/521348d0a03e961043a2aa64911f2f048e0469b2/doc/styleguide.txt#L112-L117). Note that the existing names such as `_init_completion` and `_filedir` will probably be kept as compatibility names.

```suggestion
_comp_cmd_stunnel()
```

Actually, there is an ongoing discussion at #539, and the names of the external completion functions might possibly be changed to [`_comp_ext_CommandName`](https://github.com/scop/bash-completion/discussions/539#discussioncomment-942082), but the current preferred name is `_comp_cmd_CommandName`.

If there would be any change in the future, we will submit a PR. For that purpose, we maintain a list of external completion on the Wiki page *[Externally available completions](https://github.com/scop/bash-completion/wiki/Externally-available-completions)*. Could you also update the Wiki page, if you have included the completion functions in the upstream repositories?

completions/stunnel
```diff
@@ -0,0 +1,25 @@
...
+    esac
+
+    if [[ ${cur} == -* ]] ; then
+        COMPREPLY=($(compgen -W "${opt}" -- ${cur}))
+        return
+    fi
+
+    _filedir '@(cnf|conf)'
+} &&
+    complete -F _stunnel stunnel
```
akinomyoga, line 23:
```suggestion
    complete -F _comp_cmd_stunnel stunnel
```

akinomyoga (commented):
@mtrojnar Great! Before including them in the upstream repositories, I have a comment. I'd suggest similar changes to the `osslsigncode` completion.

akinomyoga:
Now it seems to be included in `stunnel-5.62` https://github.com/mtrojnar/stunnel/blob/master/tools/stunnel.bash. We may update the list https://github.com/scop/bash-completion/wiki/Externally-available-completions.

scop:
Done, closing here. Thanks!

https://github.com/scop/bash-completion/wiki/Externally-available-completions/_compare/1cab67579732b80b95093d9a2117f42e54c8c917...f7b9e0a1318199292f8957315e59812aed3066e3

PR #681, opened 2022-01-04
Remove redundante variable docs from README

scop:
While working on #678, I noticed we have some redundant and inaccurate descriptions in README for some variables. I'm sure some of these are not FAQ's these days any more either.

This removes the redundant descriptions from README, moves it elsewhere for `tar`, and generally makes the info a bit less scattered. I think we should make the main README.md even smaller going forward, but that's something for another PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,30 +188,6 @@ A. Install it in one of the directories pointed to by
      ${BASH_COMPLETION_COMPLETIONSDIR})
    ```
 
-**Q. I use CVS in combination with passwordless SSH access to my remote
-   repository. How can I have the `cvs` command complete on remotely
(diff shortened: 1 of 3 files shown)


akinomyoga:
Maybe we can change the extension of `doc/bash_completion.txt` from `.txt` to `.adoc` so that GitHub renders the content of the file (just like Markdown files). Then we can put a link to the file in `README.md`.

akinomyoga (approved):

scop:
Yup, either that, or change all docs to markdown. Markdown covers our needs fine, and more people are familiar with it than with asciidoc.

PR #682, opened 2022-01-07
feat(feh): complete on WebP images

Maelan:
(trivial change)

`feh` (as well as ImageMagick) has been supporting WebP for a while now. WebP has already gained much traction as a successor to PNG and JPEG, so it’s important than completion lets us open them. :-)


scop:
Thanks!

PR #683, opened 2022-01-09
rsync: improve completion by using _parse_help

sbraz:
Hi,
I often need this option and I find the name rather long, it would be nice to have it completed.

In order to limit the size of the diff, I didn't try to optimize the layout of the options in the `COMPREPLY` variable. Let me know if that is a problem.

### DIFF
--- a/completions/rsync
+++ b/completions/rsync
@@ -37,32 +37,15 @@ _rsync()
 
     case $cur in
         -*)
-            COMPREPLY=($(compgen -W '--verbose --quiet --no-motd --checksum
-                --archive --recursive --relative --no-implied-dirs
-                --backup --backup-dir= --suffix= --update --inplace --append
-                --append-verify --dirs --old-dirs --links --copy-links
-                --copy-unsafe-links --safe-links --copy-dirlinks
-                --keep-dirlinks --hard-links --perms --executability --chmod=
-                --acls --xattrs --owner --group --devices --copy-devices
-                --specials --times --omit-dir-times --super --fake-super
-                --sparse --dry-run --whole-file --no-whole-file
-                --one-file-system --block-size= --rsh= --rsync-path=
-                --existing --ignore-existing --remove-source-files --delete
-                --delete-before --delete-during --delete-delay --delete-after
-                --delete-excluded --ignore-errors --force --max-delete=
-                --max-size= --min-size= --partial --partial-dir=
-                --delay-updates --prune-empty-dirs --numeric-ids --timeout=
-                --contimeout= --ignore-times --size-only --modify-window=
-                --temp-dir= --fuzzy --compare-dest= --copy-dest= --link-dest=
-                --compress --compress-level= --skip-compress= --cvs-exclude
-                --filter= --exclude= --exclude-from= --include= --include-from=
-                --files-from= --from0 --protect-args --address= --port=
-                --sockopts= --blocking-io --no-blocking-io --stats
-                --8-bit-output --human-readable --progress --itemize-changes
-                --out-format= --log-file= --log-file-format= --password-file=
-                --list-only --bwlimit= --write-batch= --only-write-batch=
-                --read-batch= --protocol= --iconv= --ipv4 --ipv6 --version
-                --help --daemon --config= --no-detach' -- "$cur"))
+            COMPREPLY=($(
+                # Account for the fact that older rsync versions (before cba00be6, meaning before v3.2.0)
+                # contain the following unusual line in --help:
+                # "(-h) --help                  show this help (-h is --help only if used alone)"
+                compgen -W '$(
+                "$1" --help 2>&1 | sed -e "s/^([^)]*)//" | _parse_help -)
+                --daemon --old-d{,irs}
+                --no-{blocking-io,detach,whole-file,inc-recursive,i-r}' -X '--no-OPTION' -- "$cur"
+            ))
             [[ ${COMPREPLY-} == *= ]] || compopt +o nospace
             ;;
         *:*)
--- a/test/t/test_rsync.py
+++ b/test/t/test_rsync.py
@@ -14,3 +14,7 @@ class TestRsync:
     @pytest.mark.complete("rsync --rsh=")
     def test_3(self, completion):
         assert completion == "rsh ssh".split()
+
+    @pytest.mark.complete("rsync --", require_cmd=True)
+    def test_4(self, completion):
+        assert "--help" in completion
--- a/test/test-cmd-list.txt
+++ b/test/test-cmd-list.txt
@@ -318,6 +318,7 @@ rmdir
 rmmod
 rpm
 rpmbuild
+rsync
 rtcwake
 sbopkg
 screen


completions/rsync
```diff
@@ -38,6 +38,7 @@ _rsync()
         -*)
             COMPREPLY=($(compgen -W '--verbose --quiet --no-motd --checksum
                 --archive --recursive --relative --no-implied-dirs
+                --no-inc-recursive --no-i-r
                 --backup --backup-dir= --suffix= --update --inplace --append
                 --append-verify --dirs --old-dirs --links --copy-links
                 --copy-unsafe-links --safe-links --copy-dirlinks
```
akinomyoga, line 44:
```suggestion
            COMPREPLY=($(compgen -W '$(_parse_help "$1")
                --config --daemon --old-d{,irs}
                --no-{blocking-io,detach,whole-file,inc-recursive,i-r}'))
```

akinomyoga (commented):
`_parse_help rsync` shows still other options not listed in `completions/rsync`. Why didn't we use `_parse_help` for `rsync`?

```
--atimes
--checksum-choice=
--checksum-seed=
--compress-choice=
--copy-as=
--crtimes
--debug=
--del
--delete-missing-args
--early-input=
--groupmap=
--ignore-missing-args
--info=
--max-alloc=
--mkpath
--munge-links
--omit-link-times
--open-noatime
--outbuf=
```

There seems to be another short name `--old-d` (which is a synonym of `--old-dirs`).

On the other hand, there are some options that cannot be extracted by `_parse_help`. Maybe  we can use `_parse_help` and supplement the options that cannot be extracted by `_parse_help`:

```bash
            COMPREPLY=($(compgen -W '$(_parse_help "$1")
                --config --daemon --old-d{,irs}
                --no-{blocking-io,detach,whole-file,inc-recursive,i-r}'))
```

sbraz:
If it doesn't affect performance, I'm all for it.

akinomyoga:
Thanks for your reply! I'm currently thinking that maybe we can also generate `--no-*` options automatically. Not tested but something like

```bash
            COMPREPLY=($(compgen -W '
                $(_parse_help "$1" | awk '\''{print; if (!/^--no-|=$/ && sub(/^--/, "--no-")) print; }'\'')
                --config --daemon --old-d{,irs} --no-{detach,inc-recursive,i-r}' \
                -X '--no-OPTION' -- "$cur"))
```

sbraz:
Isn't that a bit extreme? I'm afraid we might suggest options that don't exist if we do this.
The manpage says
> Not all options may be prefixed with a "no-": only options that are implied by other options (e.g. --no-D, --no-perms) or have different defaults in various circumstances (e.g. --no-whole-file, --no-blocking-io, --no-dirs)

akinomyoga:
Yeah, I guess so.

I searched and found the list at [`rsync/options.c`](https://github.com/WayneD/rsync/blob/ee9199b54274aed05776f55806dc18b3e2f20a32/options.c#L585-L831). I assumed that all the boolean-type options (which do not take option arguments) would have their counterparts, but it seems only a part of the boolean options (probably less than 50%) have `--no-*` counterparts. Interestingly, there seems to be the option `--i-r`, which means that `--no-i-r` was actually a counterpart of `--i-r`.

scop:
> Why didn't we use `_parse_help` for `rsync`?

No idea, but I think we should start doing it from here on.

By the way, some options such as `--config=` are valid in daemon mode only, so we should look into if `--daemon` has been given and `_parse_help` using `--daemon --help` to get those options if it is instead of `--help`. But this can be left to another PR if we don't want to expand the scope of this one to cover that, too. Before doing this separation, I suggest sticking with what `--help` outputs and not hardcoding any additions such as the mentioned `--config=` which could be wrong (or at least being careful with what we hardcode).

akinomyoga:
> I suggest sticking with what `--help` outputs and not hardcoding any additions such as the mentioned `--config=` which could be wrong (or at least being careful with what we hardcode).

Thank you. @sbraz Could you update the PR as described above?

sbraz:
> > I suggest sticking with what `--help` outputs and not hardcoding any additions such as the mentioned `--config=` which could be wrong (or at least being careful with what we hardcode).
> 
> Thank you. @sbraz Could you update the PR as described above?

So `_parse_help` + the options you mentioned in your first comment?

akinomyoga:
> So `_parse_help` + the options you mentioned in your first comment?

The suggested change seems to be actually

> not hardcoding any additions such as the mentioned `--config=`

but I think you may just drop `--config` from my first comment since the other options seem to be available anyway.

----

Although maybe it should be processed in a separate PR, there seem to be still many options not listed in `rsync --help`. Maybe the hardcoded options can be filtered by `grep -Fxe "$(strings "$1" | sed 's/^/--/')"` to check whether `rsync` installed in the user's system really supports them. Also, when `--daemon` is specified, we need to use `$(_parse_help "$1" '--daemon --help')` instead of `$(_parse_help "$1")`.

sbraz:
I kept with the first suggestion, I only changed whitespace a little bit. I have renamed the PR and edited the commit message.

completions/rsync
```diff
@@ -36,32 +36,10 @@ _rsync()
...
-                --filter= --exclude= --exclude-from= --include= --include-from=
-                --files-from= --from0 --protect-args --address= --port=
-                --sockopts= --blocking-io --no-blocking-io --stats
-                --8-bit-output --human-readable --progress --itemize-changes
-                --out-format= --log-file= --log-file-format= --password-file=
-                --list-only --bwlimit= --write-batch= --only-write-batch=
-                --read-batch= --protocol= --iconv= --ipv4 --ipv6 --version
-                --help --daemon --config= --no-detach' -- "$cur"))
+            COMPREPLY=($(compgen -W '$(_parse_help "$1")
+                --config --daemon --old-d{,irs}
```
akinomyoga, line 40: Thank you for updating the PR! Could you drop `--config` from here, or do you think we can keep `--config` in this PR?

sbraz: Meh my bad, I removed it, then I copied some old code again. Let me fix that.

akinomyoga: No problem, thanks.

akinomyoga:
It seems `pre-commit` fails for the commit message and some code formatting. Could you set up `pre-commit` and update the codes and the commit message?

cf [CONTRIBUTING.md](https://github.com/scop/bash-completion/blob/master/CONTRIBUTING.md)

> - We use [Conventional Commits](https://www.conventionalcommits.org/)
>   to format commit messages, with types and most other details from [commitlint's config-conventional](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional). `gitlint` in our pre-commit config checks commit messages for conformance with these rules.

> - Install pre-commit and set it up, see <https://pre-commit.com/>. That'll run a bunch of linters and the like, the same as the
  bash-completion CI does. Running it locally and fixing found issues before commit/push/PR reduces some roundtrips with the review. After installing it, enable it for stages we use it with like:
>
>   ```shell
>   pre-commit install --hook-type pre-commit --hook-type commit-msg
>   ```

In order to run pre-commit on the existing commits, you can run, e.g.,

```bash
$ pre-commit run --from-ref origin/master --to-ref @
```

@scop Is the above description in CONTRIBUTING.md a sufficient description of installing `pre-commit` hook? In my environment, I somehow need to also run the following commands (as described in [ci.yaml](https://github.com/scop/bash-completion/blob/a98845fae7a9e58a719255be37930a41e7460894/.github/workflows/ci.yaml#L25-L27)).

```bash
$ python3 -m venv venv
$ source venv/bin/activate
$ python3 -m pip install -Ur test/requirements-dev.txt
```

and also need to run the Docker daemon and grant permissions to the user to access `/var/run/docker.sock`.

sbraz:
I've fixed the shfmt error. I don't understand the gitlint thing, is it really required? I don't see a lot of commit messages  that "start with one of fix, feat, chore, docs, style, refactor, perf, test, revert, ci, build" as the error indicates.

akinomyoga:
Thank you! `gitlint` *requires* that the commit message starts with one of these keywords which is not satisfied currently, i.e., you can update it to something like `fix(rsync): improve completion with _parse_help` for the present case.

akinomyoga:
Conventional commits are introduced after #534.

akinomyoga:
OK! The tests are all green. LGTM

akinomyoga:
@scop This is also a PR that I think is now ready to merge. Could you check it?

A better support for `rsync --daemon --[TAB]` will be addressed in a separate PR.

scop (changes requested):
Sorry for the delay. LGTM otherwise, but now that we're scraping and parsing the output, I think we should add at least a basic test case to `test/t/test_rsync.py` with `require_cmd=True` which checks that we receive at least something for `rsync --<TAB>` which is not one of the hardcoded options here.

sbraz:
I've added one. Please let me know if that is OK.

sbraz:
Err my bad, my test is for one of the hardcoded options, let me fix that.

sbraz:
Fixed, I've pushed a test for a parsed option.

scop (changes requested):
In addition to the comment, the `required=True` will cause `rsync` to be added to `test/test-cmd-list.txt` and CI will fail if it's not there. `test/update-test-cmd-list` will take care of adding it, ditto pre-commit which runs that; please include the change to that file in the PR.

test/t/test_rsync.py
```diff
@@ -14,3 +14,7 @@ def test_2(self, completion):
     @pytest.mark.complete("rsync --rsh=")
     def test_3(self, completion):
         assert completion == "rsh ssh".split()
+
+    @pytest.mark.complete("rsync --log-file", require_cmd=True)
```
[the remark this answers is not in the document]
sbraz: You're absolutely right, I've fixed the test accordingly.

completions/rsync
```diff
@@ -37,32 +37,11 @@ _rsync()
...
-                --filter= --exclude= --exclude-from= --include= --include-from=
-                --files-from= --from0 --protect-args --address= --port=
-                --sockopts= --blocking-io --no-blocking-io --stats
-                --8-bit-output --human-readable --progress --itemize-changes
-                --out-format= --log-file= --log-file-format= --password-file=
-                --list-only --bwlimit= --write-batch= --only-write-batch=
-                --read-batch= --protocol= --iconv= --ipv4 --ipv6 --version
-                --help --daemon --config= --no-detach' -- "$cur"))
+            COMPREPLY=($(
+                compgen -W '$(_parse_help "$1")
```
scop, line 41: Heh, interestingly, the line containing `--help` is formatted differently from everything else here, throwing off `_parse_help` :P. One way to fix it:

```suggestion
                # sed preprocessing for lines like '(-h) --help ...', seen
                # e.g. with rsync 3.1.3 on Ubuntu 20.04
                compgen -W '$(
                    rsync --help 2>&1 | sed -e "s/^([^)]*)//" | _parse_help -
                )
```

Another would be to generalize the fix by doing it in `__parse_options` (likely in a separate PR).

sbraz: Any idea why I can't reproduce this locally when I run `pytest -vv test/t/test_rsync.py::TestRsync`? Tests pass which seems to indicate that `--help, -h (*)` is not a problem and I don't understand why they do if you say we need to pre-process the output.

akinomyoga: It depends on the version of `rsync`.

```bash
[murase@padparadscha 0 ~]$ rsync --version | head -2
rsync  version 3.1.3  protocol version 31
Copyright (C) 1996-2018 by Andrew Tridgell, Wayne Davison, and others.
[murase@padparadscha 0 ~]$ rsync --help | grep help
(-h) --help                  show this help (-h is --help only if used alone)
Use "rsync --daemon --help" to see the daemon-mode command-line options.
```


```bash
[murase@chatoyancy 0 ~]$ rsync --version | head -2
rsync  version 3.2.3  protocol version 31
Copyright (C) 1996-2020 by Andrew Tridgell, Wayne Davison, and others.
[murase@chatoyancy 0 ~]$ rsync --help | grep help
--help, -h (*)           show this help (* -h is help only on its own)
Use "rsync --daemon --help" to see the daemon-mode command-line options.
```

sbraz: Ah thanks, I'd somehow misread the regex and I thought the newest version had the problem. Looks like https://github.com/WayneD/rsync/commit/cba00be6 fixed it.

I've fixed the problem in my latest push.

sbraz:
@scop are we good to go?

scop:
Yup! Thanks for the work and patience here :+1:

Issue #573, opened 2021-08-21
Rework test suite to use captured outputs


scop:
## Describe the feature/solution

Currently we pull a bunch of large containers for running the test suite, and run them against real commands in them. While this is useful in the sense that it tests against the very real things, and also in situations where unexpected errors in commands might occur, and tracks changes in developing distros, it is:
- slow
- limited to what containers we have available

It would be great if we had a corpus of command outputs for different command versions on different platforms we would primarily test against. Secondarily we could (perhaps on schedule, bi-weekly or so) run the live tests at least against developing distros such as Fedora and Alpine bleeding edges, perhaps Debian too.

## Maintenance (please complete the following information)

- [ ] This is a request for new completion
- [ ] Link to upstream project query about shipping the completion:

N/A

akinomyoga:
I'm interested in how you think this should be achieved. For example, are target commands supposed to be replaced by the function of the same name in the tests? How are we going to manage the "corpus" of the command outputs? For example, if we save them in the git tree, I'm afraid that the repository size would become unreasonably large, so should we manage the "corpus" externally? Maybe we can automatically generate/update the corpus with CI.

scop:
I was primarily thinking of adding mock executables, but guess functions could work as well. We have a few examples of the former in the tree, `ls test/**/bin`

Command outputs could be saved as text files, and the mocked executables/functions would basically `cat` them.

Sure, the tree would become largish, but I'm not sure how much of a problem that would be. But if it is, and an external storage would solve it, I'm all for it.

Generating in CI or some automated way would be nice. Probably need a way to make manual additions work too, though.

Issue #430, opened 2020-07-11
Various tests are flaky when run in parallel


Patagonicus:
I've noticed that tests are very likely to fail on ARM machines when run in parallel with xdist. I was also able to reproduce this on a x86_64 machine by introducing some load. I was unable to reproduce this on either machine when running the tests serially, even with a high load (stress -c 16 on a machine with 4 logical threads).

Please see the https://github.com/NixOS/nixpkgs/issues/91706 for more information.

scop:
Yep, I'm aware of this, and it happens every now and then in Travis too. Would like to see it fixed and not brushed under the carpet by turning parallelism off, as I think the test suite is too slow already even with modest parallel runners. And I think keeping a problem visible makes it more likely that something will be done about it.

Issue #139, opened 2017-07-02
test: service unresolved/times out


mgorny:
Originally it was unresolved but when I try to repro it via `./run` I get timeout very quickly, so that's probably some weird issue with expect.

```
Test run by mgorny on Sun Jul  2 10:20:13 2017
Native configuration is x86_64-pc-linux-gnu

		=== completion tests ===

Schedule of variations:
    unix

Running target unix
Using /usr/share/dejagnu/baseboards/unix.exp as board description file for target.
Using /usr/share/dejagnu/config/unix.exp as generic interface file for target.
Using ./config/default.exp as tool-and-target-specific interface file.
Running ./completion/service.exp ...
exp_spawn bash --rcfile ./config/bashrc
/@echo $?
0
/@BASH_COMPLETION_COMPAT_DIR="$SRCDIR/fixtures/shared/empty_dir"
/@echo $?
0
/@source $(cd "$SRCDIR/.."; pwd)/bash_completion
/@echo $?
0
/@printf "%s" "$COMP_WORDBREAKS"
 	
"'><=;|&(:/@echo $?
0
/@printf "%s " "${BASH_VERSINFO[@]}"
4 4 12 1 release x86_64-pc-linux-gnu /@echo $?
0
/@printf "%s" "$BASH_VERSION"
4.4.12(1)-release/@echo $?
0
/@printf "%s" "$TESTDIR"
/home/mgorny/git/bash-completion/test/@echo $?
0
/@eval $(locale); printf "%s" "$LC_CTYPE"
en_US.utf8/@echo $?
0
/@exec 6>'xtrace.log'
/@echo $?
0
/@BASH_XTRACEFD=6
/@echo $?
0
/@set -o xtrace
/@echo $?
0
/@type service &> /dev/null && echo -n 0 || echo -n 1
0/@__load_completion service ; complete -p service &> /dev/null && echo -n 0 || echo -n 1
0/@{ (set -o posix ; set); declare -F; shopt -p; set -o; } > "$TESTDIR/tmp/env.env1~"
/@echo $?
0
/@service 
6to4                                              acpid
agetty                                            alsa-restore
alsasound                                         alsa-state
auditd                                            avahi-daemon
avahi-dnsconfd                                    awesfx
binfmt                                            bluetooth
boa                                               bootmisc
busybox-ntpd                                      busybox-watchdog
carbon-aggregator                                 carbon-cache
carbon-relay                                      consolefont
cups                                              cups-browsed
[...]

Issue #137, opened 2017-07-02
test: py.test fails due to timeout (cold cache)


mgorny:
```
Test run by mgorny on Sun Jul  2 10:09:40 2017
Native configuration is x86_64-pc-linux-gnu

		=== completion tests ===

Schedule of variations:
    unix

Running target unix
Using /usr/share/dejagnu/baseboards/unix.exp as board description file for target.
Using /usr/share/dejagnu/config/unix.exp as generic interface file for target.
Using ./config/default.exp as tool-and-target-specific interface file.
Running ./completion/py.test.exp ...
exp_spawn bash --rcfile ./config/bashrc
/@echo $?
0
/@BASH_COMPLETION_COMPAT_DIR="$SRCDIR/fixtures/shared/empty_dir"
/@echo $?
0
/@source $(cd "$SRCDIR/.."; pwd)/bash_completion
/@echo $?
0
/@printf "%s" "$COMP_WORDBREAKS"
 	
"'><=;|&(:/@echo $?
0
/@printf "%s " "${BASH_VERSINFO[@]}"
4 4 12 1 release x86_64-pc-linux-gnu /@echo $?
0
/@printf "%s" "$BASH_VERSION"
4.4.12(1)-release/@echo $?
0
/@printf "%s" "$TESTDIR"
/home/mgorny/git/bash-completion/test/@echo $?
0
/@eval $(locale); printf "%s" "$LC_CTYPE"
en_US.utf8/@echo $?
0
/@exec 6>'xtrace.log'
/@echo $?
0
/@BASH_XTRACEFD=6
/@echo $?
0
/@set -o xtrace
/@echo $?
0
/@type py.test &> /dev/null && echo -n 0 || echo -n 1
0/@__load_completion py.test ; complete -p py.test &> /dev/null && echo -n 0 || echo -n 1
0/@{ (set -o posix ; set); declare -F; shopt -p; set -o; } > "$TESTDIR/tmp/env.env1~"
/@echo $?
0
/@py.test 
completion/ config/     docker/     fixtures/   lib/        log/        log-full/   log-old/    tmp/        unit/
/@py.test PASS: py.test  should show completions
^C
/@py.test -FAIL: py.test - should show completions at timeout
FAIL: Sync after INT at timeout
{ (set -o posix ; set); declare -F; shopt -p; set -o; } > "$TESTDIR/tmp/env.env2~"
bash: błąd składni przy nieoczekiwanym znaczniku `('
[...]

PR #687, opened 2022-01-22
Properly handle environments with unset IFS

sudonym1:
From the bash man page:

    Word Splitting
    ...snip...
    The shell treats each character of IFS as a delimiter, and splits the
    results of the other expansions into words using these characters as
    field terminators.  If IFS is unset, or its value is exactly
    <space><tab><newline>, the  default, then  sequences of <space>, <tab>,
    and <newline> at the beginning and end of the results of the previous
    expansions are ignored, and any sequence of IFS characters not at the
    beginning or end serves to delimit words.  If IFS has a value other than
    the default, then sequences of the whitespace characters space, tab, and
    newline are ignored at the beginning and end of the word, as long as the
    whitespace character is in the value of IFS (an IFS whitespace
    character).  Any character  in  IFS that is not IFS whitespace, along
    with any adjacent IFS whitespace characters, delimits a field.  A
    sequence of IFS whitespace characters is also treated as a delimiter.
    If the value of IFS is null, no word splitting occurs.

Per this paragraph, IFS can have one of three states:

1. IFS is set to some value, eg $' \t\n'
2. IFS is unset, in which case it behaves like $' \t\n'
3. IFS is null, in this case no splitting occurs.

Prior to this commit, the current state of `IFS` was stored into a local
variable `ifs`. Later, IFS is restored by setting `IFS=$ifs`. If the
user starts with either a set `IFS` or a null `IFS`, then this behavior
is correct. If the user starts with an unset `IFS`, then `IFS=$ifs`
results in a NULL `IFS`, which is not the original state. Most of the
time a null `IFS` is not desired, and indeed it breaks some things
inside bash-completions. eg

```
unset IFS
local ifs=$IFS IFS=$' \t\n'
restore=$(shopt -po noglob)
IFS=$ifs
$restore
```

Above the line `IFS=$ifs` puts us in a NULL `IFS` state, which is not
where we started. The line `$restore` fails with a confusing error
message `-bash: set -o noglob: command not found`.

In addition, this pattern does not make good use of bash shadowing
rules. When a `local IFS` shadows the global `IFS`, all that is needed
to restore the global IFS is to unset the local one.

With this commit, all instances of `local ifs=$IFS` have been removed.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -149,6 +149,23 @@ dequote()
     eval printf %s "$1" 2>/dev/null
 }
 
+# Unset the given variables across a scope boundary. Useful for unshadowing
+# global scoped variables. Note that simply calling unset on a local variable
+# will not unshadow the global variable. Rather, the result will be a local
+# variable in an unset state.
+# Usage: local IFS='|'; _comp_unlocal IFS
+# Param: $* Variable names to be unset
+_comp_unlocal()
+{
+    if ((BASH_VERSINFO[0] >= 5)) && shopt -q localvar_unset; then
+        shopt -u localvar_unset
+        unset -v "$@"
+        shopt -s localvar_unset
+    else
+        unset -v "$@"
+    fi
+}
+
 # Assign variable one scope above the caller
 # Usage: local "$1" && _upvar $1 "value(s)"
 # Param: $1  Variable name to assign value to
@@ -721,12 +738,12 @@ _comp_delimited()
         # glob char escaping issues to deal with. Do removals by hand instead.
         COMPREPLY=($(compgen "$@"))
         local -a existing
-        local x i ifs=$IFS IFS=$delimiter
+        local x i IFS=$delimiter
         existing=($cur)
         # Do not remove the last from existing if it's not followed by the
         # delimiter so we get space appended.
         [[ ! $cur || $cur == *"$delimiter" ]] || unset -v "existing[${#existing[@]}-1]"
-        IFS=$ifs
+        _comp_unlocal IFS
         if ((${#COMPREPLY[@]})); then
             for x in ${existing+"${existing[@]}"}; do
                 for i in "${!COMPREPLY[@]}"; do
@@ -1253,14 +1270,13 @@ else
         if [[ ${1-} == -s ]]; then
             procs=($(command ps ax -o comm | command sed -e 1d))
         else
-            local line i=-1 ifs=$IFS
-            IFS=$'\n'
+            local line i=-1 IFS=$'\n'
             # Some versions of ps don't support "command", but do "comm", e.g.
             # some busybox ones. Fall back
             local -a psout=($({
                 command ps ax -o command= || command ps ax -o comm=
             } 2>/dev/null))
-            IFS=$ifs
+            _comp_unlocal IFS
             for line in "${psout[@]}"; do
                 if ((i == -1)); then
                     # First line, see if it has COMMAND column header. For
@@ -1751,7 +1767,7 @@ _included_ssh_config_files()
 # Return: Completions, starting with CWORD, are added to COMPREPLY[]
 _known_hosts_real()
 {
-    local configfile flag prefix="" ifs=$IFS
+    local configfile flag prefix=""
     local cur suffix="" aliases i host ipv4 ipv6
     local -a kh tmpkh=() khd=() config=()
 
@@ -1817,7 +1833,7 @@ _known_hosts_real()
         #          spaces in their name work (watch out for ~ expansion
         #          breakage! Alioth#311595)
         tmpkh=($(awk 'sub("^[ \t]*([Gg][Ll][Oo][Bb][Aa][Ll]|[Uu][Ss][Ee][Rr])[Kk][Nn][Oo][Ww][Nn][Hh][Oo][Ss][Tt][Ss][Ff][Ii][Ll][Ee][ \t=]+", "") { print $0 }' "${config[@]}" | sort -u))
-        IFS=$ifs
+        _comp_unlocal IFS
     fi
     if ((${#tmpkh[@]} != 0)); then
         local j
@@ -1873,7 +1889,7 @@ _known_hosts_real()
                         # Add host to candidates
                         COMPREPLY+=($host)
                     done
-                    IFS=$ifs
+                    _comp_unlocal IFS
                 done <"$i"
             done
             ((${#COMPREPLY[@]})) &&
@@ -1935,7 +1951,6 @@ _known_hosts_real()
             $(compgen -A hostname -P "$prefix" -S "$suffix" -- "$cur"))
     fi
 
-    IFS=$' \t\n'
     $reset
 
     if ((${#COMPREPLY[@]})); then
@@ -2349,12 +2364,12 @@ complete -F _minimal ''
 __load_completion()
 {
     local -a dirs=(${BASH_COMPLETION_USER_DIR:-${XDG_DATA_HOME:-$HOME/.local/share}/bash-completion}/completions)
-    local ifs=$IFS IFS=: dir cmd="${1##*/}" compfile
+    local IFS=: dir cmd="${1##*/}" compfile
     [[ -n $cmd ]] || return 1
     for dir in ${XDG_DATA_DIRS:-/usr/local/share:/usr/share}; do
         dirs+=($dir/bash-completion/completions)
     done
-    IFS=$ifs
+    _comp_unlocal IFS
 
     if [[ $BASH_SOURCE == */* ]]; then
         dirs+=("${BASH_SOURCE%/*}/completions")
--- a/completions/7z
+++ b/completions/7z
@@ -49,9 +49,9 @@ _7z()
             local reset=$(shopt -po noglob)
             set -o noglob
             compopt -o filenames
-            local ifs=$IFS IFS=$'\n'
+            local IFS=$'\n'
             COMPREPLY=($(compgen -d -P${cur:0:2} -S/ -- "${cur:2}"))
-            IFS=$ifs
+            _comp_unlocal IFS
             $reset
             compopt -o nospace
             return
--- a/completions/_umount.linux
+++ b/completions/_umount.linux
@@ -25,9 +25,9 @@ _reply_compgen_array()
     ecur=${ecur//\'/\\\'}
 
     # Actually generate completions.
-    local ifs=$IFS
-    IFS=$'\n' eval 'COMPREPLY=(`compgen -W "$wlist" -- "${ecur}"`)'
-    IFS=$ifs
+    local IFS=$'\n'
+    COMPREPLY=($(compgen -W "$wlist" -- "${ecur}"))
+    _comp_unlocal IFS
 }
 
 # Unescape strings in the linux fstab(5) format (with octal escapes).
@@ -49,24 +49,23 @@ _linux_fstab()
 
     # Read and unescape values into COMPREPLY
     local fs_spec fs_file fs_other
-    local ifs="$IFS"
     while read -r fs_spec fs_file fs_other; do
         if [[ $fs_spec == [#]* ]]; then continue; fi
         if [[ ${1-} == -L ]]; then
             local fs_label=${fs_spec/#LABEL=/}
             if [[ $fs_label != "$fs_spec" ]]; then
                 __linux_fstab_unescape fs_label
-                IFS=$'\0'
+                local IFS=$'\0'
                 COMPREPLY+=("$fs_label")
-                IFS=$ifs
+                _comp_unlocal IFS
             fi
         else
             __linux_fstab_unescape fs_spec
             __linux_fstab_unescape fs_file
-            IFS=$'\0'
+            local IFS=$'\0'
             [[ $fs_spec == */* ]] && COMPREPLY+=("$fs_spec")
             [[ $fs_file == */* ]] && COMPREPLY+=("$fs_file")
-            IFS=$ifs
+            _comp_unlocal IFS
         fi
     done
 
--- a/completions/svcadm
+++ b/completions/svcadm
@@ -44,10 +44,9 @@ _smf_complete_fmri()
             # we generate all possibles abbreviations for the FMRI
             # no need to have a generic loop as we will have a finite
             # number of components
-            local ifs="$IFS"
-            IFS="/"
+            local IFS="/"
             set -- $fmri
-            IFS=$ifs
+            _comp_unlocal IFS
             case $# in
                 1) fmri_part_list=" $1" ;;
                 2) fmri_part_list=" $2 $1/$2" ;;
--- a/test/t/unit/test_unit_unlocal.py
+++ b/test/t/unit/test_unit_unlocal.py
@@ -1,0 +1,18 @@
+import pytest
+
+from conftest import assert_bash_exec
+
+
+@pytest.mark.bashcomp(cmd=None, ignore_env=r"^\+(VAR=|declare -f foo)")
+class TestUnlocal:
+    def test_1(self, bash):
+        cmd = (
+            "foo() { "
+            "local VAR=inner; "
+            "_comp_unlocal VAR; "
+            "echo $VAR; "
+            "}; "
+            "VAR=outer; foo; "
+        )
+        res = assert_bash_exec(bash, cmd, want_output=True).strip()
+        assert res == "outer"


akinomyoga:
Thank you for the PR! I think we should finally merge this PR. There are two things that should be addressed.

### 1. unset -v

We are making it explicit whether the `unset` target is a variable or a function by specifying the option `-v` or `-f`. In this case, we want to use `unset -v IFS`.

### 2. dynamic unset

`unset` in the same scope as the variable doesn't recover the original value but just makes it unset state (local unset).

```bash
$ function func { declare -p IFS; local IFS=b; declare -p IFS; unset -v IFS; declare -p IFS; }
$ IFS=a; func
declare -- IFS="a"
declare -- IFS="b"
declare -- IFS
```

In order to remove the local variable of the current scope and to make the variable of the previous context visible (dynamic unset), one needs to let `unset` find the variable through dynamic scoping, i.e., to call `unset` in a function:

```bash
$ function _comp_unlocal { unset -v "$@"; }
$ function func { declare -p IFS; local IFS=b; declare -p IFS; _comp_unlocal IFS; declare -p IFS; }
$ IFS=a; func
declare -- IFS="a"
declare -- IFS="b"
declare -- IFS="a"
```

In Bash 5.0+, one additionally needs to care about `shopt -s localvar_unset` in the above `_comp_unlocal`. You can add the function `_comp_unlocal` before the existing function `_upvar` since they are both basic utilities and utilize the behavior of dynamic unset.

sudonym1:
Mind blown on that _comp_unlocal function. Let me see if I understand why it works:

1. `func` declares some local variable.
2. `_comp_unlocal` inherits the local version of the variable.
3. `_comp_unlocal` unsets that variable.
4. We leave the scope of `_comp_unlocal` and bash sees the variable as unset, and so unshadows the global variable.

Is this correct understanding? Is there somewhere in the docs I can read more?

akinomyoga:
Ah, I think I first should have explained that the variable scoping in Bash is [dynamic scoping](https://en.wikipedia.org/wiki/Scope_(computer_science)#Lexical_scope_vs._dynamic_scope) like Perl's `local` or Emacs Lisp's `let`, etc. The dynamic scoping is explained in [Bash Reference Manual §3.3](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Functions)/¶11:

> The shell uses *dynamic scoping* to control a variable’s visibility within functions. With dynamic scoping, visible variables and their values are a result of the sequence of function calls that caused execution to reach the current function. The value of a variable that a function sees depends on its value within its caller, if any, whether that caller is the "global" scope or another shell function. This is also the value that a local variable declaration "shadows", and the value that is restored when the function returns.

Also, it should be noted that we need to differentiate three different situations for a variable in a specific scope: 1) a value is set, 2) the variable is unset, and 3) the variable does not exist. Case 2 is different from case 1 with an empty value `''` and also from case 3 where the variable of the same name of the previous context is visible.

Then the behavior of `unset` is explained in ¶14 in the same section:

> The `unset` builtin also acts using the same dynamic scope: if a variable is local to the current scope, `unset` will unset it; otherwise the `unset` will refer to the variable found in any calling scope as described above. If a variable at the current local scope is unset, it will remain so until it is reset in that scope or until the function returns. Once the function returns, any instance of the variable at a previous scope will become visible. If the unset acts on a variable at a previous scope, any instance of a variable with that name that had been shadowed will become visible.

In short, when the `unset` just finds the variable in the local scope (i.e., the same scope as the call of `unset`), `unset` performs "localvar-unset" which makes the variable of the local scope the unset state. When the `unset` finds the variable using dynamic scoping, `unset` performs "dynamic-unset" which removes the variable so that the variable in the second previous scope becomes visible.

sudonym1:
I misunderstood the key sentence: ```If the unset acts on a variable at a previous scope, any instance of a variable with that name that had been shadowed will become visible.```. Thank you for clarifying.

Looking at the source code, there is more explanation:

```
  /* If we're unsetting a local variable and we're still executing inside
     the function, just mark the variable as invisible.  The function
     eventually called by pop_var_context() will clean it up later.  This
     must be done so that if the variable is subsequently assigned a new
     value inside the function, the `local' attribute is still present.
     We also need to add it back into the correct hash table. */
  if (old_var && local_p (old_var) &&
	(old_var->context == variable_context || (localvar_unset && old_var->context < variable_context)))
    {
         // keep the variable around
    } else {
        // actually get rid of it
````

So how stable is this? Would it be better to just handle the unset global IFS case explicitly?

**Comment by akinomyoga** [comment]
> So how stable is this?

AFAIK, the behavior is consistent at least from Bash 3.0 except for the recent addition of `shopt -s localvar_unset`. I have now checked the behavior of Bash 2.* and it seems the behavior doesn't change at least since Bash 2.0. I haven't checked Bash 1.* since I don't have working binaries of Bash 1.*.

You can find a long discussion at [`bug-bash/2018-02/#65`](https://lists.gnu.org/archive/html/bug-bash/2018-02/threads.html#00065) [`bug-bash/2018-03/#0`](https://lists.gnu.org/archive/html/bug-bash/2018-03/threads.html#00000), which introduced `localvar_unset` and `localvar_inherit` IIRC.

> Would it be better to just handle the unset global IFS case explicitly?

Could you describe in more detail about what you mean by "*the unset global IFS case*"?

**Comment by sudonym1** [comment]
Sounds stable enough!

**Comment by sudonym1** [comment]
Updated. Is there a better place to slap that unit test?

**Comment by sudonym1** [comment]
Oops. Messed up the version check. Fixed.

File: completions/7z [hunk 1]
```diff
@@ -49,9 +49,9 @@ _7z()
             local reset=$(shopt -po noglob)
             set -o noglob
             compopt -o filenames
-            local ifs=$IFS IFS=$'\n'
+            local IFS=$'\n'
             COMPREPLY=($(compgen -d -P${cur:0:2} -S/ -- "${cur:2}"))
-            IFS=$ifs
+            _comp_unlocal
```
akinomyoga, line 54: missing IFS?
```suggestion
            _comp_unlocal IFS
```

completions/_umount.linux
```diff
@@ -25,9 +25,7 @@ _reply_compgen_array()
     ecur=${ecur//\'/\\\'}
 
     # Actually generate completions.
-    local ifs=$IFS
     IFS=$'\n' eval 'COMPREPLY=(`compgen -W "$wlist" -- "${ecur}"`)'
```
akinomyoga, line 28: In POSIX mode, the effect of the variable assignment before special builtins (including `eval`) remains after the execution of the command.

```bash
$ bash --norc
$ set -o posix
$ declare -p IFS
declare -- IFS="
"
$ IFS=: eval 'x=($SHELLOPTS)'
$ declare -p IFS x
declare -x IFS=":"
```

So we actually need to explicitly restore the original IFS. In this sense, the original code is a bit strange. I think we may rewrite it as

```suggestion
    local IFS=$'\n'
    COMPREPLY=($(compgen -W "$wlist" -- "${ecur}"))
    _comp_unlocal IFS
```

What do you think?

bash_completion
```diff
@@ -149,6 +149,22 @@ dequote()
...
+# Usage: local IFS=$'|'; _comp_unlocal IFS
+# Param: $* Variable names to be unset
+_comp_unlocal()
+{
+    local reset=$([[ ${BASH_VERSINFO[0]} -ge 5 ]] && {
+                      shopt -p localvar_unset
+                      shopt -u localvar_unset
+                  })
+    unset -v "$@"
+    $reset
```
akinomyoga, line 165: I think we here assume `IFS=$' \t\n'` or unset `IFS` to make `$reset` work, but we don't know the actual value of the global `IFS`. Can you use a different approach? Also, I actually don't like the subshell `$()` which has the fork cost.

Actually, there are the same problems in other places, which should be resolved in the future but that doesn't need to be addressed in this PR. Maybe we can open a discussion about what would be a better replacement for the `reset=$(shopt -p)` / `$reset` idiom currently used in many places.

completions/_umount.linux
```diff
@@ -49,24 +47,23 @@ _linux_fstab()
...
         if [[ $fs_spec == [#]* ]]; then continue; fi
         if [[ ${1-} == -L ]]; then
             local fs_label=${fs_spec/#LABEL=/}
             if [[ $fs_label != "$fs_spec" ]]; then
                 __linux_fstab_unescape fs_label
-                IFS=$'\0'
+                local IFS=$'\0'
                 COMPREPLY+=("$fs_label")
-                IFS=$ifs
+                _comp_unlocal
```
akinomyoga, line 58: missing IFS. By the way, I don't see what is the intent of the original code: Does anything change here in the behavior by setting `IFS=$'\0'`? Do you have an idea?

completions/_umount.linux
```diff
@@ -49,24 +47,23 @@ _linux_fstab()
...
             fi
         else
             __linux_fstab_unescape fs_spec
             __linux_fstab_unescape fs_file
-            IFS=$'\0'
+            local IFS=$'\0'
             [[ $fs_spec == */* ]] && COMPREPLY+=("$fs_spec")
             [[ $fs_file == */* ]] && COMPREPLY+=("$fs_file")
-            IFS=$ifs
+            _comp_unlocal
```
akinomyoga, line 66: missing IFS

completions/svcadm
```diff
@@ -44,10 +44,9 @@ _smf_complete_fmri()
             # we generate all possibles abbreviations for the FMRI
             # no need to have a generic loop as we will have a finite
             # number of components
-            local ifs="$IFS"
-            IFS="/"
+            local IFS="/"
             set -- $fmri
-            IFS=$ifs
+            _comp_unlocal
```
akinomyoga, line 49: missing IFS

test/t/test_bash.py
```diff
@@ -1,7 +1,19 @@
...
+    def test_comp_unlocal(self, bash):
+        cmd = ( 'foo() { '
+                      'local VAR=inner; '
+                      '_comp_unlocal VAR; '
+                      'echo $VAR; '
+                '}; '
+                'VAR=outer; foo; unset VAR foo; '
+        )
+        res = assert_bash_exec(bash, cmd, want_output=True).strip()
+        assert res == "outer"
```
akinomyoga, line 19: Thank you for adding the test! The test for the utilities defined in the main `bash_completion` should be placed in `test/t/unit` i.e. `test/t/unit/test_unit_unlocal.py` for this case.

akinomyoga (changes requested):
Thank you for updating! Could you see the following comments?

completions/_umount.linux
sudonym1: I think I am learning a lot about shell! I will apply this suggestion.

completions/_umount.linux
sudonym1: There is some attempt at escaping quotes. The NULs are injected above in __linux_fstab_unescape.

bash_completion
sudonym1: Perhaps something leveraging a map, and `shopt -q` could be a general solution here. Let me go try something.

sudonym1:
```
store_options() {
    declare -n state=$1
    shift
    for opt in $@
    do
        shopt -q $opt && state[$opt]="-s" || state[$opt]="-u"
    done
}

restore_options() {
    declare -n state=$1
    shift
    for opt in ${!state[@]}
    do
        shopt ${state[$opt]} $opt
    done
}

some_function() {
    declare -A opts
    set -x
    store_options opts localvar_unset

    restore_options opts
    set +x
}
```
What about this?

completions/_umount.linux
akinomyoga: Hmm, I still don't understand why they are needed. I don't have an idea why `__linux_fstab_unescape` can be related to `IFS=$'\0'`. First of all, we cannot include NUL in the variable values in Bash. Next, even if we could include NUL in the variables, `COMPREPLY+=("$fs_label")` will add the value as *a single array element* regardless of `IFS` because it is quited by `"..."`.

I have checked the change history of the relevant code section.

- https://github.com/scop/bash-completion/commit/99b289474ead02cc8c629ceec1192ce9f0b4f84e#diff-a33e10856930590d5777b4de609f66b088b7477e930a200bb2439e83724ef328L51-R87
- https://github.com/scop/bash-completion/commit/1c4b461882c94febb39a550e0fe3065208139757#diff-a33e10856930590d5777b4de609f66b088b7477e930a200bb2439e83724ef328R55-R60
- https://github.com/scop/bash-completion/commit/2f61acd068a67e630dd28ce04157908398dd8798#diff-a33e10856930590d5777b4de609f66b088b7477e930a200bb2439e83724ef328R50-R53
- https://github.com/scop/bash-completion/commit/4375c4b94ea7bdd370cd771aeb3483f36a42bd9f#diff-a33e10856930590d5777b4de609f66b088b7477e930a200bb2439e83724ef328L50-R62

It seems the code tries to evaluate the octal sequences of `\ooo` and split the field with the resulting NULs (encoded as `\000` in the original data). But, if that is the original intent, the implementation is actually broken from the beginning. It has never worked. We need to re-implement it using `read -d ''` which can be used for the NUL-separated data.

bash_completion
akinomyoga: Maybe we can just save the value of `$BASHOPTS`, and then later test it using `[[ :$saved_bashopts: == *:"$opt":* ]]`. This is actually what I do in my personal scripts.


```bash
_comp_restore_shopt() {
    local bashopts=$1; shift
    local opt
    for opt; do
        if [[ :$bashopts: == *:"$opt":* ]]; then
            shopt -s "$opt"
        else
            shopt -u "$opt"
        fi
    done
}
some_function() {
    local bashopts=$BASHOPTS
    set -x
    shopt -u localvar_unset

    _comp_restore_shopt "$bashopts" localvar_unset
    set +x
}
```

bash_completion
```diff
@@ -149,6 +149,24 @@ dequote()
...
 
+# Unset the given variables across a scope boundary. Useful for unshadowing
+# global scoped variables. Note that simply calling unset on a local variable
+# will not unshadow the global variable. Rather, the result will be a local
+# variable in an unset state.
+# Usage: local IFS=$'|'; _comp_unlocal IFS
+# Param: $* Variable names to be unset
+_comp_unlocal()
+{
+    if [[ ${BASH_VERSINFO[0]} -ge 5 ]] && shopt -q localvar_unset
```
akinomyoga, line 160:
```suggestion
    if ((BASH_VERSINFO[0] >= 5 )) && shopt -q localvar_unset
```

completions/_umount.linux
```diff
@@ -25,9 +25,9 @@ _reply_compgen_array()
     ecur=${ecur//\'/\\\'}
 
     # Actually generate completions.
-    local ifs=$IFS
-    IFS=$'\n' eval 'COMPREPLY=(`compgen -W "$wlist" -- "${ecur}"`)'
-    IFS=$ifs
+    local IFS=$'\n'
+    eval 'COMPREPLY=(`compgen -W "$wlist" -- "${ecur}"`)'
```
akinomyoga, line 29: The argument of `eval` is a fixed string, so we actually don't need to use `eval`. Also, we prefer `$(...)` to the older form ``` `...` ``` deprecated by POSIX.

```suggestion
    COMPREPLY=($(compgen -W "$wlist" -- "${ecur}"))
```

akinomyoga (commented):
Just suggestions of cosmetic changes

bash_completion
```diff
@@ -149,6 +149,24 @@ dequote()
     eval printf %s "$1" 2>/dev/null
 }
 
+# Unset the given variables across a scope boundary. Useful for unshadowing
+# global scoped variables. Note that simply calling unset on a local variable
+# will not unshadow the global variable. Rather, the result will be a local
+# variable in an unset state.
+# Usage: local IFS=$'|'; _comp_unlocal IFS
```
akinomyoga, line 156:
```suggestion
# Usage: local IFS='|'; _comp_unlocal IFS
```

bash_completion
sudonym1: I like this approach overall, and it is straightforward. The only concern I have is that it is easy to forget an argument when calling `_comp_restore_shopt`. That could be avoided by looping over all options like this:

```
_comp_restore_shopt() {
    local bashopts=$1; shift
    local IFS=:
    for opt in $BASHOPTS $bashopts; do
        if [[ :$bashopts: == *:"$opt":* ]]; then
            shopt -s "$opt"
        else
            shopt -u "$opt"
        fi
    done
}
```

I think this approach is likely not worth the loss of efficiency compared with your approach.

sudonym1:
Ok, I think I have all of the feedback folded in. I held off on the _comp_shopt_restore stuff, as that makes more sense to me as a separate commit.

akinomyoga (approved):

akinomyoga:
Thank you! Now it LGTM. Let's run the CI test now.

Can I reopen these two unresolved conversations although they should be addressed in separate PRs? This is just because I wanted to later work on them and would like to keep attention to them. Otherwise, I can possibly forget them.

akinomyoga (changes requested):
@sudonym1 It seems the `pre-commit` hook fails. Can you set `pre-commit` up and fix the issues? The way to set it up is a bit non-trivial but you can reference the comment at https://github.com/scop/bash-completion/pull/683#issuecomment-1010579492.

sudonym1:
Oops. Went and fixed the lints before I saw your comment. Will recheck with the tool you mention.

sudonym1:
Ok, my environment doesn't support docker ATM so I ran the shfmt tool manually. Everything else seems to pass now.

akinomyoga:
OK, I rerun the CI test.

sudonym1:
> Can I reopen these two unresolved conversations although they should be addressed in separate PRs?

Of course!

akinomyoga (approved):

akinomyoga:
Somehow the test in `fedoradev` failed but it did pass in [the previous time](https://github.com/scop/bash-completion/actions/runs/1737487561). The difference between the previous time is just formatting made by `shfmt`  87d1672c...f2412520, so I don't think that change generated the error. Another error in `debian10, true, none` in the previous test is caused by the timeout we often see. Now I think this PR is ready to merge.

@scop Do you have any comments?

sudonym1:
> is caused by the timeout we often see

Is there a ticket for this? I wouldn't mind taking a look this upcoming weekend.

akinomyoga:
Thanks, I have merged it.

> Is there a ticket for this? I wouldn't mind taking a look this upcoming weekend.

Thank you! I don't remember any. I have quickly searched, but there seems none explicitly mentioning this issue. #136 and #137 seem to be the timeouts caused by the test case itself. I'm not sure whether the cause of #139 is the test case itself or our test framework. There is a report #430 saying that the test *fails* or *is flaky* in ARM architecture with high loads but there is no further information. #573 is related to refactoring the test suite. I couldn't find any issue explicitly mentioning the timeout failure.

scop:
Timeouts are a quite general failure mode in the test suite.

Sometimes they occur when tests fail to process expected/unexpected output or the lack of it, kind of masking the root cause. These are fairly reproducible and straightforward to fix.

Sometimes they occur due to something else, perhaps a command actually stalling every now and then, this would be hard to reproduce and fix.

But the most "interesting" thing to me (and the one I guess we're referring to) here is that I think we may still have something wrong in our test suite that triggers this in unknown situations that is not caused by the things mentioned here. Would be awesome to track these down and fix this category of issues. I believe enabling parallel tests provokes the failures more often -- at least it does provoke _some_ failures more. Parallel running happens by default in local dockerized test runs, and can be enabled on the host without docker e.g. like
```shellsession
$ PYTESTFLAGS="--numprocesses=auto --dist=loadfile" make check
```

sudonym1:
It seems I have been getting (un)lucky. I've run the fedora docker container ~10 times with parallel pytests, and no failures so far. I will try a different approach. It could be these issues are related to the performance characteristics of the host computer.

akinomyoga:
I guess the problem is caused by some race conditions or the synchronization problem of input and output pipes in the communications between `pexpect` and the Bash process. There is already some [`sleep`](https://github.com/scop/bash-completion/blob/7e7928d0e8bf2071e09749e4e005a972acf2aa53/test/t/conftest.py#L758-L759) that I guess is related.

Actually, I'm not sure if the patterns specified [here](https://github.com/scop/bash-completion/blob/7e7928d0e8bf2071e09749e4e005a972acf2aa53/test/t/conftest.py#L763-L779) covers all the possible cases of the Bash outputs. What happens when some garbage outputs remained in the stdout before writing something to stdin (though I'm not even sure whether that can happen because we are checking MAGIC_MARK between every roundtrip)? The accepted patterns in the current testing code are

- `/\r\n<PROMPT><COMMAND>.*<MAGIC_MARK>/`
- `/^<MAGIC_MARK>/`
- `/^([^\r]+)<MAGIC_MARK>$/`
- `/^([^\r].*)<MAGIC_MARK>$/`

I feel we should just accept `[\s\S]*<MAGIC_MARK>` and then examine the extracted outputs, or at least we can add another fallback pattern in the above list in order to check what `pexpect` actually captured.

completions/_umount.linux
akinomyoga: I have opened a PR at #690 for this, so let me close the conversation.

bash_completion
akinomyoga: I have created an issue at #691.

akinomyoga:
Maybe we can create a new issue or a Draft PR for this timeout issue of the test suite.

PR #692, opened 2022-01-31
test(ubuntu14): switch Python to Pyston 2.3.1

scop:
deadsnakes packages for Ubuntu 14.04 and 16.04 no longer exist, switch to portable Pyston 2.3.1. Go with Pyston instead of PyPy for (hopefully) better compatibility with CPython.

- https://github.com/deadsnakes/issues/issues/195
- https://github.com/scop/bash-completion/runs/4995567978?check_suite_focus=true#step:4:548

The primary reasons Ubuntu 14 is still kept around in the test matrix are that it's our only bash 4.3 environment at the moment, and it's nice to be able to test against various older versions of tools -- the coverage with e.g. CentOS 7 isn't as big.

### DIFF
--- a/test/docker/ubuntu14/Dockerfile
+++ b/test/docker/ubuntu14/Dockerfile
@@ -9,12 +9,8 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         make \
         software-properties-common \
         xvfb \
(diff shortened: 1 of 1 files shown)


scop:
Pyston 2.3.2 no longer works with Ubuntu < 18.04, so stick with 2.3.1 for now.

PR #702, opened 2022-02-20
feat: gpg option arg completion improvements

scop:
Contains a somewhat unrelated style change which I'm too lazy to submit separately :P

### DIFF
--- a/completions/gpg
+++ b/completions/gpg
@@ -10,7 +10,9 @@ _gpg()
             _filedir
             return
             ;;
-        --export | --sign-key | --lsign-key | --nrsign-key | --nrlsign-key | --edit-key)
+        --export | --sign-key | --lsign-key | --nrsign-key | --nrlsign-key | \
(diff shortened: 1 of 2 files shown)


akinomyoga:
I'm not familiar with these GPG keys but it seems there are more options that receive keys as arguments.

- `--list-keys` / `-k` / `--list-public-keys`, `--list-secret-keys` / `-K` ... optionally receive keys as arguments?
- `--locate-external-keys` ... synonym to `--locate-keys`
- `--delete-secret-keys`
- `--search-keys`
- `--export-secret-keys`, `--export-secret-subkeys`, `--export-ssh-key` ... some variants of `--export`, but it is not explicit whether they accept keys in arguements.
- `--try-secret-key`

scop:
Thanks! Added all but `--export-ssh-key`, as I didn't have any keys around it would work with.

akinomyoga (approved):

PR #708, opened 2022-02-25
feat(_comp_expand_glob): Add a utility to safely perform pathname expansions (Fix #705)

### LINKED ISSUE #705: shopt failglob causes bash_completiion to give error in _included_ssh_config_files
## Describe the bug

Running the shell with `shopt -s failglob` causes the `ssh` config file glob expansion in `_included_ssh_config_files()` to fail with an error.

## To reproduce
~~~~~
bash-5.0$ shopt -s failglob
bash-5.0$ ssh <TAB>bash: no match: /etc/ssh/ssh_config.d/*.conf
~~~~~
## Expected behavior

Completion scripts should work without error even with `failglob` set.

- [`GNU/linux/Ubuntu 20.04.4 LTS`] Operating system name/distribution and version:
- [`5.0.17(1)-release`] bash version, `echo "$BASH_VERSION"`:
- [2.11.0] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`:

## Additional context

Latest completion scripts downloaded via `git clone` and copied into Ubuntu.

## Debug trace
~~~~~
+ _included_ssh_config_files /etc/ssh/ssh_config
+ (( 1 < 1 ))
+ local configfile i f
+ configfile=/etc/ssh/ssh_config
++ shopt -po noglob
+ local 'IFS= 
' 'reset=set -o noglob'
+ set -o noglob
+ included=($(command sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' "${configfile}"))
++ command sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' /etc/ssh/ssh_config
++ sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' /etc/ssh/ssh_config
+ local included
+ set -o noglob
+ [[ -n /etc/ssh/ssh_config.d/*.conf ]]
+ for i in "${included[@]}"
+ [[ /etc/ssh/ssh_config.d/*.conf =~ ^~.*|^/.* ]]
+ __expand_tilde_by_ref i
+ [[ /etc/ssh/ssh_config.d/*.conf == \~* ]]
+ set +o noglob
bash: no match: /etc/ssh/ssh_config.d/*.conf
~~~~~

akinomyoga:
Fixes #705 and similar problems

## Description

Pathname expansions are used in many places in the codebase, but it seems only a small part of these pathname expansions considers the setting of `nullglob`, and only `mysql` considers the setting of `failglob`. All the pathname expansions except for that in `mysql` can possibly fail due to `failglob`. Actually, there are even other settings that we need to care about for the pathname expansions, such as `nounset`, `dotglob`, `GLOBIGNORE`, and `nocaseglob`.

We don't want to always write codes to check all these settings in-place, so I suggest introducing a shell function `_comp_expand_glob` to perform the pathname expansions.

## Commits

This PR contains several changes.

- ecc384df This is an extension to `conftest.py (bash_env_saved)` that allows an arbitrary change to the variables and settings within the `bash_env` context. This is used in [a test of `_comp_expand_glob`](https://github.com/scop/bash-completion/commit/fae52a81dd996bb601dec5bf2842ebcb34913846#diff-7b154d46af2d8ced2f735ba1a2ee1bf1b133a0e5fd7ee5c953056f4bec44f5e0R99).
- f0a4028d This is also an extension to `conftest.py (bash_env_saved)` that allows saving/restoring the settings of `set -o NAME` and checking the unintended changes to them (just like the existing ones for `shopt` and variables). This is also used in [a test of `_comp_expand_glob`](https://github.com/scop/bash-completion/commit/fae52a81dd996bb601dec5bf2842ebcb34913846#diff-7b154d46af2d8ced2f735ba1a2ee1bf1b133a0e5fd7ee5c953056f4bec44f5e0R58).
- 59ec4602 This is the function to perform the pathname expansions. (This function is actually a simplified version of my functions [`ble/util/eval-pathname-expansion`](https://github.com/akinomyoga/ble.sh/blob/b9fdaabd4c551177ea0857ac1c258832889ad447/src/util.sh#L3597-L3642) and [`ble/complete/util/eval-pathname-expansion`](https://github.com/akinomyoga/ble.sh/blob/master/lib/core-complete.sh#L2429-L2509)).
- 36a540f5 Fixes #705.
- 89dc92ea This fixes the other pathname expansions that I could find (I guess there are still other places that use the pathname expansions).
- 3b858481 This is a fix of a problem of `mysql` completion uncovered by the new `_comp_expand_glob`, which has originally happened with `shopt -s nullglob`.
- 5d9d3547 This is another fix to `mysql` for a problem that has existed for a long time.

## Discussion

There are several things that I want to discuss for the detailed design.

### 1. `dotglob`

While performing the pathname expansions, should we turn `dotglob` on or turn it off? Currently, It's turned off as that is the default of Bash. But maybe it is better to be turned on in some cases.

### 2. `nocaseglob`

Currently, `_comp_expand_glob` doesn't change the settings for `nocaseglob` for its internal expansions, but maybe we should also normalize the behavior of the case matching regardless of the current user settings of `nocaseglob`.

### 3. Naming of `_comp_expand_glob`

Other possibilities include
- `_comp{,_util}_{,_}expand_{pathname{,s},glob}`

### 4. `extglob`

I think we are requiring that `extglob` is turned on, but maybe we can explicitly adjust `extglob` as well as the other options.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1148,15 +1148,20 @@ _mac_addresses()
 #
 _configured_interfaces()
 {
+    local -a files
     if [[ -f /etc/debian_version ]]; then
         # Debian system
+        _comp_expand_glob files '/etc/network/interfaces /etc/network/interfaces.d/*'
+        ((${#files[@]})) || return 0
         COMPREPLY=($(compgen -W "$(command sed -ne 's|^iface \([^ ]\{1,\}\).*$|\1|p' \
-            /etc/network/interfaces /etc/network/interfaces.d/* 2>/dev/null)" \
+            "${files[@]}" 2>/dev/null)" \
             -- "$cur"))
     elif [[ -f /etc/SuSE-release ]]; then
         # SuSE system
+        _comp_expand_glob files '/etc/sysconfig/network/ifcfg-*'
+        ((${#files[@]})) || return 0
         COMPREPLY=($(compgen -W "$(printf '%s\n' \
-            /etc/sysconfig/network/ifcfg-* |
+            "${files[@]}" |
             command sed -ne 's|.*ifcfg-\([^*].*\)$|\1|p')" -- "$cur"))
     elif [[ -f /etc/pld-release ]]; then
         # PLD Linux
@@ -1165,8 +1170,10 @@ _configured_interfaces()
             command sed -ne 's|.*ifcfg-\([^*].*\)$|\1|p')" -- "$cur"))
     else
         # Assume Red Hat
+        _comp_expand_glob files '/etc/sysconfig/network-scripts/ifcfg-*'
+        ((${#files[@]})) || return 0
         COMPREPLY=($(compgen -W "$(printf '%s\n' \
-            /etc/sysconfig/network-scripts/ifcfg-* |
+            "${files[@]}" |
             command sed -ne 's|.*ifcfg-\([^*].*\)$|\1|p')" -- "$cur"))
     fi
 }
@@ -1422,12 +1429,12 @@ _xinetd_services()
 {
     local xinetddir=${_comp__test_xinetd_dir:-/etc/xinetd.d}
     if [[ -d $xinetddir ]]; then
-        local IFS=$' \t\n' reset=$(shopt -p nullglob)
-        shopt -s nullglob
-        local -a svcs=($xinetddir/!($_comp_backup_glob))
-        $reset
-        ((!${#svcs[@]})) ||
+        local -a svcs
+        _comp_expand_glob svcs '$xinetddir/!($_comp_backup_glob)'
+        if ((${#svcs[@]})); then
+            local IFS=$'\n'
             COMPREPLY+=($(compgen -W '"${svcs[@]#$xinetddir/}"' -- "${cur-}"))
+        fi
     fi
 }
 
@@ -1438,12 +1445,9 @@ _services()
     local sysvdirs
     _sysvdirs
 
-    local IFS=$' \t\n' reset=$(shopt -p nullglob)
-    shopt -s nullglob
-    COMPREPLY=(
-        $(printf '%s\n' ${sysvdirs[0]}/!($_comp_backup_glob|functions|README)))
-    $reset
+    _comp_expand_glob COMPREPLY '${sysvdirs[0]}/!($_comp_backup_glob|functions|README)'
 
+    local IFS=$'\n'
     COMPREPLY+=($({
         systemctl list-units --full --all ||
             systemctl list-unit-files
@@ -1474,7 +1478,7 @@ _service()
         _services
         [[ -e /etc/mandrake-release ]] && _xinetd_services
     else
-        local sysvdirs
+        local IFS=$'\n' sysvdirs
         _sysvdirs
         COMPREPLY=($(compgen -W '`command sed -e "y/|/ /" \
             -ne "s/^.*\(U\|msg_u\)sage.*{\(.*\)}.*$/\2/p" \
@@ -1726,7 +1730,9 @@ _terms()
         {
             toe -a || toe
         } | awk '{ print $1 }'
-        find /{etc,lib,usr/lib,usr/share}/terminfo/? -type f -maxdepth 1 |
+        _comp_expand_glob dirs '/{etc,lib,usr/lib,usr/share}/terminfo/?'
+        ((${#dirs[@]})) &&
+            find "${dirs[@]}" -type f -maxdepth 1 |
             awk -F/ '{ print $NF }'
     } 2>/dev/null)" -- "$cur"))
 }
--- a/bash_completion
+++ b/bash_completion
@@ -1786,7 +1786,7 @@ _included_ssh_config_files()
 {
     (($# < 1)) &&
         echo "bash_completion: $FUNCNAME: missing mandatory argument CONFIG" >&2
-    local configfile i f
+    local configfile i files f
     configfile=$1
 
     local IFS=$' \t\n' reset=$(shopt -po noglob)
@@ -1809,15 +1809,16 @@ _included_ssh_config_files()
         fi
         __expand_tilde_by_ref i
         # In case the expanded variable contains multiple paths
-        set +o noglob
-        for f in $i; do
-            if [[ -r $f ]]; then
-                config+=("$f")
-                # The Included file is processed to look for Included files in itself
-                _included_ssh_config_files $f
-            fi
-        done
-        $reset
+        _comp_expand_glob files '$i'
+        if ((${#files[@]})); then
+            for f in "${files[@]}"; do
+                if [[ -r $f ]]; then
+                    config+=("$f")
+                    # The Included file is processed to look for Included files in itself
+                    _included_ssh_config_files $f
+                fi
+            done
+        fi
     done
 } # _included_ssh_config_files()
 
--- a/bash_completion
+++ b/bash_completion
@@ -260,6 +260,55 @@ _upvars()
     done
 }
 
+# Get the list of filenames that match with the specified glob pattern.
+# This function does the globbing in a controlled environment, avoiding
+# interference from user's shell options/settings or environment variables.
+# @param $1 array_name  Array name
+#   The array name should not start with the double underscores "__".  The
+#   array name should not be "GLOBIGNORE".
+# @param $2 pattern     Pattern string to be evaluated.
+#   This pattern string will be evaluated using "eval", so brace expansions,
+#   parameter expansions, command substitutions, and other expansions will be
+#   processed.  The user-provided strings should not be directly specified to
+#   this argument.
+_comp_expand_glob()
+{
+    if (($# != 2)); then
+        printf 'bash-completion: %s: unexpected number of arguments\n' "$FUNCNAME" >&2
+        printf 'usage: %s ARRAY_NAME PATTERN\n' "$FUNCNAME" >&2
+        return 2
+    elif [[ $1 == @(GLOBIGNORE|__*|*[^_a-zA-Z0-9]*|[0-9]*|'') ]]; then
+        printf 'bash-completion: %s: invalid array name "%s"\n' "$FUNCNAME" "$1" >&2
+        return 2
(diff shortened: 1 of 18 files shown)


bash_completion
```diff
@@ -1082,15 +1132,19 @@ _mac_addresses()
 #
 _configured_interfaces()
 {
+    local -a files
     if [[ -f /etc/debian_version ]]; then
         # Debian system
+        _comp_expand_glob files '/etc/network/interfaces /etc/network/interfaces.d/*'
         COMPREPLY=($(compgen -W "$(command sed -ne 's|^iface \([^ ]\{1,\}\).*$|\1|p' \
-            /etc/network/interfaces /etc/network/interfaces.d/* 2>/dev/null)" \
+            "${files[@]}" 2>/dev/null)" \
```
scop, line 1140: `${#files[@]}` check (bash 4.2 nounset protection) missing?

test/t/conftest.py
```diff
@@ -512,7 +513,33 @@ def _save_cwd(self):
             self.cwd_changed = True
             self._copy_variable("PWD", "%s_OLDPWD" % self.prefix)
 
+    def _check_set(self, name: str):
+        if self.saved_set[name] != 1:
```
scop, line 517: Let's see if we can find more meaningful values for this than magic 1 and 2 ints. An enum?

(saved_shopt and saved_variables on the other hand look like they would be better done as `[str, bool]`, but that's out of scope here.)

scop (changes requested):
TBH, I wish we didn't have to do something like this :/ But I guess we do want it.

bash_completion
```diff
@@ -245,6 +245,56 @@ _upvars()
...
+#   array name should not be "GLOBIGNORE".
+# @param $2 pattern     Pattern string to be evaluated.
+#   This pattern string will be evaluated using "eval", so brace expansions,
+#   parameter expansions, command substitutions, and other expansinos will be
+#   processed.  The user-provided strings should not be directly specified to
+#   this argument.
+_comp_expand_glob()
+{
+    if (($# != 2)); then
+        printf '%s\n' "bash-completion: $FUNCNAME: unexpected number of arguments." >&2
```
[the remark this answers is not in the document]
akinomyoga: Thank you for the review! I will address them one by one.

> just simply `echo`,

Simple `echo` has a problem with `shopt -s xpg_echo` being turned on, so I don't want to use `echo` with parameter expansions. But maybe it is fine for the present case because `$FUNCNAME` will not contain backslashes unless the user unsets `FUNCNAME` and rewrites it. I will just apply the suggested change.

bash_completion
```diff
@@ -245,6 +245,56 @@ _upvars()
...
+    # Save and adjust the settings.  Note: BASHOPTS is available in bash-4.1+.
+    local __original_opts=$SHELLOPTS:$BASHOPTS
+    set +o noglob
+    shopt -s nullglob
+    shopt -u failglob dotglob
+
+    # Also the user's GLOBIGNORE may affect the result of pathname expansions.
+    # Note: bash-3.0 and 3.1 have bugs that fails to revert the effect of local
+    # GLOBIGNORE after the function returns.  We now assume bash-4.2+ so can
+    # use "local GLOBIGNORE" without caring it.
```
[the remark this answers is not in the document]
akinomyoga: Thanks. I will also remove these explanations. I think I was afraid that some users learning Bash by reading bash-completion might apply these techniques without knowing the limitations, but I guess that's unnecessary fear.

bash_completion
akinomyoga: Right. That's my oversight. Thank you.

akinomyoga:
I have added fixes 5d9d3547...3f86a094 and rebased & squashed changes.

test/t/conftest.py
akinomyoga: I have introduced an enum in 34fdc3435ad7cae8fffe993db6e73135bf506b12

bash_completion
akinomyoga: fixed in 3f86a094a0aa76cec2783f333a1fe2f20933b75e

bash_completion
```diff
@@ -245,6 +245,56 @@ _upvars()
     done
 }
 
+# Get the list of filenames that match with the specified glob pattern.
+# @param $1 array_name  Array name
+#   The array name should not start with the double underscores "__".  The
+#   array name should not be "GLOBIGNORE".
+# @param $2 pattern     Pattern string to be evaluated.
+#   This pattern string will be evaluated using "eval", so brace expansions,
+#   parameter expansions, command substitutions, and other expansinos will be
```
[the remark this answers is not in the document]
akinomyoga: Applied in 9f3aae24

bash_completion
```diff
@@ -245,6 +245,56 @@ _upvars()
...
+    if (($# != 2)); then
+        printf '%s\n' "bash-completion: $FUNCNAME: unexpected number of arguments." >&2
+        printf '%s\n' "usage: $FUNCNAME ARRAY_NAME PATTERN." >&2
+        return 2
+    elif [[ $1 == @(GLOBIGNORE|__*|*[^_a-zA-Z0-9]*|[0-9]*|'') ]]; then
+        printf '%s\n' "bash-completion: $FUNCNAME: invalid array name '$1'." >&2
+        return 2
+    fi
+
+    # Save and adjust the settings.  Note: BASHOPTS is available in bash-4.1+.
```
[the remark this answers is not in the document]
akinomyoga: Applied in 9f3aae24

bash_completion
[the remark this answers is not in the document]
akinomyoga: Applied in 9f3aae24

bash_completion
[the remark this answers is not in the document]
akinomyoga: Updated in 9f3aae24

bash_completion
```diff
@@ -1368,12 +1420,9 @@ _services()
...
     _sysvdirs
 
-    local IFS=$' \t\n' reset=$(shopt -p nullglob)
-    shopt -s nullglob
-    COMPREPLY=(
-        $(printf '%s\n' ${sysvdirs[0]}/!($_backup_glob|functions|README)))
-    $reset
+    _comp_expand_glob COMPREPLY '${sysvdirs[0]}/!($_backup_glob|functions|README)'
 
+    local IFS=$' \t\n'
```
scop, line 1425: I think this may have been added originally for `$reset` to work and thus might not be needed any longer -- at least this PR is removing it with another similar change in `_xinetd_services` above and elsewhere.

completions/ipmitool
```diff
@@ -16,11 +16,14 @@ _ipmitool()
...
-                command ls -d /dev/ipmi* /dev/ipmi/* /dev/ipmidev/* \
-                    2>/dev/null | command sed -ne 's/^[^0-9]*\([0-9]\{1,\}\)/\1/p'
-            )" \
-                -- "$cur"))
+            local -a files
+            _comp_expand_glob files '/dev/ipmi* /dev/ipmi/* /dev/ipmidev/*'
+            ((${#files[@]})) &&
+                COMPREPLY=($(compgen -W "$(
+                    printf '%s\n' "${files[@]}" |
+                        command sed -ne 's/^[^0-9]*\([0-9]\{1,\}\)/\1/p'
```
scop, line 24: Now that we have the candidates in a bash array, I suppose we could loop over and filter them in pure bash instead of doing this. Not a biggie, leaving up to you to decide what to do about it or to leave as is.

bash_completion
```diff
@@ -245,6 +245,53 @@ _upvars()
     done
 }
 
+# Get the list of filenames that match with the specified glob pattern.
```
scop, line 248: I think a little reminder here why we should be using this this would be a nice touch.

```suggestion
# Get the list of filenames that match with the specified glob pattern.
# This function does the globbing in a controlled environment, avoiding
# interference from user's shell options/settings or environment variables.
```

scop (commented):
Barring a few small comments, and resolving conflicts, looks good to me; feel free to merge after addressing (or as always, feel free to bounce back for more opinions if you want).

I'm not sure if the hooks stuff discussed elsewhere will eventually lead into a generic ability to make the environment/options/settings what we want and reset them back at start and end of all our completions (e.g. `_init_completion` and `trap` trickery). If they do, this could become obsolete, but that stuff might take some time and might never actually realize, so I think this is fair enough at least in the meantime.

bash_completion
akinomyoga: Sorry for the delay. I think I kept it so that it doesn't affect the later word splitting `COMPREPLY+=($(...))`, and `IFS` removal in `_xinetd_services` is an oversight. 8b267ebf. I have also added `IFS=$'\n'` to protect `_service` for the splitting of `compgen` results.

completions/ipmitool
akinomyoga: Refactored in 24e095b9. We can actually implement it using `compgen -X` without a loop.

bash_completion
akinomyoga: Thank you! I have applied it 62e2440f.

akinomyoga:
> I'm not sure if the hooks stuff discussed elsewhere will eventually lead into a generic ability to make the environment/options/settings what we want and reset them back at start and end of all our completions (e.g. `_init_completion` and `trap` trickery). If they do, this could become obsolete, but that stuff might take some time and might never actually realize, so I think this is fair enough at least in the meantime.

Yeah, I have also thought about it. Something we need to care about is whether the scope of these shell settings for the pathname expansions can be safely expanded to the entire completion. For example, if there are any completion functions that generate the completions depending on the user's current shell settings, expanding the scope of the adjusted shell settings can affect the results. But I'm not sure if there are any existing ones.

akinomyoga:
Merging

PR #724, opened 2022-04-06
host exclusion patterns

### LINKED ISSUE #720: add mechanism to filter out known hostnames
Hey.

It would be nice to have a simple way to filter out known hostnames that bash-completion gathers from files like `/etc/hosts` or SSH’s various config files.

My use case is that I have some rather complex SSH config with special settings for many single hosts. Typically I don't use/need these hostsnames so I wouldn't want them to show up in completions as they break fast completions of those that I really use all the time, e.g. consider:
* `*.rarelyused.example.org`
* `*.oftenused.example.org`
maybe with pairs belonging together, one being the actual host, the other it's service processor, e.g. `host.srv.example.org` and `host.sp.example.org`.

The idea would be something like a regular expression that is stored in some file (which is considered by bash-completion) and that can be used to remove any unwanted names, e.g. 
`^.+\.rarelyused\.example\.org$`
would clear all the unwanted one.

A simple script could then add/remove that regexp and act as a switch,... so when I want to use these host names, I switch them on, otherwise off.

Such file could go into `.config/bash-completion/` which may perhaps even be used for further purposes, e.g. a subdir `.config/hosts.d/*.conf` which contains files with further hostnames (one per line per file),... which shall be completed but do not occur elsewhere.

Cheers,
Chris.

calestyo:
This would fix #720.

See commit messages for details… and please double check all my code (I'm no `bash-completion` expert).

Thanks! 😊

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1978,7 +1978,7 @@ _known_hosts_real()
         local host_exclusion_pattern="$( printf '%s' "${BASH_COMPLETION_HOST_EXCLUSION_PATTERN-}"  |  command -p awk 1 - "${BASH_COMPLETION_HOST_EXCLUSION_PATTERN_FILES[@]-}" 2>/dev/null  |  grep -E -v '^[[:space:]]*(#|$)'  ||  true )"
 
         if [ -n "${host_exclusion_pattern}" ]; then
-            COMPREPLY=( $(  printf '%s\n' "${COMPREPLY[@]}"  |  command -p grep ${re_language} -v "${host_exclusion_pattern}" 2>/dev/null  ||  true  ) )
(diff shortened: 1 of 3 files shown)


calestyo:
btw: One think that you may [not?😇] like is that 3af7347eb3c7aca189554cae78f3b0753d2b2fcb introduces `.config/bash-completion/` …

`bash_completion` (as a file) is already used in `.config/` and `/etc/`, which I think is a bit unfortunate as it would be better to have one umbrella dir, where everything bash-completion related should (eventually) go in.

I didn't want to add yet another "top-level" dir + file, just for exclusion patterns.

Hope you can either live with it, or have some even better idea.

calestyo:
Also, it may make sense to add some `/etc/bash-completion/host-exclusion-patterns` counterpart to `bash_completion.sh.in` .. didn't do it so far, cause I wanted to wait whether you'd be okay with adding that as a bash-completion "main" directory.

calestyo:
> btw: There's still some subtle bug in this... still looking for the reason.

fixed

PR #728, opened 2022-04-09
docs: convert to markdown, remove some no longer relevant sections

scop:
This is mostly just a mechanical conversion, contentwise a bunch of things remain to be added and updated.

Should likely go in before #678 and paves way for https://github.com/scop/bash-completion/discussions/539#discussioncomment-2240641


doc/configuration.md
```diff
@@ -0,0 +1,75 @@
...
+`~/.bash_completion` if unset or null.
+
+### `$XDG_CONFIG_HOME/bash_completion`
+
+Sourced by the `bash_completion.sh` `profile.d` script. This file is
+suitable for definitions of all `COMP*` environment variables below.
+If `$XDG_CONFIG_HOME` is unset or null,`~/.config` is
+used instead of it.
+
+## Environment variables
```
akinomyoga, line 19: Maybe this should be processed in a separate PR, but I feel this should be "Shell variables" because these variables do not have to have the export attribute.

akinomyoga (approved):

doc/configuration.md
scop: Oh! I've never thought that there would be a difference. Changed.

PR #729, opened 2022-04-10
refactor: completion internal variable renames

scop:
Per https://github.com/scop/bash-completion/discussions/539#discussioncomment-864857

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -419,7 +419,7 @@ _sftp()
 
 # things we want to backslash escape in scp paths
 # shellcheck disable=SC2089
-_scp_path_esc='[][(){}<>"'"'"',:;^&!$=?`\\|[:space:]]'
+_comp_cmd_scp__path_esc='[][(){}<>"'"'"',:;^&!$=?`\\|[:space:]]'
 
 # Complete remote files with ssh.  If the first arg is -d, complete on dirs
 # only.  Returns paths escaped with three backslashes.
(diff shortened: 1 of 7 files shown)


akinomyoga:
- Should we also rename `_backup_glob` (though some external completions might use it)?
- Some other temporary variables in `bash_completion` might also be renamed in a consistent way: `_comp_compat_dir` `_comp_file` `_comp_init_blacklist_glob` `_comp_init_blacklist_glob` to e.g. `_comp__init_*`.
- I think `BASH_COMPLETION_ORIGINAL_V_VALUE` (used in `bash_completion`) can also be turned into an internal variable like `_comp__init_original_shopt_verbose`.

scop:
All good points. However, my intention with this was that I'd address "completion internal" things only, i.e. `completions/*` stuff, that's what I was trying to convey with the PR title. No particular preference if we want to expand this to cover the mentioned things in `bash_completion` as well, or to leave them for another PR.

I'll leave it for you to decide, and feel also free to push commits to this branch/PR implementing the above if expanding this is what you'd prefer.

akinomyoga:
> All good points. However, my intention with this was that I'd address "completion internal" things only, i.e. `completions/*` stuff,

Ah, OK! Thank you for the clarification. I have been just misunderstanding that.

I will make a separate PR because these changes do not overlap with each other, and also there are some points that I would like to discuss that might take a time.

akinomyoga (approved):
I think this is ready to be merged.

PR #731, opened 2022-04-11
docs: add API and naming

scop:
- https://github.com/scop/bash-completion/discussions/537
- https://github.com/scop/bash-completion/discussions/539

### DIFF
--- a/doc/api-and-naming.md
+++ b/doc/api-and-naming.md
@@ -3,13 +3,14 @@
 Due to its nature, bash-completion adds a number of functions and variables in
 the shell's environment.
 
-|                                | `bash_completion`   | `completions/*`                                                            |
-| :----------------------------- | :------------------ | :------------------------------------------------------------------------- |
-| public configuration variables | `BASH_COMPLETION_*` | `BASH_COMPLETION_CMD_${Command^^}_${Config^^}`                             |
-| non-local internal variables   | `_comp__*`          | `_comp_cmd_${Command}__${Data}`                                            |
-| public/exported functions      | `_comp_*`           | `_comp_cmd_${Command}` (functions for `complete -F`)                       |
-|                                |                     | `_comp_xfunc_${Command}_${Utility}` (functions for use with `_comp_xfunc`) |
-| private/internal functions     | `_comp__*`          | `_comp_cmd_${Command}__${Utility}` (utility functions)                     |
+|                                   | `bash_completion`   | `completions/*`                                                            |
+| :-------------------------------- | :------------------ | :------------------------------------------------------------------------- |
+| public configuration variables    | `BASH_COMPLETION_*` | `BASH_COMPLETION_CMD_${Command^^}_${Config^^}`                             |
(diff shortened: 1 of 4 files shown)


akinomyoga:
I think the descriptions of the existing rules are fine. In addition, I have questions about the naming rules for the following cases (I have created separate threads in #539):

- More flexible positions of double underscores in the names? -> [\#539 (thread)](https://github.com/scop/bash-completion/discussions/539#discussioncomment-2552482)
- Naming of local variables in the function that exports variables of dynamic names? -> [\#539 (thread)](https://github.com/scop/bash-completion/discussions/539#discussioncomment-2552484)

akinomyoga:
> https://github.com/scop/bash-completion/discussions/539#discussioncomment-3541381
>
> I think we have a conclusion here and in https://github.com/scop/bash-completion/pull/791. Could you document the convention in https://github.com/scop/bash-completion/pull/731? TIA :)

I'm not sure if I correctly get your intent, but I added a commit 4f6e4241865ffdbc7789df214173348452643701 that describes the rule of `_mut_`.

I also suggest using the word "private" instead of "internal" for the symmetry with "public" in the counterpart (see commit 5435b7ce49f1be7290e9e1fe9b6097363d78b4fa), though I am not sure if you like it. If not, we can drop this commit any time.

scop:
Both 4f6e424 and 5435b7c look good to me, thanks!

akinomyoga:
Thanks for checking!

doc/api-and-naming.md
```diff
@@ -0,0 +1,94 @@
...
+them. The most common one of these is named `ret`. Consult the commentary
+before each function in the source to find out the specific names.
+`local`izing output variables before invoking a function that populates them
+is the caller's responsibility.
+Note that if calling multiple functions that assign output to the same variable
+during one completion function run, each result should be copied to another
+variable between the calls to avoid it possibly being overwritten and lost on
+the next call. Also, the variables should also be ensured to be clear before
+each call, typically by `unset -v`ing them when multiple such calls are used,
+to avoid them interfering with each other.
```
akinomyoga, line 19: Do we always require to clear the state by `unset -v`? I feel we can require it only when the function called next will use the state of the variable (i.e., the value, the existence of the value, and/or the existence of the variable). Otherwise, we always need to save the values including the following case:

```bash
# I would like to allow writing this way as far as
# _func1/_func2 doesn't reference the original state of ret.
local ret
_func1 "$1"
_func2 "$ret"

# This might be much safer but feels redundant
local ret
unset -v ret
_func1 "$1"
local intermediate=$ret
unset -v ret
_func2 "$intermediate"
```

scop: Agreed, addressed in 1fb425f827629f7419433bf1e6a84cda5fe38897 I hope.

akinomyoga: Thanks!

scop:
Link to CONTRIBUTING.md added, cheers.

akinomyoga:
Thanks!

PR #733, opened 2022-04-11
test(pymarkdown): enable inline HTML check

scop:
Blocking issue fixed in v0.9.1.


scop:
Doing in #617 instead.

Issue #542, opened 2021-06-06
Add deprecation warnings with option to turn off


scop:
## Describe the feature/solution

Related to #537, would be good to emit deprecation warnings for deprecated things. Currently a bunch of our deprecations is just in code comments.

Because the deprecation warnings are something end users (as opposed to external completion authors) will see, a setting to turn those off should be added, and referred to in those deprecation warnings. Maybe a global envvar for consistency, see doc/bash_completion.txt

## Maintenance (please complete the following information)

- [ ] This is a request for new completion
- [ ] Link to upstream project query about shipping the completion:

N/A

akinomyoga:
> Because the deprecation warnings are something end users (as opposed to external completion authors) will see, a setting to turn those off should be added, and referred to in those deprecation warnings.

Maybe we can output the warnings only when the deprecated interfaces are first used in the session.

scop:
> Maybe we can output the warnings only when the deprecated interfaces are first used in the session.

Possibly, _in addition_ to the ability to turn them off altogether. An end user who just uses e.g. packaged software shouldn't be modifying files from the packages, but rather preferably report a bug somewhere and then turn the warnings off so they're not bothered by it even once a session after that. Maybe even a possibility to turn each individual warning off separately, so people who want to help with stuff can continue finding new deprecations instead of turning off everything when they see the first one. But dunno if there's any actual audience for such functionality, or if it's just added code/complexity for very little or no gain.

akinomyoga:
> Maybe even a possibility to turn each individual warning off separately, so people who want to help with stuff can continue finding new deprecations instead of turning off everything when they see the first one. But dunno if there's any actual audience for such functionality, or if it's just added code/complexity for very little or no gain.

For those who are willing to continue to help us find new depcreations, I think we may provide an option to specify a log file instead of providing options to turn off each warning. Suppose the variable name to be `COMP_WARNING_FILE`, the default `COMP_WARNING_FILE=` (or unset) means the warnings are output to `stderr`, `COMP_WARNING_FILE=/dev/null` means completely suppressing the error messages, and `COMP_WARNING_FILE=/other/path` means appending warning messages into the specified file.

akinomyoga:
It seems [spf13/cobra](https://github.com/spf13/cobra) has a similar variable `BASH_COMP_DEBUG_FILE` to specify the debug output target. The debug output can be passed to the following function:

```bash
__xxxx_debug()
{
[...]

PR #734, opened 2022-04-11
feat: add `_comp_xfunc`

### LINKED ISSUE #732: Implement new `_comp_xfunc` and naming
Per https://github.com/scop/bash-completion/discussions/539#discussioncomment-2213575

New `_comp_xfunc` implementation, renaming of target functions, docs in `doc/api-and-naming.md`

akinomyoga:
Resolve #732

- 2c7ca225 add `_comp_xfunc`
- 2b4bb190 use `_comp_xfunc` instead of `_xfunc`
- fc0b4b87 add `_comp_deprecate_func` for backward-compatibility names

The function `_comp_deprecate_func` is introduced to later automatically detect the use of old function names for #542. To implement #542, some codes can be added in the automatically generated function body in `_comp_deprecate_func`.

### DIFF
--- a/completions/cvs
+++ b/completions/cvs
@@ -35,7 +35,7 @@ _cvs_kflags()
     COMPREPLY=($(compgen -W 'kv kvl k o b v' -- "$cur"))
 }
 
-_cvs_roots()
+_comp_xfunc_cvs_roots()
 {
     local -a cvsroots
     [[ -v CVSROOT ]] && cvsroots=("$CVSROOT")
@@ -44,6 +44,8 @@ _cvs_roots()
     COMPREPLY=($(compgen -W '${cvsroots[@]}' -- "$cur"))
     __ltrim_colon_completions "$cur"
 }
+
+_comp_deprecate_func _cvs_roots _comp_xfunc_cvs_roots
 
 _cvs()
 {
@@ -273,7 +275,7 @@ _cvs()
             fi
             ;;
         cvsroot)
-            _cvs_roots
+            _comp_xfunc_cvs_roots
             ;;
         diff | log | status)
             if [[ $cur == -* ]]; then
--- a/completions/ssh
+++ b/completions/ssh
@@ -2,7 +2,7 @@
 
 _ssh_queries()
 {
-    local queries=$(_ssh_query "$1" help)
+    local queries=$(_comp_xfunc_ssh_query "$1" help)
     [[ $queries ]] || queries="cipher cipher-auth mac kex key key-cert
          key-plain key-sig protocol-version compression sig ciphers macs
          kexalgorithms pubkeyacceptedkeytypes hostkeyalgorithms
@@ -10,14 +10,16 @@ _ssh_queries()
     COMPREPLY+=($(compgen -W "$queries help" -- "${cur,,}"))
 }
 
-_ssh_query()
+_comp_xfunc_ssh_query()
 {
     ${1:-ssh} -Q $2 2>/dev/null
 }
 
+_comp_deprecate_func _ssh_query _comp_xfunc_ssh_query
+
 _ssh_ciphers()
 {
-    local ciphers=$(_ssh_query "$1" cipher)
+    local ciphers=$(_comp_xfunc_ssh_query "$1" cipher)
     [[ $ciphers ]] || ciphers="3des-cbc aes128-cbc aes192-cbc aes256-cbc
         aes128-ctr aes192-ctr aes256-ctr arcfour128 arcfour256 arcfour
         blowfish-cbc cast128-cbc"
@@ -26,13 +28,13 @@ _ssh_ciphers()
 
 _ssh_macs()
 {
-    local macs=$(_ssh_query "$1" mac)
+    local macs=$(_comp_xfunc_ssh_query "$1" mac)
     [[ $macs ]] || macs="hmac-md5 hmac-sha1 umac-64@openssh.com hmac-ripemd160
         hmac-sha1-96 hmac-md5-96"
     COMPREPLY+=($(compgen -W "$macs" -- "$cur"))
 }
 
-_ssh_options()
+_comp_xfunc_ssh_options()
 {
     # curl --silent https://raw.githubusercontent.com/openssh/openssh-portable/master/ssh_config.5 | awk '$1==".It" && $2=="Cm" && $3!="Host" && $3!="Match" {print "        "$3}' | sort
     local opts=(
@@ -65,7 +67,7 @@ _ssh_options()
         GSSAPIKeyExchange GSSAPIRenewalForcesRekey GSSAPIServerIdentity
         GSSAPITrustDns SmartcardDevice UsePrivilegedPort
     )
-    local protocols=$(_ssh_query "${1:-ssh}" protocol-version)
+    local protocols=$(_comp_xfunc_ssh_query "${1:-ssh}" protocol-version)
     if [[ ! $protocols || $protocols == *1* ]]; then
         opts+=(Cipher CompressionLevel Protocol RhostsRSAAuthentication
             RSAAuthentication)
@@ -80,6 +82,8 @@ _ssh_options()
     done))
     $reset
 }
+
+_comp_deprecate_func _ssh_options _comp_xfunc_ssh_options
 
 # Complete a ssh suboption (like ForwardAgent=y<tab>)
 # Two parameters: the string to complete including the equal sign, and
@@ -121,14 +125,14 @@ _ssh_suboption()
             COMPREPLY=($(compgen -W 'yes no always' -- "$cur"))
             ;;
         identityfile)
-            _ssh_identityfile
+            _comp_xfunc_ssh_identityfile
             ;;
         *file | identityagent | include | controlpath | revokedhostkeys | \
             xauthlocation)
             _filedir
             ;;
         casignaturealgorithms)
-            COMPREPLY=($(compgen -W '$(_ssh_query "$2" sig)' -- "$cur"))
+            COMPREPLY=($(compgen -W '$(_comp_xfunc_ssh_query "$2" sig)' -- "$cur"))
             ;;
         cipher)
             COMPREPLY=($(compgen -W 'blowfish des 3des' -- "$cur"))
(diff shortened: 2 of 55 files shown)


bash_completion
```diff
@@ -2432,6 +2457,28 @@ _xfunc()
...
+_comp_deprecate_func _mysql_character_sets _comp_xfunc_mysql_character_sets
+_comp_deprecate_func _python_modules _comp_xfunc_python_modules
+_comp_deprecate_func _python_warning_actions _comp_xfunc_python_warning_actions
+_comp_deprecate_func _rpm_installed_packages _comp_xfunc_rpm_installed_packages
+_comp_deprecate_func _scp_local_files _comp_xfunc_ssh_scp_local_files
+_comp_deprecate_func _scp_remote_files _comp_xfunc_ssh_scp_remote_files
+_comp_deprecate_func _ssh_identityfile _comp_xfunc_ssh_identityfile
+_comp_deprecate_func _ssh_options _comp_xfunc_ssh_options
+_comp_deprecate_func _ssh_query _comp_xfunc_ssh_query
+_comp_deprecate_func _ssh_suboption_check _comp_xfunc_ssh_suboption_check
```
akinomyoga, line 2478: @scop has suggested in https://github.com/scop/bash-completion/issues/732#issuecomment-1094831476 that the backward-compatibility names might be defined in a separated file (e.g. in `/etc/bash_completion.d`), but currently these compatibility names are tentatively included in `bash_completion`.
- If we put the file in `/etc/bash_completion.d`, something unwanted might occur when the user sets the custom `BASH_COMPLETION_COMPAT_DIR`, so it might be better to be in another directory such as `helpers`?
- Another option is to put these compatibility names in the same file as the new API. For example, define the compatibility name in _apt_cache_packages.

akinomyoga: I now noticed that this causes the problem when the new name is not yet defined when the compatibility name is used. So, the compatibility name should be defined after the new name is defined (for example in the same file as the new name), or the function of the compatibility name should call the function of the new name through `_comp_xfunc`.

test/t/unit/test_unit_xfunc.py
```diff
@@ -0,0 +1,50 @@
...
+                want_output=True,
+            )
+            assert output.strip() == "util1[<a b><cde><fgh>]"
+
+    def test_2(self, bash):
+        with bash_env_saved(bash) as bash_env:
+            bash_env.write_variable(
+                "BASH_COMPLETION_USER_DIR", "$PWD/_comp_xfunc", quote=False
+            )
+
```
scop, line 35: Similarly here, perhaps assert that `_comp_xfunc_non_standard_name` is not yet defined?

bash_completion
```diff
@@ -2419,6 +2419,31 @@ _completion_loader()
...
+# Declare a compatibility function name
+# @param $1 Old function name
+# @param $2 New function name
+_comp_deprecate_func()
+{
+    eval -- "$1() { $2 \"\$@\"; }"
+}
+
+#---------------------------------------
+# The following functions are deprecated / backward-compatible names
```
scop, line 2445: I'd suggest adding a `@deprecated` to each deprecated function's commentary instead (see existing examples) -- that way I think people are more likely to notice than if/when this section becomes longish.

bash_completion
```diff
@@ -2432,6 +2457,8 @@ _xfunc()
     "$@"
 }
 
+#---------------------------------------
```
scop, line 2460: This can likely go if we add per function `@deprecated` markers.

completions/cvs
```diff
@@ -1,5 +1,10 @@
 # cvs(1) completion                                        -*- shell-script -*-
 
+_comp_xfunc_cvs_roots()
+{
+    _cvs_roots
+}
+
```
scop, line 7: Should we not rename + `_comp_deprecate_func` `_cvs_roots` to this instead?

completions/ssh
```diff
@@ -1,8 +1,38 @@
...
+_comp_xfunc_ssh_scp_remote_files()
+{
+    _scp_remote_files "$@"
+}
+
+_comp_xfunc_ssh_scp_local_files()
+{
+    _scp_local_files "$@"
+}
+
```
scop, line 32: Should we not rename + `_comp_deprecated_func` the originals instead?

bash_completion
```diff
@@ -2419,6 +2419,31 @@ _completion_loader()
...
     complete -D -F _completion_loader
 
+# Function for loading and calling functions from dynamically loaded
+# completion files that may not have been sourced yet.
+# @param $1 completion file to load function from in case it is missing
+# @param $2 the xfunc name.  When it does not have the form `_comp_xfunc_*',
+# `_comp_xfunc_$1_$2' is used for the actual name of the shell function.
+_comp_xfunc()
+{
+    local __xfunc_name=$2
```
scop, line 2429: Cosmetic, but I suppose we could use `xfunc_name` for this local variable's name here, no need for the underscores?

scop (approved):
Looks good to me, thanks!

Left a bunch of rather trivial comments, pre-approved with those addressed, but feel free to bounce back if you think they deserve more discussion or I missed something.

bash_completion
akinomyoga: Thanks! I haven't noticed the existing ones.

bash_completion
akinomyoga: Yeah, this is actually related to https://github.com/scop/bash-completion/discussions/539#discussioncomment-2552484. But in this case, there is much less possibility of the variable name clash. I'll just follow your suggestion! Thanks.

completions/cvs
akinomyoga: I have kept the original function `_cvs_roots` for internal use by `cvs` and just added an interface here because there seem to be many other functions similar to `_cvs_roots` below (i.e., _cvs_entries, _cvs_modules, _cvs_commands. etc.)

- Nevertheless, I may directly rename only `_cvs_roots` in the subsequent code.
- Another option is to make all the similar functions *xfunc* for consistency.

There are similar cases also in several other files.

completions/ssh
akinomyoga: I will treat them the same as `_cvs_roots` after the treatment of `_cvs_roots` is settled.

test/t/unit/test_unit_xfunc.py
```diff
@@ -0,0 +1,50 @@
...
+
+@pytest.mark.bashcomp(cmd=None)
+class TestUnitXfunc:
+    def test_1(self, bash):
+        with bash_env_saved(bash) as bash_env:
+            bash_env.write_variable(
+                "BASH_COMPLETION_USER_DIR", "$PWD/_comp_xfunc", quote=False
+            )
+
```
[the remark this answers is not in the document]
akinomyoga: Good suggestion!

test/t/unit/test_unit_xfunc.py
akinomyoga: Yes! I will process it in the other [(conversation)](https://github.com/scop/bash-completion/pull/734/files#r848820292),

test/t/unit/test_unit_xfunc.py
```diff
@@ -0,0 +1,50 @@
...
+            # first invocation (completion/xfunc-test1 is sourced)
+            output = assert_bash_exec(
+                bash,
+                "_comp_xfunc xfunc-test1 utility1 'a b' cde fgh",
+                want_output=True,
+            )
+            assert output.strip() == "util1[<a b><cde><fgh>]"
+
+            # first invocation (completion/xfunc-test1 is not sourced)
```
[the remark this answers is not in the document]
akinomyoga: f623b9d8

test/t/unit/test_unit_xfunc.py
```diff
@@ -0,0 +1,50 @@
...
+
+            # first invocation (completion/xfunc-test2 is sourced)
+            output = assert_bash_exec(
+                bash,
+                "_comp_xfunc xfunc-test2 _comp_xfunc_non_standard_name 'a b' cde fgh",
+                want_output=True,
+            )
+            assert output.strip() == "util2[<a b><cde><fgh>]"
+
+            # first invocation (completion/xfunc-test2 is not sourced)
```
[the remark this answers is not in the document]
akinomyoga: f623b9d8

test/t/unit/test_unit_xfunc.py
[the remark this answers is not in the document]
akinomyoga: 573e0db0 I have added also the test that the function is defined before the second invocation. If that addition is not needed, please let me know.

test/t/unit/test_unit_xfunc.py
akinomyoga: 573e0db0

bash_completion
akinomyoga: 2585250f

bash_completion
akinomyoga: 2585250f

bash_completion
akinomyoga: c056c2d8

akinomyoga:
I have rebased and squashed the changes. For the additional changes compared to the previous version, I have commented on the commit hashes in each conversation.

completions/cvs
scop: Hmm, I see. But then again all cases except this and ssh in this PR take the rename+deprecate approach. None of the choices strikes me obviously better than all others, but I think I'd prefer the consistency by not leaving wrappers like this behind, and making things xfunc only when we know we have xfunc use for them from somewhere else.

akinomyoga: OK! So, if I don't misunderstand, ccf13887?

akinomyoga:
Now I noticed that there can be a case where an external completion calls any function that is not marked *xfunc* by us through `_xfunc`. Maybe we need to still allow such cases for backward compatibility as 35fbf57c, or do we recommend continuing to use the deprecated `_xfunc` in such a case? Hmm, what do you think?

scop:
35fbf57cbc59e9879e59bd19dc044b1bc5f43af0 makes sense to me. With that, couldn't we just `_comp_deprecate_func _xfunc _comp_xfunc` considering that our separate completion files don't have any functions that would _not_ start with an underscore?

completions/cvs
scop: Yes, that's what I meant, thanks.

akinomyoga:
Thank you!
- ce9b010a I have supplemented 35fbf57c with an appropriate update of the description
- ffb59f43 I have moved `_comp_deprecate_func` to the beginning of the file to use it for functions defined in the main `bash_completion`, and applied it to `_comp_xfunc`
- f8a6a096 I have added input checks to `_comp_deprecate_func`

----

I have rebased and squashed.

akinomyoga:
Thanks!

PR #735, opened 2022-04-11
refactor(bash_completion): rename functions `_userland`, `_sysdirs`, `_have`, and `_rl_enabled`

akinomyoga:
- This is still a draft; only a small part of the functions are processed.
- This is based on #734 (`_comp_xfunc`)
- I open this PR to reference the related PRs because some of my existing branches already contain some part of refactoring names and behaviors, so I'd like to process them at the same time.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2073,8 +2073,9 @@ _cd()
         return
     fi
 
-    local -r mark_dirs=$(_comp_readline_variable_on mark-directories && echo y)
-    local -r mark_symdirs=$(_comp_readline_variable_on mark-symlinked-directories && echo y)
+    local mark_dirs='' mark_symdirs=''
+    _comp_readline_variable_on mark-directories && mark_dirs=y
+    _comp_readline_variable_on mark-symlinked-directories && mark_symdirs=y
 
     # we have a CDPATH, so loop on its contents
     for i in ${CDPATH//:/$'\n'}; do
--- a/bash_completion
+++ b/bash_completion
@@ -94,12 +94,14 @@ complete -b builtin
 
 # Check if we're running on the given userland
 # @param $1 userland to check for
-_userland()
+_comp_userland()
 {
     local userland=$(uname -s)
     [[ $userland == @(Linux|GNU/*) ]] && userland=GNU
     [[ $userland == "$1" ]]
 }
+
+_comp_deprecate_func _userland _comp_userland
 
 # This function sets correct SysV init directories
 #
--- a/bash_completion
+++ b/bash_completion
@@ -119,20 +119,22 @@ _comp_deprecate_func _sysvdirs _comp_sysvdirs
 
 # This function checks whether we have a given program on the system.
 #
-_have()
+_comp_have_command()
 {
     # Completions for system administrator commands are installed as well in
     # case completion is attempted via `sudo command ...'.
     PATH=$PATH:/usr/sbin:/sbin:/usr/local/sbin type $1 &>/dev/null
 }
 
+_comp_deprecate_func _have _comp_have_command
+
 # Backwards compatibility for compat completions that use have().
 # @deprecated should no longer be used; generally not needed with dynamically
-#             loaded completions, and _have is suitable for runtime use.
+#             loaded completions, and _comp_have_command is suitable for runtime use.
 have()
 {
     unset -v have
-    _have $1 && have=yes
+    _comp_have_command $1 && have=yes
 }
 
 # This function checks whether a given readline variable
--- a/bash_completion
+++ b/bash_completion
@@ -140,10 +140,12 @@ have()
 # This function checks whether a given readline variable
 # is `on'.
 #
-_rl_enabled()
+_comp_readline_variable_on()
 {
     [[ $(bind -v) == *$1+([[:space:]])on* ]]
 }
+
+_comp_deprecate_func _rl_enabled _comp_readline_variable_on
 
 # This function shell-quotes the argument
 quote()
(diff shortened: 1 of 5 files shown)


bash_completion
```diff
@@ -98,16 +98,18 @@ complete -b builtin
 
 # Check if we're running on the given userland
 # @param $1 userland to check for
-_userland()
+_comp_get_userland()
```
scop, line 101:
```suggestion
_comp_userland()
```

I feel the `get_` here in the name is redundant, and inconsistent with what we have in the codebase already. For example, none of the `_comp_*` in `completions/` have it, even though they are "getters". And this actually is not even a getter -- `is_` would be more appropriate, but I think without either is fine.

bash_completion
```diff
@@ -98,16 +98,18 @@ complete -b builtin
...
     [[ $userland == @(Linux|GNU/*) ]] && userland=GNU
     [[ $userland == "$1" ]]
 }
 
+_comp_deprecate_func _userland _comp_get_userland
+
 # This function sets correct SysV init directories
 #
-_sysvdirs()
+_comp_get_sysvdirs()
```
scop, line 112:
```suggestion
_comp_sysvdirs()
```

bash_completion
```diff
@@ -98,16 +98,18 @@ complete -b builtin
...
 # @param $1 userland to check for
-_userland()
+_comp_get_userland()
 {
     local userland=$(uname -s)
     [[ $userland == @(Linux|GNU/*) ]] && userland=GNU
     [[ $userland == "$1" ]]
 }
 
+_comp_deprecate_func _userland _comp_get_userland
```
scop, line 108:
```suggestion
_comp_deprecate_func _userland _comp_userland
```

bash_completion
```diff
@@ -1443,7 +1451,7 @@ _xinetd_services()
 _services()
 {
     local sysvdirs
-    _sysvdirs
+    _comp_get_sysvdirs
```
scop, line 1454:
```suggestion
    _comp_sysvdirs
```

bash_completion
```diff
@@ -1479,7 +1487,7 @@ _service()
         [[ -e /etc/mandrake-release ]] && _xinetd_services
     else
         local IFS=$'\n' sysvdirs
-        _sysvdirs
+        _comp_get_sysvdirs
```
scop, line 1490:
```suggestion
        _comp_sysvdirs
```

bash_completion
```diff
@@ -1490,7 +1498,7 @@ _service()
 _comp__init_set_up_service_completions()
 {
     local sysvdirs svc svcdir
-    _sysvdirs
+    _comp_get_sysvdirs
```
scop, line 1501:
```suggestion
    _comp_sysvdirs
```

completions/ifup
```diff
@@ -1,6 +1,6 @@
 # Red Hat & Debian GNU/Linux if{up,down} completion        -*- shell-script -*-
 
-_userland GNU || return 1
+_comp_get_userland GNU || return 1
```
scop, line 3:
```suggestion
_comp_userland GNU || return 1
```

scop (changes requested):

scop:
> This is still a draft; only a small part of the functions are processed.

I wouldn't worry about that -- we can process things in small chunks, that way they are easier to review.

akinomyoga:
Thank you for the review!

> I wouldn't worry about that -- we can process things in small chunks, that way they are easier to review.

Now, I agree with it because I seem to have many things to discuss in the refactoring. I think I'd like to separate the PRs. I will change the PR subject.

Maybe I could have just changed the function names by blindly prefixing `_comp_`, but I feel it would be better to reconsider the interface (print to stdout vs assigning to variables) and the naming of the functions rather than introducing another breaking change later. But, if you think we can just prefix `_comp_` to the existing function names keeping the interface for the fast release of 2.11, I can follow the suggestion.

bash_completion
akinomyoga: Thank you! Actually, these naming are something that I wanted to discuss when the PR is ready for review. I will adopt these suggestions.

akinomyoga:
I have rebased it. I have decided to narrow the scope of this PR to these 5 functions and update the PR subject. Now this is ready for review.

Another thing related to the function naming is about `_comp_has_command` (previously `_have`). If the original function name shouldn't be changed too much or there are any better suggestions, I can adjust the name.

bash_completion
```diff
@@ -113,32 +115,38 @@ _sysvdirs()
     return 0
 }
 
+_comp_deprecate_func _sysvdirs _comp_sysvdirs
+
 # This function checks whether we have a given program on the system.
 #
-_have()
+_comp_has_command()
```
scop, line 122: I would personally go with `_comp_have_command` for this -- the "has" makes it read like asking if completion has a command which is not what we're "asking" with this. `_comp_have_command` sounds just less odd to me. But not a biggie/blocker, leaving up to you.

bash_completion
```diff
@@ -113,32 +115,38 @@ _sysvdirs()
...
     unset -v have
-    _have $1 && have=yes
+    _comp_has_command $1 && have=yes
 }
 
 # This function checks whether a given readline variable
 # is `on'.
 #
-_rl_enabled()
+_comp_rl_enabled()
```
scop, line 143: I was contemplating `_comp_is_rl_enabled` for this, but I think it's better as `_comp_rl_enabled`.

Then again there's little reason to keep it this short, perhaps `_comp_readline_variable_on` or `_comp_readline_variable_enabled` (or "variable" shortened to "var" in either) would be better.

Again, leaving up to you to decide.

scop (approved):
Left a few comments, feel free to address the way you prefer or not, pre-approved either way.

bash_completion
akinomyoga: Thank you for the comment. I have switched to `_comp_have_command`

bash_completion
akinomyoga: I adopted `_comp_readline_variable_on`.

akinomyoga:
> Left a few comments, feel free to address the way you prefer or not, pre-approved either way.

Maybe does *pre-approved* mean that I'm supposed to merge it when I decided on these comments?

scop:
> Maybe does pre-approved mean that I'm supposed to merge it when I decided on the points raised in these comments?

Yes, that's what I meant, apologies for being unclear (and again taking such a long time to get back to this!). I'll go ahead and merge now, thanks!

akinomyoga:
Thank you! I'll keep it in mind from now on!

PR #736, opened 2022-04-11
refactor(bash_completion): refactor `{ => _comp_}{dequote,quote}`

akinomyoga:
In #735, I'd like to refactor the function names `quote` and `dequote`, but another branch that I created two months ago contains an improved version `_comp_dequote` and improved uses of `quote`, so I'd like to process this branch before #735 if possible.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1099,12 +1099,12 @@ _parse_help()
 
     local cmd=$1
     local line rc=1
-    {
+    (
         case $cmd in
-            -) cat ;;
-            *) LC_ALL=C "$(dequote "$cmd")" ${2:---help} 2>&1 ;;
+            -) exec cat ;;
+            *) _comp_dequote "$cmd" && LC_ALL=C "$ret" ${2:---help} 2>&1 ;;
         esac
-    } |
+    ) |
         while read -r line; do
 
             [[ $line == *([[:blank:]])-* ]] || continue
@@ -1136,12 +1136,12 @@ _parse_usage()
 
     local cmd=$1
     local line match option i char rc=1
-    {
+    (
         case $cmd in
-            -) cat ;;
-            *) LC_ALL=C "$(dequote "$cmd")" ${2:---usage} 2>&1 ;;
+            -) exec cat ;;
+            *) _comp_dequote "$cmd" && LC_ALL=C "$ret" ${2:---usage} 2>&1 ;;
         esac
-    } |
+    ) |
         while read -r line; do
 
             while [[ $line =~ \[[[:space:]]*(-[^]]+)[[:space:]]*\] ]]; do
--- a/bash_completion
+++ b/bash_completion
@@ -148,6 +148,16 @@ _comp_readline_variable_on()
 _comp_deprecate_func _rl_enabled _comp_readline_variable_on
 
 # This function shell-quotes the argument
+# @param    $1  String to be quoted
+# @var[out] ret Resulting string
+_comp_quote()
+{
+    ret=\'${1//\'/\'\\\'\'}\'
+}
+
+# This function shell-quotes the argument
+# @deprecated Use `_comp_quote` instead.  Note that `_comp_quote` stores
+#   the results in the variable `ret` instead of writing them to stdout.
 quote()
 {
     local quoted=${1//\'/\'\\\'\'}
--- a/bash_completion
+++ b/bash_completion
@@ -162,10 +162,77 @@ quote_readline()
     printf %s "$ret"
 } # quote_readline()
 
+# shellcheck disable=SC1003
+_comp_dequote__initialize()
+{
+    unset -f "$FUNCNAME"
+    local regex_param='\$([_a-zA-Z][_a-zA-Z0-9]*|[-*@#?$!0-9_])|\$\{[!#]?([_a-zA-Z][_a-zA-Z0-9]*(\[([0-9]+|[*@])\])?|[-*@#?$!0-9_])\}'
+    local regex_quoted='\\.|'\''[^'\'']*'\''|\$?"([^\"$`!]|'$regex_param'|\\.)*"|\$'\''([^\'\'']|\\.)*'\'''
+    _comp_dequote__regex_safe_word='^([^\'\''"$`;&|<>()!]|'$regex_quoted'|'$regex_param')*$'
+}
+_comp_dequote__initialize
+
+# This function expands a word using `eval` in a safe way.  This function can
+# be typically used to get the expanded value of `${word[i]}` as
+# `_comp_dequote "${word[i]}"`.  When the word contains unquoted shell special
+# characters, command substitutions, and other unsafe strings, the function
+# call fails before applying `eval`.  Otherwise, `eval` is applied to the
+# string to generate the result.
+#
+# @param    $1  String to be expanded.  A safe word consists of the following
+#               sequence of substrings:
+#
+#               - Shell non-special characaters: [^\'"$`;&|<>()!].
+#               - Parameter expansions of the forms $PARAM, ${!PARAM},
+#                 ${#PARAM}, ${NAME[INDEX]}, ${!NAME[INDEX]}, ${#NAME[INDEX]}
+#                 where INDEX is an integer, `*` or `@`, NAME is a valid
+#                 variable name [_a-zA-Z][_a-zA-Z0-9]*, and PARAM is NAME or a
+#                 parameter [-*@#?$!0-9_].
+#               - Quotes \?, '...', "...", $'...', and $"...".  In the double
+#                 quotations, parameter expansions are allowed.
+#
+# @var[out] ret Array that contains the expanded results.  Multiple words or no
+#               words may be generated through pathname expansions.
+#
+# Note: This function allows parameter expansions as safe strings, which might
+# cause unexpected results:
+#
+# * This allows execution of arbitrary commands through extra expansions of
+#   array subscripts in name references. For example,
+#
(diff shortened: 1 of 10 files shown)


test/t/unit/test_unit_dequote.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # def test_glob_noglob(self, bash):
+    #     with bash_env_saved(bash) as bash_env:
+    #         bash_env.set("noglob", True)
+    #         output = assert_bash_exec(
+    #             bash,
+    #             "__tester 'non-existent-*.txt'",
+    #             want_output=True,
+    #         )
+    #         assert output.strip() == "<non-existent-*.txt>"
```
akinomyoga, line 145: This test is currently commented out because it requires f0a4028d87f3506f2091dbafef91771794632c94 contained in #708.
- (a) #708 may be processed before this one, or
- (b) this code section may be uncommented after merging #708, or
- (c) f0a4028 and related commits may be cherry-picked into this PR.

scop: Cool, let's finalize #708 first, I just left some last comments there.

akinomyoga: I have uncommented it (8c3ae045) and rebased/squashed it.

bash_completion
```diff
@@ -158,10 +158,31 @@ quote_readline()
...
+_comp_dequote__initialize()
+{
+    unset -f "$FUNCNAME"
+    local param='\$([_a-zA-Z][_a-zA-Z0-9]*|[-*@#?$!0-9_])|\$\{[!#]?([_a-zA-Z][_a-zA-Z0-9]*(\[([0-9]+|[*@])\])?|[-*@#?$!0-9_])\}'
+    local quote='\\.|'\''[^'\'']*'\''|"([^\"$`!]|'$param'|\\.)*"|\$'\''([^\'\'']|\\.)*'\'''
+    _comp_dequote__regex_safe_word='^([^\'\''"$`;&|<>()!]|'$quote'|'$param')*$'
+}
+_comp_dequote__initialize
+
+_comp_dequote()
```
scop, line 171: Can we add a few lines of documentation here, e.g. in what kind of cases one should look into using this, is it a safe replacement for expansion with `eval`, etc?

bash_completion
```diff
@@ -158,10 +158,31 @@ quote_readline()
...
+{
+    ret=() # fallback value for unsafe word and failglob
+    [[ $1 =~ $_comp_dequote__regex_safe_word ]] || return 1
+    eval "ret=($1)" 2>/dev/null # may produce failglob
+}
+
 # This function shell-dequotes the argument
 dequote()
```
scop, line 179: I'd prefer to deprecate `dequote` and `quote`, and remove their uses from the codebase altogether (at least one `quote` instance remains in `_export`). They are in the shell's global command namespace, look like commands the users are supposed to execute interactively, while never having been intended as such I believe.

scop (changes requested):

bash_completion
akinomyoga: Thank you for the suggestion! I'm sorry that this branch seems to be somehow incomplete because I initially created it two months ago and forgot that I left it alone in an incomplete state. I have added the description b39640b1. Also, as I have described in the added description, I think I need to explain the caveats/decisions of the current implementation.

- In the new implementation, the "*unsafe*" words that contain command substitutions or other arbitrary command execution will not be expanded.
- Nevertheless, the parameter expansions are expanded although it might contain command substitutions through the name references. I think the name references should be generally designed to respect a certain kind of referential transparency even if that involves some command executions, so that shouldn't be the problem as far as with the sane setups.

bash_completion
akinomyoga: - I have marked `dequote` with `@deprecated` bf77c2f1.
- I have also added `_comp_quote` that is supposed replace `quote` and marked `quote` with `@deprecated` in 6b2b97e1.
- I have updated the uses of `quote` in `_reply_compgen_array` (`completions/_umount.linux`) and `_export` (`completions/export`) in 1cd1d671. Maybe I could have updated these functions so that they do not perform any quoting, but I have preserved the current behavior.

akinomyoga:
Also, it should be noted that, in these commits, I have changed the interface of `dequote` and `quote` writing their results to `stdout` so that the new `_comp_dequote` and `_comp_quote` write the result in the variable `ret`. If you have any suggestions, please feel free to tell me that.

Another thing that needs to be discussed is the annotations like `@param` in the code documentation. I wanted to describe the variables that are supposed to be referenced/modified so used `@var`, but this might cause problems if these annotations are processed by some documentation programs that don't support `@var`. Are these annotations intended to be processed by a certain annotation language? I have used `@var` inspired by Doxygen, but I found by searching on the internet that the existing documentation seems to follow JavaDoc (which I'm not familiar with)?

akinomyoga:
With other small fixes bf77c2f1 3f49e1b8, I have rebased/squashed the related fixup commits.

scop (approved):
LGTM, thanks! Left a few trivial comments, feel free to merge after addressing.

bash_completion
```diff
@@ -154,10 +164,77 @@ quote_readline()
...
+{
+    unset -f "$FUNCNAME"
+    local param='\$([_a-zA-Z][_a-zA-Z0-9]*|[-*@#?$!0-9_])|\$\{[!#]?([_a-zA-Z][_a-zA-Z0-9]*(\[([0-9]+|[*@])\])?|[-*@#?$!0-9_])\}'
+    local quote='\\.|'\''[^'\'']*'\''|\$?"([^\"$`!]|'$param'|\\.)*"|\$'\''([^\'\'']|\\.)*'\'''
+    _comp_dequote__regex_safe_word='^([^\'\''"$`;&|<>()!]|'$quote'|'$param')*$'
+}
+_comp_dequote__initialize
+
+# This function expands a word using `eval` in a safe way.  This function can
+# be typically used to get the expqanded value of `${word[i]}` as
```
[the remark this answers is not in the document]
akinomyoga: Thank you for your review! I'll apply the suggestions.

completions/make
```diff
@@ -135,8 +135,11 @@ _make()
         # with -C/--directory
         for ((i = 1; i < ${#words[@]}; i++)); do
             if [[ ${words[i]} == -@(C|-directory) ]]; then
-                # eval for tilde expansion
-                eval "makef_dir=( -C \"${words[i + 1]}\" )"
+                # Expand tilde expansion
+                local ret
+                _comp_dequote "${words[i + 1]-}" &&
+                    [[ -d ${ret-} ]] &&
+                    makef_dir=(-C "${ret-}")
```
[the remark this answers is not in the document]
akinomyoga: Ah, right... Thanks for pointing them out!

akinomyoga:
`shellcheck` somehow considers the value of the local variable `quoted` (previously `quote`) is used in other functions that have the variables of the same name.

```
In bash_completion line 172:
    local quoted='\\.|'\''[^'\'']*'\''|\$?"([^\"$`!]|'$param'|\\.)*"|\$'\''([^\'\'']|\\.)*'\'''
                       ^-- SC2089: Quotes/backslashes will be treated literally. Use an array.


In bash_completion line 769:
        toks+=($(compgen "${opts[@]}" -- $quoted))
                                         ^-----^ SC2090: Quotes/backslashes in this variable will not be respected.


In bash_completion line 779:
            toks+=($(compgen -f ${plusdirs+"${plusdirs[@]}"} -- $quoted))
                                                                ^-----^ SC2090: Quotes/backslashes in this variable will not be respected.
```

I have decided to further change the variable names as `{ => regex_}{param,quoted}` (ebf06de2...c6bb062b). Squashed and rebased.

scop:
Hmm, in that case I wonder if the syntax we use in the variables' values here throw off shellcheck's parser. That would be unfortunate as we probably wouldn't be able to rely on it much in the file after the position of the definitions.

akinomyoga:
> Hmm, in that case I wonder if the syntax we use in the variables' values here throw off shellcheck's parser.

I think the problem is not that `shellcheck` fails to parse the value. The shellcheck error can be reproduced even by the following simple script:

```bash
$ cat test.sh
#!/usr/bin/env bash
function setv { eval "$1=\$2"; }
function func1 { local var="'"; }
function func2 { local var; setv var 1; echo $var; }
$ shellcheck test.sh

In test.sh line 3:
function func1 { local var="'"; }
                            ^-- SC2089: Quotes/backslashes will be treated literally. Use an array.


In test.sh line 4:
function func2 { local var; setv var 1; echo $var; }
                                             ^--^ SC2090: Quotes/backslashes in this variable will not be respected.

For more information:
  https://www.shellcheck.net/wiki/SC2089 -- Quotes/backslashes will be treate...
  https://www.shellcheck.net/wiki/SC2090 -- Quotes/backslashes in this variab...
```

scop:
Ok, sounds like a shellcheck bug anyway, albeit a not that worrying one, and not really introduced or made worse by this PR, so feel free to merge at will.

Would you mind reporting the shellcheck issue upstream?

akinomyoga:
Maybe this is not really a bug of shellcheck, but shellcheck tries to find the value of uninitialized `var` in `func2` in the above example. For example, if we slightly modify `func2` as follows, the errors go away:

```bash
function func2 { local var; var=1; echo $var; }
```

If shellcheck misses something, that's the fact that the local variable `var` of `func1` is not used anywhere else because `func1` doesn't call any other shell function.

----

I'll later again check and merge this PR. Thank you!

PR #740, opened 2022-04-12
style: prefer empty/`!` over `-n`/`-z`

scop:
Originally from comments in https://github.com/scop/bash-completion/pull/678.

Docs only for now; we should implement the actual change in this PR too.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -636,7 +636,7 @@ _filedir()
 
         # Try without filter if it failed to produce anything and configured to
         [[ ${BASH_COMPLETION_FILEDIR_FALLBACK-${COMP_FILEDIR_FALLBACK-}} &&
-            -n $arg && ${#toks[@]} -lt 1 ]] && {
+            $arg && ${#toks[@]} -lt 1 ]] && {
             reset=$(shopt -po noglob)
             set -o noglob
             toks+=($(compgen -f ${plusdirs+"${plusdirs[@]}"} -- $quoted))
@@ -1567,7 +1567,7 @@ _fstypes()
              $([[ -d /etc/fs ]] && command ls /etc/fs)"
     fi
 
-    [[ -n $fss ]] && COMPREPLY+=($(compgen -W "$fss" -- "$cur"))
+    [[ $fss ]] && COMPREPLY+=($(compgen -W "$fss" -- "$cur"))
 }
 
 # Get real command.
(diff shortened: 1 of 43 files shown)


scop:
Rebased and hopefully completed. This might interfere/conflict with other sweeping changes going on -- if so, we can leave this for later and process more important things first, or sneak this in in the middle, I don't mind :)

configure.ac
```diff
@@ -8,7 +8,7 @@ AC_PROG_LN_S
 AC_PROG_MKDIR_P
 AC_PROG_SED
 AC_ARG_WITH([pytest],[  --with-pytest=executable],[PYTEST="$withval"])
-if test -z "$PYTEST"; then
+if test ! "$PYTEST"; then
```
akinomyoga, line 11:
```suggestion
if test "x$PYTEST" = x; then
```

Not sure if we really need to care about (pre-POSIX) Bourne `sh`, but Autotools seem to target Bourne `sh`. I think it is completely fine to forget about the Bourne `sh`, but I just noticed this.

I have checked the behavior of Bourne `sh`, and it seems to fail (i.e., output an error message) with `test ! "$PYTEST"` when `$PYTEST` contains a unary operator like `-f`. On the other hand, `test -z "$PYTEST"` seems to work always as expected at least in the Bourne `sh` that I tested. I guess the Bourne `sh` way is `test "x$PYTEST" = x`.

akinomyoga (approved):
Thank you! I guess the conflict resolution for this is not that difficult, so we could just merge this.

configure.ac
scop: Yeah, I actually thought I'd leave this one alone, but reading the POSIX spec made me think it'd work and I failed to think about the case the variable contains an operator. Thanks for pointing it out!

akinomyoga:
Thanks! I'm fine with merging this PR. You can merge it any time now or later when you think it is useful.

bash_completion
```diff
@@ -1989,7 +1989,7 @@ _cd()
 
     # Use standard dir completion if no CDPATH or parameter starts with /,
     # ./ or ../
-    if [[ -z ${CDPATH:-} || $cur == ?(.)?(.)/* ]]; then
+    if [[ ${CDPATH:-} || $cur == ?(.)?(.)/* ]]; then
```
akinomyoga, line 1992: The test for `cd` is failing... Ah, this is a typo; `!` seems to have been missing here.

akinomyoga (changes requested):

bash_completion
scop: Ah! Fixed, merging.

PR #742, opened 2022-04-22
Add completion for bluetoothctl

audunmg:
I've used this for some months now and thought I'd share in the hope that it's useful for someone else.

Please let me know of any changes I should make to make it fit in here. I tried my best, but since it's my first completion i might have missed something.

### DIFF
--- a/completions/bluetoothctl
+++ b/completions/bluetoothctl
@@ -1,0 +1,46 @@
+# bluetoothctl(1) completion                                    -*- shell-script -*-
+
+_bluetoothctl()
+{
+    local cur=${COMP_WORDS[COMP_CWORD]}
+    local prev=${COMP_WORDS[COMP_CWORD-1]}
+    if [[ "$COMP_CWORD" == "1" ]]; then
+        # Parse help for list of commands and remove ansi colors
+        COMPREPLY=( $( compgen -W "$(bluetoothctl help | sed -n '/  /{s/\x1B\[[0-9;]*[a-zA-Z]//g;s/\x01//g;s/\x02//g;s/ .*//p}')" -- $cur))
+        return
+    fi
+    if [[ "$COMP_CWORD" == "2" ]]; then
+        case "$prev" in
+            info|pair|trust|untrust|block|unblock|remove|connect|disconnect)
+                # Return a device to complete
+                COMPREPLY=( $( compgen -W "$(bluetoothctl devices|awk '{print $2}')" -- $cur))
+                    ;;
+            show|select)
+                # Return a controller address
+                COMPREPLY=( $( compgen -W "$(bluetoothctl list|awk '{print $2}')" -- $cur))
+                ;;
+            power|pairable|discoverable|scan)
+                COMPREPLY=( $( compgen -W "on off" -- $cur))
+                ;;
+            agent)
+                COMPREPLY=( $( compgen -W "on off capability" -- $cur))
+                ;;
+            advertise)
+                COMPREPLY=( $( compgen -W "on off type" -- $cur))
+                ;;
+        esac
+    fi
+    if [[ ${COMP_WORDS[1]} == "menu" ]]; then
+        if [[ "$COMP_CWORD" == "2" ]]; then
+            # List out submenues
+            COMPREPLY=( $(compgen -W "$(bluetoothctl help | sed -n '/Submenu/{s/\x1B\[[0-9;]*[a-zA-Z]//g;s/\x02//g;s/\x01//;s/ .*//p}')" -- $cur))
+            return
+        fi
+        if [[ "$COMP_CWORD" == "3" ]]; then
+            # Parse submenu and return commands.
+            COMPREPLY=( $( compgen -W "$(bluetoothctl menu $prev | sed -n '/  /{s/\x1B\[[0-9;]*[a-zA-Z]//g;s/\x01//g;s/\x02//g;s/ .*//p}' )" -- $cur))
+        fi
+    fi
+}
+
+# ex: filetype=sh


completions/bluetoothctl
```diff
@@ -0,0 +1,46 @@
+# bluetoothctl(1) completion                                    -*- shell-script -*-
+
+_bluetoothctl()
+{
+    local cur=${COMP_WORDS[COMP_CWORD]}
+    local prev=${COMP_WORDS[COMP_CWORD-1]}
+    if [[ "$COMP_CWORD" == "1" ]]; then
```
akinomyoga, line 7:
```suggestion
    if ((COMP_CWORD == 1)); then
```

- You can use the arithmetic command.

The same applies to the later lines testing `COMP_CWORD`.

completions/bluetoothctl
```diff
@@ -0,0 +1,46 @@
...
+        # Parse help for list of commands and remove ansi colors
+        COMPREPLY=( $( compgen -W "$(bluetoothctl help | sed -n '/  /{s/\x1B\[[0-9;]*[a-zA-Z]//g;s/\x01//g;s/\x02//g;s/ .*//p}')" -- $cur))
```
akinomyoga, line 9:
```suggestion
        COMPREPLY=( $( compgen -W "$(bluetoothctl help | sed -n '/  /{s/\x1B\[[0-9;]*[a-zA-Z]//g;s/\x01//g;s/\x02//g;s/ .*//p}')" -- "$cur"))
```
- quoting

```suggestion
        COMPREPLY=( $( compgen -W "$(bluetoothctl help | sed -n '/  /{s/\x1B\[[0-9;]*[a-zA-Z]//g;s/\x01//g;s/\x02//g;s/ .*//p;}')" -- $cur))
```

- [POSIX sed](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html) requires a semicolon or newline before \<right-brace>.

```suggestion
        COMPREPLY=( $( compgen -W "$(bluetoothctl help | sed -n '/  /{s/'$'\e''\[[0-9;]*[a-zA-Z]//g;s/['$'\x01\x02'']//g;s/ .*//p}')" -- $cur))
```

- AFAIK, POSIX sed doesn't necessarily support the escape sequences like `\x1B`. Here, I'd suggest expanding the sequences using Bash's `$'...'` before passing it to `sed`.

These comments also apply to the later lines that contain the same structure.

completions/bluetoothctl
```diff
@@ -0,0 +1,46 @@
...
+        return
+    fi
+    if [[ "$COMP_CWORD" == "2" ]]; then
+        case "$prev" in
+            info|pair|trust|untrust|block|unblock|remove|connect|disconnect)
+                # Return a device to complete
+                COMPREPLY=( $( compgen -W "$(bluetoothctl devices|awk '{print $2}')" -- $cur))
+                    ;;
```
akinomyoga, line 17:
```suggestion
                ;;
```

- The indentation seems to be inconsistent with the other similar lines?

completions/bluetoothctl
```diff
@@ -0,0 +1,46 @@
...
+                ;;
+        esac
+    fi
+    if [[ ${COMP_WORDS[1]} == "menu" ]]; then
+        if [[ "$COMP_CWORD" == "2" ]]; then
+            # List out submenues
+            COMPREPLY=( $(compgen -W "$(bluetoothctl help | sed -n '/Submenu/{s/\x1B\[[0-9;]*[a-zA-Z]//g;s/\x02//g;s/\x01//;s/ .*//p}')" -- $cur))
+            return
+        fi
+        if [[ "$COMP_CWORD" == "3" ]]; then
```
akinomyoga, line 39:
```suggestion
        elif ((COMP_CWORD" == 3)); then
```

completions/bluetoothctl
```diff
@@ -0,0 +1,46 @@
...
+            # Parse submenu and return commands.
+            COMPREPLY=( $( compgen -W "$(bluetoothctl menu $prev | sed -n '/  /{s/\x1B\[[0-9;]*[a-zA-Z]//g;s/\x01//g;s/\x02//g;s/ .*//p}' )" -- $cur))
+        fi
+    fi
+}
+
```
akinomyoga, line 45: I think `complete -F _bluetoothctl bluetoothctl` is missing.

completions/bluetoothctl
```diff
@@ -0,0 +1,46 @@
+# bluetoothctl(1) completion                                    -*- shell-script -*-
+
+_bluetoothctl()
+{
+    local cur=${COMP_WORDS[COMP_CWORD]}
+    local prev=${COMP_WORDS[COMP_CWORD-1]}
```
akinomyoga, line 6: If you intend to use the auto-loading mechanism of bash-completion (i.e., putting the file in the `completions` directory and letting `bash-completion` load it automatically on demand), I suggest using `_init_completion` for the initialization (as done in other completions). It doesn't only provide a better extraction of the words but also the contextual completions for the variable names and redirections. It might also be used by `bash-completion` for hooks in the future.

akinomyoga (commented):
Thank you for your contribution.

Have you checked [CONTRIBUTING.md](https://github.com/scop/bash-completion/blob/master/CONTRIBUTING.md) and consulted with the upstream [BlueZ](http://www.bluez.org/) about it? [The Zsh completion seems to be already in the repository](https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/completion), so I guess they would accept your contribution to the upstream. They seem to accept patches through [the developer mailing list](http://www.bluez.org/development/lists/).

audunmg:
Good point,  have not consulted with BlueZ, I will do that, and fix the points you mentioned first.

Is it OK if I close this, and reopen if upstream prefer it here?

Thank you for your time, I appreciate it.

completions/bluetoothctl
audunmg: In that case, it might just be better to not use sed at all. I think this should be doable in bash only.

akinomyoga:
> Is it OK if I close this, and reopen if upstream prefer it here?

Sure!

completions/bluetoothctl
akinomyoga: Yeah, I think either works, so you can use what you like.

PR #744, opened 2022-05-04
Add more module dirs to kldload

Dystopian:
Current completion only shows modules from `/boot/kernel/`. There are more directories with modules now in FreeBSD e.g. local modules go usually to `/boot/modules/`. Looks like all current directories are written in `sysctl` variable `kern.module_path` (see also `kldconfig(8)`). New implementation considers all these paths.

I'm not experienced in bash programming so there can be some bugs. Tested on FreeBSD 12.3.

### DIFF
--- a/completions/kldload
+++ b/completions/kldload
@@ -7,12 +7,16 @@ _kldload()
     local cur prev words cword
     _init_completion || return
 
-    local moddir=/modules/
-    [[ -d $moddir ]] || moddir=/boot/kernel/
+    local moddirs modules i
+    IFS=';' read -ra moddirs <<< "$(sysctl -n kern.module_path)"
 
     compopt -o filenames
-    COMPREPLY=($(compgen -f "$moddir$cur"))
-    COMPREPLY=(${COMPREPLY[@]#$moddir})
+    COMPREPLY=()
+    for i in "${moddirs[@]}"; do
+        modules=($(compgen -f "$i/$cur"))
+        modules=(${modules[@]#$i/})
+        COMPREPLY=("${COMPREPLY[@]}" "${modules[@]}")
+    done
     COMPREPLY=(${COMPREPLY[@]%.ko})
 
 } &&
--- a/completions/kldload
+++ b/completions/kldload
@@ -8,7 +8,7 @@ _kldload()
     _init_completion || return
 
     local moddirs modules i IFS=";"
-    moddirs=($(sysctl -n kern.module_path))
+    moddirs=($(kldconfig -r 2>/dev/null))
     _comp_unlocal IFS
 
     compopt -o filenames
--- a/completions/kldload
+++ b/completions/kldload
@@ -7,6 +7,11 @@ _kldload()
     local cur prev words cword
     _init_completion || return
 
+    if [[ "$cur" == */* ]]; then
+        _filedir ko
+        return
+    fi
+
     local moddirs modules i IFS=";"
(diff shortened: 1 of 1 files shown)


completions/kldload
```diff
@@ -7,12 +7,16 @@ _kldload()
     local cur prev words cword
     _init_completion || return
 
-    local moddir=/modules/
-    [[ -d $moddir ]] || moddir=/boot/kernel/
+    local moddirs modules i
+    IFS=';' read -ra moddirs <<< "$(sysctl -n kern.module_path)"
```
scop, line 11: We avoid herestrings because they use temporary files under the hood. Maybe something like this (untested) instead?

```suggestion
    local moddirs modules i IFS=";"
    moddirs=($(sysctl -n kern.module_path))
    _comp_unlocal IFS
```

completions/kldload
```diff
@@ -7,12 +7,16 @@ _kldload()
...
 
     compopt -o filenames
-    COMPREPLY=($(compgen -f "$moddir$cur"))
-    COMPREPLY=(${COMPREPLY[@]#$moddir})
+    COMPREPLY=()
+    for i in "${moddirs[@]}"; do
+        modules=($(compgen -f "$i/$cur"))
+        modules=(${modules[@]#$i/})
+        COMPREPLY=("${COMPREPLY[@]}" "${modules[@]}")
```
scop, line 18:
```suggestion
        COMPREPLY+=("${modules[@]}")
```

completions/kldload
```diff
@@ -7,12 +7,16 @@ _kldload()
...
 
-    local moddir=/modules/
-    [[ -d $moddir ]] || moddir=/boot/kernel/
+    local moddirs modules i
+    IFS=';' read -ra moddirs <<< "$(sysctl -n kern.module_path)"
 
     compopt -o filenames
-    COMPREPLY=($(compgen -f "$moddir$cur"))
-    COMPREPLY=(${COMPREPLY[@]#$moddir})
+    COMPREPLY=()
```
scop, line 14: I don't think we need to initialize/empty this.

```suggestion
```

completions/kldload
scop, line 11: I know nothing about this, but could you elaborate on why use `sysctl -n kern.module_path` instead of `kldconfig -r`?

Also, should we fall back to the previous hardcoded dirs if the utility we choose to use here yields nothing? Should we output its stderr to `/dev/null` in any case?

scop (changes requested):
Thanks! Some comments follow.

completions/kldload
Dystopian: Will replace command and add error handling. Also I found `kldload(8)` should support files with absolute/relative paths, will add this.

Dystopian:
> should we fall back to the previous hardcoded dirs if the utility we choose to use here yields nothing?

Due to man `kldload(8)` won't load anything if `kldconfig -r` is empty. I think no sense to use old hardcoded dirs as fallback.

Directory handling is added too so I thinks it's done.

completions/kldload
```diff
@@ -7,14 +7,26 @@ _kldload()
     local cur prev words cword
     _init_completion || return
 
-    local moddir=/modules/
-    [[ -d $moddir ]] || moddir=/boot/kernel/
+    if [[ "$cur" == */* ]]; then
```
scop, line 10: We have some more elaborate checks for what looks like a path in the tree, will convert this to use it after the merge.

scop (approved):

PR #746, opened 2022-05-08
Consistent documentation comments

akinomyoga:
The documentation comments (doc comments) of some functions are not consistent with others. This PR fixes it.

## Discussion

Maybe we can also discuss the syntax of doc comments used in `bash-completion` here. https://github.com/scop/bash-completion/pull/736#issuecomment-1107458933 is a related comment.

- What is the set of directives of the form `@xxxxx` used in `bash-completion`?
  - Currently, `@param`, `@see`, `@deprecated`, `@return` are used in the codebase.
  - Is the current syntax in `bash-completion` doc comments defined by Javadoc?
  - Is there already any tool to process these doc comments? If not, maybe we can introduce one or may create a simple script to extract these comments.
- Can we introduce `@var` that is used to mark the variable name referenced or modified by the function?
- `@usage` might also be useful. Currently, usages are written as `Usage: ....` in the comment, e.g.,
  https://github.com/scop/bash-completion/blob/6f1bbda3c66814befa8025d49363b4070ef20008/bash_completion#L181
  https://github.com/scop/bash-completion/blob/6f1bbda3c66814befa8025d49363b4070ef20008/bash_completion#L203-L204
- `@option` might also be useful.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -165,7 +165,7 @@ dequote()
 # will not unshadow the global variable. Rather, the result will be a local
 # variable in an unset state.
 # Usage: local IFS='|'; _comp_unlocal IFS
-# Param: $* Variable names to be unset
+# @param $* Variable names to be unset
 _comp_unlocal()
 {
     if ((BASH_VERSINFO[0] >= 5)) && shopt -q localvar_unset; then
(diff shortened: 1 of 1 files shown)


akinomyoga:
> I'm not aware of a processor nor a specific set of keywords to use with the `@` prefixed markers. They were introduced by someone else a long time ago. They do look a lot like Java/PHP comments.

I see, thanks for the information!

PR #753, opened 2022-06-06
updates for reportbug

nmartensen:


scop (commented):
Thanks and sorry for taking so long to respond. I filed another PR which parses available values from the option help output, so we can stop hardcoding these lists and some more, see https://github.com/scop/bash-completion/pull/852

PR #759, opened 2022-06-14
fix(cryptsetup): update the list of actions

### LINKED ISSUE #758: [cryptsetup] luksErase is not completed
## Describe the bug

When I press the tab key in `cryptsetup luks`, luksErase is not completed.

## To reproduce

1. Type `cryptsetup luksE`

## Expected behavior

luksErase appears as a candidate for completion.

- [x] Operating system name/distribution and version: Arch Linux
- [x] bash version, `echo "$BASH_VERSION"`: 5.1.16(1)-release
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.11

akinomyoga:
Fixes #758 

@carbongreat13 Thank you for the report (#758). Can you test the behavior of the updated [compeltions/cryptsetup](https://github.com/akinomyoga/bash-completion/raw/cryptsetup-action/completions/cryptsetup)? You can download and source it as

```bash
$ curl -Lo cryptsetup https://github.com/akinomyoga/bash-completion/raw/cryptsetup-action/completions/cryptsetup
$ source cryptsetup
$ cruptsetup luksE[Tab]
```

### DIFF
--- a/completions-core/cryptsetup.bash
+++ b/completions-core/cryptsetup.bash
@@ -91,10 +91,15 @@ _comp_cmd_cryptsetup()
             _comp_compgen_help
             [[ ${COMPREPLY-} == *= ]] && compopt -o nospace
         else
-            _comp_compgen -- -W 'open close resize status benchmark repair
-                erase luksFormat luksAddKey luksRemoveKey luksChangeKey
-                luksKillSlot luksUUID isLuks luksDump tcryptDump luksSuspend
-                luksResume luksHeaderBackup luksHeaderRestore'
+            _comp_compgen -- -W 'benchmark bitlkClose bitlkDump bitlkOpen close
+                config convert create erase fvault2Close fvault2Dump
+                fvault2Open isLuks loopaesClose loopaesOpen luksAddKey
+                luksChangeKey luksClose luksConfig luksConvertKey luksDump
+                luksErase luksFormat luksHeaderBackup luksHeaderRestore
+                luksKillSlot luksOpen luksRemoveKey luksResume luksSuspend
+                luksUUID open plainClose plainOpen reencrypt refresh remove
+                repair resize status tcryptClose tcryptDump tcryptOpen
+                tcryptOpen_ token'
         fi
     fi
 
--- a/completions-core/cryptsetup.bash
+++ b/completions-core/cryptsetup.bash
@@ -15,8 +15,7 @@ _comp_cmd_cryptsetup__action()
     local REPLY IFS=$' \t\n'
     _comp_dequote "${1-}" || return 1
     local cmd=${REPLY:-cryptsetup}
-    local -a actions
-    _comp_split -l actions "$(
+    _comp_compgen_split -l -- "$(
         {
             LC_ALL=C "$cmd" --help 2>&1 |
                 command sed -n '
@@ -27,39 +26,6 @@ _comp_cmd_cryptsetup__action()
                 _comp_awk '/^[[:space:]]+[[:alnum:]_]+([[:space:]]+(-[^[:space:].]+|<[^<>]+>|\[[^][]+\]|\(.*\)|or))+$/ && $1 != "cryptsetup" {print $1}'
         } | sort -u
     )"
-
-    if ((${#actions[@]} == 0)); then
-        # The fallback action list is extracted from the following sources:
-        # - https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/src/cryptsetup.c#L3154-3208 (search for "Handle aliases")
-        # - https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/src/cryptsetup.c#L2831-2867 (search for "struct action_type")
-        # - https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/src/cryptsetup_args.h#L28-53 (see the macros "*_ACTION")
-        # - The outputs of "cryptsetup --help" and "man cryptsetup" for
-        #   cryptsetup-2.7.5: "fvault2Close", "fvault2Dump", "fvault2Open", and
-        #   "tcryptOpen_" are added.
-        actions=(benchmark bitlkClose bitlkDump bitlkOpen close config convert
-            create erase fvault2Close fvault2Dump fvault2Open isLuks
-            loopaesClose loopaesOpen luksAddKey luksChangeKey luksClose
-            luksConfig luksConvertKey luksDump luksErase luksFormat
-            luksHeaderBackup luksHeaderRestore luksKillSlot luksOpen
-            luksRemoveKey luksResume luksSuspend luksUUID open plainClose
-            plainOpen reencrypt refresh remove repair resize status tcryptClose
-            tcryptDump tcryptOpen tcryptOpen_ token)
-
-        # We attempt to filter the supported actions by the strings in the binary.
-        local path
-        if path=$(type -P -- "$cmd" 2>/dev/null || type -P -- cryptsetup 2>/dev/null); then
-            local filtering_pattern
-            printf -v filtering_pattern '%s\n' "${actions[@]}"
-            filtering_pattern=${filtering_pattern%$'\n'}
-
-            local filtered_actions
-            filtered_actions=$(strings "$path" 2>/dev/null | command grep -Fx "$filtering_pattern" | sort -u) &&
-                [[ $filtered_actions ]] &&
-                _comp_split -l actions "$filtered_actions"
-        fi
-    fi
-
-    _comp_compgen -- -W '"${actions[@]}"'
 }
 
 _comp_cmd_cryptsetup()
--- a/completions-core/cryptsetup.bash
+++ b/completions-core/cryptsetup.bash
@@ -10,6 +10,58 @@ _comp_cmd_cryptsetup__device()
     _comp_compgen -c "${cur:-/dev/}" filedir
 }
 
+_comp_cmd_cryptsetup__action()
+{
+    local REPLY IFS=$' \t\n'
+    _comp_dequote "${1-}" || return 1
+    local cmd=${REPLY:-cryptsetup}
+    local -a actions
+    _comp_split -l actions "$(
+        {
+            LC_ALL=C "$cmd" --help 2>&1 |
+                command sed -n '
+                    /^<action> is one of:$/,/^[^[:space:]]/s/^[[:space:]]\{1,\}\([^[:space:]]\{1,\}\).*/\1/p
+                    /^You can also use old <action> syntax aliases:$/,/^[^[:space:]]/{/^[[:space:]]\{1,\}/!d;s///;s/[:(),[:space:]]\{1,\}/\n/gp;}
+                '
+            LC_ALL=C man cryptsetup 2>&1 |
+                _comp_awk '/^[[:space:]]+[[:alnum:]_]+([[:space:]]+(-[^[:space:].]+|<[^<>]+>|\[[^][]+\]|\(.*\)|or))+$/ && $1 != "cryptsetup" {print $1}'
+        } | sort -u
+    )"
+
+    if ((${#actions[@]} == 0)); then
+        # The fallback action list is extracted from the following sources:
+        # - https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/src/cryptsetup.c#L3154-3208 (search for "Handle aliases")
+        # - https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/src/cryptsetup.c#L2831-2867 (search for "struct action_type")
+        # - https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/src/cryptsetup_args.h#L28-53 (see the macros "*_ACTION")
+        # - The outputs of "cryptsetup --help" and "man cryptsetup" for
+        #   cryptsetup-2.7.5: "fvault2Close", "fvault2Dump", "fvault2Open", and
+        #   "tcryptOpen_" are added.
+        actions=(benchmark bitlkClose bitlkDump bitlkOpen close config convert
+            create erase fvault2Close fvault2Dump fvault2Open isLuks
+            loopaesClose loopaesOpen luksAddKey luksChangeKey luksClose
+            luksConfig luksConvertKey luksDump luksErase luksFormat
+            luksHeaderBackup luksHeaderRestore luksKillSlot luksOpen
+            luksRemoveKey luksResume luksSuspend luksUUID open plainClose
+            plainOpen reencrypt refresh remove repair resize status tcryptClose
+            tcryptDump tcryptOpen tcryptOpen_ token)
+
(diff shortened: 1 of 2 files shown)


carbongreat13:
> @carbongreat13 Thank you for the report (#758). Can you test the behavior of the updated [compeltions/cryptsetup](https://github.com/akinomyoga/bash-completion/raw/cryptsetup-action/completions/cryptsetup)? You can download and source it as

I have tested the I was able to confirm that it works correctly.

akinomyoga:
@carbongreat13 Thanks for checking!

completions/cryptsetup
```diff
@@ -40,10 +40,14 @@ _cryptsetup()
             COMPREPLY=($(compgen -W '$(_parse_help "$1")' -- "$cur"))
             [[ ${COMPREPLY-} == *= ]] && compopt -o nospace
         else
-            COMPREPLY=($(compgen -W 'open close resize status benchmark
-                repair erase luksFormat luksAddKey luksRemoveKey luksChangeKey
-                luksKillSlot luksUUID isLuks luksDump tcryptDump luksSuspend
-                luksResume luksHeaderBackup luksHeaderRestore' -- "$cur"))
+            COMPREPLY=($(compgen -W 'benchmark bitlkClose bitlkDump bitlkOpen
```
scop, line 43: If we want to keep hardcoding the list, could we add some pointers here to the source of the list, as well as how to keep the list sorted for easy maintenance? (I believe it before this change the entries were roughly in the order the man page has them, which wasn't noted either, though, but we're here to improve on things ;))

But actually, I think we should consider switching to scraping `--help` output instead of hardcoding. For example on Ubuntu 20.04, it contains
```
<action> is one of:
	open <device> [--type <type>] [<name>] - open device as <name>
	close <name> - close device (remove mapping)
	resize <name> - resize active device
	status <name> - show device status
	benchmark [--cipher <cipher>] - benchmark cipher
	repair <device> - try to repair on-disk metadata
	reencrypt <device> - reencrypt LUKS2 device
	erase <device> - erase all keyslots (remove encryption key)
	convert <device> - convert LUKS from/to LUKS2 format
	config <device> - set permanent configuration options for LUKS2
	luksFormat <device> [<new key file>] - formats a LUKS device
	luksAddKey <device> [<new key file>] - add key to LUKS device
	luksRemoveKey <device> [<key file>] - removes supplied key or key file from LUKS device
	luksChangeKey <device> [<key file>] - changes supplied key or key file of LUKS device
	luksConvertKey <device> [<key file>] - converts a key to new pbkdf parameters
	luksKillSlot <device> <key slot> - wipes key with number <key slot> from LUKS device
	luksUUID <device> - print UUID of LUKS device
	isLuks <device> - tests <device> for LUKS partition header
	luksDump <device> - dump LUKS partition information
	tcryptDump <device> - dump TCRYPT device information
	luksSuspend <device> - Suspend LUKS device and wipe key (all IOs are frozen)
	luksResume <device> - Resume suspended LUKS device
	luksHeaderBackup <device> - Backup LUKS device header and keyslots
	luksHeaderRestore <device> - Restore LUKS device header and keyslots
	token <add|remove|import|export> <device> - Manipulate LUKS2 tokens

You can also use old <action> syntax aliases:
	open: create (plainOpen), luksOpen, loopaesOpen, tcryptOpen
	close: remove (plainClose), luksClose, loopaesClose, tcryptClose
```
CentOS 7's `cryptsetup --help` emits similar looking sections, so I suppose this would be feasible to do in at least somewhat portable manner.

scop (changes requested):

completions/cryptsetup
akinomyoga: Thank you for the comment! I have updated the PR to include the analysis of `--help` (790e524). What is non-trivial is that not all the actions are found in the output of `--help`. For example, for `cryptsetup` 2.4.3 in Fedora 36, `--help` does not contain the reported `luksErase` (#758) while `man` contains the description of `luksErase`. I have decided to also analyze the `man` output.

completions/cryptsetup
```diff
@@ -11,6 +11,50 @@ _cryptsetup_device()
...
+        {
+            LC_ALL=C "$cmd" --help 2>&1 |
+                sed -n '/^<action> is one of:/,/^[^[:space:]]/s/^[[:space:]]\{1,\}\([^[:space:]]\{1,\}\).*/\1/p'
+            LC_ALL=C man cryptsetup 2>&1 |
+                awk '/^[[:space:]]+[[:alnum:]_]+([[:space:]]+(-[^[:space:].]+|<[^<>]+>|\[[^][]+\]|or))*$/ {print $1}'
+        } | sort -u
+    )
+
+    if [[ ! $actions ]]; then
+        # The fallback action list is extracted from the following source:
```
scop, line 29: I think we should drop the fallback altogether, unless we know there is an old version we want to support but cannot because its output is not feasible to parse. If there is, we should document what exactly it is.

completions/cryptsetup
```diff
@@ -11,6 +11,50 @@ _cryptsetup_device()
...
+
+        # We attempt to filter the supported actions by the strings in the binary.
+        local path
+        if path=$(type -P -- "$cmd" 2>/dev/null || type -P -- cryptsetup 2>/dev/null); then
+            local filtering_pattern
+            printf -v filtering_pattern '%s\n' $actions
+            filtering_pattern=${filtering_pattern%$'\n'}
+
+            local filtered_actions
+            filtered_actions=$(strings "$path" | command grep -Fx "$filtering_pattern" | sort -u) &&
```
scop, line 49: If we decide to keep this block or to extract this as a helper to another PR, I suppose this would be in order:
```suggestion
            filtered_actions=$(strings "$path" 2>/dev/null | command grep -Fx "$filtering_pattern" | sort -u) &&
```

completions/cryptsetup
```diff
@@ -11,6 +11,50 @@ _cryptsetup_device()
...
+        # https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/src/cryptsetup_args.h#L28-53 (see the macros "*_ACTION")
+        actions='benchmark bitlkClose bitlkDump bitlkOpen close config convert
+            create erase isLuks loopaesClose loopaesOpen luksAddKey
+            luksChangeKey luksClose luksConfig luksConvertKey luksDump
+            luksErase luksFormat luksHeaderBackup luksHeaderRestore
+            luksKillSlot luksOpen luksRemoveKey luksResume luksSuspend luksUUID
+            open plainClose plainOpen reencrypt refresh remove repair resize
+            status tcryptClose tcryptDump tcryptOpen token'
+
+        # We attempt to filter the supported actions by the strings in the binary.
```
scop, line 41: If we decide to keep the fallback, I suggest we don't do this. It's not something we do elsewhere, and frankly I think it's too much for a corner case here.

Then again, if there is a bunch of other cases we feel we should do this in other completions, then a separate helper function for doing it would be in order, in another PR. Not quite sure myself, but open to opinions.

completions/cryptsetup
```diff
@@ -11,6 +11,50 @@ _cryptsetup_device()
...
+{
+    local ret IFS=$' \t\n'
+    _comp_dequote "${1-}" || return 1
+    local cmd=${ret:-cryptsetup}
+    local actions=$(
+        {
+            LC_ALL=C "$cmd" --help 2>&1 |
+                sed -n '/^<action> is one of:/,/^[^[:space:]]/s/^[[:space:]]\{1,\}\([^[:space:]]\{1,\}\).*/\1/p'
+            LC_ALL=C man cryptsetup 2>&1 |
+                awk '/^[[:space:]]+[[:alnum:]_]+([[:space:]]+(-[^[:space:].]+|<[^<>]+>|\[[^][]+\]|or))*$/ {print $1}'
```
scop, line 24: I think we should add test cases for the parsing part here like we have for dnssec-keygen and xrandr.

scop (changes requested):

completions/cryptsetup
scop: > For example, for `cryptsetup` 2.4.3 in Fedora 36, `--help` does not contain the reported `luksErase`

Perhaps upstream would appreciate a bug report about this?

akinomyoga:
Sorry for the delay. I'm a bit busy this week, so it may delay further.

completions/cryptsetup
akinomyoga: I was just so lazy that I didn't want to check the behavior of old versions. Do we need to check the behavior of old versions of cryptsetup?

completions/cryptsetup
akinomyoga: I'm not sure if the fallback in this function finally gets into the repository, but such a function would be useful anyway where hardcoded lists are needed. I'll later consider it.

completions/cryptsetup
akinomyoga: Actually, with this extraction, there is no guarantee that we do not hit false positives. I just tested it with the version with v2.4.3 in Fedora 36. If we care about the false positives, maybe we should filter out them using `strings` for the result as well as used in the fallback.

completions/cryptsetup
scop: That's kind of my point -- if we do check old versions and want to support them and can't find a way we like to make that happen otherwise, then I'm fine with a hardcoded list and the strings + grep approach. My opinion is that if we go to such lengths, we should document the reasoning and affected versions _why_ we are doing it, why it is important here etc.

Without verifying from the sources, I believe our current standard is not to use fallback lists at all (we do have them in a few places, but it's not a standard), and the strings+grep verification is completely new, so I'm just validating if there's something exceptionally important here that warrants that stuff.

scop:
> Sorry for the delay. I'm a bit busy this week, so it may delay further.

No problem, take your time, thanks for letting me know.

akinomyoga:
This is an old PR, but I updated it.

completions/cryptsetup
akinomyoga: > Perhaps upstream would appreciate a bug report about this?

Now, I've checked the output of `man cryptsetup` in cryptsetup-2.7.5 because the `man` extraction started to fail to pick up `luksErase`. It seems that the man page started to put ` (old syntax)` after `luksErase`, so I think missing `luksErase` in the output of `cryptsetup --help` was intentional.

So I don't think we should upstream the issue of the missing action `luksErase`.

completions-core/cryptsetup.bash
```diff
@@ -11,6 +11,50 @@ _cryptsetup_device()
...
+        {
+            LC_ALL=C "$cmd" --help 2>&1 |
+                sed -n '/^<action> is one of:/,/^[^[:space:]]/s/^[[:space:]]\{1,\}\([^[:space:]]\{1,\}\).*/\1/p'
+            LC_ALL=C man cryptsetup 2>&1 |
+                awk '/^[[:space:]]+[[:alnum:]_]+([[:space:]]+(-[^[:space:].]+|<[^<>]+>|\[[^][]+\]|or))*$/ {print $1}'
+        } | sort -u
+    )
+
+    if [[ ! $actions ]]; then
+        # The fallback action list is extracted from the following source:
```
akinomyoga: The current implementation of the extraction from `cryptsetup --help` and `man cryptsetup` assumes very specific forms of the action list, so I was afraid that the actions may not be correctly extracted for older versions that I didn't test. We had been maintaining the hardcoded list, which worked regardless of `cryptsetup --help` and `man cryptsetup`. Replacing the hardcoded list with the extraction from `cryptsetup --help` and `man cryptsetup` may break the list of actions for untested versions of `cryptsetup`. I was afraid of that.

I'm not sure how much the format of `--help` and `man` is stable. I had to update it to correctly extract the actions for 2.7.5: [`a970c41..5c356f5`](https://github.com/scop/bash-completion/compare/a970c4193c66b0f0d870ae16d342c1c8d3e566fa..5c356f57e5a1a044e8b78b61d52329e83e8d6f44).

Anyway, 3.5 years have passed already after implementing the extraction, and at least the extraction is supported for v2.4.3..v2.7.5 at least. I dropped the hardcoded list with the `strings`/`grep` filtering in commit 0884492d3. In case one might want to reconsider the hardcoded list, I separated the commit to remove the hardcoded list and updated the hardcoded list in the middle commits.

completions-core/cryptsetup.bash
```diff
@@ -11,6 +11,50 @@ _cryptsetup_device()
...
+
+        # We attempt to filter the supported actions by the strings in the binary.
+        local path
+        if path=$(type -P -- "$cmd" 2>/dev/null || type -P -- cryptsetup 2>/dev/null); then
+            local filtering_pattern
+            printf -v filtering_pattern '%s\n' $actions
+            filtering_pattern=${filtering_pattern%$'\n'}
+
+            local filtered_actions
+            filtered_actions=$(strings "$path" | command grep -Fx "$filtering_pattern" | sort -u) &&
```
akinomyoga: I applied the suggested change to the middle commit, though this part is anyway removed by 0884492d3.

completions-core/cryptsetup.bash
```diff
@@ -11,6 +11,50 @@ _cryptsetup_device()
...
+        # https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/src/cryptsetup_args.h#L28-53 (see the macros "*_ACTION")
+        actions='benchmark bitlkClose bitlkDump bitlkOpen close config convert
+            create erase isLuks loopaesClose loopaesOpen luksAddKey
+            luksChangeKey luksClose luksConfig luksConvertKey luksDump
+            luksErase luksFormat luksHeaderBackup luksHeaderRestore
+            luksKillSlot luksOpen luksRemoveKey luksResume luksSuspend luksUUID
+            open plainClose plainOpen reencrypt refresh remove repair resize
+            status tcryptClose tcryptDump tcryptOpen token'
+
+        # We attempt to filter the supported actions by the strings in the binary.
```
akinomyoga: This fallback part is removed in 0884492d3.

completions-core/cryptsetup.bash
```diff
@@ -11,6 +11,50 @@ _cryptsetup_device()
...
+{
+    local ret IFS=$' \t\n'
+    _comp_dequote "${1-}" || return 1
+    local cmd=${ret:-cryptsetup}
+    local actions=$(
+        {
+            LC_ALL=C "$cmd" --help 2>&1 |
+                sed -n '/^<action> is one of:/,/^[^[:space:]]/s/^[[:space:]]\{1,\}\([^[:space:]]\{1,\}\).*/\1/p'
+            LC_ALL=C man cryptsetup 2>&1 |
+                awk '/^[[:space:]]+[[:alnum:]_]+([[:space:]]+(-[^[:space:].]+|<[^<>]+>|\[[^][]+\]|or))*$/ {print $1}'
```
akinomyoga: It is difficult to test the exact set of available actions of an arbitrary version of cryptsetup in the system. However, we started to prepare mock commands that output fixed contents of commands of specific versions, so we may prepare one for `cryptsetup` and `man` for testing purposes.

PR #762, opened 2022-06-23
Add support for 7zr and 7zz

kevinoid:
p7zip provides [7zr(1)] in addition to [7za(1)] and [7z(1)].  7zr behaves like 7z with the exception that it only supports the 7z, LZMA, and XZ formats.

Since [7-Zip 21.01 alpha], the 7-Zip project provides a command-line version of 7-Zip for Linux, [7zz(1)].  It shares nearly all of the command-line options and syntax of the p7zip port.

This PR applies the current completions for 7z to 7zr and 7zz.

**Note:** There are a few minor differences between 7zz and the others (e.g. 7zz does not support `-mh*` or `-ms*`) which will can be addressed by subsequent work.

Thanks for considering,
Kevin

[7-Zip 21.01 alpha]: https://sourceforge.net/p/sevenzip/discussion/45797/thread/d401ab2966/
[7z(1)]: https://github.com/jinfeihan57/p7zip/blob/v17.04/man1/7z.1
[7za(1)]: https://github.com/jinfeihan57/p7zip/blob/v17.04/man1/7za.1
[7zr(1)]: https://github.com/jinfeihan57/p7zip/blob/v17.04/man1/7zr.1
[7zz(1)]: https://manpages.debian.org/testing/7zip/7zz.1.en.html

### DIFF
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -485,6 +485,7 @@ CLEANFILES = \
 	7za \
 	7zr \
 	7zz \
+	7zzs \
 	aclocal-1.10 \
 	aclocal-1.11 \
 	aclocal-1.12 \
@@ -801,7 +802,8 @@ symlinks: $(DATA)
 	$(ss) 7z \
 		7za \
 		7zr \
-		7zz
+		7zz \
+		7zzs
 	$(ss) aclocal \
 		aclocal-1.10 aclocal-1.11 aclocal-1.12 aclocal-1.13 \
 		aclocal-1.14 aclocal-1.15 aclocal-1.16
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -483,6 +483,8 @@ EXTRA_DIST = $(bashcomp_DATA)
 
 CLEANFILES = \
 	7za \
+	7zr \
+	7zz \
 	aclocal-1.10 \
 	aclocal-1.11 \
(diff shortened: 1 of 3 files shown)


kevinoid:
Note: ci/pre-commit failed due to gitlint complaining about the line with https://sourceforge.net/p/sevenzip/discussion/45797/thread/d401ab2966/ being too long.  Let me know if you want me to do anything about that.

kevinoid:
I added a commit to complete `7zzs` (a statically-linked version of `7zz` included in the Linux 7-Zip downloads) as well.

completions/Makefile.am
```diff
@@ -797,7 +797,10 @@ CLEANFILES = \
 
 symlinks: $(DATA)
 	$(ss) 7z \
-		7za
+		7za \
+		7zr \
+		7zz \
+		7zzs
```
scop, line 803: We should add these to `CLEANFILES` and `.gitignore` too, git grep for 7za to find the spots.

scop (changes requested):

completions/Makefile.am
kevinoid: Good catch!  Thanks @scop.  Fix pushed.

scop:
Thanks! Added the `.gitignore` bits that I guess went unnoticed from my previous comment.

PR #763, opened 2022-06-23
feat(7z): complete h i and rn commands

kevinoid:
These were added to the help output of p7zip 15.09 and are present in all versions of 7zz.

Thanks,
Kevin

### DIFF
--- a/completions/7z
+++ b/completions/7z
@@ -11,7 +11,7 @@ _7z()
     fi
 
     local mode
-    [[ ${words[1]} == [adu] ]] && mode=w || mode=r
+    [[ ${words[1]} == @(a|d|rn|u) ]] && mode=w || mode=r
 
     case $cur in
         -ao*)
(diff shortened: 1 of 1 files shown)


completions/7z
```diff
@@ -6,7 +6,7 @@ _7z()
     _init_completion -n = || return
 
     if ((cword == 1)); then
-        COMPREPLY=($(compgen -W 'a b d e l t u x' -- "$cur"))
+        COMPREPLY=($(compgen -W 'a b d e h i l rn t u x' -- "$cur"))
```
scop, line 9: Should we (not) make `rn` a `mode=w` word a few lines below?

scop (changes requested):

completions/7z
kevinoid: Yes indeed.  Good call.  Fix pushed.

scop (approved):

scop:
Thanks!

Issue #704, opened 2022-02-23
bash completion mangles file names containing newline characters


idallen:
## Describe the bug

BASH completion cannot correctly handle file names containing newline characters.
The file names are split apart at the newlines, making them useless.

## To reproduce
~~~~~
        # First, get a shell with no completion loaded and show it working:     
        $ bash --norc
        bash-5.0$ ls -b
        one\ntwo\nthree\nfour\nfive\nsix
        bash-5.0$ xxx <TAB>'one
        two
        three
        four
        five
        six'
        bash-5.0$ touch foo
        bash-5.0$ xxx <TAB>
        foo                               one^Jtwo^Jthree^Jfour^Jfive^Jsix
        bash-5.0$ xxx o<TAB>'one
        two
        three
        four
        five
        six'

        # Now, load the completion scripts and watch it break:                  
        bash-5.0$ source /usr/share/bash-completion/bash_completion
        bash-5.0$ ls -b
        foo  one\ntwo\nthree\nfour\nfive\nsix
        bash-5.0$ xxx <TAB>
        five   foo    four   one    six    three  two
        bash-5.0$ xxx o<TAB>
        five   four   one    six    three  two
        # The last two completions are garbage.                                 
        # The file name is being split on newlines. 
~~~~~
## Expected behavior

See above, before the loading of `/usr/share/bash-completion/bash_completion`

## Versions (please complete the following information)

- [`GNU/linux/Ubuntu 20.04.4 LTS`] Operating system name/distribution and version:
- [`5.0.17(1)-release`] bash version, `echo "$BASH_VERSION"`:
- [`2.10`] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`:

## Additional context

I used the nonexistent command name `xxx` so as not to invoke any helper completion scripts.

## Debug trace

See attached  [typescript.txt](https://github.com/scop/bash-completion/files/8127550/typescript.txt)

calestyo:
Just for the records:

I think doing this properly will be very difficult. I'd guess problems like that don't show up only for files containing newlines, but also for other characters (a filename in POSIX is rather binary and may contain everything except `/` and `NUL`).

Very problematic would also be filenames with trailing newlines, as these get stripped off in any command substitutions.
All workarounds for that have their own extremely tricky problems.

calestyo:
Actually, one part of a way to fix this, could be to "simply" use shell escaped strings for any completions that contain weird characters.
[...]

PR #765, opened 2022-06-26
fix(scp): keep tilde in local paths by using _filedir instead of _expand

maxnikulin:
Do not replace `~` by `/home/user/` for local files to get behavior of `scp` consistent with other tools like `ls`. I hope, dropping `_expand` and adding `_filedir` to have file name completion will not cause issues like [Ville Skyttä \[Bash-completion-devel\] `_expand`, scp and quoting](https://alioth-lists-archive.debian.net/pipermail/bash-completion-devel/2009-February/000973.html) when a patch intended to fix [Debian #521406. bash-completion: leading tilde always expanded](https://bugs.debian.org/521406)
was reverted.


maxnikulin:
I have a little experience with bash completion, so I can easily miss important use cases. You can consider this pull request as an invitation to discuss how to properly fix the issue with expanding of tilde during completion of `scp` arguments.

I do not like when `~` is expanded by `/home/user` since it makes commands longer and more noisy. For a decade I had a hack from [1] redefining `_expand` in my `.bashrc` file. Recently I decided to try if the problem has been fixed and the snippet may be removed. For most of commands completion works accordingly to my preferences, but for some tools tilde is expanded. I checked bug reports another time and realized that a fix was committed and reverted. Due to other changes, `_longopt` widely used for various tools preserves tilde however. Maybe it is better to fix particular tools including `scp` than try to patch `_expand` (that is not used by `_filedir` called from `_longopt` and helpers specific to various tools).

The cases that this change should address:

    scp ~/Do
    scp localhost:/tmp/file ~us

[1] [lp:622403 bash-completion: readline expand-tilde not acknowledged](https://bugs.launchpad.net/bugs/622403)

akinomyoga:
The pathname completions are originally processed by [`_comp_xfunc_ssh_scp_local_files`](https://github.com/scop/bash-completion/pull/765/files#diff-ea186058eaccb2fe1f557ed71d7cf58a9220a37ed3ab7fae9d2b0c99e48029deR592). With this PR, filenames are doubly generated.

```bash
$ scp ./COPYING[TAB twice]
COPYING   COPYING
```

maxnikulin:
> ```shell
> $ scp ./COPYING[TAB twice]
> COPYING   COPYING
> ```

Mea culpa. I have added a bit more code to avoid [_comp_xfunc_ssh_scp_local_files](https://github.com/scop/bash-completion/blob/b85b88e55c15790a6e87933b7633f26992ce56e1/completions/ssh#L592) when possible, however I am not sure that I have not broken something else. Counting `COMPREPLY` options may be fragile, but I have not managed to pass variants through some function to determine if file exists and whether it is a directory, but keeping tildes. I do not consider code with a subprocess per each completion variant.

scop:
While I dislike the tilde expansion myself as well and do support efforts to get rid of it as much as possible, I don't think we should make behavioral changes here without a set of unit tests having good coverage over the possibilities. This has been a tricky and bug prone thing in the past.

The same probably applies to all other uses of `_expand` we have as well (rsync and sshfs at least -- I suspect the instance in `povray` is a false one that could be just removed, but IIRC I had no way to verify it when I tried back in the day).

maxnikulin:
> While I dislike the tilde expansion myself as well and do support efforts to get rid of it as much as possible, I don't think we should make behavioral changes here without a set of unit tests having good coverage over the possibilities. This has been a tricky and bug prone thing in the past.

It would be great to have unit tests and it can be done by an independent pull request since it would be improvement even in the case of declining of this change.

In the meanwhile I have updated the patch to avoid changing of `nospace` option. I still do not like `scp` related code. Frankly speaking, creating the pull request I was expecting reaction like "it is a legacy code, look at the command X that can do the same in a robust way". Are there some hints which files may be considered as a kind of style guide?

I am not motivated enough to create unit tests for so fragile code. I have missed an obvious issue with initial variant of my patch partially because I noticed enough glitches with original code:
- File names with newline characters (`M-/` works) #704.
- `\*2` for `*2` file unescaped in response to `[Tab]`.
- File names with trailing `@` are completed without it due to `ls`-based approach.
- `-F` as a prefix is supported, but it seems there are `-iidentity` and `-Sprogram` variants.
- IPv6 addressed are displayed as completion variants but not actually completed.
- `scp -F-F [Tab]` causes `sed` help message due to unknown `-F` option (I was intended to check `./-F` file name).
- Looking at the file list to find a source for inspiration, I discovered that `perl -x~fil[Tab]` treats tab as at the beginning of word, so completes user names, not files started with literal tilde.

PR #766, opened 2022-06-26
feat(doas): new completion

### LINKED ISSUE #505: Adding doas support.
Can someone please add support for ```doas```?

aarchetype:
Fix #505

### DIFF
--- a/completions-core/doas.bash
+++ b/completions-core/doas.bash
@@ -1,0 +1,45 @@
+# doas(1) completion                                       -*- shell-script -*-
+
+_comp_cmd_doas()
+{
+    local cur prev words cword split
+    _init_completion -s || return
+
+    local i
+
+    for ((i = 1; i <= cword; i++)); do
+        if [[ ${words[i]} != -* ]]; then
+            local PATH=$PATH:/sbin:/usr/sbin:/usr/local/sbin
+            local root_command=${words[i]}
+            _command_offset $i
+            return
+        fi
+        [[ ${words[i]} == -@(!(-*)[uCLs]) ]] &&
+            ((i++))
+    done
+
+    case "$prev" in
+        -!(-*)u)
+            COMPREPLY=($(compgen -u -- "$cur"))
+            return
+            ;;
+        -!(-*)C)
+            _filedir
+            return
+            ;;
+        -!(-*)[Ls])
+            return
+            ;;
+    esac
+
+    $split && return
+
+    if [[ $cur == -* ]]; then
+        COMPREPLY=($(compgen -W '$(_parse_usage "$1")' -- "$cur"))
+        [[ ${COMPREPLY-} == *= ]] && compopt -o nospace
+        return
+    fi
+} &&
+    complete -F _comp_cmd_doas doas
+
+# ex: filetype=sh
--- a/completions-core/doas.bash
+++ b/completions-core/doas.bash
@@ -1,45 +1,45 @@
-# doas(1) completion                                       -*- shell-script -*-
+# doas(1) completion
 
 _comp_cmd_doas()
 {
-    local cur prev words cword split
-    _init_completion -s || return
+    local cur prev words cword was_split comp_args
+    _comp_initialize -s -- "$@" || return
+
+    local noargopts='!(-*|*[uCLs]*)'
 
     local i
-
     for ((i = 1; i <= cword; i++)); do
         if [[ ${words[i]} != -* ]]; then
             local PATH=$PATH:/sbin:/usr/sbin:/usr/local/sbin
             local root_command=${words[i]}
-            _command_offset $i
+            _comp_command_offset "$i"
             return
         fi
-        [[ ${words[i]} == -@(!(-*)[uCLs]) ]] &&
+        [[ ${words[i]} == -@(${noargopts}[uCLs]) ]] &&
             ((i++))
     done
 
-    case "$prev" in
-        -!(-*)u)
-            COMPREPLY=($(compgen -u -- "$cur"))
+    # shellcheck disable=SC2254 # glob $noargopts should not be quoted
+    case $prev in
+        -${noargopts}u)
+            _comp_compgen -- -u
             return
             ;;
-        -!(-*)C)
-            _filedir
+        -${noargopts}C)
+            _comp_compgen_filedir
             return
             ;;
-        -!(-*)[Ls])
+        -${noargopts}[Ls])
             return
             ;;
     esac
 
-    $split && return
+    [[ $was_split ]] && return
 
     if [[ $cur == -* ]]; then
-        COMPREPLY=($(compgen -W '$(_parse_usage "$1")' -- "$cur"))
+        _comp_compgen_usage
         [[ ${COMPREPLY-} == *= ]] && compopt -o nospace
         return
     fi
 } &&
     complete -F _comp_cmd_doas doas
-
-# ex: filetype=sh
--- a/completions-core/doas.bash
+++ b/completions-core/doas.bash
@@ -12,6 +12,7 @@ _comp_cmd_doas()
         if [[ ${words[i]} != -* ]]; then
             local PATH=$PATH:/sbin:/usr/sbin:/usr/local/sbin
             local root_command=${words[i]}
+            local _comp_root_command=$1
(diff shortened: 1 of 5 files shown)


completions/doas
```diff
@@ -0,0 +1,45 @@
...
+    local i
+
+    for ((i = 1; i <= cword; i++)); do
+        if [[ ${words[i]} != -* ]]; then
+            local PATH=$PATH:/sbin:/usr/sbin:/usr/local/sbin
+            local root_command=${words[i]}
+            _command_offset $i
+            return
+        fi
+        [[ ${words[i]} == -@(!(-*)[uCLs]) ]] &&
```
Tachi107, line 17: What's the difference between what you wrote and my suggestion below? They both seem to work, but my suggestion seems simpler.

```suggestion
        [[ ${words[i]} == -@([uCLs]) ]] &&
```

completions/doas
```diff
@@ -0,0 +1,45 @@
...
+            ((i++))
+    done
+
+    case "$prev" in
+        -!(-*)u)
```
Tachi107, line 22: Same as above; I simply use `-u` and `-C` in my personal doas completion file.

scop: That will fail to match the case where `u` is the last in a set of bundled short options, like `doas -nu <TAB>`. Similary above.

Tachi107:
Makes sense, thanks for the explanation!

completions/doas
```diff
@@ -0,0 +1,45 @@
...
+    local cur prev words cword split
+    _init_completion -s || return
+
+    local i
+
+    for ((i = 1; i <= cword; i++)); do
+        if [[ ${words[i]} != -* ]]; then
+            local PATH=$PATH:/sbin:/usr/sbin:/usr/local/sbin
+            local root_command=${words[i]}
+            _command_offset $i
```
akinomyoga: In a similar way to`_comp_root_command()` (defined in `bash_completion`), I think we want to enable the completion as if the current user has root privilege.

```suggestion
            local _comp_root_command=1
            _command_offset $i
```

ghost:
@Duncaen 

Please confirm that you have no intention in merging or supporting bash completions in opendoas. Otherwise bash-completions the project, refuses to merge the bash-completion profile for doas. It's stuck for 3 years now.

akinomyoga:
In case, the upstream issue is https://github.com/Duncaen/OpenDoas/issues/62. The upstream PR is https://github.com/Duncaen/OpenDoas/pull/84.

Tachi107:
The upstream issue and patch have been ignored for four years. I think that's enough evidence of lack of interest from upstream

scop:
I agree the upstream asks have been open for more than long enough. Removed the related label, this PR and review can continue now.

akinomyoga:
To be included in `bash_completion/completions-core`, we need refactorings to use API v3.

completions/doas
akinomyoga: ~~10cc5397f~~ ebb854fbf

akinomyoga:
I have rebased.

> To be included in `bash_completion/completions-core`, we need refactorings to use API v3.

22ebba4cc

completions-core/doas.bash
```diff
@@ -12,6 +12,7 @@ _comp_cmd_doas()
         if [[ ${words[i]} != -* ]]; then
             local PATH=$PATH:/sbin:/usr/sbin:/usr/local/sbin
             local root_command=${words[i]}
+            local _comp_root_command=1
```
scop: This seems susceptible to me, should it be `$1` instead of `1`?

And TBH as I don't remember it by heart, I'm having a bit of a problem tracking what setting this variable actually _does_ in this context. It seems to be only used in `_comp_as_root` which is not in the `_comp_compgen_offset` call path that I can tell, so it should not have any effect here?

Same in sudo.bash BTW.

akinomyoga: > This seems susceptible to me, should it be `$1` instead of `1`?

Yes, I modified it to `$1` ([`ebb854f..95239aa`](https://github.com/scop/bash-completion/compare/ebb854fbfd0dd0120a61908b08bc910aa02597cf..95239aa8a9de6755e1da82173fb0493b67c7572c)). Thank you for pointing this out.

Since this variable is currently only used for [the non-empty test](https://github.com/scop/bash-completion/blob/8baa60b9f806dc8a26d0fb10cfd83b1e85769a5e/bash_completion#L3194), the value can be practically any non-empty string. However, it should be set to `$1` for consistency with the other cases. Even if we were to set a fixed string, the value would be `set` instead of `1`, following the existing cases for the variables for the empty/non-empty test.

> And TBH as I don't remember it by heart, I'm having a bit of a problem tracking what setting this variable actually _does_ in this context. It seems to be only used in `_comp_as_root` which is not in the `_comp_compgen_offset` call path that I can tell, so it should not have any effect here?

The trick is that `_comp_compgen_offset` [extracts the name of the completion function](https://github.com/scop/bash-completion/blob/8baa60b9f806dc8a26d0fb10cfd83b1e85769a5e/bash_completion#L3071) for a command from [the output of `complete -p -- "$cmd"`](https://github.com/scop/bash-completion/blob/8baa60b9f806dc8a26d0fb10cfd83b1e85769a5e/bash_completion#L3037), and [call it dynamically](https://github.com/scop/bash-completion/blob/8baa60b9f806dc8a26d0fb10cfd83b1e85769a5e/bash_completion#L3073). This implies that any completion functions for arbitrary commands can be called through `_comp_command_offset`. The shell function `comp_as_root` is used by the `allowed_{users,groups}` generators, which are used by many completions:

```console
$ git grep -nE 'allowed_(users|groups)'
bash_completion:2330:            _comp_compgen -v tmp -c "${cur#*:}" allowed_groups
bash_completion:2345:            _comp_compgen -c "${cur#*:}" allowed_groups
bash_completion:2356:            _comp_compgen_allowed_users
completions-core/chgrp.bash:31:        _comp_compgen_allowed_groups
completions-core/crontab.bash:10:            _comp_compgen_allowed_users
completions-core/fprintd-delete.bash:8:    _comp_compgen_allowed_users
completions-core/fprintd-enroll.bash:33:        _comp_compgen_allowed_users
completions-core/killall.bash:21:            _comp_compgen_allowed_users
completions-core/passwd.bash:22:    _comp_compgen_allowed_users
completions-core/qemu.bash:92:            _comp_compgen_allowed_users
completions-core/strace.bash:85:                _comp_compgen_allowed_users
completions-core/tcpdump.bash:33:            _comp_compgen_allowed_users
completions-fallback/chsh.bash:37:        _comp_compgen_allowed_users
completions-fallback/newgrp.bash:14:        _comp_compgen_allowed_groups
completions-fallback/renice.bash:18:                _comp_compgen_allowed_users
startup-core/000_bash_completion_compat.bash:431:    _comp_compgen -c "${1:-$cur}" allowed_users
startup-core/000_bash_completion_compat.bash:437:    _comp_compgen -c "${1:-$cur}" allowed_groups
```

scop: Ah indeed, I arrived at those functions but stopped thinking there for some reason.

So for example, setting that makes `doas killall -u <TAB>` complete all users which is better behavior than completing only the current user.

Aside, neither the implementation here or the one in `_comp_cmd_sudo` takes the possible sudo/doas `-u` argument into account but blindly assumes root. That we can address in a followup PR if we like sometime, but that'd needs some work -- neither current user nor root allowances are the right thing to do is `-u` sets something else beside `root` or the current user.

scop (approved):
Thanks!

Merging with a commit message tweak -- we don't actually generate completions as root, but as _if we are_ root.

completions-core/doas.bash
scop: Maybe we could add a test case or few for the transitive allowed_users/groups sudo/doas behavior. Would go nicely with if we want to implement the -u awareness as in above though.

akinomyoga:
Thank you!

> Merging with a commit message tweak -- we don't actually generate completions as root, but as _if we are_ root.

Ah, right. Thanks.

completions-core/doas.bash
akinomyoga: Yeah, that can be improved. I think this can easily be forgotten, so I created an issue at #1656.

PR #773, opened 2022-07-10
fix(function): suppress error messages for non-existent functions

akinomyoga:
An error message is outputted on attempting completion after non-existent function names:

```bash
$ function aaa [TAB]bash: type: aaa: not found
() 
```

This PR suppresses the error message (c343c46e).

Also, f45fc770 removes a duplicate space after the text `()` inserted for non-existent functions. With the current implementation of `master`, two spaces are inserted.


scop:
Thanks!

PR #775, opened 2022-07-20
feat(_comp_looks_like_path): new utility, use it

scop:
Be more consistent with how we test this across the tree.

`apt-get`, `gdb`, `kldload`, `pydoc`, `pylint`, `removepkg`, `rpm`, `scp`, and `vpnc` gained better recognition of various kinds of paths as the result.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -390,6 +390,14 @@ _comp_expand_glob()
     return 0
 }
 
+# Check if the argument looks like a path.
+# @param $1 thing to check
+# @return True (0) if it does, False (> 0) otherwise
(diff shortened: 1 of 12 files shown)


scop:
https://github.com/scop/bash-completion/blob/66ec66e39620c505233e7350bfe38ee2eeea10dc/completions/ssh#L563
This one's not covered here; special treatment of the colon makes it a bit different. I can live with this exception to the rule, unless there's a nice and easy way to get it covered and use the new function, too.

scop:
Merging to get a fix for kldload style issue that's breaking the build.

Issue #506, opened 2021-03-08
Supressing output from completion script


kevinoid:
While tracking down a syntax error in a bash completion script, I noticed that [`stdout` and `stderr` are redirected to `/dev/null`](https://github.com/scop/bash-completion/blob/2.11/bash_completion#L2204) for all completion scripts.  It looks like this has been present since 20c05b43b, and probably for good reason, but it seems like a big hammer which may be papering over many errors (such as the one I eventually tracked down to a syntax error, due to a missing `;`).  Since I hadn't seen it discussed before, I'm opening this issue to discuss what problems it may be solving and whether it's the best approach.

Thanks for considering,
Kevin

scop:
That's a fair question. I think we should not mute everything, #508 has a candidate fix. Thoughts?

kevinoid:
Looks great!  Thanks for considering it and for working up a fix.  I appreciate it, and I think it'll be a positive change.

It wouldn't surprise me if some third-party bash-completion scripts inadvertently produce output.  When I get a chance, I'll see if I can work up a reasonable test to see how many bash-completion scripts in the Debian archive produce output.  That might give us an idea of how much pain the change may cause and whether any triage is required first.

From a quick look through the archive, I found 3 packages which put their completion script in a subdirectory of `/usr/share/bash-completion/completions/`:
* [`crystal-lang`](https://packages.debian.org/sid/amd64/crystal-lang/filelist) - `/usr/share/bash-completion/completions/crystal-lang/completion.bash`
* [`libgadap-dev`](https://packages.debian.org/sid/amd64/libgadap-dev/filelist) - `/usr/share/bash-completion/completions/bash_completion.d/gadap-config`
* [`trace-cmd`](https://packages.debian.org/sid/amd64/trace-cmd/filelist) - `/usr/share/bash-completion/completions/trace-cmd/trace-cmd.bash`

I assume all of these are buggy and will investigate and file bugs shortly.

kevinoid:
Yep, they were all packaging bugs:

* [#984955](https://bugs.debian.org/984955) for `crystal-lang`.
* [#984957](https://bugs.debian.org/984957) for `libgadap-dev`.
* [#984959](https://bugs.debian.org/984959) for `trace-cmd`.

I included a trivial patch for each.  Hopefully they will be fixed quickly.

kevinoid:
[...]

PR #776, opened 2022-07-22
fix(__load_completion): do not warn when completing `.` first time

scop:
ca361be02a92e30a3d26d376b30faf268853e836 enables warnings about trying to source directories when loading completions. That's useful, but triggers the errors in the common case of trying to source something, `. `<kbd>Tab</kbd>

Special case `.` (and for completeness, even though not that interesting, `..`) so we don't issue the warning about them.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2560,7 +2560,9 @@ __load_completion()
             compfile="$dir/$compfile"
             # Avoid trying to source dirs; https://bugzilla.redhat.com/903540
             if [[ -d $compfile ]]; then
-                echo "bash_completion: $compfile: is a directory" >&2
+                # Do not warn with . or .. (especially the former is common)
+                [[ $compfile == */.?(.) ]] ||
+                    echo "bash_completion: $compfile: is a directory" >&2
             elif [[ -e $compfile ]] && . "$compfile"; then
                 [[ $backslash ]] && $(complete -p "$cmd") "\\$cmd"
                 return 0


akinomyoga:
I have a naive question. Do we need to output warnings here? Maybe we can just ignore the directories because the warning message disturbs the cursor positions in the terminal.

scop:
https://github.com/scop/bash-completion/issues/506 has the discussion related to this.

I'm not that concerned about messing with the output; this is a broken setup that should be fixed and not masked, consistent with what happens with a completion file with invalid syntax, or an unreadable one.

akinomyoga:
Thank you for your explanation. Then, that's fine to keep the error message.

Maybe we can also explicitly mention in the error message that putting subdirectories in `completions` directories is the wrong setup?

bash_completion
```diff
@@ -2560,7 +2560,9 @@ __load_completion()
             compfile="$dir/$compfile"
             # Avoid trying to source dirs; https://bugzilla.redhat.com/903540
             if [[ -d $compfile ]]; then
-                echo "bash_completion: $compfile: is a directory" >&2
+                # Do not warn with . or .. (especially the former is common)
+                [[ $compfile == */.?(.) ]] ||
+                    echo "bash_completion: $compfile: is a directory" >&2
```
akinomyoga, line 2565: This is just a comment so if you think it's unneeded, please merge it as is.

For example,

```suggestion
                if [[ $compfile != */.?(.) ]]; then
                    echo "bash_completion: $compfile: is a directory" >&2
                    echo "bash_completion: $dir: completionsdir should not contain subdirectories" >&2
                fi
```

akinomyoga (approved):

scop:
As mentioned in the commit message, the current warning mimics what bash itself would emit when sourcing a directory. If we keep it that way, there will be minimal behavioral changes ahead for users when the day comes we can drop our own `-d` check (probably when we start to require bash >= 4.3 -- would be good to find out the exact version in a comment here and document it, https://bugzilla.redhat.com/show_bug.cgi?id=903833)

akinomyoga:
Thank you! I see the background! Please merge it as is.

scop:
Done, along with 16ce83f9fc6780c80bf1016be0384a25f04c0d87 documenting the bash version needing the check (verified locally, didn't bother to check 4.2.x patch versions).

PR #777, opened 2022-07-23
feat(fio): update --crctest, --debug, and --terse-version arg completion

scop:
Syncs with fio 3.30.

### DIFF
--- a/completions/fio
+++ b/completions/fio
@@ -20,7 +20,7 @@ _fio()
         --debug)
             _comp_delimited , -W "process file io mem blktrace verify
                 random parse diskutil job mutex profile time net rate compress
-                steadystate helperthread"
+                steadystate helperthread zbd"
             return
             ;;
         --output-format)
@@ -28,7 +28,11 @@ _fio()
             return
             ;;
         --terse-version)
-            COMPREPLY=($(compgen -W "2 3" -- "$cur"))
+            COMPREPLY=($(compgen -W "2 3 4" -- "$cur"))
+            return
+            ;;
+        --crctest)
+            COMPREPLY=($(compgen -W '$("$1" --crctest=nonexistent 2>/dev/null)' -- "$cur"))
             return
             ;;
         --cmdhelp)
--- a/completions/fio
+++ b/completions/fio
@@ -18,9 +18,9 @@ _fio()
             return
             ;;
         --debug)
-            _comp_delimited , -W "process file io mem blktrace verify
-                random parse diskutil job mutex profile time net rate compress
-                steadystate helperthread zbd"
+            local opts=$("$1" --debug=help 2>/dev/null)
+            opts=${opts##*:}
+            _comp_delimited , -W "${opts//,/ }"
             return
             ;;
         --output-format)
--- a/completions/fio
+++ b/completions/fio
@@ -28,7 +28,7 @@ _fio()
             return
             ;;
         --terse-version)
-            COMPREPLY=($(compgen -W "2 3 4" -- "$cur"))
+            COMPREPLY=($(compgen -W "2 3 4 5" -- "$cur"))
             return
             ;;
         --crctest)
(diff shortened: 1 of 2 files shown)


completions/fio
```diff
@@ -12,15 +12,19 @@ _fio()
         --debug)
             _comp_delimited , -W "process file io mem blktrace verify
                 random parse diskutil job mutex profile time net rate compress
-                steadystate helperthread"
+                steadystate helperthread zbd"
```
akinomyoga, line 15: The list seems to be dynamically available by `--debug help`, though I haven't checked which version of `fio` started to support it.

```bash
$ fio --debug help
fio: dumping debug options:process,file,io,mem,blktrace,verify,random,parse,diskutil,job,mutex,profile,time,net,rate,compress,steadystate,helperthread,zbd,all
$ fio --version
fio-3.29
```

scop: Excellent! At least the one in Ubuntu 14 and current Fedora development do, so I think we're good. Done in ce5f74512f952e0f6add9facf023095c85428dd6.

completions/fio
```diff
@@ -10,17 +10,21 @@ _fio()
...
+            _comp_delimited , -W "${opts//,/ }"
             return
             ;;
         --output-format)
             COMPREPLY=($(compgen -W "terse json json+ normal" -- "$cur"))
             return
             ;;
         --terse-version)
-            COMPREPLY=($(compgen -W "2 3" -- "$cur"))
+            COMPREPLY=($(compgen -W "2 3 4" -- "$cur"))
```
akinomyoga, line 23: It seems the latest version supports also `5`. Somehow `5` is not included in `fio --help` with fio 3.29 / Fedore 36, but the man page mentions it. In fact, `fio --terse-version=5` seems to correctly parse while `fio --terse-version=6` fails in the argument-parse phase.

akinomyoga (commented):
Maybe `--showcmd=jobfile` can also be supported?

scop:
> Maybe `--showcmd=jobfile` can also be supported?

I'm unable to find a version that would allow an argument to `--showcmd`; the one on Ubuntu 14, Debian 10, and current Fedora dev does not. Where did you come across it?

akinomyoga:
> > Maybe `--showcmd=jobfile` can also be supported?
> 
> I'm unable to find a version that would allow an argument to `--showcmd`; the one on Ubuntu 14, Debian 10, and current Fedora dev does not. Where did you come across it?

I find it on the man page of fio 3.29 in Fedora 36 though again `fio --help` doesn't seem to include it. I'm not familiar with `fio` so it might be a typo on the man page?

akinomyoga:
In addition to [the man page](https://github.com/axboe/fio/blob/d33c7846cc5f175177e194a5489282780e2a04c4/fio.1#L70), a document [HOWTO.rst](https://github.com/axboe/fio/blob/d33c7846cc5f175177e194a5489282780e2a04c4/HOWTO.rst) mentions `--showcmd=jobfile`, but [the actual source code](https://github.com/axboe/fio/blob/3ce6a3de08240b87a8c1b2a8b0d22a81b51f714f/init.c#L150-L154) doesn't seem to support the option argument. The `jobfile` seems supposed to be given as an independent command line argument. In fact, when I specify an option argument to the option, I get the following error message:

```bash
$ fio --showcmd=file
fio: option '--showcmd' doesn't allow an argument
fio: unrecognized option '--showcmd=file'
Did you mean showcmd?
```

I think this is a typo of the man page and HOWTO.rst.

scop:
Cool, fixes to the issues you found sent upstream: https://github.com/axboe/fio/pull/1460, https://github.com/axboe/fio/pull/1462

akinomyoga:
Thank you!

completions/fio
scop: `5` added in 25a84598a3c8ea7a6ea0b76205a6bbbe1f905ef0

akinomyoga (approved):

Issue #420, opened 2020-06-06
man xyz stays xy if directory with same name present


jidanni:
```
$ man apti<TAB>
```
correctly becomes
```
$ man aptitude
aptitude aptitude-create-state-bundle aptitude-curses aptitude-run-state-bundle

```

However
```
$ mkdir /tmp/aptitude
$ cd /tmp/
$ man apti<TAB>
aptitude/ aptitude/ aptitude-create-state-bundle aptitude-curses aptitude-run-state-bundle

```

1. Does not at least complete to "aptitude", but instead stays "apti".
2. Gives *two* "aptitude/"

scop:
Reproduced, interesting.

maxnikulin:
A real world example: Ubuntu switched from deb to snap for browser packages (firefox, chromium), so users almost inevitably have `~/snap/` folder. A command to manage these packages is named `snap` as well. If CWD is `~` then hitting `TAB` while typing `man snap` causes manifestation of this issue.

scop:
I don't have good ideas offhand how to fix this.

Getting rid of the duplicate would be fairly easy, but it'd get rid of the command, leaving only the dir with the trailing slash present. I think this is not what people would generally expect.

We could opt to not completing filenames if we've found some completions already. PR #787 implements that, at the expense of not doing filename completions at all if a matching man page from manpath has already been found. I'm not entirely happy with that either, because it will hurt also the vast majority of use cases where there are no problematic duplicates present. 
Then again, the use case mostly affected by this would be the second manpage argument to man and subsequent ones, which is a quite rare one.

maxnikulin:
A workaround to force filename completion is `M-/`, so `./` prefix is not necessary. However the problem of combining filename and non-filename completions is more general, see e.g. `scp` where colon without space should be added to host names while files and directories should be handled accordingly to regular rules. Unfortunately the `scp` code is not really robust.

PR #787, opened 2022-08-20
fix(man): avoid duplicated man page and dir completions

scop:
If a directory with the same name as an already found man page candidate from manpath is present, they both end up in completions as a duplicate with the trailing slash included.

To work around, skip file based completion altogether when a man page from manpath is among found candidates. A workaround for the workaround, i.e. "forcing" file based completion, is to prefix the argument with a `./`.

Closes https://github.com/scop/bash-completion/issues/420

### DIFF
--- a/completions/man
+++ b/completions/man
@@ -99,12 +99,12 @@ _man()
     fi
 
     # shellcheck disable=SC2053
-    if [[ $prev != $mansect ]]; then
-        # File based completion for the rest, prepending ./ if needed
+    if [[ $prev != $mansect && -z ${COMPREPLY-} ]]; then
+        # File based completion, prepending ./ if needed
         # (man 1.6f needs that for man pages in current dir)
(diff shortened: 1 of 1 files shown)


scop:
Draft: I'm not sure I'm happy with this. Let's discuss in #420.

completions/man
```diff
@@ -99,12 +99,12 @@ _man()
     fi
 
     # shellcheck disable=SC2053
-    if [[ $prev != $mansect ]]; then
-        # File based completion for the rest, prepending ./ if needed
+    if [[ $prev != $mansect && -z ${COMPREPLY-} ]]; then
```
scop, line 102: We could instead loop through the completions found before `start` below and check and remove duplicates from file based ones, but that'd be very slow when there are lots of man page completions from man path (e.g. `man <TAB>` gives 9000+ on the system I'm looking at).

PR #792, opened 2022-08-27
chore(.gitignore): add more autotool files

akinomyoga:
The filenames `compfile`, `depcomp`, `ltmain.sh`, and `config.{guess,sub}` are added in `.gitignore`. Also, I collected the autotool-related entries in a section.


scop (changes requested):
I've nothing against this in principle, but I'd like to understand which versions or scenarios cause the files `compile`, `depcomp`, `config.guess`, `config.sub`, and `ltmain.sh` to appear. They do not for me in Ubuntu 20.04 or Fedora dev, `autoreconf -i; ./configure; make`.

akinomyoga:
Ah, OK. I do somehow have these files but couldn't reproduce them in a clean `bash-completion` repository (with autotools from Fedora 36), but now I have identified why I have these files. To investigate the historical behavior of `cryptsetup --help` & `man cryptsetup` for #759, I cloned a repository of [cryptsetup](https://gitlab.com/cryptsetup/cryptsetup) in a subdirectory and tried to build it. It somehow seems to create the related files at an upper level. In short, I could reproduce them by the following steps.

```bash
bash-completion$ git clone git@gitlab.com:cryptsetup/cryptsetup.git
bash-completion$ cd cryptsetup
cryptsetup$ ./autogen.sh
```

Running `autoreconf -i` in the `cryptsetup` directory instead of `autogen.sh` also rewrites the files of the parent directory. I did not know `autoreconf -i` could rewrite the files outside the directory.

akinomyoga:
OK, I think there is no reason to add these files created by `autoreconf -i` of external projects in `.gitignore`. I just dropped these files. I keep this PR open as it still has some changes though it is now reduced to just a cosmetic update.

scop (approved):

scop:
Thanks!

PR #793, opened 2022-09-01
fix(_comp_command_offset): Add support for complete -C

### LINKED ISSUE #631: Use `sudo` with `complete -C` commands
## Describe the bug

When a command is completed with `complete -C`, sudo complete is not working.

## To reproduce

```
$ complete -C 'echo' foo
$ sudo foo <tab>
```

Gives warning:

```
-bash: compgen: warning: -C option may not work as you expect
```

## Expected behavior

No warning message thrown, and `sudo foo` completes the same way as `foo`.

- [linux] Operating system name/distribution and version:
- [5.1] bash version, `echo "$BASH_VERSION"`:
- [2.11] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`:

## Additional context

`man bash` says about `compgen`:

>   When using the -F or -C options, the various shell variables set by the programmable completion facilities, while available, will not have useful values.

`_command_offset` knows about `-F` but not `-C`:

```
...
if [[ ${cspec#* -F } != "$cspec" ]]; then
...
```

elyscape:
Fixes #631.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2764,28 +2764,71 @@ _comp_command_offset()
                     #     *something* for every command thrown at it ($cspec != empty)
                     _comp_complete_minimal "${args[@]}"
                 fi
-            elif [[ $cspec == *' -F '* ]]; then
-                # complete -F <function>
+            elif [[ $cspec == *\ -[CF]\ * ]]; then
+                if [[ $cspec == *' -F '* ]]; then
+                    # complete -F <function>
 
-                # get function name
-                local func=${cspec#* -F }
-                func=${func%% *}
-                $func "${args[@]}"
+                    # get function name
+                    local func=${cspec#* -F }
+                    func=${func%% *}
+                    $func "${args[@]}"
 
-                # restart completion (once) if function exited with 124
-                if (($? == 124 && retry_count++ == 0)); then
-                    # Note: When the completion function returns 124, the state
-                    # of COMPREPLY is discarded.
+                    # restart completion (once) if function exited with 124
+                    if (($? == 124 && retry_count++ == 0)); then
+                        # Note: When the completion function returns 124, the
+                        # state of COMPREPLY is discarded.
+                        COMPREPLY=()
+
+                        cspec=$(complete -p "$compcmd" 2>/dev/null)
+
+                        # Note: When completion spec is removed after 124, we
+                        # do not generate any completions including the default
+                        # ones. This is the behavior of the original Bash
+                        # progcomp.
+                        [[ $cspec ]] || break
+
+                        continue
+                    fi
+                else
+                    # complete -C <command>
+
+                    # get command name
+                    local completer=${cspec#* -C \'}
+
+                    # completer commands are always single-quoted
+                    if ! _comp_dequote "'$completer"; then
+                        _minimal "${args[@]}"
+                        break
+                    fi
+                    completer=${REPLY[0]}
+
+                    local -a suggestions
+
+                    local IFS=$' \t\n'
+                    local reset_monitor=$(shopt -po monitor) reset_lastpipe=$(shopt -p lastpipe) reset_noglob=$(shopt -po noglob)
+                    set +o monitor
+                    shopt -s lastpipe
+                    set -o noglob
+
+                    COMP_KEY="$COMP_KEY" COMP_LINE="$COMP_LINE" \
+                        COMP_POINT="$COMP_POINT" COMP_TYPE="$COMP_TYPE" \
+                        $completer "${args[@]}" | mapfile -t suggestions
+
+                    $reset_monitor
+                    $reset_lastpipe
+                    $reset_noglob
+                    _comp_unlocal IFS
+
+                    local suggestion
+                    local i=0
                     COMPREPLY=()
+                    for suggestion in "${suggestions[@]}"; do
+                        COMPREPLY[i]+=${COMPREPLY[i]+$'\n'}$suggestion
 
-                    cspec=$(complete -p "$compcmd" 2>/dev/null)
-
-                    # Note: When completion spec is removed after 124, we do
-                    # not generate any completions including the default ones.
-                    # This is the behavior of the original Bash progcomp.
-                    [[ $cspec ]] || break
-
-                    continue
+                        if [[ $suggestion != *\\ ]]; then
+                            ((i++))
+                        fi
+                    done
                 fi
 
                 # restore initial compopts
--- a/test/fixtures/_command_offset/completer
+++ b/test/fixtures/_command_offset/completer
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+case "${2-}" in
+    b|ba|bar)
+        echo bar
+        ;;
+    cont1*)
+        echo cont10
+        echo cont11\\
+        ;;
+    f|fo|foo)
+        echo foo
+        ;;
+    l)
+        echo line\\
+        echo two
+        echo long
+        ;;
+    li*)
+        echo line\\
+        echo two
+        ;;
+    lo*)
+        echo long
+        ;;
+    *)
+        echo bar
+        echo foo
+        ;;
+esac
--- a/test/t/unit/test_unit_command_offset.py
+++ b/test/t/unit/test_unit_command_offset.py
@@ -12,7 +12,8 @@ def join(words):
 
 @pytest.mark.bashcomp(
     cmd=None,
-    ignore_env=r"^[+-]COMPREPLY=",
+    cwd="_command_offset",
+    ignore_env=r"^[+-](COMPREPLY|REPLY)=",
 )
 class TestUnitCommandOffset:
     wordlist = sorted(["foo", "bar"])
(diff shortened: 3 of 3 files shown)


test/t/unit/test_unit_command_offset.py
```diff
@@ -51,6 +56,7 @@ def test_1(self, bash, functions):
             ("cmd3", wordlist),
             ("cmd4", []),
             ("cmd5", ["0"]),
+            ("cmd6", wordlist),
```
akinomyoga, line 59: Could you separate the test from `test_2`? As in the doc comment, `test_2` is the test template for the completion retry requested by the exit status 124.

Also, the test seems to fail in my environment currently. I wondered how this could get passed [CI test #601](https://github.com/scop/bash-completion/actions/runs/2968137118), but I noticed that actually the test `test_unit_command_offset.py` isn't run in the CI test because it is not included in the [EXTRA_DIST](https://github.com/scop/bash-completion/blob/40aa4dc3c75791f43f850871466039768b3048ad/test/t/Makefile.am#L3)...

akinomyoga (changes requested):
LGTM except for the following minor points.

Also, I think we can add test cases for the line continuation by the trailing `\`.

bash_completion
```diff
@@ -2274,33 +2274,85 @@ _command_offset()
...
+                        # strip trailing newlines manually
+                        suggestion="${suggestion%$'\n'}"
+
+                        COMPREPLY[i]="${COMPREPLY[i]-}${suggestion%\\}"
+
+                        if [[ $suggestion == *\\ ]]; then
+                            COMPREPLY[i]+=$'\n'
+                        else
+                            ((i++))
+                        fi
```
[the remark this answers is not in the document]
elyscape: Done.

test/t/unit/test_unit_command_offset.py
elyscape: Separated, and tests enhanced to actually work.

elyscape:
I wasn't able to figure out a way to test for completion after the newline is inserted. That being said, it does seem to function as expected, and I expect that it doesn't actually come up that often in practice.

test/t/unit/test_unit_command_offset.py
```diff
@@ -30,9 +31,12 @@ def functions(self, bash):
             'complete -W %s "${COMP_WORDS[0]}"' % quote(join(self.wordlist)),
             'COMPREPLY=(dummy); complete -r "${COMP_WORDS[0]}"',
             "COMPREPLY+=(${#COMPREPLY[@]})",
+            'complete -C ./completer "${COMP_WORDS[0]}"',
         ]
 
         for idx, comp in enumerate(completions, 2):
+            if comp is None:
+                continue
```
akinomyoga, line 39: This is now unused.

test/t/unit/test_unit_command_offset.py
```diff
@@ -30,9 +31,12 @@ def functions(self, bash):
             'complete -W %s "${COMP_WORDS[0]}"' % quote(join(self.wordlist)),
             'COMPREPLY=(dummy); complete -r "${COMP_WORDS[0]}"',
             "COMPREPLY+=(${#COMPREPLY[@]})",
+            'complete -C ./completer "${COMP_WORDS[0]}"',
```
akinomyoga, line 34: This array is also for 124. I think the following line should be added at the end of `functions(self,bash)` instead:

```bash
        assert_bash_exec(bash, "complete -C ./completer cmd6")
```

test/fixtures/_command_offset/completer
```diff
@@ -0,0 +1,26 @@
...
+        echo two
+        echo long
+        ;;
+    li*)
+        echo line\\
+        echo two
+        ;;
+    lo*)
+        echo long
+        ;;
```
akinomyoga, line 21: I'd like to add the corner case where the last line is terminated by `\`. What do you think?

```suggestion
        ;;
    cont1*)
        echo cont10
        echo cont11\\
        ;;
```

test/t/unit/test_unit_command_offset.py
```diff
@@ -72,3 +76,17 @@ def test_2(self, bash, functions, cmd, expected_completion):
...
         """
         assert assert_complete(bash, "meta %s " % cmd) == expected_completion
+
+    @pytest.mark.parametrize(
+        "cmd,expected_completion",
+        [
+            ("cmd6 ", wordlist),
+            ("cmd6 l", [r"line\^Jtwo", "long"]),
+            ("cmd6 lo", ["ng"]),
+            ("cmd6 line", [r"\^Jtwo"]),
```
akinomyoga, line 86:
```suggestion
            ("cmd6 line", [r"\^Jtwo"]),
            ("cmd6 cont1", ["cont10", "cont11\\"]),
```

akinomyoga (changes requested):
Thank you for updating! These are the remaining points.

test/fixtures/_command_offset/completer
eliyoung-ethos: Assuming that's how bash handles it, sure.

eliyoung-ethos:
Changes applied.

test/fixtures/_command_offset/completer
elyscape: Assuming that's how bash handles it, sure.

elyscape:
Requested changes applied! Let me know if there's anything else I should change.

test/t/unit/test_unit_command_offset.py
akinomyoga: Thank you!

You have put the above line in `test_3` instead of in `functions`. I'm interested in if there is a reason that this line is put directly in `test_3`. If we put it in `test_3`, it will be executed every time for each `(cmd,expected_result)` pair. But I think this common setup can be executed just once, which is one of the reasons that one defines completion settings with `@pytest.fixture`.

elyscape: Whoops! I got confused and put it in the wrong place. Fixing now.

akinomyoga:
Thank you! Now I think everything is OK. Let's wait for @scop's check and decision!

akinomyoga (approved):

elyscape:
I went ahead and rebased onto master to pick up the changes in #794.

elyscape:
Rebased onto master to handle changes from #797.

elyscape:
Just checking in case this fell through the cracks.

bash_completion
```diff
@@ -2307,33 +2307,83 @@ _comp_command_offset()
...
-                    continue
+                    $reset_monitor
+                    $reset_lastpipe
+
+                    local suggestion
+                    local i=0
+                    COMPREPLY=()
+                    for suggestion in "${suggestions[@]}"; do
+                        # `mapfile -t` was added in bash 4.4, so we have to
+                        # strip trailing newlines manually
```
scop, line 2378: Are we sure about this? We've used `mapfile -t` in the `cvs` completion since 2011 (e35eabb791733996078aae9161fc01ab5f94a9e3), I think 5 years before bash 4.4 was released.

bash_completion
```diff
@@ -2307,33 +2307,83 @@ _comp_command_offset()
...
+                    # completer commands are always single-quoted
+                    if ! _comp_dequote "'$completer"; then
+                        _minimal
+                        break
+                    fi
+                    completer=${ret[0]}
+
+                    local -a suggestions
+
+                    local reset_monitor=$(shopt -po monitor) reset_lastpipe=$(shopt -p lastpipe)
```
scop, line 2355: Do we need protection against unusual `$IFS` here (at least for eventually invoking the resets)? Cf. `git grep -F ' reset'`.

scop (changes requested):
Sorry for sitting on this for so long. Added a few comments + looks like we need another rebase.

elyscape:
I'll take a look this weekend.

bash_completion
elyscape: The `NEWS` file indicates that [it was added in 4.4][news-4.4], but doing some digging, that's it looks like that notice was misleading. Looking at the source code, it looks like the flag did exist starting with bash 4.0, but originally [only removed trailing newlines][mapfile-4.0]; bash 4.4 added the `-d` flag and updated `-t` to remove the specified delimiter when both flags are used. I'll update this accordingly.

[news-4.4]: https://github.com/bminor/bash/blob/bash-4.4/NEWS#L17-L19
[mapfile-4.0]: https://github.com/bminor/bash/blame/bash-4.0/builtins/mapfile.def#L37

bash_completion
elyscape: Good point. Will fix.

elyscape:
Rebased again and addressed comments. We should be good to go now.

elyscape:
@scop Just checking in on this.

akinomyoga:
rebased.

akinomyoga (approved):

elyscape:
@scop Pinging on this.

scop (approved):
LGTM. Thanks for your patience, this took way too long, my bad.

Issue #602, opened 2021-09-03
Add test suite runs with failglob/nullglob set


scop:
## Describe the feature/solution

Would be good to add `failglob` and `nullglob` enabled setups to the test matrix we run, as shown by #600 and #601.

## Maintenance (please complete the following information)

- [ ] This is a request for new completion
- [ ] Link to upstream project query about shipping the completion:

PR #794, opened 2022-09-01
test(Makefile.am): update EXTRA_DIST in Makefile.am

akinomyoga:
I noticed that the tests in #793 and #791 are actually not run in CI. I found that also other files are not included in `Makefile.am`.

### DIFF
--- a/test/t/Makefile.am
+++ b/test/t/Makefile.am
@@ -21,6 +21,7 @@ EXTRA_DIST = \
 	test_apt_build.py \
 	test_apt_cache.py \
 	test_apt_get.py \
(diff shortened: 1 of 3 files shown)


akinomyoga:
I have fixed an issue of the test code found by [CI #602](https://github.com/scop/bash-completion/runs/8126714799?check_suite_focus=true). The tempenv for the shell functions (i.e., `tempenv=value func`) persists after the function call in the POSIX mode in Bash 4.2..5.0. Instead of using temporary environments, I changed to use local variables.

scop:
Nice catch, thanks!

Issue #790, opened 2022-08-23
bash command completion has a problem with sudo command


mug896:
## Describe the bug

When the command completion function starts running
there are three variables automatically setting `$1`, `$2`, `$3`.
and `$2` is not always the same as `${COMP_WORDS[COMP_CWORD]}`.

1. if `$2` value is a member of `$COMP_WORDBREAKS` variable
   then it's value is empty by default

   ex)  command --foo=[tab]
           `$2` : (empty)     `${COMP_WORDS[COMP_CWORD]}` : `=`

2. if `$2` value start with `"` or `'` (single or double quote)
   then it's quote removed ( `"foo` --> `foo` )

   ex)  command --foo "bar[tab]
           `$2` : `bar`         `${COMP_WORDS[COMP_CWORD]}` : `"bar`

this is an intended work for making completion function simpler
but if i use sudo in front of a command then
`$2` value is always the same as `${COMP_WORDS[COMP_CWORD]}`
This is a different behavior to the bash default and problematic.
and ends up can not use `$2` variable and force to use
`${COMP_WORDS[COMP_CWORD]}`  to support sudo

## Versions 

- [ ] Operating system name/distribution and version:
         Operating System: Ubuntu 22.04.1 LTS              
          Kernel: Linux 5.15.0-43-generic
         Architecture: x86-64
- [ ] bash version: 5.1.16(1)-release
- [ ] bash-completion version: 2.11

akinomyoga:
The call of the completion functions for the commands after another command (such as `sudo`) is emulated by [`_command_offset`](https://github.com/scop/bash-completion/blob/40aa4dc3c75791f43f850871466039768b3048ad/bash_completion#L2214-L2323). In particular, the arguments passed to the completion function is determined [here](https://github.com/scop/bash-completion/blob/40aa4dc3c75791f43f850871466039768b3048ad/bash_completion#L2284-L2288).

> * if `$2` value is a member of `$COMP_WORDBREAKS` variable
>   then it's value is empty by default
>   ex)  command --foo=[tab]
>   `$2` : (empty)     `${COMP_WORDS[COMP_CWORD]}` : `=`
> * if `$2` value start with `"` or `'` (single or double quote)
>   then it's quote removed ( `"foo` --> `foo` )
>   ex)  command --foo "bar[tab]
>   `$2` : `bar`         `${COMP_WORDS[COMP_CWORD]}` : `"bar`

[...]

PR #797, opened 2022-09-03
refactor(_command_offset): refactor the adjustment of `COMP_WORDS`

akinomyoga:
Extracted from ~~#790~~ #791. This is also actually independent.

> - d2c0b4ee **simplify logic**: The original code tries to manually perform a string manipulation that can be achieved by a single parameter expansion. Also, it tries to manually perform the array compaction that can be achieved by a single parameter expansion. The behaviors of the new codes slightly differ from the original ones, but I believe the behaviors of the new codes are better:
>   - In the original code, `COMP_LINE` becomes empty when `${COMP_WORDS[i]}` is not contained in `$COMP_LINE`, though I don't think that happens as far as these variables are modified by external codes. In the new code, such a word `${COMP_WORDS[i]}` is just ignored.
>   - In the original code, the words in `COMP_WORDS` after `COMP_CWORD` is not modified, but just the word at `COMP_CWORD` and the words preceding it are shifted. For example, `COMP_WORDS=([0]=sudo [1]=A [2]=B [3]=C [4]=D)` with `COMP_CWORD=2` becomes `COMP_WORDS=([0]=A [1]=B [3]=C [4]=D)` instead of `COMP_WORDS=([0]=A [1]=B [2]=C [3]=D)`. In the new code, all the words are shifted.
> - 995db594 **localize changes to `COMP_*`**: The original implementation of `_comp_offset` breaks the state of `COMP_*` after the  completion of `_comp_offset`. Currently, it doesn't cause a problem because all the completion functions in bash-completion that use `_comp_offset` seem to return from the function immediately after the call of `_comp_offset`. But I still think we should localize the changes to `COMP_*` within `_command_offset` and functions called from it.
>   - This change might possibly break existing external completion functions if there are completion functions that use the modified values of `COMP_*` after calling `_comp_offset`.
>   - The test seems to ignore also other variables `cur`, `words`, and `cword`, and they do not seem to be modified by the current implementation of `_command_offset`.

The description is the same as was in #791.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2220,6 +2220,13 @@ _command_offset()
     # rewrite current completion context before invoking
     # actual command completion
 
+    # make changes to COMP_* local.  Note that bash-4.3..5.0 have a
+    # bug that `local -a arr=("${arr[@]}")` fails.  We instead first
+    # assign the values of `COMP_WORDS` to another array `comp_words`.
+    local COMP_LINE=$COMP_LINE COMP_POINT=$COMP_POINT COMP_CWORD=$COMP_CWORD
+    local -a comp_words=("${COMP_WORDS[@]}")
+    local -a COMP_WORDS=("${comp_words[@]}")
+
     # find new first word position, then
     # rewrite COMP_LINE and adjust COMP_POINT
     local word_offset=$1 i tail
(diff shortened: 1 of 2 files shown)


scop:
Thanks!

PR #801, opened 2022-09-04
fix(_command_offset): support quoting of the command name in `complete -p` (bash 5.2)

akinomyoga:
Bash 5.2 starts to quote the command name in `complete -p` if necessary.


scop (commented):
Would it be possible/feasible to amend the test suite with a case that fails without this change?

akinomyoga:
Thank you for the suggestion! I have added test cases.

scop:
Good stuff, thanks!

PR #802, opened 2022-09-04
fix(_command_offset): fix offset mismatching for words vs COMP_WORDS (use COMP_WORDS)

### LINKED ISSUE #722: Completing sudo after --user= doesn't suggest commands
## Describe the bug

When trying to complete `sudo --user=foobar <HIT TAB HERE>`, it suggests the files in the current directory instead of commands

## To reproduce

1. Type `sudo --user=<A USER ON YOUR SYSTEM>`
2. Try to complete
3. See problem

## Expected behavior

Trying to complete `sudo --user=foo` should suggest command names, not files in the current directory

- [ ] Operating system name/distribution and version: Fedora 35 Silverblue
- [ ] bash version, `echo "$BASH_VERSION"`: 5.1.8(1)-release
- [ ] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.11

## Additional context

I discovered this when trying to [write a completion](https://gitlab.com/carbonOS/build-meta/-/blob/a0ff8032a40baeded34c73da0a87dde3c31f03db/files/polkit/pkexec.completion) for `pkexec`. There's another bug in sudo's completion that mishandles `sudo --user foobar`, so this context doesn't apply to sudo. My completion doesn't have this bug, so I can get this additional behavior: `pkexec --user=foobar <TAB>` completes the directory, but `pkexec --user foobar <TAB>` completes commands. I can confirm that, at least with my completion, `_command_offset` is being executed. So I suspect soething in _command_offset can't deal with `--arg=data` arguments?

## Debug trace

[bash-completion.log](https://github.com/scop/bash-completion/files/8387366/bash-completion.log)

akinomyoga:
Fixes #722

This PR tries to fix it so that `_command_offset` receives the offset in `COMP_WORDS` but not the offset in `words`. However, another option is to make `command_offset` receive the offset in `words` and convert it to the one in `COMP_WORDS` inside `_command_offset`. I have also made the patch for that option (acf6411c).

This PR also includes a fix (addf7233) and an improvement (6755e497) on codes calculating the offset.

### DIFF
--- a/completions/_ionice
+++ b/completions/_ionice
@@ -9,8 +9,8 @@ _comp_cmd_ionice()
     _comp_initialize -- "$@" || return
 
     local offset=0 i
-    for ((i = 1; i <= cword; i++)); do
-        case ${words[i]} in
+    for ((i = 1; i <= COMP_CWORD; i++)); do
+        case ${COMP_WORDS[i]} in
             -h)
                 return
                 ;;
@@ -26,6 +26,7 @@ _comp_cmd_ionice()
                 continue
                 ;;
         esac
+        [[ ${COMP_WORDS[i - 1]} == = ]] && continue
         offset=$i
         break
     done
--- a/completions/find
+++ b/completions/find
@@ -8,9 +8,9 @@ _comp_cmd_find()
     _comp_initialize -- "$@" || return
 
     local i
-    for i in ${!words[*]}; do
-        if [[ ${words[i]} == -@(exec|ok)?(dir) ]]; then
-            ((cword > i)) || break
+    for i in ${!COMP_WORDS[*]}; do
+        if [[ ${COMP_WORDS[i]} == -@(exec|ok)?(dir) ]]; then
+            ((COMP_CWORD > i)) || break
             _comp_command_offset $((i + 1))
             return
         fi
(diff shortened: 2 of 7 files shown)


scop:
In principle, I think I like the approach in acf6411c43a50842ecf10eeeb0274c57ea192524 more, because it removes a special case implementers need to remember regarding using `COMP_WORDS` vs `words`. However, converting the indexes isn't that "clean" and I think I need to stare more at the code in acf6411c43a50842ecf10eeeb0274c57ea192524 to understand how it even works :) Is that commit part of some other PR we have open? Which approach do you prefer?

akinomyoga:
> Is that commit part of some other PR we have open? 

No, it's not. I'll submit it as a separate PR.

> Which approach do you prefer?

This is a difficult question. As you say, for the consistency of the interface, we should accept an index in the array `words`, but the approach doesn't seem to be so clean.

I think what I cared about when I submitted this PR was the compatibility for the external completions. This function `_command_offset` has been accepted the index in `COMP_WORDS`, so I didn't want to change the behavior. However, now we switched from the name `_command_offset` to `_comp_command_offset`, so probably we do not have to care about it.

akinomyoga:
I made this to be a draft PR in favor of the other approach #860.

akinomyoga:
Now the problem is fixed by #860, so let me close this PR.

PR #803, opened 2022-09-04
feat(_comp_split): add a function to split a string into an array

akinomyoga:
I have three branches depending on this new utility (the branch of #696, and unsubmitted [akinomyoga/failglob-1](https://github.com/scop/bash-completion/compare/master...akinomyoga:failglob-1?expand=1) and [akinomyoga/failglob-3](https://github.com/scop/bash-completion/compare/master...akinomyoga:failglob-3?expand=1). I would like to have this utility merged first to avoid complicated dependency handling.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -390,6 +390,56 @@ _comp_expand_glob()
     return 0
 }
 
+# Split a string and assign to an array.  This function basically performs
+# `IFS=<sep>; <array_name>=(<text>)` but properly handles saving/restoring the
+# state of `IFS` and the shell option `noglob`.  A naive splitting by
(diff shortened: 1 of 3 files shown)


bash_completion
```diff
@@ -390,6 +390,52 @@ _comp_expand_glob()
     return 0
 }
 
+# Split a string and assign to an array.
```
scop, line 393: Is the rationale for this function mostly to avoid repeated manual management of `noglob` and `$IFS`, or is there something else to it? Perhaps add a few words of reasoning here why/when this should be used.

akinomyoga: Thank you. I have updated the comment.

PR #805, opened 2022-09-09
feat(fio): more option and arg completions

scop:
The majority of options are actually not available from `--help`, but rather `--cmdhelp=all`.

Support some option arguments "natively", some per `--cmdhelp=COMMAND` output, and engines from `--enghelp` output.

### DIFF
--- a/completions/fio
+++ b/completions/fio
@@ -31,13 +31,13 @@ _fio()
             return
             ;;
         --cmdhelp)
-            ret=($("$1" --cmdhelp=all 2>/dev/null | awk '{print $1}'))
-            COMPREPLY=($(compgen -W '${ret[@]} all' -- "$cur"))
+            ret=($("$1" --cmdhelp=all 2>/dev/null | awk '{print $1}') all)
+            COMPREPLY=($(compgen -W '"${ret[@]}"' -- "$cur"))
             return
             ;;
         --enghelp)
-            _comp_cmd_fio__engines "$1"
-            COMPREPLY=($(compgen -W '${ret[@]}' -- "$cur"))
+            _comp_cmd_fio__engines "$1" &&
+                COMPREPLY=($(compgen -W '"${ret[@]}"' -- "$cur"))
             return
             ;;
         --eta)
@@ -98,7 +98,7 @@ _fio()
             local IFS=$'\n'
             local cmdhelp=($("$1" --cmdhelp=${prev#--} 2>/dev/null))
             _comp_unlocal IFS
-            case ${cmdhelp[*]} in
+            case ${cmdhelp[*]-} in
                 *"showing closest match"*)
                     # ignore
                     ;;
@@ -111,6 +111,7 @@ _fio()
                     #        valid values: 1024       [...]
                     #                    : 1000       [...]
                     local line="" in_values=false
+                    ret=()
                     for line in "${cmdhelp[@]}"; do
                         if $in_values; then
                             if [[ $line =~ ^[[:space:]]*:[[:space:]]*([^[:space:]]+) ]]; then
@@ -123,7 +124,7 @@ _fio()
                             ret+=("${BASH_REMATCH[1]}")
                         fi
                     done
-                    COMPREPLY=($(compgen -W '${ret[@]}' -- "$cur"))
+                    COMPREPLY=($(compgen -W '"${ret[@]}"' -- "$cur"))
                     return
                     ;;
             esac
--- a/completions/fio
+++ b/completions/fio
@@ -94,6 +94,41 @@ _fio()
             _gids
             return
             ;;
+        --?*)
+            local IFS=$'\n'
+            local cmdhelp=($("$1" --cmdhelp=${prev#--} 2>/dev/null))
+            _comp_unlocal IFS
+            case ${cmdhelp[*]} in
+                *"showing closest match"*)
+                    # ignore
+                    ;;
+                *" type: boolean "* | *" type: empty or boolean "*)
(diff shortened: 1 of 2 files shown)


completions/fio
```diff
@@ -24,13 +31,13 @@ _fio()
             return
             ;;
         --cmdhelp)
-            # TODO more commands?
-            COMPREPLY=($(compgen -W "all" -- "$cur"))
+            ret=($("$1" --cmdhelp=all 2>/dev/null | awk '{print $1}'))
+            COMPREPLY=($(compgen -W '${ret[@]} all' -- "$cur"))
```
akinomyoga, line 35: - Avoid the `nounset` error in bash 4.2 and 4.3 for `${ret[@]}` in case the command fails.
- Avoid redundant word splitting by quoting `"..."`(because the necessary word splitting should have been performed in the previous line).

```suggestion
            ret=($("$1" --cmdhelp=all 2>/dev/null | awk '{print $1}') all)
            COMPREPLY=($(compgen -W '"${ret[@]}"' -- "$cur"))
```

completions/fio
```diff
@@ -70,12 +77,70 @@ _fio()
...
+            ;;
+        --gid)
+            _gids
+            return
+            ;;
+        --?*)
+            local IFS=$'\n'
+            local cmdhelp=($("$1" --cmdhelp=${prev#--} 2>/dev/null))
+            _comp_unlocal IFS
+            case ${cmdhelp[*]} in
```
akinomyoga, line 101: nounset in bash-4.2, 4.3.

```suggestion
            case ${cmdhelp[*]-} in
```

completions/fio
```diff
@@ -70,12 +77,70 @@ _fio()
...
+                    ;;
+                *" type: boolean "* | *" type: empty or boolean "*)
+                    COMPREPLY=($(compgen -W '0 1' -- "$cur"))
+                    return
+                    ;;
+                *" valid values:"*)
+                    # For example, for --kb_base=:
+                    #        valid values: 1024       [...]
+                    #                    : 1000       [...]
+                    local line="" in_values=false
```
akinomyoga, line 113: - work around localvar_inherit

```suggestion
                    local line="" in_values=false
                    ret=()
```

completions/fio
```diff
@@ -70,12 +77,70 @@ _fio()
...
+                                ret+=("${BASH_REMATCH[1]}")
+                            else
+                                break
+                            fi
+                        elif [[ $line =~ ^[[:space:]]*valid\ values:[[:space:]]*([^[:space:]]+) ]]; then
+                            in_values=true
+                            ret+=("${BASH_REMATCH[1]}")
+                        fi
+                    done
+                    COMPREPLY=($(compgen -W '${ret[@]}' -- "$cur"))
```
akinomyoga, line 126:
```suggestion
                    COMPREPLY=($(compgen -W '"${ret[@]}"' -- "$cur"))
```

It is not necessary in the present case because the elements of `ret` do not contain spaces, but that is what we would know only after we carefully read the code.

completions/fio
```diff
@@ -24,13 +31,13 @@ _fio()
...
-            COMPREPLY=($(compgen -W "all" -- "$cur"))
+            ret=($("$1" --cmdhelp=all 2>/dev/null | awk '{print $1}'))
+            COMPREPLY=($(compgen -W '${ret[@]} all' -- "$cur"))
             return
             ;;
         --enghelp)
-            # TODO print ioengine help, or list available ioengines
-            # TODO engine,help arg
+            _comp_cmd_fio__engines "$1"
+            COMPREPLY=($(compgen -W '${ret[@]}' -- "$cur"))
```
akinomyoga, line 40:
```suggestion
            _comp_cmd_fio__engines "$1"
            ((${#ret[@]})) && COMPREPLY=($(compgen -W '"${ret[@]}"' -- "$cur"))
```

- Avoid redundant word splitting. Actually, I don't think we need to suppress word splitting in this particular case because the engine names are unlikely to include spaces, but I don't want to rely on this kind of an implicit assumption about the values of the variables.
- Avoid the `nounset` error in Bash 4.2 and 4.3 in case it fails. Another approach might be to let `_comp_cmd_fio__engines` return through its exit status whether it generated at least one engine name and
  ```suggestion
              _comp_cmd_fio__engines "$1" &&
                  COMPREPLY=($(compgen -W '"${ret[@]}"' -- "$cur"))
  ```

completions/fio
```diff
@@ -70,12 +77,70 @@ _fio()
             _filedir -d
             return
             ;;
+        --ioengine)
+            _comp_cmd_fio__engines "$1"
+            COMPREPLY=($(compgen -W '${ret[@]}' -- "$cur"))
```
akinomyoga, line 82: The same as the previous comment.

akinomyoga (commented):
I have small suggestions.

completions/fio
scop: Ah, clever way to avoid that nounset ret :+1:

completions/fio
scop: Nice, I'll go with the latter approach for slightly less repeated code.

completions/fio
scop: Cool, I guess this will fix the CentOS 7 / Ubuntu 14 test failures I hadn't found time to track down.

scop:
Very nice, thanks. Applied all suggestions, will squash the eventual merge to clean up the history.

scop:
Only gitlint fails so with the squash intent, I believe we are good to go here.

akinomyoga:
Thanks!

akinomyoga (approved):

PR #808, opened 2022-09-10
feat(help): support options, improve quoting

scop:
The simple `-A helptopic` completion for does not support completion of options, and does not do a too good job with items containing parens, such as `(( ... ))` and `foo ((`.

### DIFF
--- a/completions/help
+++ b/completions/help
@@ -1,0 +1,17 @@
+# bash completion for help                                 -*- shell-script -*-
+
+_comp_cmd_help()
+{
+    local cur prev words cword
+    _init_completion || return
+
+    if [[ $cur == -* ]]; then
+        COMPREPLY=($(compgen -W '$(_parse_usage "$1" "-s $1")' -- "$cur"))
+        return
+    fi
+
+    local IFS=$'\t\n'
+    COMPREPLY=($(compgen -A helptopic -- "$cur"))
+    ((${#COMPREPLY[*]} != 1)) || printf -v "COMPREPLY[0]" %q "$COMPREPLY"
+}
+complete -F _comp_cmd_help help
--- a/completions/help
+++ b/completions/help
@@ -13,5 +13,5 @@ _comp_cmd_help()
     local IFS=$'\t\n'
     COMPREPLY=($(compgen -A helptopic -- "$cur"))
     ((${#COMPREPLY[*]} != 1)) || printf -v "COMPREPLY[0]" %q "$COMPREPLY"
-}
-complete -F _comp_cmd_help help
+} &&
+    complete -F _comp_cmd_help help
--- a/test/t/test_help.py
+++ b/test/t/test_help.py
@@ -1,0 +1,18 @@
+import pytest
+
+
+class TestHelp:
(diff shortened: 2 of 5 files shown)


completions/help
```diff
@@ -0,0 +1,17 @@
...
+    if [[ $cur == -* ]]; then
+        COMPREPLY=($(compgen -W '$(_parse_usage "$1" "-s $1")' -- "$cur"))
+        return
+    fi
+
+    local IFS=$'\t\n'
+    COMPREPLY=($(compgen -A helptopic -- "$cur"))
+    ((${#COMPREPLY[*]} != 1)) || printf -v "COMPREPLY[0]" %q "$COMPREPLY"
+}
+complete -F _comp_cmd_help help
```
akinomyoga, line 17: Maybe this is unrelated to this PR, but I have been wondering for a long time what would be the background of the idiom used in bash-completion, `_cmd() { ... } && complete -F _cmd cmd`. In this new `help` completion, this idiom is not used, so I am interested in its background if any. I'd like to learn it for my future contributions.

I'm not sure if it is related, but as far as I know, the function definition fails when extended glob is used without `shopt -s extglob` or there is an existing shell function of the same name and marked as readonly by `declare -fr _cmd`. Are there other cases of failing function definitions?

akinomyoga (approved):

test/t/test_help.py
```diff
@@ -0,0 +1,18 @@
...
+
+    @pytest.mark.complete("help -")
+    def test_options(self, completion):
+        assert completion
+
+    @pytest.mark.complete(r"help \(")
+    def test_parens(self, completion):
+        # Assumption: an item like "(( ... ))" exists in the output
+        assert any(
+            x.startswith(r"\(") and x.endswith(r"\)\)") for x in completion
```
akinomyoga, line 17: bash-4.2 doesn't seem to produce `(( ... ))` for `compgen -A helptopic`:

```bash
$ bash-4.3 -c 'compgen -A helptopic' | grep '('
(( ... ))
for ((
$ bash-4.2 -c 'compgen -A helptopic' | grep '('
$
```

completions/help
scop: That's an old practice, ISTR it's related to the function definition failing but the name ending up being used as the completion function anyway, causing `bash: completion: function [...] not found` when invoked. I don't remember the exact case that provoked sprinkling the `&&`s everywhere, so for now this is mostly cargo culted as far as I'm concerned. I wouldn't mind dropping that practice if we don't see it being useful any longer.

But in this particular case, leaving it out is an oversight, will fix.

akinomyoga: OK! Thank you for your explanation! As far as it is consistent over the codebase, I don't mind if it's *mostly cargo-culted*.

scop: b9db334eded61c1538337b8b8f35919faaa35cc4

test/t/test_help.py
scop: 7b21a8f167e6ab28c2c4d399909cf9ce5155eec9

akinomyoga (approved):

PR #814, opened 2022-09-11
refactor(_command_offset): save Bash's $2 passed to completion functions

### LINKED ISSUE #790: bash command completion has a problem with sudo command

akinomyoga:
Fixes #790

This is a fix alternative to #791. Instead of imitating the arguments passed by Bash, we may save the original arguments in `_init_completion` (renamed to `_comp_initialize`). However, this involves not only the change of `_init_completion` itself but also the interface change of functions that call `_init_completion` internally---such as `_filedir_xspec`, `_longopt`, `_minimal`, `_known_hosts`, etc. Also, this doesn't work when `_comp_command_offset` is used without calling `_init_completion` (some cases are found in external completions). We need to request external completion writers carefully update their code.


bash_completion
```diff
@@ -1010,6 +1024,8 @@ _init_completion()
                 ;;
         esac
     done
+    shift "$((OPTIND - 1))"
+    (($#)) && comp_args=("$@")
```
scop, line 1028: Would it make sense to do this only if `comp_args` is defined? Otherwise I suppose completions not prepared for the new behavior will leak a global by that name unless I'm missing something?

```suggestion
    if [[ -v comp_args ]]; then
        shift "$((OPTIND - 1))"
        (($#)) && comp_args=("$@")
    fi
```

bash_completion
```diff
@@ -987,13 +987,27 @@ _comp_variable_assignments()
...
+#                     The first argument $1 is command name.  The second
+#                     argument $2 is the string before the cursor in the
+#                     current word.  The third argument $3 is the previous
+#                     word.
+# @var[out] cur       Reconstructed current word
+# @var[out] prev      Reconstructed previous word
+# @var[out] words     Reconstructed words
+# @var[out] cword     Current word index in `words`
+# @var[out] comp_args Original arguments specified to the completion function
+#                     are saved.
```
scop, line 1000:
```suggestion
#                     are saved in this array, if the variable is defined.
```

scop:
Even though this PR is huge and there is an interface change involved, I feel the code here is much easier to understand than in #791; from that point of view I have a better feeling about this one.

If I understand correctly, the API change is such that it will mostly degrade gracefully, i.e. if the original arguments are not available, we'll basically do what we used to do before this change? Or will this flat out break some scenarios if not adjusted by affected completions?

External completion authors will likely be using `_init_completion` for a long time still. Luckily the current implementation (before this change) should be fine if it receives positional arguments, so I suppose externals can switch basically to
```bash
    local cur prev words cword split comp_args
    _init_completion -- "$@" || return
```
...in order to support earlier bash completion versions and to take advantage of the new stuff.

bash_completion
```diff
@@ -987,13 +987,27 @@ _comp_variable_assignments()
...
+# @param $1...$3 args Original arguments specified to the completion function.
+#                     The first argument $1 is command name.  The second
+#                     argument $2 is the string before the cursor in the
+#                     current word.  The third argument $3 is the previous
+#                     word.
+# @var[out] cur       Reconstructed current word
+# @var[out] prev      Reconstructed previous word
+# @var[out] words     Reconstructed words
+# @var[out] cword     Current word index in `words`
+# @var[out] comp_args Original arguments specified to the completion function
```
scop, line 999: I have mixed feelings about this name: it's nice and concise, but it somehow slightly annoys me that the name does not make it crystal clear that these are the _original_ completion arguments. Dunno what to do about it if anything, not a huge fan of `comp_orig_args` (or `orig_args` or the same with `original` instead of `orig`)...

akinomyoga:
> Even though this PR is huge and there is an interface change involved, I feel the code here is much easier to understand than in #791; from that point of view I have a better feeling about this one.

OK! Then, let us consider this direction!

> If I understand correctly, the API change is such that it will mostly degrade gracefully, i.e. if the original arguments are not available, we'll basically do what we used to do before this change?

Yes, it is basically right. Actually, it slightly improves the original behavior. Originally, we passed the entire word `${COMP_WORDS[COMP_CWORD]}` as the second argument, but now we are passing the part of the word after the current cursor position, i.e., `$cur`, as Bash does for its built-in programmable completion.

> External completion authors will likely be using `_init_completion` for a long time still. Luckily the current implementation (before this change) should be fine if it receives positional arguments, so I suppose externals can switch basically to
> 
> ```shell
>     local cur prev words cword split comp_args
>     _init_completion -- "$@" || return
> ```
> 
> ...in order to support earlier bash completion versions and to take advantage of the new stuff.

Ah, this is a good point. I hadn't noticed that point by myself. Thank you.

bash_completion
akinomyoga: `local ... comp_arg` doesn't make the variable *set state*. For `[[ -v comp_args ]]` to success, an actual value should be assigned to `comp_arg` prior to the test. So,

1. one way is to request writing like `local ... comp_arg=`.
2. Another way is to test the existence of the variable by `declare comp_args &>/dev/null`.

But I'm not sure if we need to test it explicitly because I think we can assume that `comp_args` is declared when extra arguments are specified to `_comp_initialize` or `_init_completion`. This is tested by `(($#)) &&`.

bash_completion
akinomyoga: It was actually the same for me. I also thought of different names such as `original_args`, `comp_original_args`, `old_args`, `args`, etc., but some seemed too long, and the other seemed too general. I thought `prev` and `words` are already general names so initially thought of using `args`, but it turned out `args` are already widely used by completion functions for other purposes. If we wouldn't use `comp_args`, my second candidate is `original_args`.

scop: Ok, let's go with `comp_args`.

scop (approved):
Ok, let's go with this. Thanks!

PR #810, opened 2022-09-10
feat(alias): add option completion support

scop:

PR #815, opened 2022-09-11
test(conftest): merge `assert_complete_at_point` to `assert_complete`

akinomyoga:
`assert_complete_at_point` does not properly clean up the command line. This may cause the failure of subsequent tests. The same cleanups as `assert_complete` should be performed for `assert_complete_at_point`. Instead of repeating the cleanup code in `assert_complete_at_point`, I have merged the special treatment of `assert_complete_at_point` in `assert_complete` as suggested in a `TODO` in the code comment. The *unintended* test failures in #810 and #811 are fixed by this PR.


scop:
Nice!

Issue #807, opened 2022-09-09
apt-get not completing package names on Ubuntu 22.04


leoarnold:
Hi there,

just installed Kubuntu 22.04.1 Desktop and noticed that

```
$ sudo apt-get insta<TAB>
```

would autocomplete, but

```
$ sudo apt-get install thunderb<TAB>
```

would not.

Reporting it here as a reference. May already be fixed in current development state.

## Describe the feature/solution

I looked into the code of [the version packaged with Ubuntu 18.04](https://packages.ubuntu.com/bionic/bash-completion) and found the line

```
COMPREPLY=( $( apt-cache --no-generate pkgnames "$cur" \
```

So I ran

```
$ apt-cache --no-generate pkgnames thunderb
E: The package cache file is corrupted, it has the wrong hash
```

which was weird because `sudo apt-get update` passed without error.

The [solution](https://ostechnix.com/fix-e-the-package-cache-file-is-corrupted-it-has-the-wrong-hash-error-in-ubuntu/) was to purge the cache and refresh:

```
$ sudo rm -rf /var/lib/apt/lists/*
$ sudo apt-get update
```

and tab completion worked again.

Would be nice if that error message was **not silently swallowed** by the completion though.

## Maintenance (please complete the following information)

- [ ] This is a request for new completion
- [ ] Link to upstream project query about shipping the completion:

scop:
Suppressing error output from underlying invoked commands is a "cheap" general practice we tend to follow, because letting it through may cause an interleaved error/completion mess, the errors may make no sense to users, and/or presence of various commands or support for args we invoke them with might not be essential for the completion to work (graceful degradation).

I'm not against improving on this practice, but we should establish a way to do it so it's clear what and where is suppressed intentionally, and what's just an oversight. Maybe this means just some documentation and comments on the matter. Maybe it would be useful to make suppression optional, e.g. skip it in a "debug mode". Anyway it will require careful examination of each case separately.

For example, related to this particular case: for better or for worse, we still support the version of `apt-cache` that is present on Ubuntu 14.04. It doesn't understand the `--no-generate` option, so if we just let the error output through, it'll be mostly just an annoyance to users. Such a version of `apt-cache` isn't probably that relevant these days as far as new releases of bash-completion go, and our support for it is mostly a side effect of us running tests on Ubuntu 14.04 to cover oldish bash versions, but to illustrate.

PR #818, opened 2022-09-12
fix(_known_hosts_real): exclude directories from the config files

### LINKED ISSUE #817: SCP "Is a Directory"
## Describe the bug

When using completion on scp -r something/ tabtab got this error
```
++ command sed -ne 's/^[[:blank:]]*[Hh][Oo][Ss][Tt][[:blank:]]\(.*\)$/\1/p' /etc/ssh/ssh_config /home/xxxx/.ssh/config /home/xxxx/.ssh/xxxx.ssh-config/xxxx/config /home/xxxx/.ssh/xxxx.ssh-config/xxxx-xxxx/config /home/xxxx/.ssh/xxxx.ssh-config/team-xxxx/config /home/xxxx/.ssh/xxxx.ssh-config/team-xxxx/config /home/xxxx/.ssh/xxxx.ssh-config/team-xxxx/config /home/bdupuis/.ssh/xxxx.ssh-config/team-xxxx/config /home/xxxx/.ssh/local/config /home/xxxx/.ssh/local
sed: read error on /home/xxxx/.ssh/local: Is a directory
```

## To reproduce

1. Create configuration folders in .ssh
2. scp -r something_in_your_home_folder tabtab # note that scp  -r ~/something_in_your_home doesn't bug

## Expected behavior

No error message

```
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy
```
- [ ] bash version, `echo "$BASH_VERSION"`: 5.1.16(1)-release
- [ ] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.11

## Debug trace

```
++ command sed -ne 's/^[[:blank:]]*[Hh][Oo][Ss][Tt][[:blank:]]\(.*\)$/\1/p' /etc/ssh/ssh_config /home/xxxx/.ssh/config /home/xxxx/.ssh/xxxx.ssh-config/xxxx/config /home/xxxx/.ssh/xxxx.ssh-config/xxxx-xxxx/config /home/xxxx/.ssh/xxxx.ssh-config/team-xxxx/config /home/xxxx/.ssh/xxxx.ssh-config/team-xxxx/config /home/xxxx/.ssh/xxxx.ssh-config/team-xxxx/config /home/bdupuis/.ssh/xxxx.ssh-config/team-xxxx/config /home/xxxx/.ssh/local/config /home/xxxx/.ssh/local
sed: read error on /home/xxxx/.ssh/local: Is a directory
```

akinomyoga:
Fixes #817

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1905,7 +1905,7 @@ _included_ssh_config_files()
         _comp_expand_glob files '$i'
         if ((${#files[@]})); then
             for f in "${files[@]}"; do
-                if [[ -r $f ]]; then
+                if [[ -r $f && ! -d $f ]]; then
                     config+=("$f")
                     # The Included file is processed to look for Included files in itself
                     _included_ssh_config_files $f
@@ -1973,10 +1973,10 @@ _known_hosts_real()
 
     # ssh config files
     if [[ -v configfile ]]; then
-        [[ -r $configfile ]] && config+=("$configfile")
+        [[ -r $configfile && ! -d $configfile ]] && config+=("$configfile")
     else
         for i in /etc/ssh/ssh_config ~/.ssh/config ~/.ssh2/config; do
-            [[ -r $i ]] && config+=("$i")
+            [[ -r $i && ! -d $i ]] && config+=("$i")
         done
     fi
 
@@ -2023,7 +2023,7 @@ _known_hosts_real()
         for i in /etc/ssh/ssh_known_hosts /etc/ssh/ssh_known_hosts2 \
             /etc/known_hosts /etc/known_hosts2 ~/.ssh/known_hosts \
             ~/.ssh/known_hosts2; do
-            [[ -r $i ]] && kh+=("$i")
+            [[ -r $i && ! -d $i ]] && kh+=("$i")
         done
         for i in /etc/ssh2/knownhosts ~/.ssh2/hostkeys; do
             [[ -d $i ]] && khd+=("$i"/*pub)
--- a/test/fixtures/_known_hosts_real/config_include
+++ b/test/fixtures/_known_hosts_real/config_include
@@ -5,3 +5,5 @@ Include ~/config_full_path
 Include config_relative_path
 # Include with wildcards, and more than one on same row
 Include config_asterisk* config_?uestion_mark
+# Include a directory name.  This is a misconfiguration, but ssh ignores it without errors.
+Include ../../_known_hosts_real


bash_completion
```diff
@@ -2023,7 +2023,7 @@ _known_hosts_real()
         for i in /etc/ssh/ssh_known_hosts /etc/ssh/ssh_known_hosts2 \
             /etc/known_hosts /etc/known_hosts2 ~/.ssh/known_hosts \
             ~/.ssh/known_hosts2; do
-            [[ -r $i ]] && kh+=("$i")
+            [[ -r $i && ! -r $i ]] && kh+=("$i")
```
scop, line 2026: I suppose

```suggestion
            [[ -r $i && ! -d $i ]] && kh+=("$i")
```

scop (changes requested):
Let's add some dirs to the mix in the test suite.

bash_completion
akinomyoga: Ah, thank you for pointing out!

akinomyoga:
> Let's add some dirs to the mix in the test suite.

Thanks! I have added a `Include` directive in `config_include` for the test of ssh_config inclusions.

However, directories at e.g. `~/.ssh/config` and `Include <directory>` seem to be just misconfigurations. I have checked the man page `man ssh_config`, but `Include` directive does not seem to accept a directory name. I have tried it in my actual ssh configuration, but `Include <directory>` seems to be just ignored. In this sense, maybe we shouldn't fix it but leave the error messages, or we can output a more friendly message that explains the issue. But then, a question is why we have been ignoring unreadable files by testing `[[ -r $file ]]`. If we take the `Include` directive as "include-if-exists", we can consider a case an unrelated directory is created at the same path, which should be ignored.

scop:
Parts of the discussion about error messages at https://github.com/scop/bash-completion/issues/807#issuecomment-1242687669 applies here too. Silently suppressing isn't nice, but neither is making a mess on the console, producing errors that make no sense to users, or coding and maintaining precise assumptions about undocumented details of external tool behavior into bash-completion. Whether silently suppressing is the best option or not, at least we are quite consistent with it currently. I'm open to discussing the general approach, but that's probably better done elsewhere.

akinomyoga:
Thanks!

> Parts of the discussion about error messages at [#807 (comment)](https://github.com/scop/bash-completion/issues/807#issuecomment-1242687669) applies here too.

I also remember the discussion at https://github.com/scop/bash-completion/pull/776#issuecomment-1221233985.

> Whether silently suppressing is the best option or not, at least we are quite consistent with it currently. I'm open to discussing the general approach, but that's probably better done elsewhere.

OK! Thank you!

PR #819, opened 2022-09-13
feat(firefox): Add svg file completion

kasbah:
Firefox is a very good SVG viewer and I use it for that purpose a lot. 

(Thanks for all the completions!)


scop:
Thanks!

PR #821, opened 2022-09-15
dpkg: Replace grep|grep|awk with single invocation of awk

jkaivo:
Instead of piping one grep to another, then piping that to awk, simply make one call to awk with appropriate matching. This makes the completion slightly more efficient (only one process is spawned instead of three), and removes a dependency on GNU grep's non-standard -A and -B options.

### DIFF
--- a/completions/dpkg
+++ b/completions/dpkg
@@ -13,12 +13,18 @@ if _comp_have_command grep-status; then
 else
     _comp_xfunc_dpkg_installed_packages()
     {
-	command awk -F'\n' '-vRS=' "\$1 ~ /Package: $1/ && (\$2 ~ /ok installed|half-installed|unpacked|half-configured/ || \$2 ~ /^Essential: yes/) {print(substr(\$1, 10)); }" /var/lib/dpkg/status 2>/dev/null
+	command awk -F '\n' -v RS='' "
+		\$1 ~ /Package: $1/ &&
+		\$2 ~ /ok installed|half-installed|unpacked|half-configured|^Essential: yes/
+		{ print(substr(\$1, 10)); }" /var/lib/dpkg/status 2>/dev/null
     }
 
     _comp_xfunc_dpkg_purgeable_packages()
     {
-	command awk awk -F'\n' '-vRS=' "\$1 ~ /Package: $1/ && (\$2 ~ /ok installed|half-installed|unpacked|half-configured|config-files/ || \$2 ~ /^Essential: yes/) {print(substr(\$1, 10)); }" /var/lib/dpkg/status 2>/dev/null
+	command awk -F '\n' -v RS='' "
+		\$1 ~ /Package: $1/ &&
+		\$2 ~ /ok installed|half-installed|unpacked|half-configured|config-files|^Essential: yes/
+		{ print(substr(\$1, 10)); }" /var/lib/dpkg/status 2>/dev/null
     }
 fi
 
--- a/completions/dpkg
+++ b/completions/dpkg
@@ -13,18 +13,20 @@ if _comp_have_command grep-status; then
 else
     _comp_xfunc_dpkg_installed_packages()
     {
-	command awk -F '\n' -v RS='' "
-		\$1 ~ /Package: $1/ &&
-		\$2 ~ /ok installed|half-installed|unpacked|half-configured|^Essential: yes/
-		{ print(substr(\$1, 10)); }" /var/lib/dpkg/status 2>/dev/null
+        command awk -F '\n' -v RS='' "
+            \$1 ~ /Package: $1/ &&
+            \$2 ~ /ok installed|half-installed|unpacked|half-configured|^Essential: yes/ {
+                print(substr(\$1, 10));
+            }" /var/lib/dpkg/status 2>/dev/null
     }
 
     _comp_xfunc_dpkg_purgeable_packages()
     {
-	command awk -F '\n' -v RS='' "
-		\$1 ~ /Package: $1/ &&
-		\$2 ~ /ok installed|half-installed|unpacked|half-configured|config-files|^Essential: yes/
-		{ print(substr(\$1, 10)); }" /var/lib/dpkg/status 2>/dev/null
+        command awk -F'\n' -vRS='' "
+            \$1 ~ /Package: $1/ &&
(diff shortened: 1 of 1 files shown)


completions/dpkg
```diff
@@ -13,20 +13,12 @@ if _comp_have_command grep-status; then
 else
     _comp_xfunc_dpkg_installed_packages()
     {
-        command grep -A 1 "Package: ${1-}" /var/lib/dpkg/status 2>/dev/null |
-            command grep -B 1 -Ee "ok installed|half-installed|unpacked| \
-            half-configured" \
-                -Ee "^Essential: yes" |
-            awk "/Package: ${1-}/ { print \$2 }" 2>/dev/null
+	command awk -F'\n' '-vRS=\n\n' "\$1 ~ /Package: $1/ && (\$2 ~ /ok installed|half-installed|unpacked|half-configured/ || \$2 ~ /^Essential: yes/) {print(substr(\$1, 10)); }" /var/lib/dpkg/status 2>/dev/null
```
akinomyoga, line 16: Thank you for your contribution. I generally agree with the idea of making it a single invocation of an external command, but `RS` with more than one character is unspecified by POSIX.

akinomyoga (changes requested):

jkaivo:
Fair point. It works with GNU Awk and Mawk, but now I see that it doesn't work with original-awk (or, apparently GNU Awk with POSIXLY_CORRECT set).

jkaivo:
Modified to use a null string for RS, which POSIX specifies as having the desired behavior: "If RS is null, then records are separated by sequences consisting of a <newline> plus one or more blank lines, leading or trailing blank lines shall not result in empty records at the beginning or end of the input, and a <newline> shall always be a field separator, no matter what the value of FS is."

completions/dpkg
```diff
@@ -13,20 +13,12 @@ if _comp_have_command grep-status; then
 else
     _comp_xfunc_dpkg_installed_packages()
     {
-        command grep -A 1 "Package: ${1-}" /var/lib/dpkg/status 2>/dev/null |
-            command grep -B 1 -Ee "ok installed|half-installed|unpacked| \
-            half-configured" \
-                -Ee "^Essential: yes" |
-            awk "/Package: ${1-}/ { print \$2 }" 2>/dev/null
+	command awk -F'\n' '-vRS=' "\$1 ~ /Package: $1/ && (\$2 ~ /ok installed|half-installed|unpacked|half-configured/ || \$2 ~ /^Essential: yes/) {print(substr(\$1, 10)); }" /var/lib/dpkg/status 2>/dev/null
```
akinomyoga, line 16: Thank you!

- Can you fold the long line like this?

```suggestion
	command awk -F'\n' '-vRS=' "
	    \$1 ~ /Package: $1/ && \
	        (\$2 ~ /ok installed|half-installed|unpacked|half-configured/ || \$2 ~ /^Essential: yes/) \
	        { print(substr(\$1, 10)); }" /var/lib/dpkg/status 2>/dev/null
```

- Can you combine `(\$2 ~ /re1/ || \$2 ~ /re2/)` into like `\$2 ~ /re1|re2/`?
- Now I think the single quotation of `'-vRS='` is unneeded (or `-vRS=''` would read better).
- Also, I think POSIX requires a space between the mandatory option-argument as `-v RS=` (cf [XBD 12.1 & 12.2 Guideline 6](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html)),

akinomyoga (changes requested):

completions/dpkg
jkaivo: Done, although I only went to one level of indentation deeper than the start of the command.

akinomyoga: Thanks!

akinomyoga: > although I only went to one level of indentation deeper than the start of the command.

The [lint check](https://github.com/scop/bash-completion/actions/runs/3063164431/jobs/4945021363) fails.

We use four spaces for one level of indentation and do not use a tab to replace 8 spaces. Another lint error is about the commit messages not following [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/). We also restrict the header line of the commit message to 70 characters.

jkaivo: OK, I'm cancelling this PR and creating a new one from a separate branch that follows the Conventional Commits style.

akinomyoga: Ah, OK. That also works, but the PR hasn't needed to be recreated. The normal procedure for this case is to rebase and force-push to the PR branch.

```bash
$ git rebase -i upstream/master
```

Then a commit list is opened with a text editor.

```
pick 4d7f499a replace grep|grep|awk with single invocation of awk
pick ad25d28a Replace multicharacter RS with null string RS
pick 7d24d2bd adjust formatting and merge $2 patterns
pick 2f2211f5 fix style and wrapping

# Rebase 50527025..2f2211f5 onto 50527025 (4 commands)
#
# Commands:
# p, pick <commit> = use commit
# r, reword <commit> = use commit, but edit the commit message
# e, edit <commit> = use commit, but stop for amending
# s, squash <commit> = use commit, but meld into previous commit
# f, fixup [-C | -c] <commit> = like "squash" but keep only the previous
#                    commit's log message, unless -C is used, in which case
#                    keep only this commit's message; -c is same as -C but
#                    opens the editor
# x, exec <command> = run command (the rest of the line) using shell
# b, break = stop here (continue rebase later with 'git rebase --continue')
# d, drop <commit> = remove commit
# l, label <label> = label current HEAD with a name
# t, reset <label> = reset HEAD to a label
# m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]
# .       create a merge commit using the original merge commit's
# .       message (or the oneline, if no original merge commit was
# .       specified); use -c <commit> to reword the commit message
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
#
# However, if you remove everything, the rebase will be aborted.
#
```

You can rewrite `pick` at the beginning of the lines to `reword`, save the text file, and close the text editor. Then, the next text editor will be opened to edit the commit messages in order.

akinomyoga: I think we can proceed with the new PR #822, but if you would like to practice it, please feel free to reopen this PR and try force-pushing the reworded commits to the branch. After that, you can close it again.

jkaivo: I'll just leave it with the other PR.

Issue #706, opened 2022-02-24
test: Intermittent test failures by timeouts, etc. with high loadavg


akinomyoga:
Let me create an issue for the discussion made in #687 before we forget it.

> ### Reply https://github.com/scop/bash-completion/pull/687#issuecomment-1021580514 by @sudonym1 
>
> > is caused by the timeout we often see
> 
> Is there a ticket for this? I wouldn't mind taking a look this upcoming weekend.
>
> ### Reply https://github.com/scop/bash-completion/pull/687#issuecomment-1021630333 by @akinomyoga 
>
> Thank you! I don't remember any. I have quickly searched, but there seems none explicitly mentioning this issue. #136 and #137 seem to be the timeouts caused by the test case itself. I'm not sure whether the cause of #139 is the test case itself or our test framework. There is a report #430 saying that the test _fails_ or _is flaky_ in ARM architecture with high loads but there is no further information. #573 is related to refactoring the test suite. I couldn't find any issue explicitly mentioning the timeout failure.
>
> ### Reply https://github.com/scop/bash-completion/pull/687#issuecomment-1025089837 by @scop
>
> Timeouts are a quite general failure mode in the test suite.
> 
> Sometimes they occur when tests fail to process expected/unexpected output or the lack of it, kind of masking the root cause. These are fairly reproducible and straightforward to fix.
> 
> Sometimes they occur due to something else, perhaps a command actually stalling every now and then, this would be hard to reproduce and fix.
> 
> But the most "interesting" thing to me (and the one I guess we're referring to) here is that I think we may still have something wrong in our test suite that triggers this in unknown situations that is not caused by the things mentioned here. Would be awesome to track these down and fix this category of issues. I believe enabling parallel tests provokes the failures more often -- at least it does provoke _some_ failures more. Parallel running happens by default in local dockerized test runs, and can be enabled on the host without docker e.g. like
> 
> ```
> $ PYTESTFLAGS="--numprocesses=auto --dist=loadfile" make check
> ```
>
> ### https://github.com/scop/bash-completion/pull/687#issuecomment-1025321800 by @sudonym1 
>
> It seems I have been getting (un)lucky. I've run the fedora docker container ~10 times with parallel pytests, and no failures so far. I will try a different approach. It could be these issues are related to the performance characteristics of the host computer.
>
> ### https://github.com/scop/bash-completion/pull/687#issuecomment-1025351856 by @akinomyoga 
>
[...]

PR #822, opened 2022-09-15
refactor(dpkg): replace grep|grep|awk with pure awk

jkaivo:
Instead of piping one grep to another, then piping that to awk, simply make one call to awk with appropriate matching. This makes the completion slightly more efficient (only one process is spawned instead of three), and removes a dependency on GNU grep's non-standard -A and -B options.

### DIFF
--- a/completions/dpkg
+++ b/completions/dpkg
@@ -13,20 +13,20 @@ if _comp_have_command grep-status; then
 else
     _comp_xfunc_dpkg_installed_packages()
     {
-        command grep -A 1 "Package: ${1-}" /var/lib/dpkg/status 2>/dev/null |
-            command grep -B 1 -Ee "ok installed|half-installed|unpacked| \
-            half-configured" \
-                -Ee "^Essential: yes" |
-            awk "/Package: ${1-}/ { print \$2 }" 2>/dev/null
+        command awk -F '\n' -v RS='' "
+            \$1 ~ /Package: $1/ &&
+            \$2 ~ /ok installed|half-installed|unpacked|half-configured|^Essential: yes/ {
+                print(substr(\$1, 10));
+            }" /var/lib/dpkg/status 2>/dev/null
     }
 
     _comp_xfunc_dpkg_purgeable_packages()
     {
-        command grep -A 1 "Package: ${1-}" /var/lib/dpkg/status 2>/dev/null |
-            command grep -B 1 -Ee "ok installed|half-installed|unpacked| \
-            half-configured|config-files" \
-                -Ee "^Essential: yes" |
-            awk "/Package: ${1-}/ { print \$2 }" 2>/dev/null
+        command awk -F'\n' -vRS='' "
+            \$1 ~ /Package: $1/ &&
+            \$2 ~ /ok installed|half-installed|unpacked|half-configured|config-files|^Essential: yes/ {
+                print(substr(\$1, 10));
+            }" /var/lib/dpkg/status 2>/dev/null
     }
 fi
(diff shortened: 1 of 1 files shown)


akinomyoga (approved):
We see a CI error, but this is just a timeout error of the test framework as discussed in #706.

completions/dpkg
```diff
@@ -13,20 +13,20 @@ if _comp_have_command grep-status; then
 else
     _comp_xfunc_dpkg_installed_packages()
     {
-        command grep -A 1 "Package: ${1-}" /var/lib/dpkg/status 2>/dev/null |
-            command grep -B 1 -Ee "ok installed|half-installed|unpacked| \
-            half-configured" \
-                -Ee "^Essential: yes" |
-            awk "/Package: ${1-}/ { print \$2 }" 2>/dev/null
+        command awk -F '\n' -v RS='' "
+            \$1 ~ /Package: $1/ &&
```
scop, line 17: Would likely be a regression, error in `nounset` mode.
```suggestion
            \$1 ~ /Package: ${1-}/ &&
```

completions/dpkg
```diff
@@ -13,20 +13,20 @@ if _comp_have_command grep-status; then
...
 
     _comp_xfunc_dpkg_purgeable_packages()
     {
-        command grep -A 1 "Package: ${1-}" /var/lib/dpkg/status 2>/dev/null |
-            command grep -B 1 -Ee "ok installed|half-installed|unpacked| \
-            half-configured|config-files" \
-                -Ee "^Essential: yes" |
-            awk "/Package: ${1-}/ { print \$2 }" 2>/dev/null
+        command awk -F'\n' -vRS='' "
+            \$1 ~ /Package: $1/ &&
```
scop, line 26: As above.
```suggestion
            index(\$1, \"Package: ${1-}\") == 1 &&
```

completions/dpkg
```diff
@@ -13,20 +13,20 @@ if _comp_have_command grep-status; then
...
     }
 
     _comp_xfunc_dpkg_purgeable_packages()
     {
-        command grep -A 1 "Package: ${1-}" /var/lib/dpkg/status 2>/dev/null |
-            command grep -B 1 -Ee "ok installed|half-installed|unpacked| \
-            half-configured|config-files" \
-                -Ee "^Essential: yes" |
-            awk "/Package: ${1-}/ { print \$2 }" 2>/dev/null
+        command awk -F'\n' -vRS='' "
```
scop, line 25: Stylistic nit, we have `-F` and `-v` here without a following space, whereas in the above we have them with the space. Let's be consistent with one of the approaches.

completions/dpkg
scop: Even better, to protect against metachars in the input, we could do away with the regex here:

```suggestion
            index(\$1, \"Package: ${1-}\") == 1 &&
```
(Not a regression in this PR, but I think we could improve this while at it.)

scop (changes requested):
Added a few comments.

Additionally, while we're at it, it seems the two functions are identical with the exception of `|config-files` being in the latter's status match. I wouldn't mind refactoring these to use one underlying function that gets the status matcher as an argument (but this can be left out if we don't feel like doing it yet).

completions/dpkg
jkaivo: It should have the spaces here, as well. I must have missed this line.

```suggestion
        command awk -F '\n' -v RS='' "
```

scop:
Thanks!

PR #823, opened 2022-09-18
feat(__load_completion): load in-tree before system data dirs

scop:
This swaps the lookup order of in-tree (relative to the main script) and system data dirs. By checking in-tree completions first, before system data dirs, we support setups running directly from a git clone better, and have a more robust setup in any case.

The main script and in-tree completions should stay in sync. There might for example be internal API changes between releases which wouldn't work if the main `bash_completion` and the set of completions we ship were from different versions, e.g. from git and a release.

Refs https://github.com/scop/bash-completion/discussions/537#discussioncomment-3617912

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2587,19 +2587,28 @@ complete -F _minimal ''
 
 __load_completion()
 {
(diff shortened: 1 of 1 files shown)


akinomyoga:
I think a part of this change overlaps with #696.

scop:
Oh, indeed, I had forgotten about that one. Sorry, and thanks for picking up these changes there!

PR #824, opened 2022-09-21
feat: set up drill to be completed by _known_hosts

foutrelis:
Included with the ldns library, the drill tool is similar to dig and is used for DNS lookups. As such, it greatly benefits from host completion.


scop:
Thanks!

Issue #786, opened 2022-08-08
Aborting a completion attempt leaves `set -o noglob` in effect


abbeyj:
## Describe the bug

Certain completions turn off globbing with `set -o noglob`, perform some operation, and then attempt to reset globbing back to its original state.  See for example https://github.com/scop/bash-completion/blob/36ceb272ddf7ef70b7fa79c5c3686080b1510054/bash_completion#L758-L763 but this pattern is used in more than one place.

Unfortunately if the user presses Ctrl+C in the middle of this then the reset code is never run and the shell is left with `set -o noglob` in effect.  This will likely be very disconcerting for the user.  For instance, running `ls *` reports `ls: cannot access *: No such file or directory`.  That's the message you'd normally expect to see if the directory is empty but now you'll see it reported for all directories, giving the impression that all of your files have been deleted.

A user could recover from this by running `set +o noglob` but they'd have to realize what was wrong first.  I'm assuming in most cases people will just think their terminal is "broken" and be forced to close it and open a new one.

See also #691 .

## To reproduce

You'll need a directory somewhere that's very slow to enumerate.  One option would be to mount an NFS directory from some slow, distant machine.  Another would be to create a local directory with an absurdly large number of files in it and have a cold cache (possibly something like `echo 3 > /proc/sys/vm/drop_caches` might help on Linux).  Or maybe https://serverfault.com/a/954175.  I can personally reproduce this with a large directory on NFS.

```
# Verify that globbing is turned on to start
$ shopt -po noglob
set +o noglob

$ cd /path/to/your/directory/<TAB><Ctrl+C>

# Globbing is now unexpectedly turned off
$ shopt -po noglob
set -o noglob
```

## Expected behavior

The `noglob` setting should retain the previous value.

## Versions (please complete the following information)

- [X] Operating system name/distribution and version: CentOS Linux release 7.9.2009 (Core)
- [X] bash version, `echo "$BASH_VERSION"`: 4.2.46(2)-release
- [X] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: This command outputs "2.11.0", but I'm actually using git commit 36ceb272ddf7ef70b7fa79c5c3686080b1510054.

## Additional context

Maybe this could be handled by running the compgen in a subshell so that the setting would not have to be remembered and then explicitly restored?

## Debug trace

[...]

PR #827, opened 2022-09-30
feat(xdg-mime): MIME type and arg completion improvements

scop:


completions/xdg-mime
```diff
@@ -2,9 +2,18 @@
 
 _xdg_mime_mimetype()
 {
-    COMPREPLY+=($(compgen -S / -W 'application audio font image message model
-                                    multipart text video' -- "$cur"))
-    [[ ${COMPREPLY-} == */ ]] && compopt -o nospace
+    local d i
+    for d in /usr/share/mime /usr/local/share/mime; do
+        command cd /usr/share/mime 2>/dev/null || continue
+        trap "command cd - &>/dev/null" RETURN
```
akinomyoga, line 8: I feel the handling of RETURN could be subtle due to the same problem as discussed in https://github.com/scop/bash-completion/issues/786#issuecomment-1221243215.

In theory, we can combine it with trapping SIGINT by saving/restoring the users' SIGINT handler. However, in experimenting it with #739, I found that Bash has a bug that the internal readline state is broken when we use `trap ... INT` inside the completion functions called through `complete -F`. I have sent [a patch](https://lists.gnu.org/archive/html/bug-bash/2022-09/msg00005.html) to Bash, but Bash 4.3..5.2 are affected at least, and these older versions will not be fixed.

Instead, I would propose a new option `-C dir` of `_filedir`. `cd $dir` may be performed in the subshell of `_filedir` as `toks=($(cd "$dir" && compgen ...))`.

completions/xdg-mime
```diff
@@ -2,9 +2,18 @@
 
 _xdg_mime_mimetype()
 {
-    COMPREPLY+=($(compgen -S / -W 'application audio font image message model
-                                    multipart text video' -- "$cur"))
-    [[ ${COMPREPLY-} == */ ]] && compopt -o nospace
+    local d i
```
akinomyoga, line 5: The change to `OLDPWD` can be localized (if we still perform `cd` in the main shell).

```suggestion
    local d i OLDPWD=$OLDPWD
```

completions/xdg-mime
scop: Damn, I had completely forgotten about that. That's a bummer, and would be great to have something to remind, because I'm bound to forget about this again (it feels so natural when one works with Go and its `defer`).

`_filedir -C` makes sense though; we have a few completions that could make use of it.

(Aside, I kind of wouldn't want to introduce more reliance on `_filedir` but rather to get rid of it some day, but the day where that would be feasible might never come...)

scop: #847 has some `_filedir -C` experimentation.

scop: 4fc5c89a462d22fdc685af79c8eade8ae476dba4 switches to an approach not involving `_filedir` or `trap`, lifting draft.

completions/xdg-mime
scop: 4fc5c89a462d22fdc685af79c8eade8ae476dba4 no longer `cd`s in the main shell, removed `ignore_env` from the test.

completions/xdg-mime
```diff
@@ -2,9 +2,24 @@
...
+            case ${arr[i]} in
+                packages*) unset -v "arr[i]" ;; # not a MIME type dir
+                *.xml) arr[i]=${arr[i]%.xml} ;;
+                */*) ;;
+                *) arr[i]+=/ ;;
+            esac
+        done
+        COMPREPLY+=("${arr[@]}")
+    done
+    [[ $COMPREPLY != */ ]] || compopt -o nospace
```
akinomyoga, line 22:
```suggestion
    [[ ${COMPREPLY-} != */ ]] || compopt -o nospace
```
In case there is no matching completion.

akinomyoga (approved):

scop:
Thanks!

PR #828, opened 2022-09-30
feat(ip): more option arg (non-)completions

scop:


completions/ip
```diff
@@ -7,13 +7,22 @@ _iproute2_etc()
...
+}
+
 _ip()
 {
     local cur prev words cword
     _init_completion || return
 
     case $prev in
-        -V | -Version | -rc | -rcvbuf)
+        -V | -Version | -rc | -rcvbuf | -l | loops)
```
akinomyoga, line 25:
```suggestion
        -V | -Version | -rc | -rcvbuf | -l | -loops)
```

akinomyoga (approved):

completions/ip
scop: Thanks!

Issue #515, opened 2021-04-08
Error while completing ssh: "bash: set -o noglob: command not found"


mattelacchiato:
## Describe the bug

When I want to auto complete known hosts for ssh, e.g. `ssh dev.<tab>`, the output is cluttered with:
```bash
ssh dev.bash: set -o noglob: command not found
bash: set -o noglob: command not found
bash: set +o noglob: command not found
myserver.com
```

## To reproduce

1. type `ssh <tab>`
2. Output is cluttered

## Expected behavior

No failure output and successful completion.

## Versions (please complete the following information)

- [x] Operating system name/distribution and version: Manjaro 21.0
- [x] bash version, `echo "$BASH_VERSION"`: 5.1.0(1)-release
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.11

## Additional context

It seems that `$reset` contains the string `'set -o noglob'`, which will get executed, but of course does not exist.

## Debug trace

[bash_completion.txt](https://github.com/scop/bash-completion/files/6276595/bash_completion.txt)
Added only the trace to the first occurrence because I don't want to mask all >500 host entries. If you need more output, let me know...

scop:
From the debug trace:
> + local configfile flag prefix= 'ifs=
> '

Do I interpret correctly that you have `$IFS` set to the newline only? That would explain it, and I can reproduce with that.

If so, are you setting it to newline only yourself, or does something in your system deliberately do that? I'm asking because we should work in such setups (will post a candidate fix for review), but it's a nonzero chance that we might be also leaking such an `$IFS` because we need to fiddle with it here and there (e.g. to support completions containing spaces), and maybe there's a code path in which we fail to `local`ize it.

scop:
Candidate fix in #519, would you mind trying it out?

I also had a peek at our IFS settings but didn't spot a place where we'd leak one.

mattelacchiato:
> Do I interpret correctly that you have $IFS set to the newline only?
Yes. I'm setting this in my `.bashrc`.

If I set `IFS=$' \t\n'` manually, ssh completion works for me, as well as if I patch my `/usr/share/bash-completion/bash_completion` manually. 

Thank you so much :heart:

scop:
>> Do I interpret correctly that you have $IFS set to the newline only?

> Yes. I'm setting this in my `.bashrc`.

[...]

PR #829, opened 2022-10-02
fix(_known_hosts_real): work around custom IFS

akinomyoga:
Fixes https://github.com/scop/bash-completion/issues/515#issuecomment-1264741656


scop (approved):
Thanks!

PR #834, opened 2022-10-17
chore(typos): add config, filter out false positives

scop:
Unfortunately `typos` lacks ability to filter out false positives on per file (or line) basis at the moment, so we end up going global with all exclusions.


scop:
The pre-commit error is unrelated to this fix; it's about pyupgrade. Will need to take a look at fixing that, ditto the fedoradev one.

PR #837, opened 2022-10-24
Add doas to _comp_root_command commands

tpikonen:
Allows correct completions when using 'doas' as a sudo replacement.


tpikonen:
Ok, so there already was a PR for doas, sorry for the noise.

PR #841, opened 2022-10-30
refactor: more `_comp_initialize` use

scop:


akinomyoga (approved):
Ah, this is my oversight. Thanks!

scop:
(Not that it matters at all, but IIRC the help completion was introduced after this one was branched, so no oversight there.)

akinomyoga:
I see. Thank you!

PR #846, opened 2022-11-20
dolphin: add completion

EmilySeville7cfg:
Simple completion for [`dolphin`](https://manned.org/dolphin.1).

### DIFF
--- a/completions/dolphin
+++ b/completions/dolphin
@@ -1,0 +1,1 @@
+complete -W '--help --help-qt --help-kde --help-all --author -v --version --license' dolphin


completions/dolphin
```diff
@@ -0,0 +1 @@
+complete -W '--help --help-qt --help-kde --help-all --author -v --version --license' dolphin
```
akinomyoga, line 1: This also overwrites the default completion of bash-completion. We would like to handle basic completions provided by `_comp_initialize` also for this command. By default, these are provided as the default completion `_minimal` of `bash-completion` but will be lost by this setting.

akinomyoga (changes requested):

PR #847, opened 2022-11-21
feat(_filedir): add `-C`, use it

scop:
This attempts to implement `_filedir -C` as discussed in #827.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -789,10 +789,11 @@ _quote_readline_by_ref()
 
 # This function performs file and directory completion. It's better than
 # simply using 'compgen -f', because it honours spaces in filenames.
-# @param $1  If `-d', complete only on directories.  Otherwise filter/pick only
-#            completions with `.$1' and the uppercase version of it as file
-#            extension.
-#
+# @param $1 Complete filenames matching `.$1' and the uppercase version of it.
+#           Ignored with `-d`.
+# OPTIONS
+#   -C dir  Change to dir before generating completions
+#   -d      Complete only on directories
 _filedir()
 {
     local IFS=$'\n'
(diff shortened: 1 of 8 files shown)


bash_completion
```diff
@@ -789,23 +789,38 @@ _quote_readline_by_ref()
...
-# @param $1  If `-d', complete only on directories.  Otherwise filter/pick only
-#            completions with `.$1' and the uppercase version of it as file
-#            extension.
-#
+# @param $1 Complete filenames matching `.$1' and the uppercase version of it.
+#           Ignored with `-d`.
+# OPTIONS
+#   -C dir  Change to dir before generating completions
+#   -d      Complete only on directories
 _filedir()
```
scop, line 797: Problem here is that the `-o filenames` at end of the function won't take effect as we are no longer in the dir we changed to. So e.g. appending slashes to dir names does not happen. Not sure what to do about that.

(Commenting here because it seems the UI here won't let me comment that far away from changed lines.)

scop: Doing the `compopt -o filenames` inside each `compgen` subshell doesn't seem to make any difference.

akinomyoga: Hmm, this is a good point. But after another thinking of it, I guess this issue also existed in the other approach in #827, so I think this is not an issue specific to this approach.

If we are to solve this issue within this PR, maybe we need to manually suffix a slash in the shell code.

- One way is to test the filename in a `for` loop suffix / if it is a directory name.
- Another way is to use `compgen -S / -d` for directory names, and get non-directory filenames using another way (`find`, etc.).
- Or we may switch between the above two based on the number of files in the specified directory.

Maybe we don't have to too much care about it, but the opposite issue is that when a filename in the specified directory has the same name as a directory in the current directory, it would be unexpectedly suffixed by `/`.

scop: Yes, the problem existed in the other implementation in #827, I don't know why I didn't notice that. In my later tests I also noticed that it ends up changing the dir to /usr/share/mime in various scenarios, something I didn't notice originally either for some reason.

I'm going to try out yet another different approach for #827 before returning to this.

scop: Tweaked some, tests pass but there are still issues:
```
$ xx() { _filedir -C foo; }
$ mkdir -p foo/bar foo/quux
$ complete -F xx x
$ x <TAB>
bar/   quux/   # <-- this is all good
$ x ba<TAB>
bar/   quux/   # <-- this is not 
```

The problem of filename elsewhere matching a dir in current dir you mentioned also persists.

akinomyoga:
Let me reopen this. The problem that the suffix slash is not added to directory names for the paths generated in another directory is not solved.

Also, I'd like to use the feature to suffix slashes to directory names to use it in another PR. There are many cases where the suffix slash is not added because the generated word does not exactly match a directory name in the current directory. In addition to the cases we change the directory, this also happens when the generated words are suffixed by `xxx,yyy,`, `@`, etc.

akinomyoga:
Rebased and added a fix.

scop:
Fixed a doc typo and rebased. Being the original author I can't approve this, but LGTM. Can merge and include in 2.17.0 if @akinomyoga you think this is good to go?

scop:
(And/or if @yedayak this looks good to you.)

akinomyoga:
I'd be happy if this is merged. As I declared in https://github.com/scop/bash-completion/pull/847#issuecomment-1732539966, I wanted to use this for another PR to improve the behavior of other filename generation, but it's already good to include this alone in 2.17.0, even if subsequent PRs will not be included in 2.17.0.

akinomyoga:
Thank you!

PR #852, opened 2022-12-06
feat(reportbug): parse various option args from option help

scop:
Closes https://github.com/scop/bash-completion/pull/753

It's unfortunate that the help output goes to stderr, but I suppose this is good enough.

### DIFF
--- a/completions/reportbug
+++ b/completions/reportbug
@@ -27,39 +27,17 @@ _reportbug()
                 done)' -- "$cur"))
             return
             ;;
-        --bts | -!(-*)B)
-            COMPREPLY=($(compgen -W "debian guug kde mandrake help" -- \
-                "$cur"))
+        --tag | --ui | --interface | --type | --bts | --severity | --mode | -!(-*)[TutBS])
+            COMPREPLY+=($(
+                compgen -W \
+                    '$("$1" $prev help 2>&1 | sed -ne /^[[:space:]]/p)' \
+                    -- "$cur"
+            ))
             return
             ;;
         --editor | --mua | --mbox-reader-cmd | -!(-*)e)
             compopt -o filenames
             COMPREPLY=($(compgen -c -- "$cur"))
-            return
-            ;;
-        --mode)
-            COMPREPLY=($(compgen -W "novice standard expert" -- "$cur"))
-            return
-            ;;
-        --severity | -!(-*)S)
-            COMPREPLY=($(compgen -W "grave serious important normal minor
-                wishlist" -- "$cur"))
-            return
-            ;;
-        --ui | --interface | -!(-*)u)
-            COMPREPLY=($(compgen -W "newt text gnome" -- "$cur"))
-            return
-            ;;
-        --type | -!(-*)t)
-            COMPREPLY=($(compgen -W "gnats debbugs" -- "$cur"))
-            return
-            ;;
-        --tag | -!(-*)T)
-            COMPREPLY=($(compgen -W "none woody potato sarge sarge-ignore
-                etch etch-ignore lenny lenny-ignore sid experimental confirmed
-                d-i fixed fixed-in-experimental fixed-upstream help l10n
-                moreinfo patch pending security unreproducible upstream wontfix
-                ipv6 lfs" -- "$cur"))
             return
             ;;
         --from-buildd)
--- a/completions/reportbug
+++ b/completions/reportbug
@@ -30,7 +30,7 @@ _reportbug()
         --tag | --ui | --interface | --type | --bts | --severity | --mode | -!(-*)[TutBS])
             COMPREPLY+=($(
                 compgen -W \
-                    '$("$1" $prev help 2>&1 | sed -ne /^[[:space:]]/p)' \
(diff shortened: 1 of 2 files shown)


scop:
CI fail is unrelated, https://github.com/scop/bash-completion/pull/853

completions/reportbug
```diff
@@ -27,41 +27,19 @@ _reportbug()
                 done)' -- "$cur"))
             return
             ;;
-        --bts | -!(-*)B)
-            COMPREPLY=($(compgen -W "debian guug kde mandrake help" -- \
-                "$cur"))
+        --tag | --ui | --interface | --type | --bts | --severity | --mode | -!(-*)[TutBS])
+            COMPREPLY+=($(
+                compgen -W \
+                    '$("$1" $prev help 2>&1 | sed -ne /^[[:space:]]/p)' \
```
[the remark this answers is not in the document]
akinomyoga: Thank you for checking the behavior. I have installed `reportbug` in my virtual machine and have checked the behavior. I think we can explicitly exclude a case like `prev=-utext` (which is interpreted as `-u text`), though `reportbug -utext help` seems to anyway fail for this particular case.

completions/reportbug
```diff
@@ -27,41 +27,19 @@ _reportbug()
                 done)' -- "$cur"))
             return
             ;;
-        --bts | -!(-*)B)
-            COMPREPLY=($(compgen -W "debian guug kde mandrake help" -- \
-                "$cur"))
+        --tag | --ui | --interface | --type | --bts | --severity | --mode | -!(-*)[TutBS])
```
akinomyoga, line 30: For example, like this

```suggestion
        --tag | --ui | --interface | --type | --bts | --severity | --mode | -!(-*|*[CefKHPsoiATjVuQtBS]*)[TutBS])
```

Hmm, but there seem to be many similar `case` patterns for other options. Maybe we do not have to care about it too much.

completions/reportbug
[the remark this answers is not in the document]
akinomyoga: I tried `result=$(reportbug -utext help)` in Debian, but it seems to try to start an interactive session for a bug report. I now believe we should exclude the cases that `prev` is an option flag plus an option argument like `-utext` (`-u` + `text`) as suggested in https://github.com/scop/bash-completion/pull/852#discussion_r1041399560.

Note: In the previous reply, I actually tried `reportbug` in Ubuntu. I reported that `reportbug -utext help` fails anyway, but it was just because `reportbug` didn't support Ubuntu.

completions/reportbug
akinomyoga: I now think we should exclude the patterns of an option flag plus an option argument. I also think we should update all the similar patterns for e.g. the following case:

```
$ reportbug -tdebbugs -[TAB]
```

This should complete into a list of possible options, but I think the current completion function tries to complete it to a list of possible option arguments for `-s` because `-tdebbugs` matches `-!(-*)s` and fails.

scop: I'm torn on this. While I agree it would be the right thing to do, we have a lot of affected completions throughout the tree. If we are to fix them, it would be good to do it in a consistent manner (for example to use the same variable name to hold appropriate flags and reuse it for all affected arg completion cases within the file), and even more so to take care of them ~everywhere instead of just singling it out here. Also, taking care of this would be quite a bit of work for a relatively rare use case/bug.

akinomyoga: I see. I agree that we can do it in a consistent manner for the whole tree, but the actual problem here is that the following hangs:

```bash
compgen -W '$(reportbug -utext help 2>&1 | sed /^[[:space:]]/p)'
```

because `reportbug` tries to start an interactive session and wait for the user input. I think an alternative workaround is to redirect stdin to `/dev/null` so `reportbug` gives up starting the interactive session.

```bash
compgen -W '$(reportbug -utext help 2>&1 </dev/null | sed /^[[:space:]]/p)'
```

completions/reportbug
akinomyoga, line 33:
```suggestion
                    '$("$1" $prev help 2>&1 </dev/null | sed -ne "/^[[:space:]]/p")' \
```

I also suggest quoting the regex passed to `sed`. I noticed that `sed -ne /^[[:space:]]/p` produces an error message and outputs the help text of `sed` in my Debian 10. This is caused by `nullglob`: the word `/^[[:space:]]/p` contains a glob pattern `[[:space:]]` matching nothing, so `sed` receives just an option `-ne` as its commandline arguments.

scop: Good catch, thanks.

completions/reportbug
scop: Addressed below in 3f9970daf5c84878cd67576268888cd7d82d14fb

akinomyoga: Thanks!

akinomyoga (approved):

completions/reportbug
akinomyoga: I'm sorry. This workaround turned out to fail to produce the candidates. I expected that `help` should output candidates even without stdin, but it seems to fail when stdin is redirected `/dev/null`. I'll think about an alternative workaround.

PR #861, opened 2022-12-18
fix: variable quoting fixes

scop:

PR #862, opened 2022-12-19
fix(completions/*): skip `prev` of the form `-oOPTARG`

akinomyoga:
I'm not sure if this is what was intended, but this is a possible fix to https://github.com/scop/bash-completion/pull/852#discussion_r1052164007 (which is causing the test failures in #861 and #546). This properly handles the completion of the form `-oOPTARG [TAB]`, yet this does not handle the completion of the form `-oOPTAR[TAB]`.

I also thought about the possibility of modifying `__reassemble_comp_words_by_ref` so that `-o` and `OPTARG` in a single word `-oOPTARG` would be generated as separate words in `words`, but this can make the processing more complicated. For example, we need to prefix `-o` for the completion of `OPTARG` if `-o` and `OPTARG` originally come from a single word, etc.

**Edit**:

- Commit 1fcbe442aecc61686901b0f375dd9302ca0dd344 is an additional change to the `pytest` completion.

### DIFF
--- a/completions/a2x
+++ b/completions/a2x
@@ -5,16 +5,18 @@ _a2x()
     local cur prev words cword split comp_args
     _comp_initialize -s -- "$@" || return
 
+    local noargopts='!(-*|*[aDd]*)'
+    # shellcheck disable=SC2254
     case $prev in
         --attribute | --asciidoc-opts | --dblatex-opts | --fop-opts | --help | \
-            --version | --xsltproc-opts | -!(-*)[ah])
+            --version | --xsltproc-opts | -${noargopts}[ah])
             return
             ;;
-        --destination-dir | --icons-dir | -!(-*)D)
+        --destination-dir | --icons-dir | -${noargopts}D)
             _filedir -d
             return
             ;;
-        --doctype | -!(-*)d)
+        --doctype | -${noargopts}d)
             _comp_xfunc asciidoc doctype
             return
             ;;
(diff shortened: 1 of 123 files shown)


scop (changes requested):
Wow, that's just a huge amount of work you've put into this, thanks! And yes, this is what I meant in #852.

I guess the only thing I'd change here would be to use something else than `otherflags` for the variable name, because it's not "other" as the set of flag names in it will contain also the one being processed (e.g. for `-${otherflags}d` there should be a `d` also in `${otherflags}`), and it doesn't capture the reason for its existence in the name.

`noargopts` springs to mind. This would read nicely like `-${noargopts}d` -- "dash followed by options that do not take an arg, followed by d".

akinomyoga:
Thank you! I have switched to `noargopts`.

I actually initially thought about `flags` (to me, which imply the options that do not take option arguments), but the variable name was already used. Another possibility might `flagopts`, but it is not explicit that they do not receive option arguments.

scop:
Nice! Merged.

Very subtle, but the reason I have a slight preference on "opts" instead of "flags" here is that the glob we use for this contains also the prevention to match long options (and I don't know if "long flags" is a thing, at least it sounds odd to me).

akinomyoga:
Thanks!

>  the glob we use for this contains also the prevention to match long options

I see. That makes sense.

PR #863, opened 2022-12-23
feat(jq): support jsonl files as well as json

polm:
`jq` support `jsonl` files just as well as plain JSON, so this just adds support for that file extension. This follows the pattern I saw in other completions, though it is my first commit here, so just let me know if something is off (or change it). 

I am not sure this counts as a "feature" but no other prefix seemed obviously appropriate.

### DIFF
--- a/completions/jq
+++ b/completions/jq
@@ -31,7 +31,7 @@ _jq()
                 return
                 ;;
             --slurpfile | --argfile)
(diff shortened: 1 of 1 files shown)


scop:
Tweaked the commit message to meet conventional commits conventions, other than that seems spot on to me. Thanks!

PR #864, opened 2023-01-09
fix typo

uniontech-lilinjie:
Signed-off-by: lilinjie <lilinjie@uniontech.com>
fix typo


scop:
CHANGELOG.md is automatically generated/updated from git logs by release-please, so I'm not sure if changes to it will persist. But we'll see now :)

PR #866, opened 2023-01-11
Brctl

steelman:
Alternatively `addbr` could also get completion of available bridges to enable avoiding clashes.

### DIFF
--- a/completions/brctl
+++ b/completions/brctl
@@ -21,11 +21,11 @@ _comp_cmd_brctl()
             ;;
         2)
             case $command in
-                show) ;;
+                addbr) ;;
 
                 *)
                     _comp_compgen_split -- "$("$1" show |
-                        _comp_awk 'NR>1 {print $1}')"
+                        _comp_awk '(NR>1 && !/^\t/) {print $1}')"
                     ;;
             esac
             ;;
--- a/completions/brctl
+++ b/completions/brctl
@@ -1,5 +1,11 @@
 # bash completion for brctl                                -*- shell-script -*-
 
+_comp_cmd_brctl__interfaces()
+{
+    _comp_compgen_split -- "$(${1:-brctl} show ${2:+"$2"} 2>/dev/null | _comp_awk \
+        '(NR == 1) { next }; (/^\t/) { print $1; next }; { print $4 }')"
+}
+
 _comp_cmd_brctl()
 {
     local cur prev words cword comp_args
@@ -25,8 +31,11 @@ _comp_cmd_brctl()
             ;;
         3)
             case $command in
-                addif | delif)
+                addif)
                     _comp_compgen_available_interfaces
+                    ;;
+                delif)
+                    _comp_cmd_brctl__interfaces "$1" "$prev"
                     ;;
                 stp)
                     _comp_compgen -- -W 'on off'
--- a/completions/brctl
+++ b/completions/brctl
@@ -26,7 +26,7 @@ _comp_cmd_brctl()
         3)
             case $command in
                 addif | delif)
-                    _comp_compgen_configured_interfaces
+                    _comp_compgen_available_interfaces
                     ;;
                 stp)
                     _comp_compgen -- -W 'on off'
--- a/completions/brctl
+++ b/completions/brctl
@@ -35,6 +35,6 @@ _comp_cmd_brctl()
             ;;
     esac
 } &&
-    complete -F _comp_cmd_brctl -o default brctl
+    complete -F _comp_cmd_brctl brctl
 
 # ex: filetype=sh


steelman:
[Why?](https://github.com/scop/bash-completion/actions/runs/3895816344/jobs/6651595057#step:6:123) With `brctl show $1` (w/o quotes) I (rightfully so) got an error massage about possible splitting.

completions/brctl
```diff
@@ -1,4 +1,19 @@
 # bash completion for brctl                                -*- shell-script -*-
+_brctl_interfaces()
+{
+    local PATH=$PATH:/sbin
+    COMPREPLY=($({
+        if [[ -z "$1" ]]; then
```
scop, line 6: With `brctl show $1` splitting and unintentional globbing can happen. With `[[ -z $1 ]]` it can't, shfmt is correct in doing the cleanup.

completions/brctl
```diff
@@ -1,4 +1,19 @@
 # bash completion for brctl                                -*- shell-script -*-
+_brctl_interfaces()
```
scop, line 2: We have some new naming conventions for functions in effect, please refer to https://github.com/scop/bash-completion/blob/master/doc/api-and-naming.md#naming

steelman: So this one's going to be `_comp_xfunc_brctl_interfaces`? I mean, there is no other consumer now, but makes perfect sense to have a function that returns a list of bridged interfaces. Then, I belive, a check like `[[ -x brctl ]] || return` needs to be added in case brctl isn't installed.

scop: Maybe, but then again we can also do that when there is a known consumer for that function. No strong opinions.

The current implementation of the function already redirects stderr to `/dev/null`, so I don't think a separate existence/executability check is required. We tend to not sprinkle those unless there's a specific reason to.

Also, the brctl to invoke should be passed to the function as a parameter. Random in-tree example:
- https://github.com/scop/bash-completion/blob/8490bf806a86a7960f0b9a14e16421ea7bd05ef5/completions/python#L6
- https://github.com/scop/bash-completion/blob/8490bf806a86a7960f0b9a14e16421ea7bd05ef5/completions/python#L39

completions/brctl
```diff
@@ -1,4 +1,12 @@
 # bash completion for brctl                                -*- shell-script -*-
+_comp_xfunc_brctl_interfaces()
+{
+    _comp_split COMPREPLY "$(${1:-brctl} show ${2:+"$2"} 2>/dev/null | _comp_awk \
+        '(NR == 1) { next }; (/^\t/) { print $1; next }; { print $4 }'))"
+
+    ((${#COMPREPLY[@]})) &&
+        _comp_split COMPREPLY "$(compgen -W '"${COMPREPLY[@]/%[[:punct:]]/}"' -- "$cur")"
```
akinomyoga, line 8:
```suggestion
    _comp_compgen_split -- "$(${1:-brctl} show ${2:+"$2"} 2>/dev/null | _comp_awk \
        '(NR == 1) { next }; (/^\t/) { sub(/[[:punct:]]$/, "", $1); print $1; next }
        { sub(/[[:punct:]]$/, "", $4); print $4 }'))"
```

completions/brctl
akinomyoga: Is this function intended to be exposed to other completions? If not, the function name should be `_comp_cmd_brctl__interfaces` or `_comp_cmd_brctl__network_interfaces`.

completions/brctl
steelman: Done. `sub(/[[:punct:]]$/, "", $1);` isn't actually necessary.

akinomyoga: Thanks!

> Done. `sub(/[[:punct:]]$/, "", $1);` isn't actually necessary.

Great, my suggestion was just an equivalent transform, without considering the output format of `brctl show`. So we can skip that if it is unnecessary.

completions/brctl
```diff
@@ -1,4 +1,9 @@
 # bash completion for brctl                                -*- shell-script -*-
+_comp_xfunc_brctl_interfaces()
```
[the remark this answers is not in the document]
steelman: Done.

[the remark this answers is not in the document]
akinomyoga: Thanks!

completions/brctl
steelman: OK, I'm fixing it. Multitasking isn't my strong suit today, sorry.

akinomyoga: Thank you!

scop (approved):
Thanks! Unfortunately this just missed the 2.16.0 release as it had fallen too far in the list of open PR's for me to notice. Will be included in the next one.

akinomyoga:
Sorry, I had a "pending" review that I forgot to submit. I added it in #1296.

PR #867, opened 2023-01-12
feat(openssl): recognize `storeutl`

vladislav-sharapov:
Add `storeutl` to commands list. [Man page](https://www.openssl.org/docs/man1.1.1/man1/openssl-storeutl.html) says this command was added in OpenSSL 1.1.1.


scop:
Thanks!

PR #883, opened 2023-02-21
refactor([0-9a-c]*): apply naming guidelines where API is as desired

scop:


completions/convert
```diff
@@ -1,6 +1,6 @@
 # bash completion for ImageMagick                          -*- shell-script -*-
 
-_ImageMagick()
+_comp_cmd__imagemagick()
```
scop, line 3: This seems to be a case which we don't have a completely applicable naming rule for. Could have made it e.g. `_comp_cmd_convert__*` but that would have felt somewhat wrong as it's for many commands here. `_comp_cmd__imagemagick` at least conveys the privacy without that problem.

akinomyoga: This is indeed tricky. Either `_comp_cmd_convert__*` or `_comp_cmd__imagemagick` seems fine to me.

When we regard `convert` as the *module* name or the file name, `_comp_cmd_convert__*` could be understood as private functions intended to be used from the other function in the file `completions/convert`. But here, there is a conflict between `_comp_cmd_${CommandName}` vs `_comp_cmd_${ModuleName}` when we define completions for multiple commands in one file.

On the other hand, `_comp_cmd__imagemagick` is more consistent with the current naming rule because the function is called from the completion functions of different commands `_comp_cmd_*`, and also the function is not intended to be directly used externally or to be specified to `complete -F func`. But this naming implies that this function may be called by arbitrary `_comp_cmd_*` in arbitrary `completions/*` files.

I guess it might be easier to introduce an additional explicit rule of the *file scope* with `_comp_cmd_${FileName}`, where `FileName` is the basename `*` of `completions/*` (dots, hyphens, etc. are replaced with `_`).

scop: Good point the `_comp_cmd__imagemagick` "visibility" implication, I think that's serious enough to rule that option out.

With regards to the file scope, I'm not sure I follow -- wouldn't there be a conflict pretty much for every command, because complefion files' basename is typically the corresponding command's basename, i.e. there would be no way to know if a prefix is for a command or a file? Or did you think of the file scope as just a "thinking aid" to justify use of one of the command names of multiple possibly defined in the file as the namespace? Or perhaps `_comp_file_${FileName}` (not `_comp_cmd_${FileName}`). With filename based ones there's the "issue" of completion file symlinks (not a biggie, but mentioning for completeness).

Changed to `_comp_cmd_convert__common_options` in 72b7f3c1bde9ec3694757ba0aa8e3bfc694d2fab. It's true that there's a potential conflict within the file, but as these are private functions, we can manage them as we like within the file so it should not be a problem. Not a biggie for xfunc ones either, we have a precedent for that with `_comp_xfunc_ssh_scp_(local|remote)_files` already. So I think this approach is fine.

akinomyoga: > Or did you think of the file scope as just a "thinking aid" to justify use of one of the command names of multiple possibly defined in the file as the namespace?

I think this is close to my original thinking.



> Changed to `_comp_cmd_convert__common_options` in [72b7f3c](https://github.com/scop/bash-completion/commit/72b7f3c1bde9ec3694757ba0aa8e3bfc694d2fab).

Thank you. This is what I thought.

completions/ant
```diff
@@ -25,7 +25,7 @@ _ant_parse_targets()
     done <"$1"
 }
 
-_ant()
+_comp_cmd_ant()
```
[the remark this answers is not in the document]
scop: Cheers, addressed.

PR #870, opened 2023-01-15
refactor: ship depreated things in a separate file

scop:
As discussed at https://github.com/scop/bash-completion/discussions/537#discussioncomment-3613857

Draft/TODO:
- [ ] note "deprecated in" bash-completion version in `@deprecated` comments
- [ ] add new parameter to `_comp_deprecate_func` to specify the "deprecated in" bash-completion version (could use this in warning messages if we ever start to emit them)
- [x] load eagerly (additionally) from in-tree bash_completion.d dir for in-tree setups
- [ ] deprecations in `completions/*`
- ...probably some more

PR #886, opened 2023-02-22
refactor(bash_completion): rename `_get_comp_words_by_ref`, etc.

akinomyoga:
I had some pending changes for the naming convention in `bash_completion`, but let me submit it now.

One thing that I haven't yet decided is whether to strip `_by_ref` from the new names or not. I think the suffix `_by_ref` is added to differentiate them from the older versions that output the result to stdout. Now we deprecated these older versions and wouldn't try to prepare the new names `_comp_*` for these deprecated functions, so there is no need for the differentiation anymore. But perhaps it might still be better to keep the suffix to remind us of the unique convention of these functions that receive variable names in the arguments.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
     hooks:
       - id: shfmt-docker
         types: [text]
-        files: ^(bash_completion|completions/.+|test/(config/bashrc|update-test-cmd-list)|.+\.sh(\.in)?)$
+        files: ^(bash_completion(\.d/[0-9]{3}_.+)?|completions/.+|test/(config/bashrc|update-test-cmd-list)|.+\.sh(\.in)?)$
         exclude: ^completions/(\.gitignore|Makefile.*)$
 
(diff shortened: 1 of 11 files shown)


akinomyoga:
I think this PR introduces conflicts in #870. I guess it is easier to first merge the change in the present version of PR #870, and then process this PR and other TODOs of #870 in separate PRs.

scop:
Thanks for this and for considering #870. I completed the "load eagerly (additionally) from in-tree bash_completion.d dir for in-tree setups" TODO in it so we don't lose any functionality temporarily. Agreed the remaining TODOs in it can be addressed separately afterwards.

akinomyoga:
I'll make this until finishing #870. After #870 is merged, I'll rebase this PR.

scop (approved):
Nice!

scop:
> or another option could be to rename `000_bash_completion_compat{ => .sh or .bash}` and enable the files of the name `bash_completion.d/*.{bash,sh}`.

I like this, implementation with just `*.bash` in https://github.com/scop/bash-completion/pull/899

PR #887, opened 2023-02-23
refactor(completions/*): ensure quoting for `compgen -W '${arr[@]}'`

akinomyoga:
There are also other fixes I noticed while rewriting `compgen -W '${arr[@]}'`. The descriptions are found in the commit messages.

### DIFF
--- a/completions/minicom
+++ b/completions/minicom
@@ -34,7 +34,7 @@ _minicom()
     fi
 
     local -a files
-    _comp_expand_glob files '/etc/minirc.* /etc/minicom/minirc.* ~/.minirc.*'
+    _comp_expand_glob files '{/etc/,/etc/minicom/,~/.}minirc.?*'
     ((${#files[@]})) &&
         COMPREPLY=($(compgen -W '"${files[@]##*minirc.}"' -- "$cur"))
 } &&
--- a/completions/minicom
+++ b/completions/minicom
@@ -19,7 +19,7 @@ _minicom()
         --ptty | -${noargopts}p)
             _comp_expand_glob COMPREPLY '/dev/tty*'
             ((${#COMPREPLY[@]})) &&
-                COMPREPLY=($(compgen -W '"${COMPREPLY[@]}" "${COMPREPLY[@]#/dev/}}' \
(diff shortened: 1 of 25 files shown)


completions/minicom
```diff
@@ -34,7 +34,7 @@ _minicom()
     fi
 
     local -a files
-    _comp_expand_glob files '/etc/minirc.* /etc/minicom/minirc.* ~/.minirc.*'
+    _comp_expand_glob files '{/etc/,/etc/minicom/,~/.}minirc.?*'
```
scop, line 37: If I read correctly, the empty completions would occur (only) with files having basename `minirc.`. If that's the case, could note that in the commit message.

scop (commented):
This seems the right thing to do in a sense, but given the uses are mostly in cases like `COMPREPLY=($(...))` that just end up splitting everything to the array by whitespace anyway, I feel in practice this doesn't make much of a practical difference at the moment, or at least it doesn't generally protect against whitespace embedded in individual completion items. But perhaps I'm missing something?

akinomyoga:
> This seems the right thing to do in a sense, but given the uses are mostly in cases like `COMPREPLY=($(...))` that just end up splitting everything to the array by whitespace anyway, I feel in practice this doesn't make much of a practical difference at the moment, or at least it doesn't generally protect against whitespace embedded in individual completion items. But perhaps I'm missing something?

Ah, yes. I actually intend to submit a separate PR for the mass change of `COMPREPLY=($(...))` for the protection from random `IFS` (025736d42e7bb4f8b022b979d459546985bc49b0 is a WIP commit), but it's blocked by #870 and even another planned PR.

Another reason is that it is not evident at a glance whether these codes are kept unquoted because they are harmless or just not yet checked. It is noisy for me because I tend to check whether it is safe or not every time I witness such a code.

completions/minicom
akinomyoga: Sure!

akinomyoga: I added it in 3597289e71be0ddedc9a1eea590e72a67db9204f.

scop:
Just a ping here about the conflict so it doesn't go unnoticed, no rush.

akinomyoga:
I have rebased it.

scop (approved):

PR #888, opened 2023-02-23
fix(java): fix for failglob and refactor

akinomyoga:


completions/java
```diff
@@ -120,8 +120,6 @@ _java_packages()
     if ((${#COMPREPLY[@]} != 0)); then
         # keep only packages with the package suffix `/` being removed
         _comp_split -l COMPREPLY "$(printf '%s\n' "${COMPREPLY[@]}" | command sed -n 's,/$,,p')"
-        # convert path syntax to package syntax
-        cur="${COMPREPLY[*]//\//.}"
```
akinomyoga, line 122: Something that I don't understand is these lines in the original code. It seems strange to replace `cur` with the *list* of package names. I guessed it intended to revert the change in `cur` made in [this line](https://github.com/scop/bash-completion/pull/888/commits/98a9f8cd3eb5302f16e1f86d82f3e3a97f110308#diff-20440e71a6a7ec0be45bf8bec35417781bcdc464f87f1701f577036237621032L110), but it's still strange why it is performed only when there are non-zero packages.

Also, this function `_java_packages` is only used from `_javadoc` in the codebase, where the function is called at the end of the function. This means that the value of `cur` after the call of `_java_packages` is unused now.

In this commit 98a9f8cd3eb5302f16e1f86d82f3e3a97f110308, I just made `cur` local to this function as it doesn't seem to be used by caller functions.

akinomyoga: I checked the change history of this line.

> but it's still strange why it is performed only when there are non-zero packages.

Originally, it seems to have been always performed. In commit 1a633f3a551154ac3c70bfdc5263d73d8b3d9f3f, the entire part including the change to `cur` seems to be put inside the `if` statement.

> It seems strange to replace cur with the list of package names.

If I further track back the history, the line seems to be introduced by this commit: [`ff01750d` (`bash_completion:2818`)](https://github.com/scop/bash-completion/commit/ff01750d2ff63520649549f7d269a44c7ca65dab#diff-a4757074ff650000804fd3eaabe9b0a9e02e33040ca5b8afd4c0275fc5f3e136R2818). I don't find the corresponding line prior to this commit, but there is [a similar line](https://github.com/scop/bash-completion/commit/ff01750d2ff63520649549f7d269a44c7ca65dab#diff-a4757074ff650000804fd3eaabe9b0a9e02e33040ca5b8afd4c0275fc5f3e136R2793) that is introduced at the same time, which has [the corresponding line](https://github.com/scop/bash-completion/commit/ff01750d2ff63520649549f7d269a44c7ca65dab#diff-a4757074ff650000804fd3eaabe9b0a9e02e33040ca5b8afd4c0275fc5f3e136L2765) before the commit. But this similar assigns the results to `COMPREPLY` but not to `cur`. I suspect this is a bug introduced in this commit. It is reasonable to think that this line intended to assign the values to `COMPREPLY` but not `cur`.

I then checked the man page of the `javadoc` command

> *Quoted from [Javadoc Command](https://docs.oracle.com/javase/9/javadoc/javadoc-command.htm#JSJAV-GUID-3051737B-FFF9-45F9-85D1-5F07D6703592)*:
>
> ### Synopsis
> ```
> javadoc  [packages|source-files] [options][@files]
> ```
>
> *packages*
>
> Names of packages that you want to document, separated by spaces, for example `java.lang java.lang.reflect java.awt`. If you want to also document the subpackages, then use the `-subpackages` option to specify them.

It requires the package name of the form `xxx.yyy.zzz` delimited by periods. So, I conclude it is a bug that tries to assign the results to `cur` instead of `COMPRELY`.

akinomyoga: I fixed it in commit c85af251298ca95e67f7171ddf85fed31d48ce1b and described the background in its commit message.

scop:
Looks like there is a problem with nonexistent sourcepath as well as a `nounset` related one triggered by the same recipe:
```
$ javadoc -sourcepath /non/existent<TAB>bash_completion: _comp_split: unexpected number of arguments.
usage: _comp_split [-a] [-F SEP] ARRAY_NAME TEXT
bash: sourcepath: unbound variable
```

Could we add some test cases for the changed bits? Doesn't have to be anything elaborate, even basic ones would do.

scop (changes requested):
Looks like there is a problem with nonexistent sourcepath as well as a `nounset` related one triggered by the same recipe:
```
$ javadoc -sourcepath /non/existent<TAB>bash_completion: _comp_split: unexpected number of arguments.
usage: _comp_split [-a] [-F SEP] ARRAY_NAME TEXT
bash: sourcepath: unbound variable
```

Also, could we add some kind of a test case for the changed bits? Doesn't have to be anything elaborate, even basic ones would do.

akinomyoga:
rebased

scop:
Just to clarify, this is still intentionally a draft, perhaps due to some tests not being added yet?

akinomyoga:
Ah, yes. Sorry for confusing you. I'll later add tests.

akinomyoga:
I added tests for each fix (a commit for failing tests and a commit for the fix is paired for each).

I also picked changes in #942 and modified (53d98e844d9cb140e16e0ff2f943950c8b1cb0a4).

scop (approved):
Thanks! Fixed a typo in the 2nd last commit message, merging with that.

PR #891, opened 2023-02-23
fix: add fixes for `localvar_inherit`

akinomyoga:
I think this is not the complete list of the places that need a fix for `localvar_inherit`, but this pending branch starts to conflict with others frequently. I'm not planning to soon scan through all the codebase, so I decided to submit it as is now.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -396,8 +396,7 @@ _comp_looks_like_path()
 #
 __reassemble_comp_words_by_ref()
 {
-    local exclude i j line ref
-    unset -v exclude # workaround for localvar_inherit
+    local exclude="" i j line ref
     # Exclude word separator characters?
     if [[ $1 ]]; then
         # Yes, exclude word separator characters;
@@ -408,7 +407,7 @@ __reassemble_comp_words_by_ref()
     # Default to cword unchanged
     printf -v "$3" %s "$COMP_CWORD"
     # Are characters excluded which were former included?
-    if [[ -v exclude ]]; then
+    if [[ $exclude ]]; then
         # Yes, list of word completion separators has shrunk;
         line=$COMP_LINE
         # Re-assemble words to complete
@@ -532,13 +531,18 @@ __get_cword_at_cursor_by_ref()
 #
 _get_comp_words_by_ref()
 {
-    local exclude flag i OPTIND=1
+    local exclude="" flag i OPTIND=1
     local cur cword words=()
-    local upargs=() upvars=() vcur vcword vprev vwords
-    unset -v vcur vcword vprev vwords # workaround for localvar_inherit
+    local upargs=() upvars=() vcur="" vcword="" vprev="" vwords=""
 
     while getopts "c:i:n:p:w:" flag "$@"; do
         case $flag in
+            [cipw])
+                if [[ $OPTARG != [a-zA-Z_]*([a-zA-Z_0-9])?(\[*\]) ]]; then
+                    echo "bash_completion: $FUNCNAME: -$flag: invalid variable name \`$OPTARG'" >&2
+                    return 1
+                fi
+                ;;&
             c) vcur=$OPTARG ;;
             i) vcword=$OPTARG ;;
             n) exclude=$OPTARG ;;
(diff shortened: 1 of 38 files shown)


scop (commented):
Looks good in a way to me, but to be honest, having to add one line per such case along with a comment why it's done feels worse than if we'd instead initialize to the empty string on define and test with `[[ $foo ]]`.

What do you think?

akinomyoga:
I'm just afraid that it might break functions that would like to distinguish an empty value `''`and the unset state. I can carefully check one by one if each function wants to distinguish an empty value and the unset state or not. Let me see.

akinomyoga:
I've checked each of the codes that use `local var; ...; [[ -v var ]]`. In checking them, I found three suspicious places so first fixed them (57d1cb73d6f18a106d33f9ef7837592ae1cd5ba4, 2c39e49fdf0e340861c615a6a4b266eb4d42aff1, 4b41777788047c5068d4751e0f0a9357eccadc9e). Then, I rewrote the codes in 7034483034ce22d103fcea38017e6708d7856ce8 and  3ba87cfb4686aa0ebc4f9eeaf1ffda9898368269.

scop (approved):
Nice, looks better to me now.

A couple of points about the `$has_` variables (no need to block this one for that, we can look into addressing them in followup PR's if we like). Leaving up to you to decide if we postpone them or do some of it here.

1) I suppose many of the cases where a `$has_*` was introduced here are because of bugs in existing implementations that fail to consider the empty string appropriately. Would be good to clean those up.

2) We have some precedent in `_split_longopt` and `$split` (and some other cases I think) in making vars used for booleans to have `true` or `false` values which can be tested by executing instead of doing the variable emptiness check (i.e. testing e.g. `if $foo; ...` instead of `if [[ $foo ]]; ...`. I think that reads cleaner, we might want to have a look at doing that for the cases remaining if/after we weed out the cases where the separate boolean is not required in the first place.

akinomyoga:
> 1. I suppose many of the cases where a `$has_*` was introduced here are because of bugs in existing implementations that fail to consider the empty string appropriately. Would be good to clean those up.

Currently, I do not have a good idea of how to handle these cases, so I think I'd postpone the resolution of these cases.

The remaining `$has_*` are all related to empty arguments stored in `words`, which originally come from `COMP_WORDS`. We usually don't have an empty element in `COMP_WORDS` (e.g., the *empty* argument `''` or `""` is usually stored in `COMP_WORDS` literally as `''` or `""`), but `COMP_WORDS` can still contain empty elements in some cases (e.g., `cmd arg [TAB]`). Having an empty element in `words` itself is not a bug because the empty element is coming from the user content, and we cannot output error or warning messages when we find the empty element.

----

> 2\. We have some precedent in `_split_longopt` and `$split` (and some other cases I think)

These are the other cases (excluding `$split`):

```bash
./bash_completion:823:    if $deduplicate; then
./completions/export:55:            if ! $remove && [[ $action != function ]]; then
./completions/find:74:    if ! $exprfound && [[ $cur != [-\(\),\!]* ]]; then
./completions/fio:121:                        if $in_values; then
./completions/jarsigner:43:    if ! $jar; then
./completions/lzip:38:    if $decompress; then
./completions/pack200:56:    if ! $pack; then
./completions/pack200:68:    elif ! $jar; then
./completions/pytest:115:            if $in_class && [[ $line =~ ^[[:space:]]+(async[[:space:]]+)?def[[:space:]]+(test_[A-Za-z0-9_]+) ]]; then
./completions/route:23:        $found || COMPREPLY+=("$opt")
./completions/ssh:490:    if $dirsonly; then
./completions/timeout:11:            if $found; then
./completions/unpack200:36:    if ! $pack; then
./completions/unpack200:44:    elif ! $jar; then
```

Another approach seems to prepare a variable that has a value 0 or 1, and test it with `((var == 0))`, etc.

```bash
./completions/mcrypt:56:                decrypt=1
./completions/postcat:22:        [[ $idx == -q ]] && qfile=1 && break
./completions/svcadm:32:        exact_mode=1
./completions/tar:174:        old_opt_progress=1
./completions/tar:179:        old_opt_used=1
./completions/tar:204:    first_arg=1
./completions/upgradepkg:20:        [[ ${COMPREPLY-} ]] || nofiles=1
./completions/wodim:72:            track_mode=1
```

The other boolean flags have been assigning a non-empty value and testing it with `[[ -v $var ]]`, which were now converted to `[[ $var ]]`. This is the approach that I used for these new `has_*` variables.

```bash
./bash_completion:1909:            4) ipv4=1 ;;
./bash_completion:1910:            6) ipv6=1 ;;
./completions/tar:694:    basic_tar=yes
```

akinomyoga:
I have switched `has_xxx` to use `true` / `false` and `"$has_xxx"` / `! "$has_xxx"` ([`cfe3487b..ace9cd66`](https://github.com/scop/bash-completion/compare/cfe3487b1defc8d48577bde8d92a5e86452974c3..ace9cd66be8ad76729a8aafdd02d0109f7f28054)). I kept existing cases.

PR #892, opened 2023-02-23
chore(pre-commit/update-deprecated-links): exclude untracked files

akinomyoga:
The description is in the commit message 36573340e0f06b1da97bfb0ff2aa2596a3871b1a.

### DIFF
--- a/test/deprecated/update-deprecated-links
+++ b/test/deprecated/update-deprecated-links
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 is_tracked_by_git()
--- a/test/deprecated/update-deprecated-links
+++ b/test/deprecated/update-deprecated-links
@@ -1,14 +1,20 @@
 #!/bin/bash
 set -euo pipefail
 
+is_tracked_by_git()
+{
+    git ls-files --error-unmatch "$1" &>/dev/null
+}
+
 cd "$(dirname "$0")/completions"
 for f in *; do
-    case $f in
-        Makefile* | README.md) ;;
-        *) git rm -f $f ;;
-    esac
+    if [[ -h $f ]] && is_tracked_by_git "$f"; then
+        git rm -f "$f"
+    fi
 done
 for f in ../../../completions/_*; do
-    ln -sf $f ${f##*/_}
-    git add --verbose ${f##*/_}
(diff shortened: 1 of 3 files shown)


test/deprecated/update-deprecated-links
```diff
@@ -1,14 +1,21 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
scop, line 1: I think this could use some explanation/rationale in the commit message. We don't do that for any other bash script of ours, and for example `test/update-test-cmd-list` which is also invoked by pre-commit has a `/bin/bash` one.

test/deprecated/update-deprecated-links
```diff
@@ -1,14 +1,21 @@
...
-        *) git rm -f $f ;;
-    esac
+    if [[ -h $f ]] && _is_tracked_by_git "$f"; then
+        git rm -f "$f"
+    fi
 done
-for f in ../../../completions/_*; do
-    ln -sf $f ${f##*/_}
-    git add --verbose ${f##*/_}
+for f in ../../../completions/_!(*.swp|*~); do
```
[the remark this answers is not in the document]
akinomyoga: Thanks. That makes sense. I actually initially excluded these backup files and did not check whether the file is tracked. The git check was added later. Now we can remove the exclusion by `!()`. I'll remove it.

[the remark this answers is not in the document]
akinomyoga: I have removed the `!()` exclusion. This is the updated version cfc64cad5ab5f9f765f0a10b3ba9ff887f89e70e.

test/deprecated/update-deprecated-links
akinomyoga: I have also updated `test/update-test-cmd-list` and `runLint`. I added reasoning in the commit message, but if you think we don't have to change the shebang, I can drop the commit from the PR as I'm using Linux and do not have the issue with the path of Bash.

I kept the shebang of `test/docker/*/install-packages.sh` because these are scripts for Linux, but these might also be updated for consistency.

scop: This is fine with me. Could fine tune the message of the latter commit now that it includes more than just `update-deprecated-links`, pre-approved with that change. Thanks!

scop (approved):

PR #905, opened 2023-03-23
feat: add 3rd party generated completion fallbacks

scop:
Per https://github.com/scop/bash-completion/discussions/904


completions/_gh
```diff
@@ -0,0 +1,18 @@
+# 3rd party completion loader for commands emitting        -*- shell-script -*-
+# their completion using "$cmd completion --shell bash".
+#
+# This serves as a fallback in case the completion is not installed otherwise.
+
+# shellcheck disable=SC2168 # "local" is ok, assume sourced by __load_completion
+local _comp__load_3rdparty_reset_shopt=$(shopt -po posix)
+set +o posix
+
+. <("$1" completion --shell bash 2>/dev/null)
```
scop, line 10: Regarding the discussion on non-posix mode in https://github.com/scop/bash-completion/discussions/904#discussioncomment-5397380: my use case was primarily for the process substitution here.

The suggested eval might work instead of it, did not consider that, thanks! But I think it's too much to assume that the completions we load here are posix mode safe, hence setting non-posix mode serves that purpose too.

Then again, if the shell _is_ in posix mode, I'm not sure if there are any guarantees that the function that was loaded without posix mode in effect will actually work in posix mode when invoked, so this isn't bulletproof.

bash_completion
```diff
@@ -2567,7 +2567,7 @@ __load_completion()
             # Do not warn with . or .. (especially the former is common)
             [[ $compfile == */.?(.) ]] ||
                 echo "bash_completion: $compfile: is a directory" >&2
-        elif [[ -e $compfile ]] && . "$compfile"; then
+        elif [[ -e $compfile ]] && . "$compfile" "$cmd"; then
```
scop, line 2570: I guess this is something we could consider adding a test for, but I'm not sure if it's worth the trouble.

Also not sure about adding tests for the the actual completions as their result varies depending on if the generating command is installed or not.

completions/_gh
akinomyoga: > Then again, if the shell _is_ in posix mode, I'm not sure if there are any guarantees that the function that was loaded without posix mode in effect will actually work in posix mode when invoked, so this isn't bulletproof.

Hmm, this is also a good point.

~~It seems cobra uses process substitutions in its completions (for [V1](https://github.com/spf13/cobra/blob/4dd4b25de38418174a6e859e8a32eaccca32dccc/bash_completions.go#L165) and [V2](https://github.com/spf13/cobra/blob/4dd4b25de38418174a6e859e8a32eaccca32dccc/bash_completionsV2.go#L235)). Ah, but you have actually already reported it [there](https://github.com/spf13/cobra/pull/1321#issuecomment-874279523), though there doesn't seem to be any reply to that comment. I think we can submit a PR to fix it to use `<<< "$(...)"` instead, though I'm not sure if process substitutions are the only POSIX-incompatible part of the cobra completions.~~

~~The problem here is that, if we start to load the completions provided by the command itself automatically, the user doesn't have the option to enable the POSIX mode and intentionally *does not* load the completions provided by the command to avoid errors. Maybe we can load the completions by Cobra only when the POSIX mode is turned off, but it might be too much if the Cobra completions will be fixed to support the POSIX mode.~~

Now I tested it with Bash 4.2 and 5.0. The POSIX mode of Bash seems to complain about the process substitutions only in the parsing stage. Once the function is defined (in the non-POSIX mode), the function seems to be able to be used without errors even in the POSIX mode. So, for the process substitutions used inside the functions, it is enough to turn off the POSIX mode only when we load the completion.

So I now think we can turn off the POSIX mode only when we load the completion as we are doing here now.

----

By the way, although it is not a completion setting, I remember that Starship also uses a process substitution in the startup script that it outputs for `eval "$(starship init bash)"`. There seems to be a report at https://github.com/starship/starship/issues/1674, but it's not solved.

akinomyoga: I now remembered another issue of toggling the POSIX mode described in CONTRIBUTING.md:

> Unfortunately due to a bash < 5.1 bug, toggling POSIX mode interferes with keybindings and should not be done. This rules out use of process substitution which causes syntax errors in POSIX mode.

scop: GAH, thanks a lot for the sanity checks. I wonder if shellcheck could be taught to flag it if we ever try to fiddle with POSIX mode. https://github.com/scop/bash-completion/pull/907 adds it to `runLint`.

43da208ace65a924262838aa977e9b22fcd780fc switches to using `eval`, guess that's the best alternative we have here.

scop: One thing worth noting regarding `eval`: unlike sourcing, `return` does not work with it. But I think its usage is rare in these files, and a common use case among those rare ones likely is to return early when an unsupported setup is encountered (for example like we do in `ifup`). And in that case what we want to generally do is to fall back to our default completion. In that sense, it's not a functional problem.

However, if the intent of a file is to return gracefully, that will cause `bash: return: can only 'return' from a function or sourced script` error spewage which should be harmless too, but surely annoying.

I guess we could do `eval -- "$(...)" 2>/dev/null` to avoid this, but it would throw all other stderr away as well, which is not something we do when sourcing regular completion files, and there's value in being consistent here.

Anyway, not a blocker for using `eval` as far as I'm concerned, mostly just thinking aloud here.

akinomyoga: `return` inside `eval` will be evaluated in the caller context, which means that it causes the return from the `. completions/_gh`, etc. I think error messages will not be shown in this case, but the check for `complete -p "$1" || complete -p "${1##*/}"` will be skipped. In this case, if the script returns 0 without defining the completion setting, performing `_minimal` is expected, but nothing happens.

akinomyoga (approved):
As for adding or not adding the tests, either works for me.

completions/_gh
scop: Ah, right :+1:

PR #907, opened 2023-03-26
test(runLint): add check for touching POSIX mode

scop:
Problematic because it can cause keybindings to change.

https://github.com/scop/bash-completion/issues/190

As we are not running `runLint` automatically and we aren't clean per its checks anyway, this won't be catching all of it. But I guess it's better than nothing.


akinomyoga (approved):
(Pre-)approved. Maybe we can also detect `POSIXLY_CORRECT=`, which also affects the keybinding.

scop:
Cheers, amended with some `POSIXLY_CORRECT` detection.

PR #910, opened 2023-03-29
fix(rsync,ssh): do not overescape spaces in remote filenames

### LINKED ISSUE #848: rsync ssh remote paths are double-escaped (incorrect since rsync 3.2.4)
## Describe the bug

[In 3.2.4](https://download.samba.org/pub/rsync/NEWS#3.2.4) rsync changed the default way that arguments were passed to a remote ssh host, to no longer require double-escaping.

The bash-completion still double-escapes.

## To reproduce

1. Create `/tmp/foo bar` on a remote machine (named `remote`)
2. Type `rsync remote:/tmp/foo`
3. Press tab.
4. Note that this completes to `rsync remote:/tmp/foo\\\ bar`, which is incorrect.

## Expected behaviour

The remote path should only be escaped once, not twice (unless the `--old-args` argument is passed to rsync (>= 3.2.4)

- [x] Operating system name/distribution and version: Debian testing
- [x] bash version, `echo "$BASH_VERSION"`: `5.2.2(1)-release`
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: `2.11` (Debian: `1:2.11-6`)

Rogach:
Fixes #848.

If a remote machine contains a file named `a b`, completing rsync command results in `rsync remote:a\\\ b`, which results in rsync failing to find the file.

This commit removes the extra slashes, and now completion results in `rsync remote:a\ b`.

scp somehow accepts both variants, so this change won't break it.

### DIFF
--- a/completions/rsync
+++ b/completions/rsync
@@ -1,5 +1,24 @@
 # bash completion for rsync                                -*- shell-script -*-
 
+_comp_cmd_rsync__vercomp()
+{
+    if [[ $1 == "$2" ]]; then
+        return 0
+    fi
+    local IFS=.
+    local i ver1=($1) ver2=($2)
+    local n=$((${#ver1[@]} >= ${#ver2[@]} ? ${#ver1[@]} : ${#ver2[@]}))
+    for ((i = 0; i < n; i++)); do
+        if ((10#${ver1[i]:-0} > 10#${ver2[i]:-0})); then
+            return 1
+        fi
+        if ((10#${ver1[i]:-0} < 10#${ver2[i]:-0})); then
+            return 2
+        fi
+    done
+    return 0
+}
+
 _comp_cmd_rsync()
 {
     local cur prev words cword was_split comp_args
@@ -81,7 +100,15 @@ _comp_cmd_rsync()
                     break
                 fi
             done
-            [[ $shell == ssh ]] && _comp_compgen -x scp remote_files
+            if [[ $shell == ssh ]]; then
+                local rsync_version=$("$1" --version 2>/dev/null | sed -n '1s/.*rsync *version \([0-9.]*\).*/\1/p')
+                _comp_cmd_rsync__vercomp "$rsync_version" "3.2.4"
+                if (($? == 2)); then
+                    _comp_compgen -x scp remote_files
+                else
+                    _comp_compgen -x scp remote_files -l
+                fi
+            fi
             ;;
         *)
             _comp_compgen_known_hosts -c -a -- "$cur"
--- a/completions/rsync
+++ b/completions/rsync
@@ -5,8 +5,9 @@ _comp_cmd_rsync__vercomp()
     if [[ $1 == "$2" ]]; then
         return 0
     fi
-    local IFS=.
-    local i ver1=($1) ver2=($2)
+    local i ver1 ver2
+    _comp_split -F . ver1 "$1"
+    _comp_split -F . ver2 "$2"
     local n=$((${#ver1[@]} >= ${#ver2[@]} ? ${#ver1[@]} : ${#ver2[@]}))
     for ((i = 0; i < n; i++)); do
         if ((10#${ver1[i]:-0} > 10#${ver2[i]:-0})); then
--- a/completions/ssh
+++ b/completions/ssh
@@ -462,12 +462,30 @@ _comp_cmd_sftp()
 # shellcheck disable=SC2089
 _comp_cmd_scp__path_esc='[][(){}<>"'"'"',:;^&!$=?`\\|[:space:]]'
 
-# Complete remote files with ssh.  If the first arg is -d, complete on dirs
-# only.  Returns paths escaped with three backslashes.
+# Complete remote files with ssh. Returns paths escaped with three backslashes
+# (unless -l option is provided).
+# Options:
+#     -d  Complete on dirs only.
+#     -l  Return paths escaped with one backslash instead of three.
 # @since 2.12
 # shellcheck disable=SC2120
 _comp_xfunc_scp_compgen_remote_files()
 {
+    local _dirs_only=""
+    local _less_escaping=""
+
+    local _flag OPTIND=1 OPTARG="" OPTERR=0
+    while getopts "dl" _flag "$@"; do
+        case $_flag in
+            d) _dirs_only=set ;;
+            l) _less_escaping=set ;;
+            *)
+                echo "bash_completion: $FUNCNAME: usage error: $*" >&2
+                return 1
+                ;;
+        esac
+    done
+
     # remove backslash escape from the first colon
     local cur=${cur/\\:/:}
 
@@ -483,20 +501,25 @@ _comp_xfunc_scp_compgen_remote_files()
         _path=$(ssh -o 'Batchmode yes' "$_userhost" pwd 2>/dev/null)
     fi
 
+    local _escape_replacement='\\\\\\&'
+    if [[ $_less_escaping ]]; then
+        _escape_replacement='\\&'
+    fi
+
     local _files
-    if [[ ${1-} == -d ]]; then
+    if [[ $_dirs_only ]]; then
         # escape problematic characters; remove non-dirs
         # shellcheck disable=SC2090
         _files=$(ssh -o 'Batchmode yes' "$_userhost" \
             command ls -aF1dL "$_path*" 2>/dev/null |
-            command sed -e 's/'"$_comp_cmd_scp__path_esc"'/\\&/g' -e '/[^\/]$/d')
+            command sed -e 's/'"$_comp_cmd_scp__path_esc"'/'"$_escape_replacement"'/g' -e '/[^\/]$/d')
     else
         # escape problematic characters; remove executables, aliases, pipes
         # and sockets; add space at end of file names
         # shellcheck disable=SC2090
         _files=$(ssh -o 'Batchmode yes' "$_userhost" \
             command ls -aF1dL "$_path*" 2>/dev/null |
-            command sed -e 's/'"$_comp_cmd_scp__path_esc"'/\\&/g' -e 's/[*@|=]$//g' \
+            command sed -e 's/'"$_comp_cmd_scp__path_esc"'/'"$_escape_replacement"'/g' -e 's/[*@|=]$//g' \
                 -e 's/[^\/]$/& /g')
     fi
     _comp_compgen -R split -l -- "$_files"
--- a/completions/ssh
+++ b/completions/ssh
@@ -489,14 +489,14 @@ _comp_xfunc_scp_compgen_remote_files()
         # shellcheck disable=SC2090
         _files=$(ssh -o 'Batchmode yes' "$_userhost" \
             command ls -aF1dL "$_path*" 2>/dev/null |
-            command sed -e 's/'"$_comp_cmd_scp__path_esc"'/\\\\\\&/g' -e '/[^\/]$/d')
+            command sed -e 's/'"$_comp_cmd_scp__path_esc"'/\\&/g' -e '/[^\/]$/d')
     else
         # escape problematic characters; remove executables, aliases, pipes
         # and sockets; add space at end of file names
         # shellcheck disable=SC2090
         _files=$(ssh -o 'Batchmode yes' "$_userhost" \
             command ls -aF1dL "$_path*" 2>/dev/null |
-            command sed -e 's/'"$_comp_cmd_scp__path_esc"'/\\\\\\&/g' -e 's/[*@|=]$//g' \
+            command sed -e 's/'"$_comp_cmd_scp__path_esc"'/\\&/g' -e 's/[*@|=]$//g' \
                 -e 's/[^\/]$/& /g')
     fi
     _comp_compgen -R split -l -- "$_files"
--- a/test/t/test_rsync.py
+++ b/test/t/test_rsync.py
@@ -1,5 +1,9 @@
 import pytest
 
+from conftest import assert_bash_exec
+
+LIVE_HOST = "bash_completion"
+
 
 @pytest.mark.bashcomp(ignore_env=r"^[+-]_comp_cmd_scp__path_esc=")
 class TestRsync:
@@ -18,3 +22,54 @@ class TestRsync:
     @pytest.mark.complete("rsync --", require_cmd=True)
     def test_4(self, completion):
         assert "--help" in completion
+
+    @pytest.mark.parametrize(
+        "ver1,ver2,result",
+        [
+            ("1", "1", "="),
+            ("1", "2", "<"),
+            ("2", "1", ">"),
+            ("1.1", "1.2", "<"),
+            ("1.2", "1.1", ">"),
+            ("1.1", "1.1.1", "<"),
+            ("1.1.1", "1.1", ">"),
+            ("1.1.1", "1.1.1", "="),
+            ("2.1", "2.2", "<"),
+            ("3.0.4.10", "3.0.4.2", ">"),
+            ("4.08", "4.08.01", "<"),
+            ("3.2.1.9.8144", "3.2", ">"),
+            ("3.2", "3.2.1.9.8144", "<"),
+            ("1.2", "2.1", "<"),
+            ("2.1", "1.2", ">"),
+            ("5.6.7", "5.6.7", "="),
+            ("1.01.1", "1.1.1", "="),
+            ("1.1.1", "1.01.1", "="),
+            ("1", "1.0", "="),
+            ("1.0", "1", "="),
+            ("1.0.2.0", "1.0.2", "="),
+            ("1..0", "1.0", "="),
+            ("1.0", "1..0", "="),
+        ],
+    )
+    def test_vercomp(self, bash, ver1, ver2, result):
+        output = assert_bash_exec(
+            bash,
+            f"_comp_cmd_rsync__vercomp {ver1} {ver2}; echo $?",
+            want_output=True,
+        ).strip()
+
+        if result == "=":
+            assert output == "0"
+        elif result == ">":
+            assert output == "1"
+        elif result == "<":
+            assert output == "2"
+        else:
+            raise Exception(f"Unsupported comparison result: {result}")
+
+    @pytest.mark.complete(f"rsync {LIVE_HOST}:spaces", sleep_after_tab=2)
+    def test_remote_path_with_spaces(self, completion):
+        assert (
+            completion == r"\ in\ filename.txt"
+            or completion == r"\\\ in\\\ filename.txt"
+        )
--- a/test/t/test_rsync.py
+++ b/test/t/test_rsync.py
@@ -1,8 +1,6 @@
 import pytest
 
-from conftest import assert_bash_exec
-
-LIVE_HOST = "bash_completion"
+from conftest import assert_bash_exec, assert_complete
 
 
 @pytest.mark.bashcomp(ignore_env=r"^[+-]_comp_cmd_scp__path_esc=")
@@ -67,8 +65,10 @@ class TestRsync:
         else:
             raise Exception(f"Unsupported comparison result: {result}")
 
-    @pytest.mark.complete(f"rsync {LIVE_HOST}:spaces", sleep_after_tab=2)
-    def test_remote_path_with_spaces(self, completion):
+    def test_remote_path_with_spaces(self, bash):
+        assert_bash_exec(bash, "ssh() { echo 'spaces in filename.txt'; }")
+        completion = assert_complete(bash, "rsync remote_host:spaces")
+        assert_bash_exec(bash, "unset -f ssh")
         assert (
             completion == r"\ in\ filename.txt"
             or completion == r"\\\ in\\\ filename.txt"
--- a/test/t/test_scp.py
+++ b/test/t/test_scp.py
@@ -95,3 +95,7 @@ class TestScp:
     )
     def test_remote_path_with_failglob(self, completion):
         assert not completion
+
+    @pytest.mark.complete(f"scp {LIVE_HOST}:spaces", sleep_after_tab=2)
+    def test_remote_path_with_spaces(self, completion):
+        assert completion == r"\\\ in\\\ filename.txt"
--- a/test/t/test_scp.py
+++ b/test/t/test_scp.py
@@ -2,7 +2,7 @@ from itertools import chain
 
 import pytest
 
-from conftest import assert_bash_exec
+from conftest import assert_bash_exec, assert_complete
 
 LIVE_HOST = "bash_completion"
 
@@ -96,6 +96,8 @@ class TestScp:
     def test_remote_path_with_failglob(self, completion):
         assert not completion
 
-    @pytest.mark.complete(f"scp {LIVE_HOST}:spaces", sleep_after_tab=2)
-    def test_remote_path_with_spaces(self, live_pwd, completion):
+    def test_remote_path_with_spaces(self, bash):
+        assert_bash_exec(bash, "ssh() { echo 'spaces in filename.txt'; }")
+        completion = assert_complete(bash, "scp remote_host:spaces")
+        assert_bash_exec(bash, "unset -f ssh")
         assert completion == r"\\\ in\\\ filename.txt"
--- a/test/t/test_scp.py
+++ b/test/t/test_scp.py
@@ -97,5 +97,5 @@ class TestScp:
         assert not completion
 
     @pytest.mark.complete(f"scp {LIVE_HOST}:spaces", sleep_after_tab=2)
-    def test_remote_path_with_spaces(self, completion):
+    def test_remote_path_with_spaces(self, live_pwd, completion):
         assert completion == r"\\\ in\\\ filename.txt"


Rogach:
Sure, but that'll require an extra call to `rsync --version` and an ugly extra argument to _comp_xfunc_ssh_scp_remote_files.

akinomyoga:
Do you have a better solution?

completions/rsync
```diff
@@ -1,5 +1,36 @@
...
+    local IFS=.
+    local i ver1=($1) ver2=($2)
+    # fill empty fields in ver1 with zeros
+    for ((i=${#ver1[@]}; i<${#ver2[@]}; i++))
+    do
+        ver1[i]=0
+    done
+    for ((i=0; i<${#ver1[@]}; i++))
+    do
+        if [[ -z ${ver2[i]} ]]
```
akinomyoga, line 17: https://github.com/scop/bash-completion/blob/master/doc/styleguide.md#x-and--x-vs--n-x-and--z-x

completions/rsync
```diff
@@ -1,5 +1,36 @@
 # bash completion for rsync                                -*- shell-script -*-
 
+_rsync_vercomp() {
```
akinomyoga, line 3: _comp_cmd_rsync__vercomp

https://github.com/scop/bash-completion/blob/master/doc/api-and-naming.md#naming
We are still in the middle of migration, but the new functions would be better to follow the new naming.

completions/rsync
```diff
@@ -1,5 +1,36 @@
...
+    if [[ $1 == $2 ]]
+    then
```
akinomyoga, line 5: The format doesn't seem to be the one filtered by `shfmt`. Have you set up pre-commit? It is partly described in the second last item of [CONTRIBUTING.md](https://github.com/scop/bash-completion/blob/master/CONTRIBUTING.md).

completions/ssh
```diff
@@ -455,17 +457,30 @@ _comp_xfunc_ssh_scp_remote_files()
     if [[ ${1-} == -d ]]; then
```
akinomyoga, line 457: I think it would be cleaner to switch only the replacement text specified to `sed`. For example, we may define a variable in the following way and pass the value to `sed`.

```suggestion
    local escape_replacement='\\&'
    if [[ $overescape == overescape ]]; then
        escape_replacement='\\\\\\&'
    fi
    if [[ ${1-} == -d ]]; then
```

Rogach:
I've added a commit with the rsync version check.

However, if I were a maintainer, I'd much prefer the simpler solution of just working with the latest version. By the time bash-completion 2.12 is released and propagates to the distributions, it is likely that rsync will also be updated in those distributions. It would be strange if bash-completion were updated while rsync remained on an older version.

I've done a quick check on the distributions:

* Ubuntu 22.04 LTS has rsync 3.2.7 (after security update)
* Ubuntu 22.10 has rsync 3.2.7
* Debian stable has rsync 3.2.3
* Debian testing has rsync 3.2.7
* Fedora 36-38 has rsync 3.2.7

Arch Linux, of course, has the latest version. I was actually quite surprised to find out that Arch still uses bash-completion from 2020.

From a maintenance perspective, I believe it would be much better not to check the rsync version, as this significantly complicates the code.

akinomyoga (commented):
Thank you for updating the PR.

I have a question. The function `_comp_xfunc_ssh_scp_remote_files` seems to be also called from the completions for other commands ssh and sshfs. I guess we want to keep the original behavior for these commands. Or did these commands also change its interpretation of the escapes?

The others are cosmetic.

Rogach:
> I have a question. The function _comp_xfunc_ssh_scp_remote_files seems to be also called from the completions for other commands ssh and sshfs. I guess we want to keep the original behavior for these commands. Or did these commands also change its interpretation of the escapes?

It was my understanding that they accept both versions of these arguments (tested on my versions), but I'm not sure for how long they had this behavior. I'll invert the check so that we only change the behavior for the newer rsyncs.

akinomyoga:
> From a maintenance perspective, I believe it would be much better not to check the rsync version, as this significantly complicates the code.

Ubuntu LTS 18.04 will soon expire, but LTS 20.04 (with rsync 3.1.3) will still continue to be maintained until 2025-04. That's the idea of LTS. Also, we perform CI tests under under Debian 10 (w/ rsync 3.1.3), CentOS 7 (w/ 3.1.2), and Ubuntu 14.04 (w/ rsync ???). We actually support as old Bash as 4.2, (which is the reason that we have tests in Ubuntu 14.04).

akinomyoga:
> It was my understanding that they accept both versions of these arguments (tested on my versions), but I'm not sure for how long they had this behavior. I'll invert the check so that we only change the behavior for the newer rsyncs.

Thank you. I think that's safer for now.

completions/rsync
akinomyoga: If the description in CONTRIBUTING.md does not work, you can also check https://github.com/scop/bash-completion/pull/683#issuecomment-1010579492.

Rogach: That won't work for me, unfortunately, starting with docker probably. But I've made my best effort to install and use the tools (shfmt and shellcheck found errors).

completions/ssh
Rogach: Yes, that's indeed much better.

completions/rsync
```diff
@@ -1,5 +1,35 @@
 # bash completion for rsync                                -*- shell-script -*-
 
+_comp_cmd_rsync__vercomp()
+{
+    if [[ "$1" == "$2" ]]; then
```
akinomyoga, line 5:
```suggestion
    if [[ $1 == "$2" ]]; then
```

Actually, the `==` operator of the conditional command `[[ ]]` is not symmetric. The right-hand side of `==` in the conditional command `[[ ]]` is considered a glob pattern so needs to be quoted for the literal comparison. But the left-hand side doesn't need to be quoted.

completions/ssh
```diff
@@ -451,20 +454,25 @@ _comp_xfunc_ssh_scp_remote_files()
         path=$(ssh -o 'Batchmode yes' "$userhost" pwd 2>/dev/null)
     fi
 
+    local escape_replacement='\\\\\\&'
+    if [[ "$no_overescape" == "no-overescape" ]]; then
```
akinomyoga, line 458:
```suggestion
    if [[ $no_overescape == "no-overescape" ]]; then
```

akinomyoga (changes requested):
Thank you for addressing all the points. Two points are raised by shfmt of [the CI](https://github.com/scop/bash-completion/actions/runs/4554046992/jobs/8037505102).

completions/rsync
```diff
@@ -1,5 +1,35 @@
...
+    done
+    for ((i = 0; i < ${#ver1[@]}; i++)); do
+        if [[ ! ${ver1[i]} ]]; then
+            # replace empty strings ver1 with zeros
+            ver1[i]=0
+        fi
+        if [[ ! ${ver2[i]} ]]; then
+            # fill empty fields in ver2 with zeros
+            ver2[i]=0
+        fi
```
akinomyoga, line 22: Can we maybe use `${ver1[i]:-0}` when we compare the values (such as `10#${ver1[i]:-0}`)? I'd think it is simpler. Another way is to use `10#0${ver1[i]-}`.

completions/rsync
```diff
@@ -1,5 +1,35 @@
...
+{
+    if [[ "$1" == "$2" ]]; then
+        return 0
+    fi
+    local IFS=.
+    local i ver1=($1) ver2=($2)
+    # fill empty fields in ver1 with zeros
+    for ((i = ${#ver1[@]}; i < ${#ver2[@]}; i++)); do
+        ver1[i]=0
+    done
```
akinomyoga, line 13: I prefer determining the max number first to filling zeroes.
```suggestion
    local i ver1=($1) ver2=($2)
    local n=$((${#ver1[@]} >= ${#ver2[@]} ? ${#ver1[@]} : ${#ver2[@]}))
```

Or another way might be to loop with `while [[ ${ver1[i]-}${ver2[i]-} ]]; do ... ; ((i++)); done`

akinomyoga (commented):
These are additional suggestions/comments.

akinomyoga (approved):

completions/rsync
```diff
@@ -59,7 +78,15 @@ _rsync()
                     break
                 fi
             done
-            [[ $shell == ssh ]] && _comp_xfunc ssh scp_remote_files
+            if [[ $shell == ssh ]]; then
+                local rsync_version=$(rsync --version | sed -n 's/rsync  version \([0-9\.]*\).*/\1/; 1p')
```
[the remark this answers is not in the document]
akinomyoga: Thanks for quick responses and fixes!

akinomyoga (approved):

completions/ssh
```diff
@@ -434,8 +434,11 @@ _comp_cmd_scp__path_esc='[][(){}<>"'"'"',:;^&!$=?`\\|[:space:]]'
 # Complete remote files with ssh.  If the first arg is -d, complete on dirs
 # only.  Returns paths escaped with three backslashes.
 # shellcheck disable=SC2120
+# @param  $1 `no-overescape` if we should use 1 backslash for escaping instead of 3
```
scop, line 437: Even though I agree that the original escape seems a bit overdoing it, it is the traditional/standard one that's been there for a long time, and it is still the say ssh does it -- rsync is kind of a "guest" here. In that sense I'd prefer a more neutral variable name here.

Also, requiring a positional boolean-like parameter to have a specific value like here feels somewhat unconventional. Making it an option (i.e. dash prefixed, no value required as it's a boolean) would be more in line with rest of the codebase.

completions/rsync
```diff
@@ -59,7 +78,15 @@ _rsync()
                     break
                 fi
             done
-            [[ $shell == ssh ]] && _comp_xfunc ssh scp_remote_files
+            if [[ $shell == ssh ]]; then
+                local rsync_version=$(rsync --version | sed -n '1s/.*rsync *version \([0-9.]*\).*/\1/p')
+                _comp_cmd_rsync__vercomp "$rsync_version" "3.2.4"
+                if [[ $? == "2" ]]; then
```
scop, line 84: `$?` is numeric, better to use arithmetic evaluation.
```suggestion
                if (($? == 2)); then
```

completions/ssh
```diff
@@ -451,20 +454,25 @@ _comp_xfunc_ssh_scp_remote_files()
         path=$(ssh -o 'Batchmode yes' "$userhost" pwd 2>/dev/null)
     fi
 
+    local escape_replacement='\\\\\\&'
+    if [[ $no_overescape == "no-overescape" ]]; then
```
scop, line 458: Errors in `nounset mode`, unnecessary quoting. (A different fix likely appropriate if/when we tweak how the parameter is handled.)

```suggestion
    if [[ ${no_overescape-} == no-overescape ]]; then
```

scop (changes requested):
Left some comments.

Also, I think at least `_comp_cmd_rsync__vercomp` would deserve some tests.

completions/ssh
```diff
@@ -434,8 +434,11 @@ _comp_cmd_scp__path_esc='[][(){}<>"'"'"',:;^&!$=?`\\|[:space:]]'
 # Complete remote files with ssh.  If the first arg is -d, complete on dirs
 # only.  Returns paths escaped with three backslashes.
 # shellcheck disable=SC2120
+# @param  $1 `no-overescape` if we should use 1 backslash for escaping instead of 3
 _comp_xfunc_ssh_scp_remote_files()
 {
+    local no_overescape=$1
```
akinomyoga, line 440: Actually, with the current approach, the `set -o nounset` (`set -u`) should be avoided in this line.

completions/ssh
Rogach: > I think we should test the new behavior of scp_remote_files too.

How shall I approach that? It calls ssh inside to fetch completions, so simply calling the function as it is now won't work.

Rogach:
Applied changes according to comments.

Regarding scp_remote_files test - added tests that expect LIVE_HOST remote files to include file `spaces in filename.txt`.

completions/ssh
```diff
@@ -431,11 +431,29 @@ _sftp()
...
+# Complete remote files with ssh. Returns paths escaped with three backslashes
+# (unless -l option is provided).
+# Options:
+#     -d  Complete on dirs only.
+#     -l  Return paths escaped with one backslash instead of three.
 # shellcheck disable=SC2120
 _comp_xfunc_ssh_scp_remote_files()
 {
+    local dirs_only
+    local less_escaping
```
akinomyoga, line 443: When `shopt -s localvar_inherit` is set, the declaration without values will inherit the values in the outer scope. We need to initialize them with any values. One way is to initialize them with empty values and later reference them as `[[ $dirs_only ]]`.

```suggestion
    local dirs_only=
    local less_escaping=
```

Another way is to initialize them with `false` and later reference them as `"$dirs_only"`.

```suggestion
    local dirs_only=false
    local less_escaping=false
```

We are recently using the `true`/`false` pair instead of the non-empty/empty pair.

completions/ssh
```diff
@@ -451,20 +469,25 @@ _comp_xfunc_ssh_scp_remote_files()
...
     fi
 
+    local escape_replacement='\\\\\\&'
+    if [[ ${less_escaping-} ]]; then
+        escape_replacement='\\&'
+    fi
+
     local files
-    if [[ ${1-} == -d ]]; then
+    if [[ ${dirs_only-} ]]; then
```
akinomyoga, line 478: Once it is initialized, `$dirs_only` is fine. But as far as we use the approach with `true`/`false`, one should just say `"$dirs_only"`.

```suggestion
    if "$dirs_only"; then
```

Similar for `less_escaping`.

test/t/test_rsync.py
```diff
@@ -18,3 +22,50 @@ def test_3(self, completion):
...
+        _test_vercomp("1.2", "2.1", "<")
+        _test_vercomp("2.1", "1.2", ">")
+        _test_vercomp("5.6.7", "5.6.7", "=")
+        _test_vercomp("1.01.1", "1.1.1", "=")
+        _test_vercomp("1.1.1", "1.01.1", "=")
+        _test_vercomp("1", "1.0", "=")
+        _test_vercomp("1.0", "1", "=")
+        _test_vercomp("1.0.2.0", "1.0.2", "=")
+        _test_vercomp("1..0", "1.0", "=")
+        _test_vercomp("1.0", "1..0", "=")
```
akinomyoga, line 64: Can you use `@pytest.mark.parametrize`? I haven't tested it but something like this:

```suggestion
    @pytest.mark.parametrize(
        "ver1,ver2,result",
        [
            ("1", "1", "="),
            ("1", "2", "<"),
            ("2", "1", ">"),
            ("1.1", "1.2", "<"),
            ("1.2", "1.1", ">"),
            ("1.1", "1.1.1", "<"),
            ("1.1.1", "1.1", ">"),
            ("1.1.1", "1.1.1", "="),
            ("2.1", "2.2", "<"),
            ("3.0.4.10", "3.0.4.2", ">"),
            ("4.08", "4.08.01", "<"),
            ("3.2.1.9.8144", "3.2", ">"),
            ("3.2", "3.2.1.9.8144", "<"),
            ("1.2", "2.1", "<"),
            ("2.1", "1.2", ">"),
            ("5.6.7", "5.6.7", "="),
            ("1.01.1", "1.1.1", "="),
            ("1.1.1", "1.01.1", "="),
            ("1", "1.0", "="),
            ("1.0", "1", "="),
            ("1.0.2.0", "1.0.2", "="),
            ("1..0", "1.0", "="),
            ("1.0", "1..0", "="),
        ]
    )
    def test_vercomp(self, bash, ver1, ver2, result):
        output = assert_bash_exec(
            bash,
            f"_comp_cmd_rsync__vercomp {ver1} {ver2}; echo $?",
            want_output=True,
        ).strip()
        if result == "=":
            assert output == "0"
        elif result == ">":
            assert output == "1"
        elif result == "<":
            assert output == "2"
        else:
            raise Exception(f"Unsupported comparison result: {result}")
```

akinomyoga (commented):
Thank you for updating the PR. Here are the comments on the new part.

completions/ssh
Rogach: Do we need to escape `"$dirs_only"`, if we always initialize it? Seems like this would work as well:
```
if $dirs_only; then
```

completions/rsync
akinomyoga: @Rogach Thank you for removing the double quotation, but the current version of PR still uses the conditional command `[[ ... ]]`. Could you switch from `[[ $? == 2 ]]` to the arithmetic command `(($? == 2))`?

completions/ssh
akinomyoga: In case `IFS=abcdefghijklmnopqrstuvwxyz`

akinomyoga: Anyway, we are changing ourselves to quote similar cases in a context-independent manner.

Rogach: Makes sense. I'm not accustomed to thinking about all the crazy variations IFS that might happen :)

akinomyoga: Thanks!

jucor:
I got bitten by this problem today. Would it be possible to get this merged, please?
This is also referred in https://www.reddit.com/r/pop_os/comments/124jmp1/synchronizing_rsync_so_it_only_singlequotes/ 
Thanks a lot!

completions/ssh
jucor: @scop I realise this is the current change requested that's still blocking. However, now that the latest rsync has a bigger install base, the problem (that this PR would fix) is spreading. Would you consider merging it even without this test, to pull users out of their misery, please?

akinomyoga:
rebased. The test using LIVE_HOST is not fixed.

jucor:
👋  Hi ! Any chance to get this merged, please? @scop and @akinomyoga , it your approval is needed, please. Thanks.

Rogach:
I think I understand what he is talking about, but I can't help with that - it relates to the testing infrastructure, I don't know anything about that.

akinomyoga:
The new behavior of the completion needs to be tested in a proper way. That is the only thing.

The current test fails because of the unknown setup of `LIVE_HOST`. It is unclear how `LIVE_HOST` is supposed to be set up in the CI environment, so we cannot determine what is the proper way (including whether `LIVE_HOST` is a preferable way or we could use a differenet approach).

jucor:
> The current test fails because of the unknown setup of `LIVE_HOST`. It is unclear how `LIVE_HOST` is supposed to be set up in the CI environment, so we cannot determine what is the proper way (including whether `LIVE_HOST` is a preferable way or we could use a differenet approach).

Understood, thanks.
@Rogach , how do you set up `LIVE_HOST` when you run the tests? I imagine that's an environment variable containing the IP/name of a server that runs SSH so you can test the completion?

@akinomyoga If I'm correct with the above, I guess we need a test fixture that mocks a remote server, or something of the kind. How do you normally test the existing tab completion, please? Maybe we could use that.

akinomyoga:
The existing test that uses `LIVE_HOST` is skipped (through the fixture `live_pwd`). This implies that `LIVE_HOST` is not available on the CI.

```bash
../../../../test/t/test_scp.py::TestScp::test_basic PASSED               [ 56%]
../../../../test/t/test_scp.py::TestScp::test_basic_spaced_conf PASSED   [ 56%]
../../../../test/t/test_scp.py::TestScp::test_capital_f_without_space PASSED [ 56%]
../../../../test/t/test_scp.py::TestScp::test_live SKIPPED (Live hos...) [ 56%]
../../../../test/t/test_scp.py::TestScp::test_option_arg PASSED          [ 56%]
../../../../test/t/test_scp.py::TestScp::test_remote_path_with_nullglob PASSED [ 56%]
../../../../test/t/test_scp.py::TestScp::test_remote_path_with_failglob PASSED [ 56%]
```

The added test in `rsync` should also reference `live_pwd` for the dependency. The CI error can be avoided this way, but then the test will never be performed in the CI. That's not wanted. One way is to prepare a mock command as mentioned in https://github.com/scop/bash-completion/pull/910#discussion_r1155863026. You can define a shell function `ssh() { ... }` or `rsync() { ... }` at the beginning of the test to mimic the output of those commands. Then, one can `unset -f ssh` or `unset -f rsync` after the test.

jucor:
> The existing test that uses `LIVE_HOST` appears to be skipped (through the fixture `live_pwd`) in the CI. This implies that `LIVE_HOST` is not available on the CI.


Would it be fair to suggest that we split that specific problem with LIVE_HOST into a separate issue, as it requires a different skillset and was not introduced here? Yes it does need sorting, but since (again if I understand correctly) the LIVE_HOST was already used before, maybe it shouldn't block this PR, should it?

We could skip the new test like the old one was.

We could then merge this PR.

akinomyoga:
> @akinomyoga If I'm correct with the above, I guess we need a test fixture that mocks a remote server, or something of the kind. How do you normally test the existing tab completion, please? Maybe we could use that.

Existing tab completions are tested using real commands. We currently do not use any mock commands for tests. See #573.

jucor:
> Existing tab completions are tested using real commands. We currently do not use any mock commands for tests. See #573.

Nice, thanks for the pointer. So introducing a mock would be a completely new approach here, bit of a scope creep 😅

How do you currently (before this PR) test the tab completion for paths?

akinomyoga:
> > The existing test that uses `LIVE_HOST` appears to be skipped (through the fixture `live_pwd`) in the CI. This implies that `LIVE_HOST` is not available on the CI.
>
> Would it be fair to suggest that we split that specific problem with LIVE_HOST into a separate issue, as it requires a different skillset and was not introduced here? Yes it does need sorting, but since (again if I understand correctly) the LIVE_HOST was already used before, maybe it shouldn't block this PR, should it?
>
> We could skip the new test like the old one was.

We can skip the new test, but that is not the problem. As I mentioned, the new behavior needs to be tested in the CI. The existing test using `LIVE_HOST` seems just an extra test case; even if the existing test is skipped, there are other tests independent of `LIVE_HOST`. On the other hand, this PR adds only one test, which depends on `LIVE_HOST`, but it adds no other tests that do not require `LIVE_HOST`. It's not a problem that this the added test with `LIVE_HOST` is skipped, but then we need to add other tests.

jucor:
OK, now I see where you're coming from, thanks for taking the time to explain. Let me check what's the current tests are for the tab complete for paths.

Do we agree that, as long as this PR does the same level of tests for tab complete of paths as there was for the old behaviour, this can be merged?

akinomyoga:
> Do we agree that, as long as this PR does the same level of tests for tab complete of paths as there was for the old behaviour, this can be merged?

In general, new features and behaviors should come with new tests. Speficailly for the present case, we haven't tested anything about the path completion of `rsync`/`ssh` before (probably because we didn't realize the path specification could be a problem). Then, you are suggesting adding no tests effectively, which is not useful. Why don't we add tests that do not depend on `LIVE_HOST`?

akinomyoga:
@Rogach Could you convert the test or add a new one so that it doesn't depend on `LIVE_HOST`?

jucor:
> In general, new features and behaviors should come with new tests. Speficailly for the present case, we haven't tested anything about the path completion of `rsync`/`ssh` before (probably because we didn't realize the path specification could be a problem). Then, you are suggesting adding no tests effectively, which is not useful. Why don't we add tests that do not depend on `LIVE_HOST`?

I hear you.
If we can code those new tests without having to mock a full ssh, we could.

I'm trying to be pragmatic and avoid users doing the kind of tricks referred there: https://www.reddit.com/r/pop_os/comments/124jmp1/comment/js1mdp8/

I wholeheartedly agree that more code should mean more tests.
I am also aware that adding new mocks and fixtures is possibly not within @Rogach 's wheelhouse, as illustrated by the year since this is open. Given that it was acceptable for the tab completion for paths so far to not have tests, and the particularities of this request (the hoops users are jumping through to get this feature, the broken compatibility of the status quo with rsync, the time this problem has been open, and the precedent for no test on tab completion for paths), I'm hoping that you will considerate merging with the test skipped. 

Of course, it's your project, your call, I can only advocate :)

Rogach:
Here you go, I've figured out how to make a test with a mock function.

Rogach:
>  1: T1 Title exceeds max length (99>72): "fix(rsync,ssh): do not depend on LIVE_HOST when testing remote files with spaces, use mocks instead"

I'm open to suggestions for a better commit title.

How you are supposed to fit anything useful into 72 characters, anyway?

akinomyoga:
> I'm open to suggestions for a better commit title.

Something like this? I'd use `test` for the type of commit.

```
test(rsync,ssh): test remote filenames with spaces using mock commands
```
 
> How you are supposed to fit anything useful into 72 characters, anyway?

By giving up something that is not so important. If you want to include details, you can do it in the second (or a later) paragraph (separated by an empty line).

test/t/test_rsync.py
```diff
@@ -18,3 +19,56 @@ def test_3(self, completion):
...
+            assert output == "1"
+        elif result == "<":
+            assert output == "2"
+        else:
+            raise Exception(f"Unsupported comparison result: {result}")
+
+    def test_remote_path_with_spaces(self, bash):
+        assert_bash_exec(bash, """ssh() { echo "spaces in filename.txt"; }""")
+        completion = assert_complete(bash, f"rsync remote_host:spaces")
+        assert_bash_exec(bash, """unset -f ssh""")
```
akinomyoga, line 70:
```suggestion
        assert_bash_exec(bash, 'ssh() { echo "spaces in filename.txt"; }')
        completion = assert_complete(bash, "rsync remote_host:spaces")
        assert_bash_exec(bash, "unset -f ssh")
```

test/t/test_scp.py
```diff
@@ -95,3 +95,9 @@ def test_remote_path_with_nullglob(self, completion):
     )
     def test_remote_path_with_failglob(self, completion):
         assert not completion
+
+    def test_remote_path_with_spaces(self, bash):
+        assert_bash_exec(bash, """ssh() { echo "spaces in filename.txt"; }""")
+        completion = assert_complete(bash, f"scp remote_host:spaces")
+        assert_bash_exec(bash, """unset -f ssh""")
```
akinomyoga, line 102:
```suggestion
        assert_bash_exec(bash, 'ssh() { echo "spaces in filename.txt"; }')
        completion = assert_complete(bash, "scp remote_host:spaces")
        assert_bash_exec(bash, "unset -f ssh")
```

akinomyoga (commented):
In addition to the fix for CI, I'd suggest changing the quoting of the string literal. The current quoted string is not so special that it requires triple quotations., and also, we haven't used triple quotations other than docstrings.

completions/ssh
akinomyoga: @scop To summarize the conversation, the initial test followed the `LIVE_HOST` approach that one of the `scp` tests used. The problem was that it was non-trivial how to properly set the host for the continuous integration (CI). However, the existing test case turned out to be skipped in the CI. We decided to use an approach that does not require `LIVE_HOST`: the one using mock functions that reproduces the fixed output of the command `ssh` or `rsync`.

akinomyoga (approved):

akinomyoga:
@yedayak Could you also quickly review this PR if there aren't obvious problems you notice or points to improve?

yedayak (approved):

jucor:
Woot woot, thanks everyone!

Issue #552, opened 2021-07-05
_comp_delimited loses completion on ambiguous choices


scop:
## To reproduce

Make sure `show-all-if-ambigous` is *off* for readline. Seems this doesn't happen with it on.

https://github.com/scop/bash-completion/pull/532#issuecomment-873675676

## Versions (please complete the following information)

- [ ] Operating system name/distribution and version: Ubuntu 20.04
- [ ] bash version, `echo "$BASH_VERSION"`: 5.0.17(1)-release
- [ ] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: current master

PR #911, opened 2023-03-29
feat(rsync): add completion for --info values

Rogach:


akinomyoga (changes requested):
Thank you for your contribution.

`--info` seems to accept a list of names separated by commas. bash-completion has a utility function `_comp_delimited` which can be used to fill such an option argument. Could you consider using `_comp_delimited`? You can search in the `completions` directory to see the actual examples of using `comp_delimited`.

----

We basically want to avoid hardcoding the available values (as described in the second paragraph of [`CONTRIBUTING.md`](https://github.com/scop/bash-completion/blob/master/CONTRIBUTING.md)). The `rsync` command seems to provide the list with `rsync --info=help`, but the output doesn't seem to be machine-friendly.

```bash
$ rsync --info=help
Use OPT or OPT1 for level 1 output, OPT2 for level 2, etc.; OPT0 silences.

BACKUP     Mention files backed up
COPY       Mention files copied locally on the receiving side
DEL        Mention deletions on the receiving side
FLIST      Mention file-list receiving/sending (levels 1-2)
MISC       Mention miscellaneous information (levels 1-2)
MOUNT      Mention mounts that were found or skipped
NAME       Mention 1) updated file/dir names, 2) unchanged names
NONREG     Mention skipped non-regular files (default 1, 0 disables)
PROGRESS   Mention 1) per-file progress or 2) total transfer progress
REMOVE     Mention files removed on the sending side
SKIP       Mention files skipped due to transfer overrides (levels 1-2)
STATS      Mention statistics at end of run (levels 1-3)
SYMSAFE    Mention symlinks that are unsafe

ALL        Set all --info options (e.g. all4)
NONE       Silence all --info options (same as all0)
HELP       Output this help message

Options added at each level of verbosity:
0) NONREG
1) COPY,DEL,FLIST,MISC,NAME,STATS,SYMSAFE
2) BACKUP,MISC2,MOUNT,NAME2,REMOVE,SKIP
```

Maybe the current hardcoded version is fine. Do you have any thoughts?

akinomyoga:
Thank you.

> I tried doing that, but there is a problem:
> 
> ```
> $ rsync --info=progress2,copy,del # fine up to here
> $ rsync --info=progress2,copy,del,n<TAB>
> $ rsync --info=n
> ```
> 
> Seems that if a later argument shares a prefix with other option then it deletes all the arguments before it. Maybe I'm using _comp_delimited incorrectly? Here's my diff:

This is a known issue with `bind 'set show-all-if-ambiguous off'` (Bash default) reported in #552. Hmm, this hasn't yet been fixed. We can leave this issue for this PR. If you would like to solve it, a separate PR is welcome.

> I'd vote for hardcoding. It's strictly an improvement, because previously we were not completing these arguments at all.

Thank you for your thoughts. I agree.

> Of course it's possible to parse `rsync --info=help` output, but it's quite brittle (especially those `levels 1-3` and `1)` `2)`).

Yeah, that's the tricky point.

akinomyoga:
By the way, the help output also mentions `all4` and `all0` in the description. Also, `opt0` for any `opt` seems to be able to be used to suppress the messages (though I haven't tried it). Suffixing `{,0}` to each name in the `-W` argument will produce them via brace expansions.

I think `flist` is also available in addition to `flist1` and `flist2`.

The names with explicit levels can also be contracted e.g. as `flist{,0,1,2}`.



> Seems that if a later argument shares a prefix with other option then it deletes all the arguments before it. Maybe I'm using _comp_delimited incorrectly? Here's my diff:
> 
> ```diff
> +++ b/completions/rsync
> @@ -32,13 +32,12 @@ _rsync()
>              return
>              ;;
>          --info)
> -            compopt +o nospace
> -            COMPREPLY=($(compgen -W '
> +            _comp_delimited , -W '
> ```

You can keep `compopt +o nospace` to suppress inserting a space so that the user can choose a space or a comma for the next character.

scop:
Thanks!

PR #912, opened 2023-03-29
refactor: rename `_quote_readline_by_ref => _comp_quote_compgen`

akinomyoga:
- Also, `quote_readline` is deprecated.
- `_comp_quote_compgen` always stores the result in `ret` (as in the second paragraph in [§ General API conventions](https://github.com/scop/bash-completion/blob/master/doc/api-and-naming.md#general-api-conventions)).

After this is merged, I'll submit the refactoring mentioned in https://github.com/scop/bash-completion/pull/909#issuecomment-1488328631.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -135,14 +135,6 @@ _comp_quote()
     ret=\'${1//\'/\'\\\'\'}\'
 }
 
-# @see _comp_quote_compgen()
-quote_readline()
-{
-    local ret
-    _comp_quote_compgen "$1"
-    printf %s "$ret"
-} # quote_readline()
-
 # shellcheck disable=SC1003
 _comp_dequote__initialize()
 {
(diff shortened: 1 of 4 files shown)


akinomyoga:
Shellcheck in CI is complaining about the assignment to the variable `ret`.

```
shellcheck...............................................................Failed
- hook id: shellcheck
- exit code: 1

bash_completion:636:9: warning: Variable was used as an array but is now assigned a string. [SC2[17](https://github.com/scop/bash-completion/actions/runs/4558856248/jobs/8042189130?pr=912#step:7:18)8]
```

The line is a part of the shell function `_comp_quote_compgen`, which sets the result to the variable `ret` to return the result to the caller. The function `_comp_quote_compgen` is only called from two places in the codebase, and `ret` is only used as a scalar variable there. I suspect Shellcheck considers that `_comp_quote_compgen` tries to set a global variable `ret`, which is not the case.

scop (approved):
Looks good to me, thanks!

Only thing I'm slightly wondering is if we should leave some tests in for the backwards compatibility wrappers. But I don't remember offhand us doing that for any other renamed things, so either way is fine with me, leaving up to you to make the call (approved either way).

akinomyoga:
Thank you for the review. I realized that we didn't have specific tests for `quote_readline`, so I have added tests and parametrized them for the old name `quote_readline` 8490389d40e88fb98cd0ceceefec13e6b999c66c. I'll later merge the PR.

PR #913, opened 2023-03-30
fix(_comp_delimited): prepend prefix to all compreply args

### LINKED ISSUE #552: _comp_delimited loses completion on ambiguous choices

Rogach:
Fixes #552.

Previously prefix was only prepended if COMPREPLY only contained one argument - this commit fixes it so prefix is prepended to all arguments.

The problem existed since 4138714c, in the very first version of `_comp_delimited` function. I'm not sure why the original only prepended the prefix back when there was only one completion, so I'm worried I'm missing something.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -895,6 +895,13 @@ _comp_delimited()
         _comp_compgen COMPREPLY "$@" -- "${cur##*"$delimiter"}"
     fi
 
+    # It would seem that in some specific cases we could avoid adding the
+    # prefix to all completions, thereby making the list of suggestions
+    # cleaner, and only adding it when there's exactly one completion.
+    # The cases where this opportunity has been observed involve having
+    # `show-all-if-ambiguous` on, but even that has cases where it fails
+    # and the last separator including everything before it is lost.
+    # https://github.com/scop/bash-completion/pull/913#issuecomment-1490140309
     local i
     for i in "${!COMPREPLY[@]}"; do
         COMPREPLY[i]="$prefix${COMPREPLY[i]}"
--- a/bash_completion
+++ b/bash_completion
@@ -895,7 +895,10 @@ _comp_delimited()
         _comp_compgen COMPREPLY "$@" -- "${cur##*"$delimiter"}"
     fi
 
-    ((${#COMPREPLY[@]})) && COMPREPLY=("${COMPREPLY[@]/#/"$prefix"}")
+    local i
+    for i in "${!COMPREPLY[@]}"; do
+        COMPREPLY[i]="$prefix${COMPREPLY[i]}"
+    done
 
     [[ $delimiter != : ]] || __ltrim_colon_completions "$cur"
 }
--- a/bash_completion
+++ b/bash_completion
@@ -895,7 +895,8 @@ _comp_delimited()
         _comp_compgen COMPREPLY "$@" -- "${cur##*"$delimiter"}"
     fi
 
-    ((${#COMPREPLY[@]} == 1)) && COMPREPLY=("$prefix$COMPREPLY")
+    ((${#COMPREPLY[@]})) && COMPREPLY=("${COMPREPLY[@]/#/"$prefix"}")
+
     [[ $delimiter != : ]] || __ltrim_colon_completions "$cur"
 }
 
--- a/test/t/test_pgrep.py
+++ b/test/t/test_pgrep.py
@@ -31,4 +31,3 @@ class TestPgrep:
     )
     def test_nslist_after_comma(self, completion):
         assert completion
-        assert not any("," in x for x in completion)
--- a/test/t/test_ssh_keygen.py
+++ b/test/t/test_ssh_keygen.py
@@ -27,13 +27,11 @@ class TestSshKeygen:
     @pytest.mark.complete("ssh-keygen -s foo_key -n foo,")
     def test_usernames_for_n(self, completion):
         assert completion
-        assert not any("," in x for x in completion)
         # TODO check that these are usernames
 
     @pytest.mark.complete("ssh-keygen -s foo_key -h -n foo,")
     def test_host_for_h_n(self, completion):
         assert completion
-        assert not any("," in x for x in completion)
         # TODO check that these are hostnames
 
     @pytest.mark.complete("ssh-keygen -Y foo -n ")
--- a/test/t/test_tox.py
+++ b/test/t/test_tox.py
@@ -12,7 +12,7 @@ class TestTox:
 
     @pytest.mark.complete("tox -e foo,", cwd="tox")
     def test_3(self, completion):
-        assert all(x in completion for x in "py37 ALL".split())
+        assert all("foo," + x in completion for x in "py37 ALL".split())
 
     @pytest.mark.complete("tox -e foo -- ", cwd="tox")
     def test_default_after_dashdash(self, completion):
--- a/test/t/test_tshark.py
+++ b/test/t/test_tshark.py
@@ -14,7 +14,7 @@ class TestTshark:
     @pytest.mark.complete("tshark -O foo,htt", require_cmd=True)
     def test_3(self, completion):
         # p: one completion only; http: e.g. http and http2
-        assert completion == "p" or "http" in completion
+        assert completion == "p" or "foo,http" in completion
 
     @pytest.mark.complete("tshark -o tcp", require_cmd=True)
     def test_4(self, completion):
--- a/test/t/unit/Makefile.am
+++ b/test/t/unit/Makefile.am
@@ -2,6 +2,7 @@ EXTRA_DIST = \
 	test_unit_command_offset.py \
 	test_unit_compgen.py \
 	test_unit_count_args.py \
+	test_unit_delimited.py \
 	test_unit_deprecate_func.py \
 	test_unit_dequote.py \
 	test_unit_expand.py \
--- a/test/t/unit/test_unit_delimited.py
+++ b/test/t/unit/test_unit_delimited.py
@@ -1,0 +1,42 @@
+import pytest
+
+from conftest import assert_bash_exec
+
+
+@pytest.mark.bashcomp(cmd=None)
+class TestUnitDelimited:
+    @pytest.fixture(scope="class")
+    def functions(self, request, bash):
+        assert_bash_exec(
+            bash,
+            "_comp_cmd_test_delim() {"
+            "  local cur prev words cword comp_args;"
+            "  _comp_get_words cur;"
+            "  _comp_delimited , -W 'alpha beta bravo';"
+            "};"
+            "complete -F _comp_cmd_test_delim test_delim",
+        )
+
+    @pytest.mark.complete("test_delim --opt=a")
+    def test_1(self, functions, completion):
+        assert completion == ["lpha"]
+
+    @pytest.mark.complete("test_delim --opt=b")
+    def test_2(self, functions, completion):
+        assert completion == ["beta", "bravo"]
+
+    @pytest.mark.complete("test_delim --opt=alpha,b")
+    def test_3(self, functions, completion):
+        assert completion == ["alpha,beta", "alpha,bravo"]
+
+    @pytest.mark.complete("test_delim --opt=alpha,be")
+    def test_4(self, functions, completion):
+        assert completion == ["ta"]
+
+    @pytest.mark.complete("test_delim --opt=beta,a")
+    def test_5(self, functions, completion):
+        assert completion == ["lpha"]
+
+    @pytest.mark.complete("test_delim --opt=c")
+    def test_6(self, functions, completion):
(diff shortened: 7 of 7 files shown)


bash_completion
```diff
@@ -833,7 +833,10 @@ _comp_delimited()
         COMPREPLY=($(compgen "$@" -- "${cur##*"$delimiter"}"))
     fi
 
-    ((${#COMPREPLY[@]} == 1)) && COMPREPLY=(${COMPREPLY/#/$prefix})
+    for ((i = 0; i < ${#COMPREPLY[@]}; i++)); do
+        COMPREPLY[i]=${COMPREPLY[i]/#/$prefix}
+    done
+
```
akinomyoga, line 839: Actually, this can be written in a compact form:

```suggestion
    COMPREPLY=("${COMPREPLY[@]/#/"$prefix"}")
```

Also, `$prefix` needs to be quoted as `"$prefix"` here because otherwise the special character `&` would be interpreted here in Bash 5.2 with `shopt -s patsub_replacement`. For example, if you use bash-5.2, you can try this in your terminal:

```bash
bash-5.2$ p='1&2' w=3
bash-5.2$ echo "${w/#/$p}"
123
bash-5.2$ echo "${w/#/"$p"}"
1&23
```

Rogach:
Updated with suggested changes.

Regarding "&" - not sure if it's interesting, but for me tab-completion just stops working if I add "&" as part of previous delimited argument:

```
$ ssh-keyscan -t dsa,e<TAB> # nothing happens
```

akinomyoga (changes requested):
Can you add tests in the subdirectory [`test/t/unit`](https://github.com/scop/bash-completion/tree/master/test/t/unit) with the filename `test_unit_delimited.py`? You can reference other files such as [`test_unit_ip_addresses.py`](https://github.com/scop/bash-completion/blob/master/test/t/unit/test_unit_ip_addresses.py) for how to make tests. For the implementation of a mock completion function, you may look at `_echo` in https://github.com/scop/bash-completion/pull/913#issuecomment-1490140309. To run the test, you can use the following command:

```bash
$ pytest test/t/unit/test_unit_delimited.py
```

The command name `pytest` might be `pytest-3` depending on the installation.

Rogach:
Added tests.

<hr>

> Where did you add &? & may be specified in the command line, such as 

It's late here, I'm getting tired :)

I forgot to quote the & - I had a setup like this:

```
$ ssh-keyscan ds&,e<TAB><TAB>
```

and obviously nothing was happening, because I was tring to complete `,e`, everything before was cut off.

However if I quote the "&" then I see no difference between quotingthe prefix and not quoting it:

```
$ ssh-keyscan -t "ds&",e<TAB><TAB>
"ds&",ecdsa    "ds&",ed25519  
```

There is the same result, even with `COMPREPLY=("${COMPREPLY[@]/#/$prefix}")`.

akinomyoga (approved):
Thank you! Looks perfect!

Rogach:
I finally understood what was my problem with "&" - you were talking about bash 5.2 and pathsub_replacement, while I have older bash (5.1) without that option.

akinomyoga:
> It's late here, I'm getting tired :)

Thank you for your hard work!

> ```bash
> $ ssh-keyscan -t "ds&",e<TAB><TAB>
> "ds&",ecdsa    "ds&",ed25519  
> ```

As far as I test, the above behavior seems to be different depending on the quoting of `$prefix`.

> I finally understood what was my problem with "&" - you were talking about bash 5.2 and pathsub_replacement, while I have older bash (5.1) without that option.

Ah, I see. Sorry for confusing you.

akinomyoga:
CI pre-commit has [an error](https://github.com/scop/bash-completion/actions/runs/4564416747/jobs/8054289514?pr=913). `black` suggests the following change:

```diff
diff --git a/test/t/unit/test_unit_delimited.py b/test/t/unit/test_unit_delimited.py
index 5eb74bdb..249dc1b6 100644
--- a/test/t/unit/test_unit_delimited.py
+++ b/test/t/unit/test_unit_delimited.py
@@ -2,6 +2,7 @@ import pytest

 from conftest import assert_bash_exec

+
 @pytest.mark.bashcomp(cmd=None)
 class TestUnitDelimited:
     @pytest.fixture(scope="class")
```

Rogach:
Failing tests weren't too difficult, these are all resulting from changes in completion output. However for ssh-keygen and pgrep tests I'm not sure how to actually fix the tests - currently I just removed the outdated asserions, leaving only `assert completion`.

Rogach:
Two tests are failing, but I don't understand why. They are failing only in one distribution (ubuntu14), one timeout (in seemingly innocious test), and one environment modification (where there shouldn't be any).

akinomyoga:
The second error environment modification is probably caused by the first error. There are also many errors in [`centos7`](https://github.com/scop/bash-completion/actions/runs/4565005163/jobs/8095948844) before it is canceled.

```
../../../test/t/test_fio.py::TestFio::test_3 FAILED                      [ 19%]
../../../test/t/test_fio.py::TestFio::test_cmdhelp_all PASSED            [ 19%]
../../../test/t/test_fio.py::TestFio::test_enghelp PASSED                [ 19%]
../../../test/t/test_fio.py::TestFio::test_cmdhelp_boolean PASSED        [ 19%]
../../../test/t/test_fio.py::TestFio::test_cmdhelp_valid_values PASSED   [ 19%]
../../../test/t/test_fio.py::TestFio::test_cmdhelp_nonexistent PASSED    [ 19%]
../../../test/t/test_fio.py::TestFio::test_crctest PASSED                [ 19%]
../../../test/t/test_fio.py::TestFio::test_debug FAILED                  [ 19%]

...

../../../test/t/test_man.py::TestMan::test_delimited_first FAILED        [ 39%]
../../../test/t/test_man.py::TestMan::test_delimited_after_delimiter PASSED [ 39%]
../../../test/t/test_man.py::TestMan::test_delimited_later FAILED        [ 39%]
../../../test/t/test_man.py::TestMan::test_delimited_deduplication ERROR [ 39%]
../../../test/t/test_man.py::TestMan::test_zstd_arbitrary_sectsuffix PASSED [ 39%]
../../../test/t/test_man.py::TestMan::test_zstd_arbitrary_sectsuffix ERROR [ 39%]

...

../../../test/t/test_ps.py::TestPs::test_7 FAILED                        [ 52%]

...

../../../test/t/test_ss.py::TestSs::test_3 FAILED                        [ 63%]

...

../../../test/t/test_ssh_keygen.py::TestSshKeygen::test_usernames_for_n FAILED [ 64%]
../../../test/t/test_ssh_keygen.py::TestSshKeygen::test_host_for_h_n FAILED [ 64%]

...

../../../test/t/test_tox.py::TestTox::test_2 FAILED                      [ 69%]
../../../test/t/test_tox.py::TestTox::test_3 FAILED                      [ 69%]

...

../../../test/t/test_tshark.py::TestTshark::test_3 FAILED                [ 70%]
```



I reran the CI test twice, but the same error happened in both reruns. I also ran the tests locally, and the error was reproduced with bash 4.2. The following is the error message:

```
FAILED test/t/test_man.py::TestMan::test_delimited_first - assert not <CompletionResult ['\x08""1']>
FAILED test/t/test_man.py::TestMan::test_delimited_later - assert not <CompletionResult ['\x08"1:"2']>
ERROR test/t/test_man.py::TestMan::test_delimited_deduplication - AssertionError: Unexpected output: [bash: COMPREPLY[@]: unbound variable
ERROR test/t/test_man.py::TestMan::test_zstd_arbitrary_sectsuffix - AssertionError: Environment should not be modified
```

The third line must be the cause of the error. Ah, OK. In Bash 4.2, `${arr[@]}` fails when `set -u` is set and there are no elements in the array even if the array exists. You need to check if it has any elements, such as

```bash
((${#COMPREPLY[@]})) && COMPREPLY=("${COMPREPLY[@]/#/"$prefix"}")
```

bash_completion
```diff
@@ -833,7 +833,8 @@ _comp_delimited()
         COMPREPLY=($(compgen "$@" -- "${cur##*"$delimiter"}"))
     fi
 
-    ((${#COMPREPLY[@]} == 1)) && COMPREPLY=(${COMPREPLY/#/$prefix})
+    COMPREPLY=("${COMPREPLY[@]/#/"$prefix"}")
```
akinomyoga, line 836:
```suggestion
    ((${#COMPREPLY[@]})) && COMPREPLY=("${COMPREPLY[@]/#/"$prefix"}")
```

Can you also add a test case where no completions are generated in `test_unit_delimited.py`?

akinomyoga (changes requested):

akinomyoga:
Can you add `test_unit_delimited.py` in `test/t/unit/Makefile.am`?

Rogach:
Done.

Rogach:
Something broke on centos7, but I don't know how to check it :(

akinomyoga:
I realized that the double quotations `"$prefix"` inside `"${var/#/...}"` is not processed in Bash 4.2 (while those in unquoted `${var/#/...}` are processed). I added a fix d84f1eca71ffaa6bb7638bf5cbffae256a2b9878. We finally go back to something similar to your first version (using the for loop to process each element).

Rogach:
CI seems to pass fine now. Thank you!

bash_completion
```diff
@@ -895,7 +895,11 @@ _comp_delimited()
         _comp_compgen COMPREPLY "$@" -- "${cur##*"$delimiter"}"
     fi
 
-    ((${#COMPREPLY[@]} == 1)) && COMPREPLY=("$prefix$COMPREPLY")
+    local i
+    for i in "${!COMPREPLY[@]}"; do
+        COMPREPLY[i]="$prefix${COMPREPLY[i]}"
+    done
```
scop, line 901: While this certainly seems to work, it's also somewhat unfortunate, because displaying the completions with the prefix in place makes them harder to read, and with `show-all-if-ambiguous` on it appears we wouldn't have to do it (I'm biased, I have it on). So I'm contemplating something like:

```suggestion
    # If `show-all-if-ambiguous` is off, existing completions preceding the
    # comma and the comma itself are lost if there are multiple completions
    # offered for the thing after the prefix. This does not seem to happen
    # with it on, so avoid repeating the prefix in that case for readability.
    # https://github.com/scop/bash-completion/pull/532#issuecomment-873675676
    if _comp_readline_variable_on show-all-if-ambiguous; then
        ((${#COMPREPLY[@]} == 1)) && COMPREPLY=("$prefix$COMPREPLY")
    else
        local i
        for i in "${!COMPREPLY[@]}"; do
            COMPREPLY[i]="$prefix${COMPREPLY[i]}"
        done
    fi
```

I suppose this would require some addressing/complexity in tests, too.

Thoughts?

akinomyoga: As we have discussed in the past in #546, the current test framework totally relies on `set show-all-if-ambiguous on`, so I'm afraid there is no way to test the behavior for the case `set show-all-if-ambiguous off`.

Rogach: @akinomyoga showed that `((${#COMPREPLY[@]} == 1)) && COMPREPLY=("$prefix$COMPREPLY")` isn't guaranteed to work even if `show-all-if-unmodified` is `on` (https://github.com/scop/bash-completion/pull/913#issuecomment-1490140309). Copying the snippet here:
```
$ bind 'set show-all-if-unmodified on'
$ echo x,y,he[TAB]
hello1  hello2  hello3
$ echo x,y,he         # The original word is preserved
$ echo x,y,h[TAB]
hello1  hello2  hello3
$ echo hello          # The original word is lost
```

I'm not an expert here, but it feels very brittle to be relying on some obscure undocumented interaction between options, that might break easily down the road (as showcased above).

akinomyoga: Ah, yes. I was also about to write it. I tried the change suggested by @scop with `set show-all-if-ambiguous` and confirmed that it still fails. Also, I was checking if it is possible to detect the insertion by looking at `COMP_TYPE`, but there doesn't seem to be a reliable way.

scop: Ok. Then let's just go with the implementation here that is known to work in all cases, you guys have spent more than enough time here already. Thanks!

scop: Added a note in e3840d0f9e5ed0d9c16bfb4cd1d9d63198292886 as a reminder to take stuff into account if someone is tempted to revisit this later.

scop (approved):

PR #914, opened 2023-04-03
fix: quote boolean command names such as `if "$boolean_variable"; then`

akinomyoga:
I have created this branch after https://github.com/scop/bash-completion/pull/891#issuecomment-1445250871 but left it for about one month. Now there was a related discussion in https://github.com/scop/bash-completion/pull/910#discussion_r1155867722, so I decided to submit it here.

**edit**: This branch also includes other fixes: 7b6fb48894c79a3c7c57e9a7f5cb14bf7b2ff198, e731fc4d846423e330f199f786f8c054f0e01b99, 7971f50f569a2e2af7133b9f3459f00a89c22895

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -868,7 +868,7 @@ _comp_delimited()
     fi
     [[ $cur == *$delimiter* ]] && prefix=${cur%"$delimiter"*}$delimiter
 
-    if $deduplicate; then
+    if "$deduplicate"; then
         # We could construct a -X pattern to feed to compgen, but that'd
         # conflict with possibly already set -X in $@, as well as have
         # glob char escaping issues to deal with. Do removals by hand instead.
--- a/bash_completion
+++ b/bash_completion
@@ -1930,7 +1930,7 @@ _included_ssh_config_files()
 _known_hosts_real()
 {
     local configfile="" flag prefix=""
-    local cur suffix="" aliases="" i host ipv4="" ipv6=""
+    local cur suffix="" aliases="" i host ipv4=false ipv6=false
     local -a kh tmpkh=() khd=() config=()
 
     # TODO remove trailing %foo from entries
@@ -1948,8 +1948,8 @@ _known_hosts_real()
                 configfile=$OPTARG
                 ;;
             p) prefix=$OPTARG ;;
-            4) ipv4=1 ;;
-            6) ipv6=1 ;;
+            4) ipv4=true ;;
+            6) ipv6=true ;;
             *)
                 echo "bash_completion: $FUNCNAME: usage error" >&2
                 return 1
@@ -2122,13 +2122,13 @@ _known_hosts_real()
     $reset
 
     if ((${#COMPREPLY[@]})); then
-        if [[ $ipv4 ]]; then
+        if "$ipv4"; then
             COMPREPLY=("${COMPREPLY[@]/*:*$suffix/}")
         fi
-        if [[ $ipv6 ]]; then
+        if "$ipv6"; then
             COMPREPLY=("${COMPREPLY[@]/+([0-9]).+([0-9]).+([0-9]).+([0-9])$suffix/}")
         fi
-        if [[ $ipv4 || $ipv6 ]]; then
+        if "$ipv4" || "$ipv6"; then
             for i in "${!COMPREPLY[@]}"; do
                 [[ ${COMPREPLY[i]} ]] || unset -v 'COMPREPLY[i]'
             done
--- a/completions/lzip
+++ b/completions/lzip
@@ -35,7 +35,7 @@ _lzip()
         return
     fi
 
-    if $decompress; then
+    if "$decompress"; then
         _filedir lz
         return
     fi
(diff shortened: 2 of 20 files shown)


akinomyoga:
I decided to separate the change `$split && return` to `"$split" && return` for easier review.

completions/lzip
```diff
@@ -14,14 +14,14 @@ _lzip()
             --volume-size | --data-size | -${noargopts}@([bmsSB]|[hV]*))
             return
             ;;
-        --decompress-${noargopts}d)
+        --decompress-d)
```
scop, line 17: This reads weird to me, but then again so did the original. I suppose it should be:
```suggestion
        --decompress | -${noargopts}d)
```

Similarly for `--threads` and `--output` below.

scop (changes requested):

completions/strace
```diff
@@ -40,9 +40,9 @@ _strace()
                             while read -r define syscall rest; do
                                 [[ $define == "#define" &&
                                     $syscall =~ ^__NR_(.+) ]] &&
-                                    syscalls[${BASH_REMATCH[1]}]=1
+                                    syscalls[${BASH_REMATCH[1]}]=true
                             done 2>/dev/null </usr/include/asm/unistd.h
-                            if [[ ! $syscalls ]]; then
+                            if ((${#syscalls[@]} == 0)); then
```
scop, line 45: Cherry-picked as 223aa08665959fcecc4c0191f7a5bcc44d3cf05c

bash_completion
```diff
@@ -806,7 +806,7 @@ _comp_delimited()
     fi
     [[ $cur == *$delimiter* ]] && prefix=${cur%"$delimiter"*}$delimiter
 
-    if $deduplicate; then
+    if "$deduplicate"; then
```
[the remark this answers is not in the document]
akinomyoga: > I suppose there was a PR somewhere (which I've likely failed to follow up with) that would ensure a desired environment at start of a completion -- that would sound much preferable over this approach to me.

PR #739 and Issue #786 are related. PR #739 contains unrelated long discussions, but the current situation is summarized in the following two comments:

- https://github.com/scop/bash-completion/pull/739#issuecomment-1356376393
- https://github.com/scop/bash-completion/issues/786#issuecomment-1480199375

> To me, reading `if "$something"; then` raises the question "was this supposed to be `if [[ "$something" ]]; then`" (even though the quotes within the square brackets are superfluous) and thus makes the code heavier to read.

I actually prefer `[[ $something ]]` with `something` being an empty/non-empty variable. The reason that I decided to switch to this direction was item 2 in the review comment https://github.com/scop/bash-completion/pull/891#pullrequestreview-1314581077. Might I have misread the comment? Anyway, if you would like to switch to `[[ $something ]]`, I will adjust them.

completions/strace
akinomyoga: Thank you!

completions/svcadm
```diff
@@ -24,12 +24,12 @@ _smf_complete_fmri()
 {
     local cur="$1" prefix="$2"
     local cur_prefix fmri fmri_list=""
-    local exact_mode pattern
+    local exact_mode=false pattern
```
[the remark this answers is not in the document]
akinomyoga: I have rebased it.

completions/lzip
akinomyoga: Ah, OK. This was a revert commit for https://github.com/scop/bash-completion/pull/862/files#diff-87387e55a9536f7ac2aa5fce39d9a0442c29ec2058215563388d5a284a9ac1c1, but now I realized what was the problem after checking the man page of `lzip`: `|` was actually missing in the original codes.

I modified the commit. You can also cherry-pick this commit 7467419.

scop: Thanks! Cherry-picked.

bash_completion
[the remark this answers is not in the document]
scop: > I actually prefer `[[ $something ]]` with `something` being an empty/non-empty variable. The reason that I decided to switch to this direction was item 2 in the review comment [#891 (review)](https://github.com/scop/bash-completion/pull/891#pullrequestreview-1314581077). Might I have misread the comment?

You've got it right. My opinion is still that `if $something; then` reads cleaner than `if [[ $something ]]; then`. But per reasoning here, we can't have the former if we want to protect against funky `$IFS`. And I think `if "$something"; then` with the quotes is so weird that I prefer `if [[ $something ]]; then` over that.

> Anyway, if you would like to switch to `[[ $something ]]`, I will adjust them.

Yes, please. I'd also suggest using literal `set` as the value when the variable is set, instead of `true`, `1` or the like.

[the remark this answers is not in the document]
akinomyoga: > Yes, please. I'd also suggest using literal `set` as the value when the variable is set, instead of `true`, `1` or the like.

I reworked them so decided to submit them as another PR #931.

akinomyoga:
Switched to an opposite approach #931.

PR #917, opened 2023-04-05
feat(nox): add fallback 3rd party completion loader

scop:


akinomyoga (approved):

completions/_nox
```diff
@@ -0,0 +1,9 @@
+# 3rd party completion loader for argcomplete commands.    -*- shell-script -*-
+#
+# This serves as a fallback in case the completion is not installed otherwise.
+
+eval -- "$(register-python-argcomplete --shell bash "$1" 2>/dev/null)"
```
akinomyoga, line 5: I'm not sure if it changes anything, but I thought it might be better to pick up the `register-python-argcomplete` installed in the same directory as the command in case there are multiple versions of Python in the system. (E.g., `nox` is installed to the system version of Python, but the corresponding system version of `register-python-argcomplete` is hidden by another version of `register-python-argcomplete` set up by `pyenv`, `anaconda`, etc., but I'm actually not so familiar with Python environments).

```suggestion
eval -- "$(
    PATH=$(type -p "$1" 2>/dev/null | sed 's,/[^/]*$,,')${PATH:+:$PATH} \
    register-python-argcomplete --shell bash "$1" 2>/dev/null)"
```

akinomyoga: Also, there is an article that `register-python-argcomplete` is installed as `register-python-argcomplete3` in Ubuntu 22.04. I tried it in Ubuntu 20.04 (with the package `python3-argcomplete`) and confirmed that `register-python-argcomplete` doesn't exist while `register-python-argcomplete3` exists.

scop: Good stuff, thanks!

One thing I was wondering was if we should try the `3` variant first, but per https://peps.python.org/pep-0394/ it seems fine to go as is here, it's generally likely to invoke the "preferred" variant, while in some rare/old cases the Python 2 one.

Another thought was that if this gets invoked for a nonexistent `/path/to/nonexistent/nox`, we end up setting an empty path segment in the front like `PATH=:$PATH`, but I suppose that's not a problem.

Yet another thing that occurred to me is that a consequence of 56186541f is that we never receive a full path to the command here because [the dir gets stripped as the first thing in `__load_completion`](https://github.com/scop/bash-completion/blob/a02a7bc01fa95483eca10fbd317114ddcf0b7b5c/bash_completion#L2488), making the above observation moot. But I'm not sure if this is something we need to address.

completions/_nox
```diff
@@ -0,0 +1,14 @@
+# 3rd party completion loader for argcomplete commands     -*- shell-script -*-
+# sourced using no args to `register-python-argcomplete`.
+#
+# This serves as a fallback in case the completion is not installed otherwise.
+
+eval -- "$(
+    PATH=$(type -p "$1" 2>/dev/null | command sed 's,/[^/]*$,,')${PATH:+:$PATH}
```
akinomyoga, line 7: @scop I'm sorry, I now looked at it again and noticed that this treatment has a problem when `type -p "$1"` failed. The result of `$(...)` becomes empty, and the PATH environment becomes `PATH=:<original content of PATH>`. The empty path in `PATH` results in the execution of an executable file in the current directory, which can be a vulnerability.

```suggestion
    bin_path=$(type -p "$1" 2>/dev/null | command sed 's,/[^/]*$,,')
    [[ $bin_path ]] && PATH=$bin_path${PATH:+:$PATH}
```

completions/_nox
akinomyoga: Ah, thanks. I missed this reply. I have just replied in https://github.com/scop/bash-completion/pull/917#discussion_r1161092667 about the point `PATH:$PATH`.

completions/_nox
scop: Yeah, I observed that as well in https://github.com/scop/bash-completion/pull/917#discussion_r1161092202 above, but thought that wouldn't be an issue and that `.` in `$PATH` would be required for current dir to be considered, not with an empty segment. But you seem right indeed :/ Anyway this already went in, would you mind submitting another PR (pre-approved, feel free to merge directly) for this?

completions/_nox
akinomyoga: > Yet another thing that occurred to me is that a consequence of [5618654](https://github.com/scop/bash-completion/commit/56186541f5a382056cfc25e1d08f5da8018076b8) is that we never receive a full path to the command here because [the dir gets stripped as the first thing in `__load_completion`](https://github.com/scop/bash-completion/blob/a02a7bc01fa95483eca10fbd317114ddcf0b7b5c/bash_completion#L2488), making the above observation moot. But I'm not sure if this is something we need to address.

Ah, maybe there is also my fault here. I have suggested passing `$cmd` to the completion file, but I haven't thought about the consequence of removing `*/` deeply. Does this mean for the completion `/some/path/cmd [TAB]`, where `cmd` is a command that bash-completion offers the 3rd party completion loader and `/some/path` is not included in `PATH`, it fails to find the proper completion settings and fails to complete?

completions/_nox
akinomyoga: > Anyway this already went in, would you mind submitting another PR (pre-approved, feel free to merge directly) for this?

Thank you for your reply. I'll open a PR.

akinomyoga: #920

completions/_nox
scop: Yeah, that is one consequence of it, another might be generating with a wrong command when invoked with a path to the command, but a different one from `$PATH` is used to generate. I dug into this rabbit hole today, results in #924 (and #923).

PR #918, opened 2023-04-08
fix: quote `$split` of `$split && return` for custom IFS

akinomyoga:
These are all the following simple changes but just applied to many files.

```diff
-    $split && return
+    "$split" && return
```

This was originally a part of #914, but I have separated it for easier review of non-trivial changes in #914.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2372,7 +2372,7 @@ _complete_as_root()
 
 _longopt()
 {
-    local cur prev words cword split comp_args
+    local cur prev words cword was_split comp_args
     _comp_initialize -s -- "$@" || return
 
     case "${prev,,}" in
@@ -2403,7 +2403,7 @@ _longopt()
             ;;
     esac
 
-    $split && return
+    [[ $was_split ]] && return
 
     if [[ $cur == -* ]]; then
         _comp_compgen COMPREPLY -W "$(LC_ALL=C $1 --help 2>&1 |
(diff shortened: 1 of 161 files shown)


akinomyoga:
Do we also switch to `[[ $split ]]` for `split`? This involves an interface change to `_comp_initialize`, but we have renamed it from `_init_completion` so we may write an adapter in `000_bash_completion_compat.sh`.

akinomyoga:
> *From https://github.com/scop/bash-completion/pull/931#discussion_r1174398570* by @scop 
> Now that we're doing the boolean checks differently throughout for other cases, the `_comp_initialize` `$split` case looks inconsistent. I wonder if it would make sense and be worth the trouble to introduce another variable, e.g. `$was_split` or something with the empty/`set` value space we use for the other boolean checks? Would be a big diff though. Perhaps there would be an opportunity to do something about this as we've renamed `_init_completion` to `_comp_initialize`. What do you think?

akinomyoga:
- https://github.com/scop/bash-completion/compare/master...akinomyoga:quote-split-B ... This is the version for the suggested interface change so that the variable `split` is set to an empty string or `set` by `_comp_initialize`.
- https://github.com/scop/bash-completion/compare/master...akinomyoga:quote-split-C ... This is the version with the further name change `split` => `was_split`.

scop:
Ooh :heart: thanks heaps! I think I'd prefer option C -- it reads slightly better to me, and could cause less confusion for 3rd party completion writers, in case they take some of our existing completions as a template but opt to use `_init_completion` for backwards compatibility. Which option would you prefer?

Let's just add a `@deprecated` note and elaboration to the new `_init_completion` wrapper.

akinomyoga:
I pushed option C (`was_split`).

scop (approved):
Thanks!

PR #919, opened 2023-04-08
fix(bash_completion): use `_comp_split` and new `_comp_compgen` for safer splitting

akinomyoga:
Now we have a utility function `_comp_split` to safely split strings (including the output of the `compgen` builtin and other commands) without being affected by pathname expansions, unexpected word splitting, etc. I would like to switch from the repeated `local IFS=... reset=$(shopt -po noglob); set -o noglob; ...; IFS=$' \t\n'; $reset` to `_comp_split`.

- b46b71e7 I modified `_comp_split` to return whether nonzero elements are produced in the exit status. This is because `((${#arr[@]}))` is tested in many cases after splitting strings.
- f8a8f9f6 Also, there are many cases of `_comp_split -l arr "$(compgen ARGS)"`, so it is useful to define a function `_comp_compgen arr ARGS` (in particular to reduce nested quoting and command substitutions).
- bf37c0d91cd1bd933b152174a75d9a9580976ebd and a765cc58c9dec40e5618ce2f85400c48f30ca2f8 are independent fixes of the problems I noticed in testing `_comp_compgen`, etc.
- d015b5c1e03d5ac590abdc9e7fbd517901fb1849 is the main rewrite for the switching

The changes suggested in this PR are actually the ones that I mentioned in the following PRs before.

- https://github.com/scop/bash-completion/pull/887#issuecomment-1442170763
- https://github.com/scop/bash-completion/pull/909#issuecomment-1488328631

The changes to `completions/*` are left for later PRs.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1082,8 +1082,6 @@ __parse_options()
     fi
     [[ $option ]] || return 1
 
-    local IFS=$' \t\n' # affects parsing of the regexps below...
-
     # Expand --[no]foo to --foo and --nofoo etc
     if [[ $option =~ (\[((no|dont)-?)\]). ]]; then
         option2=${option/"${BASH_REMATCH[1]}"/}
--- a/bash_completion
+++ b/bash_completion
@@ -432,9 +432,14 @@ _comp_compgen()
     elif [[ $1 == @(*[^_a-zA-Z0-9]*|[0-9]*|''|_*|IFS|OPTIND|OPTARG|OPTERR) ]]; then
         printf '%s\n' "bash_completion: $FUNCNAME: invalid array name \`$1'." >&2
         return 2
-    elif [[ ${*:2} == *\$[0-9]* || ${*:2} == *\$\{[0-9]* ]]; then
+    elif
+        local _nopt=$(($# - 1))
+        [[ ${*:$#-1:1} == -- ]] && ((_nopt -= 2)) # exclude "-- CUR" from check
+        [[ ${*:2:_nopt} == *\$[0-9]* || ${*:2:_nopt} == *\$\{[0-9]* ]]
+    then
         # Note: extglob *\$?(\{)[0-9]* can be extremely slow when the string
-        # "${*:2}" becomes longer, so we test \$[0-9] and \$\{[0-9] separately.
+        # "${*:2:_nopt}" becomes longer, so we test \$[0-9] and \$\{[0-9]
+        # separately.
         printf '%s\n' "bash_completion: $FUNCNAME: positional parameter \$1, \$2, ... do not work inside this function." >&2
         return 2
     fi
--- a/bash_completion
+++ b/bash_completion
@@ -324,24 +324,27 @@ _comp_expand_glob()
 # state of `IFS` and the shell option `noglob`.  A naive splitting by
 # `arr=(...)` suffers from unexpected IFS and pathname expansions, so one
 # should prefer this function to such naive splitting.
-# @param $1 array_name  The array name
-#   The array name should not start with an underscores "_", which is
-#   internally used.  The array name should not be either "IFS" or
-#   "OPT{IND,ARG,ERR}".
-# @param $2 text        The string to split
 # OPTIONS
 #   -a      Append to the array
 #   -F sep  Set a set of separator characters (used as IFS).  The default
 #           separator is $' \t\n'
 #   -l      The same as -F $'\n'
+# @param $1 array_name  The array name
+#   The array name should not start with an underscores "_", which is
+#   internally used.  The array name should not be either "IFS" or
+#   "OPT{IND,ARG,ERR}".
+# @param $2 text        The string to split
+# @return 2 when the usage is wrong, 0 when one or more completions are
+#   generated, or 1 when the execution succeeds but no candidates are
+#   generated.
 _comp_split()
 {
-    local _assign='=' IFS=$' \t\n'
+    local _append=false IFS=$' \t\n'
 
     local OPTIND=1 OPTARG="" OPTERR=0 _opt
     while getopts ':alF:' _opt "$@"; do
         case $_opt in
-            a) _assign='+=' ;;
+            a) _append=true ;;
             l) IFS=$'\n' ;;
             F) IFS=$OPTARG ;;
             *)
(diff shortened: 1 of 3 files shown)


bash_completion
```diff
@@ -1266,19 +1328,20 @@ _ncpus()
...
-        result=${#COMPREPLY[@]}
-        # 2>/dev/null for direct invocation, e.g. in the _tilde unit test
-        ((result > 0)) && compopt -o filenames 2>/dev/null
+        if _comp_compgen COMPREPLY -P '~' -u -- "${1#\~}"; then
+            # 2>/dev/null for direct invocation, e.g. in the _tilde unit test
+            compopt -o filenames 2>/dev/null
+            return 1
+        fi
     fi
-    return "$result"
```
scop, line 1281: eeacd68777aa993be57727726a73a7a5e451deeb cherry-picked as 177cba155a39229da986de41b2ac5bca20bf3045

scop (commented):
`_tilde` and `_known_hosts_real` fixes cherry-picked, will revisit the main content of this later. (As a first impression, I like the idea.)

bash_completion
```diff
@@ -1974,6 +2022,9 @@ _known_hosts_real()
     # If we have known_hosts files to use
     if ((${#kh[@]} + ${#khd[@]} > 0)); then
         if ((${#kh[@]} > 0)); then
+            # make sure that COMPREPLY exists
+            COMPREPLY+=()
+
```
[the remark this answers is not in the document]
akinomyoga: Thanks!

bash_completion
akinomyoga: I have rebased it.

bash_completion
```diff
@@ -324,24 +324,27 @@ _comp_expand_glob()
...
 #   -a      Append to the array
 #   -F sep  Set a set of separator characters (used as IFS).  The default
 #           separator is $' \t\n'
 #   -l      The same as -F $'\n'
+# @param $1 array_name  The array name
+#   The array name should not start with an underscores "_", which is
+#   internally used.  The array name should not be either "IFS" or
+#   "OPT{IND,ARG,ERR}".
+# @param $2 text        The string to split
+# @exit 2 when the usage is wrong, 0 when one or more completions are
```
scop, line 337:
```suggestion
# @return 2 when the usage is wrong, 0 when one or more completions are
```

bash_completion
```diff
@@ -363,10 +366,92 @@ _comp_split()
...
+#           separators in evaluating the expansions of `-W '...'`, etc.  The
+#           delimiter of the output of `compgen` is always a newline.
+#   -l      The same as -F $'\n'
+# @param $1 array_name  The array name
+#   The array name should not start with an underscores "_", which is
+#   internally used.  The array name should not be either "IFS" or
+#   "OPT{IND,ARG,ERR}".
+# @param $2... args     The arguments that are passed to compgen
+#   Note: References to positional parameters $1, $2, ... (such as -W '$1')
+#   will not work as expected because these references the arguments of
```
scop, line 406:
```suggestion
#   will not work as expected because these reference the arguments of
```

bash_completion
```diff
@@ -363,10 +366,92 @@ _comp_split()
...
+            l) IFS=$'\n' ;;
+            F) IFS=$OPTARG ;;
+            *)
+                echo "bash_completion: $FUNCNAME: usage error" >&2
+                return 2
+                ;;
+        esac
+    done
+    shift "$((OPTIND - 1))"
+    if (($# < 2)); then
```
scop, line 428: Maybe?
```suggestion
    if (($# != 2)); then
```

bash_completion
```diff
@@ -363,10 +366,92 @@ _comp_split()
...
+# This function essentially performs ARRAY=($(compgen args...)) but properly
+# handles shell options, IFS, etc. using _comp_split.  This function is
+# equivalent to `_comp_split [-a] -F sep ARRAY "$(compgen args...)"`, but this
+# pattern is frequent in the codebase and is good to separate out as a function
+# for the possible future implementation change.
+# OPTIONS
+#   -a      Append to the array
+#   -F sep  Set a set of separator characters (used as IFS in evaluating
+#           `compgen').  The default separator is $' \t\n'.  Note that this is
+#           not the separators to delimit the output of `compgen' but the
```
scop, line 396:
```suggestion
#           not the set of separators to delimit output of `compgen', but the
```

bash_completion
```diff
@@ -995,28 +1057,27 @@ _comp_initialize()
...
+                    ;;
+                -?*) [[ $option ]] || option=$i ;;
+                *) break ;;
+            esac
+        done
+    fi
     [[ $option ]] || return 1
 
-    IFS=$' \t\n' # affects parsing of the regexps below...
+    local IFS=$' \t\n' # affects parsing of the regexps below...
```
scop, line 1080: I think we could just remove this -- the comment is still true, but we are no longer modifying `$IFS` _ourselves_  in a way that it'd break those, and we're not doing this for other similar cases across the codebase either.

bash_completion
```diff
@@ -1391,24 +1449,26 @@ else
                     #
                     line=${line:i}   # take command starting from found index
                     line=${line%% *} # trim arguments
-                    procs+=($line)
+                    [[ $line ]] &&
+                        procs+=("$line")
```
scop, line 1453:
```suggestion
                    [[ $line ]] && procs+=("$line")
```

bash_completion
```diff
@@ -1391,24 +1449,26 @@ else
...
                 for line in "${psout[@]}"; do
                     if [[ $line =~ ^[[(](.+)[])]$ ]]; then
-                        procs+=(${BASH_REMATCH[1]})
+                        procs+=("${BASH_REMATCH[1]}")
                     else
                         line=${line%% *}      # trim arguments
                         line=${line##@(*/|-)} # trim leading path and -
-                        procs+=($line)
+                        [[ $line ]] &&
+                            procs+=("$line")
```
scop, line 1465:
```suggestion
                        [[ $line ]] && procs+=("$line")
```

bash_completion
```diff
@@ -2023,7 +2072,8 @@ _known_hosts_real()
                 if [[ $i == *key_22_$cur*.pub && -r $i ]]; then
                     host=${i/#*key_22_/}
                     host=${host/%.pub/}
-                    COMPREPLY+=($host)
+                    [[ $host ]] &&
+                        COMPREPLY+=("$host")
```
scop, line 2076:
```suggestion
                    [[ $host ]] && COMPREPLY+=("$host")
```

scop (approved):
Oh yes, this is great, thanks! Left some nitty comments, pre-approved with them addressed/not as you like.

bash_completion
akinomyoga: We expect two or more arguments here. We need an array name plus arguments that will be specified to `compgen`.

akinomyoga: I realized my mistake in the usage text. I think I copied the related code from `_comp_split` but forgot to update the usage text. I now fixed it in 0a267dc2bf74e9956aacc2b31d032d4d239814d6.

bash_completion
akinomyoga: Thank you. I have applied it eaeb53e8.

scop:
Added https://github.com/scop/bash-completion/projects/2#card-89052206 for the completions/* work, perhaps that's something we could try to do for 2.12.

PR #921, opened 2023-04-08
Add support for ping4

sur5r:


akinomyoga (approved):

completions/ping
```diff
@@ -65,8 +65,9 @@ _ping()
     fi
 
     [[ $1 == *6 ]] && ipvx=-6
+    [[ $1 == *4 ]] && ipvx=-4
     _known_hosts_real ${ipvx-} -- "$cur"
 } &&
-    complete -F _ping ping ping6
+    complete -F _ping ping ping4 ping6
```
scop, line 71: We should add a `ping4` symlink so the completion gets actually loaded for it, similarly as for `ping6`. For reference:
```
$ git grep -wFn ping6 | head -n 3
completions/.gitignore:154:/ping6
completions/Makefile.am:686:	ping6 \
completions/Makefile.am:961:		ping6
```

scop (changes requested):

sur5r:
Ah, truel, I missed that. Will amend the PR later today.

akinomyoga:
Also, actually, the commit message should be in Conventional Commits (as described in CONTRIBITING.md).

completions/ping
sur5r: Done.

scop (approved):

scop:
Thanks!

PR #922, opened 2023-04-10
fix(ssh): deprecated option PubkeyAcceptedKeyTypes

inconstante:
As reported by a Debian user in https://bugs.debian.org/1033642, ssh changed the option name from PubkeyAcceptedKeyTypes to PubkeyAcceptedAlgorithms. The old name still works, but no longer presented in the manpages.

Does this pull request make sense? It's been a while since I last submitted something here, but I noticed that many things changed, so I might me making some mistakes with the submission.

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -55,7 +55,7 @@ _comp_xfunc_ssh_options()
         NoHostAuthenticationForLocalhost NumberOfPasswordPrompts
         PasswordAuthentication PermitLocalCommand PKCS11Provider Port
         PreferredAuthentications ProxyCommand ProxyJump ProxyUseFdpass
-        PubkeyAcceptedKeyTypes PubkeyAuthentication RekeyLimit RemoteCommand
+        PubkeyAcceptedAlgorithms PubkeyAuthentication RekeyLimit RemoteCommand
         RemoteForward RequestTTY RevokedHostKeys SecurityKeyProvider SendEnv
         ServerAliveCountMax ServerAliveInterval SetEnv StreamLocalBindMask
         StreamLocalBindUnlink StrictHostKeyChecking SyslogFacility TCPKeepAlive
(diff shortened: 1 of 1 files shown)


completions/ssh
```diff
@@ -5,7 +5,7 @@ _ssh_queries()
     local queries=$(_comp_xfunc_ssh_query "$1" help)
     [[ $queries ]] || queries="cipher cipher-auth mac kex key key-cert
          key-plain key-sig protocol-version compression sig ciphers macs
-         kexalgorithms pubkeyacceptedkeytypes hostkeyalgorithms
+         kexalgorithms pubkeyacceptedalgorithms hostkeyalgorithms
```
scop, line 8: This is the fallback hardcoded set of query types in case `ssh -Q help` doesn't give us a set of supported ones to work with. I don't think we need to cater for the change in question here.

completions/ssh
```diff
@@ -188,7 +188,7 @@ _ssh_suboption()
         proxycommand | remotecommand | localcommand)
             COMPREPLY=($(compgen -c -- "$cur"))
             ;;
-        pubkeyacceptedkeytypes)
+        pubkeyacceptedalgorithms)
```
scop, line 191: Supporting both here should be no problem.

```suggestion
        pubkeyacceptedalgorithms | pubkeyacceptedkeytypes)
```

completions/ssh
```diff
@@ -55,7 +55,7 @@ _comp_xfunc_ssh_options()
         NoHostAuthenticationForLocalhost NumberOfPasswordPrompts
         PasswordAuthentication PermitLocalCommand PKCS11Provider Port
         PreferredAuthentications ProxyCommand ProxyJump ProxyUseFdpass
-        PubkeyAcceptedKeyTypes PubkeyAuthentication RekeyLimit RemoteCommand
+        PubkeyAcceptedAlgorithms PubkeyAuthentication RekeyLimit RemoteCommand
```
scop, line 58: The change in question seems to be a fairly recent one, so I think it would be better to add `PubkeyAcceptedKeyTypes` to the "Selected old ones" list below this one, instead of removing it completely.

scop (changes requested):

inconstante:
Thanks, Ville. I made the suggested changes.

scop (approved):
Thanks! Took the liberty to move `PubkeyAcceptedKeyTypes` a bit per an earlier suggestion, merging with that.

PR #923, opened 2023-04-10
refactor: rename `_realcommand` => `_comp_realcommand`

scop:
Some small related improvements while at it, see individual commits for details.

PR #924, opened 2023-04-10
Feat/load completion improvements

scop:
Intended to be merged after #923.

This started from looking into https://github.com/scop/bash-completion/pull/917#discussion_r1161092202 and ballooned in scope somewhat. The gist of this is that we now provide the full command name (even when not invoked with one) to the completions when sourced, to facilitate generating from the correct executable, and for improved handling of commands invoked with relative paths.

The commits are kind of intertwined, but I hope looking into them individually for details provides enough context if the end result is hard to follow.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2542,9 +2542,8 @@ __load_completion()
     [[ $realcmd ]] && paths=("${realcmd%/*}") || paths=()
     _comp_split -aF : paths "$PATH"
     for dir in "${paths[@]%/}"; do
-        if [[ -d $dir && $dir == ?*/@(bin|sbin) ]]; then
+        [[ $dir == ?*/@(bin|sbin) ]] &&
             dirs+=("${dir%/*}/share/bash-completion/completions")
-        fi
     done
 
     # 4) From XDG_DATA_DIRS or system dirs (e.g. /usr/share, /usr/local/share):
@@ -2557,10 +2556,14 @@ __load_completion()
 
     # Look up and source
     shift
-    local prefix
+    local i prefix compspec
     for prefix in "" _; do # Regular from all dirs first, then fallbacks
-        for dir in "${dirs[@]}"; do
-            [[ -d $dir ]] || continue
+        for i in ${!dirs[*]}; do
+            dir=${dirs[i]}
+            if [[ ! -d $dir ]]; then
+                unset -v 'dirs[i]'
+                continue
+            fi
             for compfile in "$prefix$cmdname" "$prefix$cmdname.bash"; do
                 compfile="$dir/$compfile"
                 # Avoid trying to source dirs as long as we support bash < 4.3
--- a/bash_completion
+++ b/bash_completion
@@ -2548,8 +2548,12 @@ __load_completion()
         dirs+=(./completions)
     fi
 
-    # 3) From bin directories extracted from path to command, and $PATH
-    [[ $cmd == */* ]] && paths=("${cmd%/*}") || paths=()
+    # 3) From bin directories extracted from the specified path to the command,
+    # the real path to the command, and $PATH
+    paths=()
+    [[ $cmd == */* ]] && paths+=("${cmd%/*}")
+    local ret
+    _comp_realcommand "$cmd" && paths+=("${ret%/*}")
     _comp_split -aF : paths "$PATH"
     for dir in "${paths[@]%/}"; do
         [[ $dir == ?*/@(bin|sbin) ]] &&
--- a/bash_completion
+++ b/bash_completion
@@ -2500,6 +2500,14 @@ __load_completion()
     local -a paths
     [[ $cmdname ]] || return 1
 
+    local backslash=
+    if [[ $cmd == \\* ]]; then
+        cmd=${cmd:1}
+        # If we already have a completion for the "real" command, use it
+        $(complete -p "$cmd" 2>/dev/null || echo false) "\\$cmd" && return 0
+        backslash=\\
+    fi
+
     local -a dirs=()
 
     # Lookup order:
@@ -2537,14 +2545,6 @@ __load_completion()
     # Completions in the system data dirs.
     _comp_split -F : paths "${XDG_DATA_DIRS:-/usr/local/share:/usr/share}"
     dirs+=("${paths[@]/%//bash-completion/completions}")
-
-    local backslash=
-    if [[ $cmdname == \\* ]]; then
-        cmdname=${cmdname:1}
-        # If we already have a completion for the "real" command, use it
-        $(complete -p "$cmdname" 2>/dev/null || echo false) "\\$cmdname" && return 0
-        backslash=\\
-    fi
 
     # For loading 3rd party completions wrapped in shopt reset
     local IFS=$' \t\n'
--- a/bash_completion
+++ b/bash_completion
@@ -2496,9 +2496,9 @@ complete -F _minimal ''
 
 __load_completion()
 {
-    local cmd="${1##*/}" dir compfile
+    local cmd=$1 cmdname=${1##*/} dir compfile
     local -a paths
-    [[ $cmd ]] || return 1
+    [[ $cmdname ]] || return 1
 
     local -a dirs=()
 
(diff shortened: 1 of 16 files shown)


bash_completion
```diff
@@ -2523,68 +2538,71 @@ __load_completion()
...
+                        # If invoked without path, that one should be set, too
+                        # ...but let's not overwrite an existing one, if any
+                        [[ $origcmd != */* ]] &&
+                            ! complete -p "$origcmd" &>/dev/null &&
+                            extspecs+=("$origcmd")
+                        ((${#extspecs[*]} != 0)) && $compspec "${extspecs[@]}"
+                        return 0
+                    fi
+                    # If not, see if we got one for $cmdname
+                    if compspec=$(complete -p "$cmdname" 2>/dev/null); then
```
akinomyoga, line 2592:
```suggestion
                    if [[ $cmdname != "$cmd" ]] && compspec=$(complete -p "$cmdname" 2>/dev/null); then
```

bash_completion
```diff
@@ -2496,9 +2496,24 @@ complete -F _minimal ''
...
+        # If we already have a completion for the "real" command, use it
+        $(complete -p "$cmd" 2>/dev/null || echo false) "\\$cmd" && return 0
+        backslash=\\
+    fi
+
+    # Try to resolve real path to $cmd, and make it absolute
+    local ret realcmd="" origcmd=$cmd
+    if _comp_realcommand "$cmd"; then
+        realcmd=$ret
+        cmd=${realcmd%/*}/$cmdname # basename could be an alias
```
akinomyoga, line 2515: I think I fail to understand this line. I naively thought this should be `cmd=$realcmd` because the command name can be different in the real location, such as `~/bin/b -> /usr/bin/bash`. There seems to be a reason related to an alias as in the code comment, but I'm not sure what it exactly is now.

akinomyoga (commented):
Thank you for working on this issue! This is great.

akinomyoga:
I think we should also fix

```diff
diff --git a/completions/_nox b/completions/_nox
index ac387ef7..8e98002a 100644
--- a/completions/_nox
+++ b/completions/_nox
@@ -9,6 +9,8 @@ eval -- "$(
         register-python-argcomplete3 --shell bash "$1" 2>/dev/null
 )"

-complete -p "$1" &>/dev/null
+{
+    complete -p "$1" || complete -p "${1##*/}"
+} &>/dev/null

 # ex: filetype=sh
```

bash_completion
scop: This was my intuition too when writing it, but because we test
```bash
if compspec=$(complete -p "$cmd" 2>/dev/null); then
```
...above and return if that holds, we have already returned before reaching here if the two are equal.

akinomyoga: I think the behavior doesn't change but we may reduce the fork. When the two are equal *and* `complete -p cmd` fails, we are going to run the subshell `$(complete -p cmd)` twice with the current version.

bash_completion
scop: "Alias" is actually not a good word to describe this, but I think you got the intent right: the basename of the resolved command can be different from the one we were invoked with.

Installing our completion for the "real" resolved name would be problematic for example in cases where the target command has different behavior depending on the basename it is invoked as. `busybox` is one example of this.

Elaborated on the commentary in 3f479c173b9bd2619962a47b120e48acd54d6c0b

scop: Hmm, good point. I suppose we should actually resolve absolute path to the command without doing any symlink resolution, essentially what `_comp_realcommand` now does, but without doing any of the realpath/readlink things?

bash_completion
scop: Indeed, applied, thanks!

bash_completion
akinomyoga: Ah, yes. I think that is the solution.

The original reason that we wanted to resolve the symbolic links is to pick up the corresponding `share/bash-completion/completions`, so the symlink-resolved `realcmd` is still needed. So the conversion of `cmd` to an absolute path should be done independently of `realcmd`.

scop: 28b587a47058395fdacb7eafe866685af1e74e3c and e4ede514249c4250dd43139f54fa3c46554a1dbc are a quick attempt at fixing this, more or less untested.

bash_completion
```diff
@@ -1659,6 +1659,25 @@ _fstypes()
     [[ $fss ]] && COMPREPLY+=($(compgen -W "$fss" -- "$cur"))
 }
 
+# Get absolute path to a file, with rudimentary canonicalization.
+# No symlink resolution or existence checks are done;
+# see `_comp_realcommand` for those.
+# @param $1 The file
+# @var[out] ret The path
+_comp_abspath()
```
scop, line 1667: This would deserve some unit tests.

bash_completion
scop: Need to call it a day now, revisiting later. But do feel free to push changes you see fit here in the meantime if you like.

bash_completion
```diff
@@ -2523,68 +2548,71 @@ __load_completion()
         dirs+=(./completions)
     fi
 
-    # 3) From bin directories extracted from $(realpath "$cmd") and PATH
-    local ret
-    _comp_realcommand "$1" && paths=("${ret%/*}") || paths=()
+    # 3) From bin directories extracted from path to command, and $PATH
+    [[ $cmd == */* ]] && paths=("${cmd%/*}") || paths=()
```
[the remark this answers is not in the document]
akinomyoga: I believe we still need to use the symlink-resolved `readlcmd` here. We wanted to find `share/bash-completion/completions` of the actual command even if it is called from a symlink. This is independent of how we resolve `cmd`.

bash_completion
akinomyoga: Maybe I have confused you with my previous reply, but I actually agreed to switch to the absolute path without resolving the symbolic links. I wanted to note that the symlink-resolved `realcmd` is still used later for another purpose.

bash_completion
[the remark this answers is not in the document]
akinomyoga: > Unsure if this check should be `== /*` instead of `== */*`. But I suppose it doesn't make much of a difference.

A relative path containing a slash can happen here only when the executable file does not exist. The question is whether we should search the corresponding completion file in the relative-path directory. I'm not sure it's worth it but added 301da7f6.

> I believe we still need to use the symlink-resolved `readlcmd` here. We wanted to find `share/bash-completion/completions` of the actual command even if it is called from a symlink. This is independent of how we resolve `cmd`. (by @akinomyoga)

I have pushed a fix for the real path bbe1ae88.

bash_completion
```diff
@@ -2496,9 +2506,24 @@ complete -F _minimal ''
...
+    if [[ $cmd == \\* ]]; then
+        cmd=${cmd:1}
+        # If we already have a completion for the "real" command, use it
+        $(complete -p "$cmd" 2>/dev/null || echo false) "\\$cmd" && return 0
+        backslash=\\
+    fi
+
+    # Resolve absolute path to $cmd
+    local ret pathcmd origcmd=$cmd
+    if pathcmd=$(type -P "$cmd"); then
```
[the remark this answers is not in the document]
akinomyoga: I think either work. I don't have a particular preference here.

bash_completion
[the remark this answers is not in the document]
scop: Hm, I'm unable to locate the mentioned commits in the repo. Were they pushed to a fork somewhere? Feel free to push them in the branch here.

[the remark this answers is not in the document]
akinomyoga: Ah, sorry. It's just my mistake. I intended to push them into the PR branch but have actually pushed them into the branch of the same name in my repository. I now pushed them.

bash_completion
scop: Added in 81132e79a9902783c019a941d9568f19a8c34de2.

bash_completion
```diff
@@ -2523,68 +2548,75 @@ __load_completion()
...
     fi
 
-    # 3) From bin directories extracted from $(realpath "$cmd") and PATH
+    # 3) From bin directories extracted from the real path to the command, the
+    # specified path to the command, and $PATH
     local ret
-    _comp_realcommand "$1" && paths=("${ret%/*}") || paths=()
+    paths=()
+    _comp_realcommand "$cmd" && paths+=("${ret%/*}")
+    [[ $cmd == /* ]] && paths+=("${cmd%/*}")
```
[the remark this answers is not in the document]
akinomyoga: Hmm, I agree.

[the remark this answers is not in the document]
akinomyoga: I modified the commit bbe1ae88 (diff [`bbe1ae88..61982cbc`](https://github.com/scop/bash-completion/compare/bbe1ae88..61982cbc)) and force-pushed.

akinomyoga (approved):

PR #930, opened 2023-04-21
refactor: mark deprecated config variables clearer

scop:
Move ones in the main `bash_completion` to the compat snippet.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -61,6 +61,36 @@ _comp_deprecate_func()
     eval -- "$2() { $3 \"\$@\"; }"
 }
 
+# Declare a compatibility variable name.
+# For bash 4.3+, a real name alias is created, allowing value changes to
+# "apply through" when the variables are set later. For bash versions earlier
+# than that, the operation is once-only; the value of the new variable
+# (if it's unset) is set to that of the old (if set) at call time.
+#
+# @param $1 Version of bash-completion where the deprecation occurred
+# @param $2 Old variable name
+# @param $3 New variable name
+# @since 2.12
+_comp_deprecate_var()
+{
+    if (($# != 3)); then
+        printf 'bash_completion: %s: usage: %s DEPRECATION_VERSION OLD_NAME NEW_NAME\n' "$FUNCNAME" "$FUNCNAME"
+        return 2
+    fi
+    if [[ $2 != [a-zA-Z_]*([a-zA-Z_0-9]) ]]; then
+        printf 'bash_completion: %s: %s\n' "$FUNCNAME" "\$2: invalid variable name '$1'" >&2
+        return 2
+    elif [[ $3 != [a-zA-Z_]*([a-zA-Z_0-9]) ]]; then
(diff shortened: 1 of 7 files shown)


bash_completion.d/000_bash_completion_compat.bash
```diff
@@ -178,6 +178,14 @@ _realcommand()
     return $rc
 }
 
+# @deprecated set the `BASH_COMPLETION_*` variables directly instead.
+[[ -v COMP_FILEDIR_FALLBACK ]] &&
+    : "${BASH_COMPLETION_FILEDIR_FALLBACK:=$COMP_FILEDIR_FALLBACK}"
```
akinomyoga, line 183:
```suggestion
    : "${BASH_COMPLETION_FILEDIR_FALLBACK=$COMP_FILEDIR_FALLBACK}"
```

When there are both settings `BASH_COMPLETION_FILEDIR_FALLBACK=` and `COMP_FILEDIR_FALLBACK=set`, I think `BASH_COMPLETION_FILEDIR_FALLBACK` should have higher precedence.

akinomyoga (commented):
Some non-trivial question is whether to support the case those old variables are changed after sourcing `bash_completion`. But anyway we should remove these special treatments for old names from the main part of the script at some point.

In `bash >= 4.3`, we may declare

```bash
declare -gn COMP_FILEDIR_FALLBACK=BASH_COMPLETION_FILEDIR_FALLBACK
```

in `000_bash_completion_compat.bash`.

bash_completion.d/000_bash_completion_compat.bash
scop: Right! Addressed in 8e855a2c3057fb745d28e23728bb2d7edd182379.

scop:
Hmm, true that. I'll mark this as a draft and let it cook for a bit.

bash_completion.d/000_bash_completion_compat.bash
akinomyoga: Thanks!

scop:
Modified this some, guess it's ready for review if this is something we'd want.

akinomyoga (approved):
I agree with merging it. Bash 4.2 is anyway old and not used in many environments.

bash_completion
```diff
@@ -2520,7 +2519,7 @@ _known_hosts_real()
 
     # Add results of normal hostname completion, unless
     # `BASH_COMPLETION_KNOWN_HOSTS_WITH_HOSTFILE' is set to an empty value.
-    if [[ ${BASH_COMPLETION_KNOWN_HOSTS_WITH_HOSTFILE-${COMP_KNOWN_HOSTS_WITH_HOSTFILE-1}} ]]; then
+    if [[ ${BASH_COMPLETION_KNOWN_HOSTS_WITH_HOSTFILE-set ]]; then
```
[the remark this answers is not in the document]
akinomyoga: Ah, please ignore this comment. This seems to have been *a pending comment* for about half an year.

PR #933, opened 2023-04-23
docs: Python dependency clarifications

scop:


CONTRIBUTING.md
```diff
@@ -199,6 +199,15 @@ Also, please bear the following coding guidelines in mind:
 - In addition to running the test suite, there are a few scripts in the test/
   dir that catch some common issues, see and use for example runLint.
 
+- Make sure you have Python 3.7 or later installed. This is required for
```
scop, line 202: We require >=3.7 already per fc141bb55. pyupgrade is no longer in use (superseded by ruff), but mypy >= 0.980 requires it anyway.

akinomyoga (approved):

PR #935, opened 2023-04-23
refactor([m-o]*): apply naming guidelines where API is as desired

scop:


completions/makepkg
```diff
@@ -23,7 +23,7 @@ _makepkg_slackware()
     _filedir
 }
 
-_makepkg_bootstrap()
+_comp_cmd_makepkg__bootstrap()
```
akinomyoga, line 26: https://github.com/scop/bash-completion/blob/3ed0e1f9bed7b7ec7950a20a60ac9345aab8f7a3/test/t/test_makepkg.py#L5

This line in the test could be also updated, but maybe this line is unneeded.

akinomyoga (approved):
pre-approve

PR #938, opened 2023-04-23
refactor: samba scan config rename, docs

scop:


scop:
Chose "smbtree" for the command name even though it's not the "primary" one in this file, because that's also included in this file, and the scans are performed with it.

akinomyoga (approved):
I think the change direction in the commit message is the opposite. Except that, LGTM.

scop:
D'oh! Reworded, thanks.

PR #939, opened 2023-04-24
refactor([t-z]*): apply naming guidelines where API is as desired

scop:


completions/vncviewer
```diff
@@ -13,11 +13,11 @@ _vncviewer_bootstrap()
...
     complete -F $fname vncviewer
     $fname "$@" # Generate completions once for now
-    unset -f _vncviewer_bootstrap
+    unset -f "$FUNCNAME"
 } &&
-    complete -F _vncviewer_bootstrap vncviewer
+    complete -F _comp_cmd_vncviewer__bootstrap vncviewer
 
-_tightvncviewer()
+_comp_cmd_tightvncviewer()
```
akinomyoga, line 20: https://github.com/scop/bash-completion/blob/3ed0e1f9bed7b7ec7950a20a60ac9345aab8f7a3/completions/vncviewer#L8-L9

These also need to be updated.

completions/vncviewer
```diff
@@ -1,6 +1,6 @@
 # bash completion for vncviewer                            -*- shell-script -*-
 
-_vncviewer_bootstrap()
+_comp_cmd_vncviewer__bootstrap()
```
akinomyoga, line 3: https://github.com/scop/bash-completion/blob/3ed0e1f9bed7b7ec7950a20a60ac9345aab8f7a3/test/t/test_vncviewer.py#L4

completions/tar
```diff
@@ -710,22 +710,22 @@ _posix_tar()
...
 
-    __tar_try_list_archive && return
+    _comp_cmd_tar__try_list_archive && return
 
     # file completion on relevant files
     _filedir
 }
 
-_tar()
+_comp_cmd_tar()
```
akinomyoga, line 719: https://github.com/scop/bash-completion/blob/3ed0e1f9bed7b7ec7950a20a60ac9345aab8f7a3/test/t/test_tar.py#L8

completions/vncviewer
scop: Thanks! Addressed in a fixup commit (gitlint will fail with that until #940 is merged). Ditto the two test ignores below.

akinomyoga (approved):

PR #941, opened 2023-04-26
refactor: apply more guidelines where API is as desired

scop:


akinomyoga (approved):
LGTM for the changes in the PR

----

By the way, can we also adjust the first line of `completions/dpkg` in this PR?

https://github.com/scop/bash-completion/blob/e3f2c69ddbaab5428b22931285a63f52e402a040/completions/dpkg#L1

First, this line is treated as if the file header, but it actually seems to be the code comments to the first functions. I think the line should be split into the file header `# dpkg completion    -*- shell-script -*-` and `# These functions are ...`

scop:
> By the way, can we also adjust the first line of `completions/dpkg` in this PR?

Sure, done in 19ccc6bd22da985b4d3ede6c2c68ad7c1447b511. I opted to remove info on what might be calling the functions in question; it was out of date, and we don't maintain that kind of info in comments in the first place. And in this case especially, as the functions are "xfunc" -- we don't even know all possible callers by design.

PR #942, opened 2023-04-26
refactor(ant,pkg-get,pkgutil,rpm,ssh): adjust towards current API

scop:


completions/ant
```diff
@@ -9,7 +9,7 @@ _ant_parse_targets()
     # parse buildfile for targets
     while read -rd '>' line; do
         if [[ $line =~ \<(target|extension-point)[[:space:]].*name=[\"\']([^\"\']+) ]]; then
-            targets+=" ${BASH_REMATCH[2]}"
+            ret+=" ${BASH_REMATCH[2]}"
```
akinomyoga, line 12: `ret` needs to be initialized in advance. Also, can we use arrays?

```suggestion
            ret+=("${BASH_REMATCH[2]}")
```

completions/ant
```diff
@@ -9,7 +9,7 @@ _ant_parse_targets()
     # parse buildfile for targets
```
akinomyoga, line 9:
```suggestion
    # parse buildfile for targets
    ret=()
```

or if we still use a space-separated result in the scalar `ret`,

```suggestion
    # parse buildfile for targets
    ret=""
```

completions/ant
```diff
@@ -87,12 +87,12 @@ _comp_cmd_ant()
...
 
-        local targets
+        local ret
 
         # fill targets
-        _ant_parse_targets "$buildfile"
+        _comp_cmd_ant__targets "$buildfile"
 
-        COMPREPLY=($(compgen -W '$targets' -- "$cur"))
+        COMPREPLY=($(compgen -W '$ret' -- "$cur"))
```
akinomyoga, line 95:
```suggestion
        COMPREPLY=($(compgen -W '"${ret[@]}"' -- "$cur"))
```

completions/java
```diff
@@ -103,15 +96,15 @@ _comp_cmd_java__classes()
...
 
     # find which sourcepath to use
-    _java_find_sourcepath
+    _comp_cmd_java__find_sourcepath
 
     # convert package syntax to path syntax
     cur=${cur//.//}
     # parse each sourcepath element for packages
-    for i in ${sourcepath//:/ }; do
+    for i in ${ret//:/ }; do
```
akinomyoga, line 107: These changes overlap with #888. I'll later check both changes.

scop: Yeah, that one's marked as draft so I wasn't quite sure of its status. But we can drop the java changes from here if you like, and address them later/elsewhere.

completions/ant
scop: Ah, sure, thanks!

completions/java
scop: Removed java from here, added pkgutil, rpm, and ssh instead.

completions/ant
scop: Actually this function is recursive, so initialization needs to happen outside of it. Addressed in the latest commit.

akinomyoga: Ah, I see. Thank you.

akinomyoga (approved):

PR #949, opened 2023-05-01
feat: more fallback 3rd party completion loaders

scop:
Mainly from https://github.com/spf13/cobra/blob/main/projects_using_cobra.md

### DIFF
--- a/completions/.gitignore
+++ b/completions/.gitignore
@@ -153,6 +153,7 @@
 /md5sum
 /mdecrypt
 /mencoder
+/_metalctl
 /micropython
 /mkisofs
 /_mmctl
--- a/completions/.gitignore
+++ b/completions/.gitignore
@@ -115,6 +115,7 @@
 /_istioctl
 /javac
 /javadoc
+/_k3s
 /_kn
 /_kool
 /kplayer
--- a/completions/.gitignore
+++ b/completions/.gitignore
@@ -116,6 +116,7 @@
 /_kn
 /_kool
 /kplayer
+/_kubeadm
 /_kubectl
 /l2ping
 /lbzip2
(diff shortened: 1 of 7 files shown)


akinomyoga (approved):
> Mainly from https://github.com/spf13/cobra/blob/main/projects_using_cobra.md

Maybe we can mention this link somewhere in our documentation (or wiki), so we can track the changes in the linked page.

scop:
Sure, added https://github.com/scop/bash-completion/wiki/Externally-available-completions/_compare/61f7f9843c340613b3a13ab57e70d9614bb7ce8d...a4e3e6f0e899a06b1bb534cd1ab9158fe77da63a for general considerations about the external completion page and 3rd party fallbacks, and the link in https://github.com/scop/bash-completion/wiki/Externally-available-completions/_compare/a4e3e6f0e899a06b1bb534cd1ab9158fe77da63a...8a638f545ea707fbaeecdffbe5aab6df8e49a09d.

scop:
Oh, forgot to mention that removing entries having 3rd party loaders from the external completions page loses the info whether the completions require bash-completion or not. In that sense I'm open to reverting the first change as well as keeping it, dunno which way I'd prefer :shrug: (feel free to modify the way you see fit)

akinomyoga:
Thank you. I think it is also good to keep the existing version information, (though I'm not sure if it will actually be used in the future so I think we do not have to collect the information actively).
- I added a separate section: [wiki - Externally-available-completions - 8a638f54...c4e651fe](https://github.com/scop/bash-completion/wiki/Externally-available-completions/_compare/8a638f545ea707fbaeecdffbe5aab6df8e49a09d...c4e651fe3490667cf52dd87f932a1d617615b78d)

scop:
Thanks!

Issue #948, opened 2023-05-01
Use _comp_split and _comp_compgen in completions/*


scop:
> Now we have a utility function _comp_split to safely split strings (including the output of the compgen builtin and other commands) without being affected by pathname expansions, unexpected word splitting, etc. [...]
> The changes to completions/* are left for later PRs.

_Originally posted by @akinomyoga in https://github.com/scop/bash-completion/pull/919

akinomyoga:
### Naming of functions that generate completions in COMPREPLY

I would like to discuss the naming of a particular class of functions (including `_filedir`, `_parse_help`, and `_parse_usage`).

For this rewriting, we will switch from `COMPREPLY=($(compgen ...))`, etc. to use `_comp_compgen ...`. One of the frequent cases among them is `COMPREPLY=($(compgen -W '$(_parse_help ...)' -- "$cur"))`. I would like to rewrite them to use new versions of `_parse_help` or `_parse_usage` that directly assign results to `COMPREPLY` (or a specified array).

Then there is a related discussion that I wanted to open. Currently, there are utility functions of the name `_<CATEGORY>` that generate and store completions of the category in `COMPREPLY`, such as `_filedir`, `_ip_addresses`, and `_pids`. The functions `_parse_help` and `_parse_usage` also behave in the same way after the suggested change to store results in arrays. If we just follow the new naming convention, the simple prefixing of `_comp` works for all of them (such as `_comp_filedir`).

- `_comp_filedir`, `_comp_ip_addresses`, `_comp_pids`, ...

But I think there is also a possibility to introduce a namespace or a naming rule for this class of functions, such as

- `_comp_generate_filedir`, `_comp_generate_ip_addresses`, `_comp_generate_pids`, ...
- `_comp_compgen_filedir`, `_comp_compgen_ip_addresses`, `_comp_compgen_pids`, ...
- `_comp_gen_filedir`, `_comp_gen_ip_addresses`, `_comp_gen_pids`, ...
- `_comp_yield_filedir`, `_comp_yield_ip_addresses`, `_comp_yield_pids`, ...
- `_comp_action_filedir`, `_comp_action_ip_addresses`, `_comp_action_pids`, ... (inspired by `compgen -A <action>`)
- or other ways

What do you think?

akinomyoga:
Also, PR #560 by @algorythmic is related to the `_parse_help` change. I have rebased #560. Could you first check #560?

akinomyoga:
[...]

PR #951, opened 2023-05-02
fix(_comp_compgen): change interface

akinomyoga:
https://github.com/scop/bash-completion/issues/948#issuecomment-1529412590

```bash
# before
_comp_compgen [-al|-F sep] ARRAY COMPGEN_ARGS... -- CUR

# after
_comp_compgen [-alR|-F sep|-c CUR|-v ARRAY] -- COMPGEN_ARGS...
```

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -383,20 +383,24 @@ _comp_split()
 
 # Call `compgen` with the specified arguments and store the results in the
 # specified array.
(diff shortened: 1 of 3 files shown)


scop (approved):
Nice! Reads better and more concise this way :+1:

akinomyoga:
Thanks!

PR #952, opened 2023-05-02
refactor: `{ => _comp_compgen}_filedir` and `_comp_compgen [opts] NAME`

akinomyoga:
I came up with a new idea for https://github.com/scop/bash-completion/issues/948#issuecomment-1529412590. This PR implements the idea using `_filedir` as an example.

1) First, the function `_filedir` is renamed to `_comp_compgen_filedir`. This function can be directly called by `_comp_compgen_filedir ...` as we have been doing so far but can also be called as `_comp_compgen filedir ...`. 

```bash
$ COMPREPLY=(1 2 3); cur=c
$ _comp_compgen_filedir; echo "${COMPREPLY[*]}"
1 2 3 configure completions config.log config.status configure~ configure.ac

$ COMPREPLY=(1 2 3); cur=c
$ _comp_compgen filedir; echo "${COMPREPLY[*]}"
1 2 3 configure completions config.log config.status configure~ configure.ac
```

In the above, `_comp_compgen` redirects the call `_comp_compgen NAME` to the shell function `_comp_compgen_NAME` in a similar way as `xfunc`. Note that this is not ambiguous with the call `_comp_compgen -- <comgpen args>` because all the arguments to the `compgen` builtin start with `-` while our function names do not contain `-`. When the first (non-option) argument of `_comp_compgen` does not start with `-`, `_comp_compgen` redirects the call to `"_comp_compgen_$1"`. Otherwise, _comp_compgen can consider the arguments to be passed to the `compgen` builtin.

2) The options specified to `_comp_compgen` will take effect in the behavior of the redirected `_comp_compgen_filedir`. For example, `_comp_compgen_filedir` stores the result in `COMPREPLY` by default, but we can instead store the results to a specified array `arr` by `_comp_compgen -v arr filedir`. This is done by recording the parsed options in local variables `_comp_compgen__*` and letting the child calls of `_comp_compgen` inside `_comp_compgen_filedir` reference the local variables.

```bash
$ COMPREPLY=(1 2 3); arr=(); cur=c
$ _comp_compgen -v arr filedir
$ echo "COMPREPLY: ${COMPREPLY[*]}"
COMPREPLY: 1 2 3
$ echo "arr: ${arr[*]}")
arr: configure completions config.log config.status configure~ configure.ac
```

Commit fcf98258c0b2b2da45c4e4c3aec5b1bbc0381ba9 implements the above two extensions of `_comp_compgen`. Actually, the implementation is simple.

----

The modification at the `_filedir` side is minimal as far as the completions are generated through `_comp_compgen`. Only the codes that directly reference `cur` and `COMPREPLY` need to be updated.

**edit**:  I also changed the behavior of `_comp_compgen_filedir` so that it replaces the existing `COMPREPLY` by default. To make it append to the array, one can specify the `-a` option as `_comp_compgen -a filedir`. Currently, many similar functions just replace the existing contents, but only a part of them try to append to the array. I'd like to make the default behaviors of similar functions consistent and allow appending as an option `-a`.


bash_completion
```diff
@@ -451,15 +492,26 @@ _comp_compgen()
...
         # Note: extglob *\$?(\{)[0-9]* can be extremely slow when the string
         # "${*:2:_nopt}" becomes longer, so we test \$[0-9] and \$\{[0-9]
         # separately.
+        [[ $* == *\$[0-9]* || $* == *\$\{[0-9]* ]]
+    then
         printf 'bash_completion: %s: positional parameter $1, $2, ... do not work inside this function.\n' "$FUNCNAME" >&2
         return 2
     fi
 
+    if [[ $1 != -* ]] && declare -f "_comp_compgen_$1" &>/dev/null; then
```
scop, line 504: Would it not be better to error out if the target function does not exist, rather than fall through?

scop (commented):
Oh, very nice!

I did not look into usages yet, but I thought I'd just ask -- in the vast majority of cases we have I believe `_filedir`'s behavior of appending to `COMPREPLY` isn't required. Usually it does no harm either though. What do you think, while we're working on this, should we also review each call site and remove the append behavior where unnecessary? I think doing so would reveal and fix some bugs, but this is quite a bit of work for a smallish gain. We can also postpone to later if seen useful but not something we'd like to do now.

bash_completion
akinomyoga: Yes, that is definitely better. I added dbdb0141551aa4eee7eabfc6929e53e184131379.

akinomyoga:
> in the vast majority of cases we have I believe `_filedir`'s behavior of appending to `COMPREPLY` isn't required and we could replace instead.

I agree with it, but we need to carefully check each case not to break existing behavior. In removing flag `-a`, we need to make sure that `COMPREPLY` is empty at that point (or we do want to discard the existing elements of `COMPREPLY`). I expect most of them would be easy, but I'm afraid that there would be not a small number of cases that it is subtle.

> What do you think, while we're working on this, should we also review each call site and remove the append behavior where unnecessary?

Actually, at the current stage, the new implementation of `_comp_compgen_filedir` is incomplete. For example, `_tilde` called from `_comp_compgen_filedir` should also be upgraded to the `_comp_compgen_tilde` style for the sake of completeness. Let me first try to make `_comp_compgen_filedir` complete by upgrading related functions.

akinomyoga:
The function `_comp_cmd_lintian` is suspicious.

https://github.com/scop/bash-completion/blob/a26a2a9d388b2f9b291b4d5cd6d60624b149a43f/completions/lintian#L70-L145

Even when completions are generated in `if [[ $prev == -* ]]; then`, it goes through the `case "$cur" in` block. For example, `_filedir` would be called twice for `lintian -p [TAB]`. Maybe

```diff
diff --git a/completions/lintian b/completions/lintian
index 0769fb4b..2175c720 100644
--- a/completions/lintian
+++ b/completions/lintian
@@ -89,28 +89,36 @@ _comp_cmd_lintian()
         case $prev in
             -C | --check-part | -X | --dont-check-part)
                 _comp_cmd_lintian__checks
+                return
                 ;;
             -T | --tags | --suppress-tags)
                 _comp_cmd_lintian__tags
+                return
                 ;;
             --tags-from-file | --suppress-tags-from-file | --cfg | -p | \
                 --packages-file)
                 _filedir
+                return
                 ;;
             --lab | --archivedir | --dist | --root)
                 _filedir -d
+                return
                 ;;
             --color)
                 COMPREPLY=($(compgen -W "never always auto html" -- "$cur"))
+                return
                 ;;
             -U | --unpack-info)
                 _comp_cmd_lintian__infos
+                return
                 ;;
             --area | --section)
                 COMPREPLY=($(compgen -W "main contrib non-free" -- "$cur"))
+                return
+                ;;
+            --arch)
+                return
                 ;;
-            --arch) ;;
-
         esac
     fi
```

akinomyoga:
The function `_comp_cmd_python` is also suspicious.

https://github.com/scop/bash-completion/blob/a26a2a9d388b2f9b291b4d5cd6d60624b149a43f/completions/python#L64-L74

Case `!(?(*/)python*([0-9.])|?(*/)pypy*([0-9.])|-?)` fall backs to the general case, then `_filedir` might be called multiple times. But I'm not sure the intent of this code, so maybe this is intended.

akinomyoga:
This branch of `_comp_cmd_smartctl` is also suspicious.

https://github.com/scop/bash-completion/blob/a26a2a9d388b2f9b291b4d5cd6d60624b149a43f/completions/smartctl#L55-L57

akinomyoga:
It is subtle to make `_comp_xfunc_ssh_identityfile` replace existing elements of `COMPREPLY` or append to them. The calls in the `bash_completion` codebase seem to be in the context where `COMPREPLY` is empty, so we can make the function replace existing elements, but possible external completions that use `_comp_xfunc_ssh_identityfile` might assume it to append the completions. Now I'm letting it append to existing elements.

https://github.com/scop/bash-completion/blob/a26a2a9d388b2f9b291b4d5cd6d60624b149a43f/completions/ssh#L251-L261

completions/lintian
```diff
@@ -95,10 +95,10 @@ _comp_cmd_lintian()
                 ;;
             --tags-from-file | --suppress-tags-from-file | --cfg | -p | \
                 --packages-file)
-                _filedir
+                _comp_compgen filedir
```
scop, line 98: Agree on the suggestion in https://github.com/scop/bash-completion/pull/952#issuecomment-1535594464

completions/python
```diff
@@ -62,15 +62,15 @@ _comp_cmd_python()
             return
             ;;
         !(?(*/)python*([0-9.])|?(*/)pypy*([0-9.])|-?))
-            [[ $cword -lt 2 || ${words[cword - 2]} != -[QWX] ]] && _filedir
+            [[ $cword -lt 2 || ${words[cword - 2]} != -[QWX] ]] && _comp_compgen filedir
```
scop, line 65: TBH, I'm having a problem understanding what this does.

So if `$prev` is not the python executable or a short option, complete on all files if the word before previous is not `-[QWX]` or we're completing the first arg to python? 

It was added in 9913362b38a01b3a59290f209645028f41b66f65. The commentary in 63ecfc38ba47335e96996e0355925411ecb228f8 provides a hint... that I may not quite get.

But perhaps the intention here is to complete the first arg following a script file with filenames?

I guess a `return` here would be in order.

completions/smartctl
```diff
@@ -40,7 +40,7 @@ _comp_cmd_smartctl__drivedb()
         prefix=+
         cur="${cur#+}"
     fi
-    _filedir h
+    _comp_compgen filedir h
     [[ $prefix ]] && COMPREPLY=("${COMPREPLY[@]/#/$prefix}")
 }
```
scop, line 46: I'd say missing `return` in `--quietmode` below is an oversight.

scop (approved):
Great stuff. Left a few comments inline in response to separate comments, pre-approved with them addressed the way you see best.

completions/pkgadd
```diff
@@ -23,14 +23,14 @@ _comp_cmd_pkgadd()
 
     case $prev in
         -d)
-            _filedir pkg
-            _filedir -d
+            _comp_compgen filedir pkg
+            _comp_compgen -a filedir -d
```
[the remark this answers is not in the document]
akinomyoga: Hmm, you are right. I'll fix it.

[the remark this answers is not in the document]
akinomyoga: One might have possibly wanted to control the ordering of the completions in the list, but it anyway does not seem to work. There are similar cases also in `dpkg-source`. Or maybe, just the author of these completions did not notice that `_filedir` and `_filedir ext` generates also the directory names.

Fixed in 4645ecdf (before applying `_filedir => _comp_compgen filedir` changes)

completions/lintian
akinomyoga: Fixed in the (forced) push [`bbb0e5f...253609a`](https://github.com/scop/bash-completion/compare/bbb0e5f16a535667a22279ce2fe6e87b2d080c2b..253609ad04003fafcca5e9fcbf44fc8af70ae44a)

completions/python
akinomyoga: I again looked at it. Maybe it tries to complete the arguments to a Python script (as in `python script.py args[TAB]`). The exclusion of `python` and `pypy` could be to test if `cword` is not the first argument. Anyway, this is just a guess, and the intent of implementing it in this way is still unclear.

The original commit 9913362 also has another missing `return` in the next for-loop, so I think this case branch also missed the `return`.

Fixed in [`bbb0e5f...253609a`](https://github.com/scop/bash-completion/compare/bbb0e5f16a535667a22279ce2fe6e87b2d080c2b..253609ad04003fafcca5e9fcbf44fc8af70ae44a)

akinomyoga:
Added partial reverts 83d6ea10 3d6cbfcf

akinomyoga:
partly squashed

Issue #432, opened 2020-07-26
Additional helper functions


qtc-de:
Hey :)

I'm new to bash-completion and wrote my completion scripts from scratch so far. Now I'm really happy that I can use some helper functions from your scripts. However, certain helper functions I used in the past are missing and I wanted to ask if you are interested in implementing them or if you can tell me alternative solution how these cases can be covered by your script (I would create a pull request, but as you will see my bash skills are really poor. Furthermore, I'm not sure if the functions are already implemented somewhere and I'm missing them).

### 1. Get n-th Parameter

As I can see, you support a ``_get_first_arg`` function. This is nice, but in the past I also need to obtain the n-th parameter from the command line. I used the following function to accomplish this:

```bash
function _comp_get_parameter() {
    # Returns the n-th parameter from the command line (not counting options
    # or the command name).
    #
    # Parameters
    #   number          (int)           Parameter to select
    #
    # Returns
    #   retval          (int)           Error / Success
    #
    # Side effects
    #   result is saved in $arg variable
    #
    local COUNT

    COUNT=0
    arg=

    for var in ${COMP_LINE}; do

	# exclude options and the toolname
	if [[ "$var" == -* ]] || [[ "$var" == "${COMP_WORDS[0]}" ]]; then
		continue;
	fi

        COUNT=$((COUNT+1))
        if [[ $COUNT -eq $1 ]]; then
            arg="$var"
            return 0
        fi

    done

    return 1
}
```

### 2. List Contains

This is not really completion specific, but it is that useful that I use it in nearly all my completion scripts. The idea is to compare two strings that contain space separated words with each other. If one word matches, it returns true, otherwise false.

```bash
function _comp_contains() {
	# Takes two strings containing space seperated words and checks if one of the
    # words in the second list matches one in the first.
    # E.g.: `_comp_contains "test test2" "no nope test"` returns true.
    #
    # Parameters
    #   list_lookup     (string)        Space seperated words to search in
    #   list_search     (string)        Space seperated words to search
    #
    # Returns
    #   retval          (int)           Error / Success
    #
    # Side effects
    #   None
    #
    for word in $2; do

        if [[ $1 =~ (^|[[:space:]])${word}($|[[:space:]]) ]]; then
            return 0
        fi

    done;

    return 1
}
```

[...]

PR #953, opened 2023-05-03
[WIP] Array utilities

akinomyoga:
I decided to move code stubs for array utilities in #739 and #950 to here.

> I guess this could end up as an invalid regex if `_parse_help` lets any strings through to `main_cmd` that make it so. The previous implementation had the same issue though, so not a regression. Leaving up to you to decide whether to address or merge as is.
> 
> Some array utilities would be nice to help with cases like this. I have this ages old unfinished patch for the main `bash_completion` in my stash, probably broken in the first place and uses namerefs so requires bash >= 4.3, but to illustrate what I was thinking back $then
> 
> ```diff
> +_bashcomp_uniq()
> [...]
> +_bashcomp_last_index() {
> [...]
> +_bashcomp_compact() {
> [...]
> +_bashcomp_index_of() {
> [...]
> ```

_Originally posted by @scop in https://github.com/scop/bash-completion/pull/950#discussion_r1183001067_

> * [bc6f87d](https://github.com/scop/bash-completion/commit/bc6f87d7f27a9df0f5b9662f44316f0426ec02fb) This implements a function `_comp_array_filter`. I'm not sure whether Ville would finally find it reasonable to include the function in bash-completion, but I tried to make the interface minimal (one function `_comp_array_filter`) yet make the function versatile as @calestyo would be likely to request. In any way, this function can be used for the array manipulations of general cases.

_Originally posted by @akinomyoga in https://github.com/scop/bash-completion/pull/739#issue-1201728747_

### DIFF
--- a/completions-core/ARRAY.bash
+++ b/completions-core/ARRAY.bash
@@ -150,3 +150,57 @@ _comp_xfunc_ARRAY_filter()
 
     [[ ! $_unset ]]
 }
+
+_bashcomp_uniq()
+{
+    local -n _bashcomp_uniq_array_=$1
+    local -A tmp
+    local -i i
+    for i in ${!_bashcomp_uniq_array_[*]}; do
+        ((tmp["${_bashcomp_uniq_array_[i]}"]++ > 0)) &&
+            unset '_bashcomp_uniq_array_[i]'
+    done
+}
+
+_bashcomp_last_index()
+{
+    local -n _bashcomp_last_index_array_=$1 _bashcomp_last_index_ret_=$2
+    local -i i
+    for i in ${!_bashcomp_last_index_array_[*]}; do :; done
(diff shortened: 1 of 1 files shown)


akinomyoga:
Ref #432

completions-core/ARRAY.bash
scop, line 1: Not sure if the intent is to leave this as a separate file, but I suppose we could think of shipping it in `startup-core` now.

akinomyoga: Actually, I'm not sure if we would finally include those utilities. Currently, those functions are placed in `completions-core/ARRAY.bash` *experimentally* so that they can be loaded through `_comp_xfunc` for delayed loading. However, they might be put in `helpers-core` with a new mechanism of helper autoloader.

Issue #958, opened 2023-05-06
Over-eager positional param check in `_comp_compgen`


scop:
## Describe the bug

Trying to use a command containing string looking like a positional parameter reference with `_comp_compgen` but which isn't one runs at odds with the check in https://github.com/scop/bash-completion/blob/77c9fd69e1d6ffa285ab5f96970a5955ed281534/bash_completion#L524-L527

I don't think it's feasible to try and detect false positives like this one. Guess I'm suggesting we either remove the check, or standardize on a workaround.

## To reproduce

```bash
_comp_compgen -- -W '$(echo foo bar | awk "{ print \$2 }")'
```

## Expected behavior

No error.

## Versions (please complete the following information)

- [ ] Operating system name/distribution and version:
- [ ] bash version, `echo "$BASH_VERSION"`:
- [ ] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`:

akinomyoga:
Actually, this is related to what I wanted to discuss finally. I have still many pending commits that I haven't submitted yet, and among them is 55facc3384f2daeeeb3214d7096dc7a8a0c355e3 related to this issue. 55facc3384f2daeeeb3214d7096dc7a8a0c355e3 is just a workaround, but I wanted to discuss how to deal with the typical `-W '$(...)'` cases.

There are three things to address.

- One is that as reported here `-W ''` has a limitation by its design. Although `awk '{print $1}'` is false-positive, a more frequent pattern `'$("$1" ... | ...)` does not work.
- There are also vulnerable cases of `-W "$(...)"` (notice that it is not single-quoted but double-quoted, where the problems can be caused when the output of command picks some strings containing `$(malicious code)` from somewhere (such as filenames).
- Another problem caused by `-W '$(...)'` is that it bypasses the check by shfmt and shellcheck. Actually, many of the fixes in #956 are revealed by converting `... -W '$(xxxx)'` to `stdout=$(xxx); ... -W '$stdout'`

All of these issues are caused by the behavior `-W` evaluating the shell expansions specified in the arguments. My current idea is to introduce a new `_comp_command split 'text'` that does not perform the expansions but just splits the arguments. Something like

```bash
_comp_compgen_split() {
    local -a input=("$@")
    local IFS=${_comp_compgen__ifs-$' \t\n'} # Note: let _comp_compgen define _ifs
    _comp_compgen -- -W '${input[*]}'
}
```

PR #961, opened 2023-05-07
refactor: dpkg xfuncs to API behavior

scop:


completions/dpkg
```diff
@@ -1,37 +1,62 @@
...
-        grep-status -P -e "^${1-}" -a -FStatus 'ok installed' -o -FStatus 'ok config-files' -n -s Package
-    }
-else
-    _comp_xfunc_dpkg_installed_packages()
-    {
+_comp_xfunc_dpkg_installed_packages()
+{
+    local pkgs
+    pkgs=$(
+        grep-status -P -e "^${cur-}" -a \
```
scop, line 7: I don't remember if we had a rule against just trying to invoke commands that might not be present, e.g. for command not found handlers. If yes, I think we should document it somewhere.

completions/dpkg
```diff
@@ -1,37 +1,62 @@
...
+            -FStatus 'ok installed' \
+            -n -s Package 2>/dev/null
+    ) || pkgs=$(
```
scop, line 10: We could check if the exit code is 127 (or 126) rather than fall back on all failures, but I suppose this would be fine.

completions/dpkg
akinomyoga: I don't remember if it is documented, but there are already many cases that try to call may-not-exist commands with `2>/dev/null`, so I guess it should be allowed.

As for the command not found handlers, we actually require the writers of `command_not_found_handler` to be responsible for it (by checking the existence of `COMP_LINE`, etc.) in README.md.

completions/dpkg
```diff
@@ -1,37 +1,65 @@
...
         command awk -F '\n' -v RS="" "
-            index(\$1, \"Package: ${1-}\") == 1 &&
+            index(\$1, \"Package: ${cur-}\") == 1 &&
             \$2 ~ /ok installed|half-installed|unpacked|half-configured|^Essential: yes/ {
                 print(substr(\$1, 10));
             }" /var/lib/dpkg/status 2>/dev/null
-    }
+    )
+    _comp_compgen -a -- -W "$pkgs"
```
akinomyoga, line 18:
```suggestion
    _comp_compgen -a -- -W '$pkgs'
```

As mentioned in https://github.com/scop/bash-completion/issues/958#issuecomment-1537214890, the `-W` option performs the expansions in the option argument, we would like to avoid the unexpected expansions that might be mixed in the outputs.

completions/dpkg
```diff
@@ -1,37 +1,65 @@
...
-    {
+# @since 2.12
+_comp_xfunc_dpkg_installed_packages()
+{
+    local pkgs
+    pkgs=$(
+        grep-status -P -e "^${cur-}" -a \
+            -FStatus 'ok installed' \
+            -n -s Package 2>/dev/null
+    ) || pkgs=$(
```
akinomyoga, line 11: I think we can simplify it as 

```suggestion
            -n -s Package 2>/dev/null ||
```

akinomyoga (commented):
I think we need to wait for the decision of #962 but left comments that I noticed for now.

scop:
> I think we need to wait for the decision of #962 but left comments that I noticed for now.

We could also go ahead with this already before #962, and address later along with others.

akinomyoga (approved):

PR #954, opened 2023-05-03
refactor: `_parse_{help,usage}` => `_comp_compgen_help`

akinomyoga:
This is based on #952 so would be kept to be Draft until PR #952 settles.

The own changes of this PR are 2ce600f6...39b15f38. This PR includes the API change of `_parse_{help,usage}` and trivial rewriting for switching from `_parse_{help,usage}`. Other non-trivial rewriting related to `_parse_{help,usage}` will be included in later PRs.

PR #964, opened 2023-05-07
style(_comp_compgen): call `_comp_compgen_filedir` directly

akinomyoga:
> I think it's easy enough to remember that if no `_comp_compgen` features are needed, calling `_comp_compgen_FOO` directly is ok (and preferred) for efficiency. So I guess that's what I'd prefer, but whatever we go with would be good to be documented somewhere, e.g. in the style guide, or the related function commentary.

_Originally posted by @scop in https://github.com/scop/bash-completion/pull/954#discussion_r1186805078_

> The change `_comp_compgen filedir` => `_comp_compgen_filedir` is another large one and makes the GitHub interface unuseful, so I decided to separate the PR for it.

_Originally posted by @akinomyoga in https://github.com/scop/bash-completion/pull/954#discussion_r1186823627_

----

c91e6c4e7bc3adb7531158d21e99af2150f8b2ea is the rewrite.

I'm including the variable name changes in `_comp_compgen_filedir` to avoid conflicts that can be introduced when `_comp_comgen -v var filedir` is used (08477b4bdbd2dd4cb1ead7ce51500450805cb453).

> but whatever we go with would be good to be documented somewhere, e.g. in the style guide, or the related function commentary.

I added a `@remarks` in the function code document (ea3a88ece00d83dcd93aabf48180b073b66713bc) but haven't added it to the style guide. It seems even usage of `xfunc` is currently undocumented in `doc/*.md`, so it is consistent to describe both `_comp_xfunc` and `_comp_compgen` if we are to add commentary of the `_comp_compgen` usage in the style guide.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -437,6 +437,11 @@ _comp_split()
 #   [-alR|-v arr|-c cur] to the child calls of `_comp_compgen` in
 #   `_comp_compgen_NAME`.
 #
+# @remarks When no options are supplied to _comp_compgen, `_comp_compgen NAME
+# args` is equivalent to the direct call `_comp_compgen_NAME args`.  As the
+# direct call is slightly more efficient, the direct call is preferred over
+# calling it through `_comp_compgen`.
+#
 # @remarks Design `_comp_compgen_NAME`: a function that produce completions can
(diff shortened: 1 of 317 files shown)


scop (approved):
Thanks!

PR #965, opened 2023-05-07
refactor: adjust to util/xfunc primary command convention

scop:
https://github.com/scop/bash-completion/discussions/959

### DIFF
--- a/completions/perl
+++ b/completions/perl
@@ -3,7 +3,7 @@
 _comp_cmd_perl__helper()
 {
     COMPREPLY=($(compgen -P "$prefix" -W \
-        "$("${2:-perl}" "${BASH_SOURCE[0]%/*}/../helpers/perl" "$1" "$cur")" \
+        "$("${1:-perl}" "${BASH_SOURCE[0]%/*}/../helpers/perl" "$2" "$cur")" \
         -- "$cur"))
     [[ $1 == functions ]] || _comp_ltrim_colon_completions "$prefix$cur"
 }
@@ -41,7 +41,7 @@ _comp_cmd_perl()
                 temp="${cur#-}"
                 prefix+=${cur%"$temp"}
                 cur="$temp"
-                _comp_cmd_perl__helper modules "$1"
+                _comp_cmd_perl__helper "$1" modules
                 return
                 ;;
             -*V)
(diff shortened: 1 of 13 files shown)


completions/rpm
```diff
@@ -244,30 +257,32 @@ _comp_cmd_rpmbuild()
...
-    # shellcheck disable=SC2254
+    # shellcheck disable=SC2119,SC2254
     case $prev in
         --buildroot | --root | --dbpath | -${noargopts}r)
             _comp_compgen filedir -d
             return
             ;;
+        --target | --eval | -${noargopts}E | --buildpolicy)
+            local pathcmd
+            pathcmd=$(type -P "$1") && local PATH=${pathcmd%/*}:$PATH
```
scop, line 269: The benefit of prepending to `$PATH` as opposed to trying to figure out the exact command is that we don't need to poke into the internals of the utility function even for file-local use, and the utility function's default command basename persists.

In some cases we may want to pass the full path, that's still possible, but I suppose the `$PATH` modification use case would be a more common one.

scop:
There are some cases where we should be using `$1` rather than a hardcoded one in the tree still, as well as others that could make use of the `$PATH` prepending as here, but I'll leave them for another followup MR. So I think this one's ready to roll, FYI @akinomyoga

akinomyoga (approved):

completions/perl
```diff
@@ -3,7 +3,7 @@
 _comp_cmd_perl__helper()
 {
     COMPREPLY=($(compgen -P "$prefix" -W \
-        "$("${2:-perl}" "${BASH_SOURCE[0]%/*}/../helpers/perl" "$1" "$cur")" \
+        "$("${1:-perl}" "${BASH_SOURCE[0]%/*}/../helpers/perl" "$2" "$cur")" \
         -- "$cur"))
     [[ $1 == functions ]] || _comp_ltrim_colon_completions "$prefix$cur"
```
akinomyoga, line 8: Hmm, I think this line should have also been updated as `[[ $2 == functions ]]`.

akinomyoga: @scop Can you include it in #988?

scop: Indeed, good catch, done in b85263a6de7f8e60f09489caac3f9a0d613e804f.

PR #967, opened 2023-05-08
refactor: more API related work

scop:
This and that on the topic of API adjustments, see individual commits for details.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -927,7 +927,7 @@ _comp__split_longopt()
 # Complete variables.
 # @return  True (0) if variables were completed,
 #          False (> 0) if not.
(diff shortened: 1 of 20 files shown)


akinomyoga (approved):

scop:
Hm, something wrong with build dependencies it seems, the scp symlink is probably not there so the function is found. Not immediately clear to me what's wrong though.

Added some missing renamings in 18442622a26a2cdd861952677489118ca629eeeb..fbc43538f55cf34ef0294df786d6b1b697c5895b but that's not the failure reason.

scop:
Will merge without the scp change and file another PR for it.

PR #968, opened 2023-05-08
refactor(ssh): remove `ssh` from `scp` xfunc names

scop:
xfuncs being part of the API, we don't want to expose the implementation detail that the `scp` completion currently resides in the `ssh` file.

Originally part of #967


scop:
https://github.com/scop/bash-completion/pull/967#issuecomment-1538710476

scop:
Merging per earlier approval in #967

PR #971, opened 2023-05-08
refactor: remaining `_comp_compgen_{help,usage}` changes

akinomyoga:
These are the remaining ones for the `_parse_{help,usage}` refactoring.

### DIFF
--- a/completions/dnssec-keygen
+++ b/completions/dnssec-keygen
@@ -37,8 +37,10 @@ _comp_cmd_dnssec_keygen()
     esac
 
     if [[ $cur == -* ]]; then
-        COMPREPLY=($(compgen -W '$(_parse_help "$1" | \
-            command sed -e "s/:\$//")' -- "$cur"))
+        _comp_compgen -R help
+        ((${#COMPREPLY[@]})) &&
+            _comp_compgen -- -W '$(printf "%s\n" "${COMPREPLY[@]}" |
+                command sed -e "s/:\$//")'
         return
     fi
 } &&
--- a/completions/dnssec-keygen
+++ b/completions/dnssec-keygen
@@ -39,8 +39,7 @@ _comp_cmd_dnssec_keygen()
     if [[ $cur == -* ]]; then
         _comp_compgen -R help
         ((${#COMPREPLY[@]})) &&
-            _comp_compgen -- -W '$(printf "%s\n" "${COMPREPLY[@]}" |
-                command sed -e "s/:\$//")'
+            _comp_compgen -- -W '"${COMPREPLY[@]%:}"'
         return
     fi
 } &&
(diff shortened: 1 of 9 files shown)


completions/dnssec-keygen
```diff
@@ -37,8 +37,10 @@ _comp_cmd_dnssec_keygen()
     esac
 
     if [[ $cur == -* ]]; then
-        COMPREPLY=($(compgen -W '$(_parse_help "$1" | \
-            command sed -e "s/:\$//")' -- "$cur"))
+        _comp_compgen -R help
+        ((${#COMPREPLY[@]})) &&
+            _comp_compgen -- -W '$(printf "%s\n" "${COMPREPLY[@]}" |
+                command sed -e "s/:\$//")'
```
scop, line 43:
```suggestion
            _comp_compgen -- -W '"${COMPREPLY[@]%:}"'
```

completions/links
```diff
@@ -80,8 +80,10 @@ _comp_cmd_links()
     esac
 
     if [[ $cur == -* ]]; then
-        COMPREPLY=($(compgen -W '$(_parse_help "$1" |
-            command grep -vF -- "->")' -- "$cur"))
+        _comp_compgen -R help
+        ((${#COMPREPLY[@]})) &&
+            _comp_compgen -- -W '$(printf "%s\n" "${COMPREPLY[@]}" |
+                command grep -vF -- "->")'
```
scop, line 86:
```suggestion
            _comp_compgen -- -W '"${COMPREPLY[@]}"' -X "->"
```

completions/zopfli
```diff
@@ -12,8 +12,10 @@ _comp_cmd_zopfli()
     esac
 
     if [[ $cur == -* ]]; then
-        COMPREPLY=($(compgen -W \
-            '$(_parse_help "$1" -h | command sed -e "s/#$//")' -- "$cur"))
+        _comp_compgen -R help -- -h
+        ((${#COMPREPLY[@]})) &&
+            _comp_compgen -- -W '$(printf "%s\n" "${COMPREPLY[@]}" |
+                command sed -e "s/#$//")' -- "$cur"
```
scop, line 18:
```suggestion
            _comp_compgen -- -W '"${COMPREPLY[@]%#}"'
```

scop (changes requested):

completions/links
akinomyoga: Thank you for these suggestions! You are right, we can do them just by the builtin feature. I would like to apply these changes, but I'd like to confirm one thing here. The above `grep -vF -- "->"` is technically equivalent to `-X '*->*'`, but does the intent actually to remove the exact matching to `->`?

akinomyoga: Thank you. I have applied the changes.

scop (approved):

PR #972, opened 2023-05-08
fix: use `_comp_compgen` for simple cases

akinomyoga:
These are the cases requiring manual editing but straightforward ones.

### DIFF
--- a/completions/bzip2
+++ b/completions/bzip2
@@ -37,7 +37,8 @@ _comp_cmd_bzip2()
     _comp_compgen_tilde && return
 
     compopt -o filenames
-    COMPREPLY=($(compgen -f -X "$xspec" -- "$cur") $(compgen -d -- "$cur"))
+    _comp_compgen -- -f -X "$xspec"
+    _comp_compgen -a -- -d
 } &&
     complete -F _comp_cmd_bzip2 bzip2 pbzip2 lbzip2
 
--- a/completions/bzip2
+++ b/completions/bzip2
@@ -37,8 +37,7 @@ _comp_cmd_bzip2()
     _comp_compgen_tilde && return
 
     compopt -o filenames
-    _comp_compgen -- -f -X "$xspec"
-    _comp_compgen -a -- -d
+    _comp_compgen -- -f -X "$xspec" -o plusdirs
 } &&
     complete -F _comp_cmd_bzip2 bzip2 pbzip2 lbzip2
 
--- a/completions/find
+++ b/completions/find
@@ -36,7 +36,7 @@ _comp_cmd_find()
             return
             ;;
         -group)
-            COMPREPLY=($(compgen -g -- "$cur" 2>/dev/null))
+            _comp_compgen -- -g 2>/dev/null
             return
             ;;
         -xtype | -type)
--- a/completions/find
+++ b/completions/find
@@ -36,7 +36,7 @@ _comp_cmd_find()
             return
             ;;
         -group)
-            _comp_compgen -- -g 2>/dev/null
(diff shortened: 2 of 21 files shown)


completions/find
```diff
@@ -36,7 +36,7 @@ _comp_cmd_find()
             return
             ;;
         -group)
-            COMPREPLY=($(compgen -g -- "$cur" 2>/dev/null))
+            _comp_compgen -- -g 2>/dev/null
```
scop, line 39: Not introduced in this PR, but I wonder what the reason for `2>/dev/null` is here.
```suggestion
            _comp_compgen -- -g
```

completions/bzip2
```diff
@@ -37,7 +37,8 @@ _comp_cmd_bzip2()
     _comp_compgen_tilde && return
 
     compopt -o filenames
-    COMPREPLY=($(compgen -f -X "$xspec" -- "$cur") $(compgen -d -- "$cur"))
+    _comp_compgen -- -f -X "$xspec"
+    _comp_compgen -a -- -d
```
scop, line 41: Maybe while at it and in similar cases throughout this PR?
```suggestion
    _comp_compgen -- -f -X "$xspec" -o plusdirs
```

completions/gcc
```diff
@@ -53,7 +53,7 @@ _comp_cmd_gcc()
         prefix_length=$((${#prefix} + 1))
         local flags=$("$1" --completion="$argument" | cut -c $prefix_length-)
         [[ ${flags} == "=*" ]] && compopt -o nospace 2>/dev/null
-        COMPREPLY=($(compgen -W "$flags" -- ""))
+        _comp_compgen -R -- -W "$flags"
```
scop, line 56: This PR changes some (but not all) cases of `-W` arg to use single quotes. Per https://github.com/scop/bash-completion/pull/961#discussion_r1186844166, should we do it everywhere?
```suggestion
        _comp_compgen -R -- -W '$flags'
```

completions/find
akinomyoga: I'm unsure about it too.

I've identified the commit ad561ac1c.

```
commit ad561ac1cabc23bc477bddc5f0abb8af92701925
Author: ianmacd <>
Date:   Sun Apr 14 22:41:04 2002 +0000

    - avoid errors when comp{gen,lete} -g don't work on an unpatched bash 2.05
```

So this was a workaround for bash-2.05. We can clean it up now.

completions/bzip2
akinomyoga: Ah, you are right. I'll later look them up in the codebase.

completions/gcc
akinomyoga: It is related also to https://github.com/scop/bash-completion/issues/958#issuecomment-1537214890. I sometimes thought we can keep the *existing* cases to address it all at once later but sometimes wanted to apply the change with another chance to the line. It is admittedly not so consistent. Probably I would like to revert it to the double quotes for now to later address it in a batch PR of similar cases.

scop: On a second thought, let me resolve those quote comments so I can see if something's ready to be merged already.

akinomyoga: I reverted it https://github.com/scop/bash-completion/compare/95eebfc73e82b15693079da1f41d2420eeac1aed..4358f1e5a1e3f9a3a292b6d2b7a7f27f1611907e

completions/bzip2
akinomyoga: Applied in 4358f1e5a1e3f9a3a292b6d2b7a7f27f1611907e

scop (approved):

PR #976, opened 2023-05-10
fix(ssh): optimize traversing ssh config includes

Rogach:
Previously each config file was read with separate `sed` command to extract `Include` directives. This can get quite slow if there are a lot of included files (on my machine it was ~0.4s for each TAB keypress).

Instead of reading each individual file, we can put all the files at a current recursion level into a single sed invocation and extract all the directives in one go.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2111,33 +2111,44 @@ _included_ssh_config_files()
     local configfile i files f
     configfile=$1
 
-    local -a included
-    _comp_split included "$(command sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' "${configfile}")" || return
+    # From man ssh_config:
+    # "Files without absolute paths are assumed to be in ~/.ssh if included
+    # in a user configuration file or /etc/ssh if included from the system
+    # configuration file."
+    # This behavior is not affected by the the including file location -
+    # if the system configuration file is included from the user's config,
+    # relative includes are still resolved in the user's ssh config directory.
+    local relative_include_base
+    if [[ $configfile == /etc/ssh* ]]; then
+        relative_include_base="/etc/ssh"
+    else
+        relative_include_base="$HOME/.ssh"
+    fi
 
-    for i in "${included[@]}"; do
-        # Check the origin of $configfile to complete relative included paths on included
-        # files according to ssh_config(5):
-        #  "[...] Files without absolute paths are assumed to be in ~/.ssh if included in a user
-        #   configuration file or /etc/ssh if included from the system configuration file.[...]"
-        if ! [[ $i =~ ^\~.*|^\/.* ]]; then
-            if [[ $configfile =~ ^\/etc\/ssh.* ]]; then
-                i="/etc/ssh/$i"
-            else
-                i="$HOME/.ssh/$i"
+    local depth=1
+    local -a included
+    local -a include_files
+    included=("$configfile")
+
+    while ((${#included[@]} > 0 && depth++ < 16)); do
+        _comp_split include_files "$(command sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' "${included[@]}")" || return
+        included=()
+        for i in "${include_files[@]}"; do
+            if [[ $i != [~/]* ]]; then
+                i="${relative_include_base}/${i}"
             fi
-        fi
-        __expand_tilde_by_ref i
-        # In case the expanded variable contains multiple paths
-        _comp_expand_glob files '$i'
-        if ((${#files[@]})); then
-            for f in "${files[@]}"; do
-                if [[ -r $f && ! -d $f ]]; then
-                    config+=("$f")
-                    # The Included file is processed to look for Included files in itself
-                    _included_ssh_config_files "$f"
-                fi
-            done
-        fi
+            __expand_tilde_by_ref i
+            # In case the expanded variable contains multiple paths
+            _comp_expand_glob files '$i'
+            if ((${#files[@]})); then
+                for f in "${files[@]}"; do
+                    if [[ -r $f && ! -d $f ]]; then
+                        config+=("$f")
+                        included+=("$f")
+                    fi
+                done
+            fi
+        done
     done
 } # _included_ssh_config_files()
 
--- a/bash_completion
+++ b/bash_completion
@@ -2130,7 +2130,10 @@ _included_ssh_config_files()
     local -a include_files
     included=("$configfile")
 
-    while ((${#included[@]} > 0 && depth++ < 16)); do
+    # Max recursion depth per openssh's READCONF_MAX_DEPTH:
+    # https://github.com/openssh/openssh-portable/blob/5ec5504f1d328d5bfa64280cd617c3efec4f78f3/readconf.c#L2240
+    local max_depth=16
+    while ((${#included[@]} > 0 && depth++ < max_depth)); do
         _comp_split include_files "$(command sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' "${included[@]}")" || return
         included=()
         for i in "${include_files[@]}"; do


bash_completion
```diff
@@ -2112,32 +2112,34 @@ _included_ssh_config_files()
...
+    local -a included_raw
+    local -a included_deeper
+    included=("$configfile")
+
+    while (("${#included[@]}" > 0)); do
+        included_deeper=()
+        _comp_split included_raw "$(command sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' "${included[@]}")" || return
+        for i in "${included_raw[@]}"; do
+            if ! [[ $i =~ ^\~.*|^\/.* ]]; then
+                if [[ $configfile =~ ^\/etc\/ssh.* ]]; then
```
akinomyoga, line 2124: Shouldn't this `configfile` be replaced by the filename where the line `$i` originates from?

Rogach: Counterintuitively, no, it's shouldn't. Relative `Include` is not resolved relatively to the including file. From the manual:

```
Files without absolute paths are assumed to be in ~/.ssh if included in a user configuration file
or /etc/ssh if included from the system configuration file.
```

And here's the supporting source code: https://github.com/openssh/openssh-portable/blob/master/readconf.c#L1878

So in fact the original implementation of `_included_ssh_config_files` was buggy since it would have attempted to resolve relative Include's in the wrong place.

akinomyoga: > Relative `Include` is not resolved relatively to the including file.

`configfile` here is not used to resolve the relative path.

> ```
> Files without absolute paths are assumed to be in ~/.ssh if included in a user configuration file
> or /etc/ssh if included from the system configuration file.
> ```

The original code seems to try to implement this by assuming that the files starting with /etc/ssh is the system configuration file and the others are the user configuration files, though I'm not sure if that assumption is correct.

But the above description in the manual is somewhat ambiguous. Do you mean that e.g. a relative `Include` directive in a system configuration file finds a file in `~/.ssh` if the system configuration file is originally included by another `Include` in a user configuration file? That's what this PR implements. Hmm, at a quick glance in `readconf.c`, it seems to be the case, though I haven't carefully read it.

Rogach: >> Relative Include is not resolved relatively to the including file.
> configfile here is not used to resolve the relative path to the including file.
Yes, but it is used to determine if the file is resolved relatively to /etc/ssh or ~/.ssh.

> Do you mean that e.g. a relative Include directive in a system configuration file finds a file in ~/.ssh if the system configuration file is originally included by another Include in a user configuration file?
Yes. The `flags` variable is set only once, at the top level, in ssh.c:567, and SSHCONF_USERCONF bit is retained when recursing.

akinomyoga: Thank you for the confirmation!

This `ssh` behavior is tricky, so I think we need to leave comments in a code comment and in the commit message. Then, the main subject of the commit becomes like `fix(ssh): resolve relative ssh_config Include correctly` instead of `perf(ssh): optimize traversing ssh config includes` (Or you may separate the commit into two).

akinomyoga (changes requested):
Thank you! These are style and minor changes.

Rogach:
@akinomyoga Implemented the requested changes.

> The option -l (linewise splitting) may be specified to _comp_split

This fails the tests, because single `Include` may specify multiple included files on the same line (separated by spaces).

akinomyoga:
> This fails the tests, because single `Include` may specify multiple included files on the same line (separated by spaces).

Thank you for pointing this out. Hmm, so it means that each element of the resulting array of `_comp_split` does not necessarily correspond to one Include directive. Then, the variable name might better to be something like `include_files` rather than `include_directives`, but it's up to you whether to change it to a better one.

Rogach:
Renamed the variable.

bash_completion
```diff
@@ -2111,33 +2111,44 @@ _included_ssh_config_files()
     local configfile i files f
     configfile=$1
 
+    # From man ssh_config:
+    # "Files without absolute paths are assumed to be in ~/.ssh if included in a user configuration
+    # file or /etc/ssh if included from the system configuration file."
+    # This behavior is not affected by the the including file location -
+    # if system configuration file is included from user's config relative includes are still
+    # resolved in user's ssh config directory.
```
akinomyoga, line 2119:
```suggestion
    # This behavior is not affected by the including file location -
    # if the system configuration file is included from the user's config, relative includes are still
    # resolved in the user's ssh config directory.
```

Also, the codes, in particular comments, are supposed to be folded on column 80 so that the line has 79 characters at maximum.

bash_completion
```diff
@@ -2111,33 +2111,44 @@ _included_ssh_config_files()
...
+    local relative_include_base
+    if [[ $configfile =~ ^\/etc\/ssh.* ]]; then
```
akinomyoga, line 2121:
```suggestion
    if [[ $configfile == /etc/ssh* ]]; then
```

bash_completion
```diff
@@ -2111,33 +2111,44 @@ _included_ssh_config_files()
...
-        if ! [[ $i =~ ^\~.*|^\/.* ]]; then
-            if [[ $configfile =~ ^\/etc\/ssh.* ]]; then
-                i="/etc/ssh/$i"
-            else
-                i="$HOME/.ssh/$i"
+    local -a include_files
+    included=("$configfile")
+
+    while (("${#included[@]}" > 0)) && ((depth < 16)); do
+        depth=$((depth + 1))
```
akinomyoga, line 2133:
```suggestion
    while ((${#included[@]} > 0 && depth++ < 16)); do
```

bash_completion
```diff
@@ -2111,33 +2111,44 @@ _included_ssh_config_files()
...
+        _comp_split include_files "$(command sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' "${included[@]}")" || return
+        included=()
+        for i in "${include_files[@]}"; do
+            if ! [[ $i =~ ^\~.*|^\/.* ]]; then
```
akinomyoga, line 2137:
```suggestion
            if [[ $i != [~/]* ]]; then
```

akinomyoga (approved):
Thank you for quick responses and updating!

These are cosmetic ones. With these addressed, I'd approve it.

Rogach:
Applied the requested changes.

bash_completion
akinomyoga: Thanks for updating! You kept the quoting `"..."` for `${#included[@]}` twice. Is there a reason?

Rogach: No, simply missed it. Will fix.

akinomyoga: Thanks

akinomyoga (approved):

Rogach:
> However, the commit message seems to talk about an actual fix in addition to a performance improvement, whereas the PR title here and the commentary seems to refer only to the perf improvement.
> If there's a fix in here, could we add a test case showcasing what was broken before this and fixed now?

Story about a fix is hidden in the resolved diffs. Basically the problem we found during discussion was that if a user config file included some config in `/etc/ssh/`, and that config used relative include to include another config, this last Include must be resolved relative to users `~/.ssh`. Previous implementation would have resolved it relatively to `/etc/ssh/`.

I don't know how to add a test for this, since the scenario involves creating an ssh config file somewhere under `/etc/ssh/` and that isn't happening.

> Or if there's no fix, could we rephrase the commit message some to clarify that? Also, if there's no actual fix here, the conventional commit type would be better set as perf instead of fix.

Original commit message was about perfomance improvement (that's probably the more interesting part anyway, since the failing scenario seems extremely unlikely). However @akinomyoga suggested that it would be changed to talk about the include fix. Should I change it back?

akinomyoga:
There is a fix. At least the commit changes the behavior, so, in my opinion, we should mark it as `fix` rather than `perf`.

Actually, the ideal way is to separate the commit into the `fix` one and the `perf` one so that we could later track the relevant changes when necessary.

Rogach:
> Actually, the ideal way is to separate the commit into the fix one and the perf one so that we could later track the relevant changes when necessary.

Ideally yes, but in this particular case it will be difficult. There are two possible commit orderings:

1. "first fix, then perf": in the "fix" commit we'll need to add an additional argument to _included_ssh_config_files, something like top_level_config, to pass around the include base. Then the "perf" commit will be rewriting recurisve version into a loop, removing this additional argument.
2. "first perf, then fix": here I'm not even sure how to keep the old behavior in the non-recursive version. Something like making another array of include bases? I might be okay with doing it in some other language, but in Bash array handling is will be too difficult for me and I might easily make a lot of mistakes.

I support marking commit as "fix". Or maybe we can mark it as "fix,perf"?

akinomyoga:
For me it's optional to separate the commit; I wouldn't request you to separate it. But if you are willing to separate the commit, either option works. If it were me, I think I'd use option 1.

Rogach:
I really don't want to separate it :)

scop:
Ok, merged as is. Thanks!

Issue #926, opened 2023-04-12
Long lag when completing the command argument of ssh


szsam:
## Describe the bug
There is a long lag (30 seconds) when completing the command argument of ssh.

## To reproduce

Type 'ssh user@host ', then hit tab. The shell will freeze for 30 seconds.

## Expected behavior

The completion should be fast. The 30-second lag is unacceptable.

## Versions (please complete the following information)

- [x] Operating system name/distribution and version: Ubuntu 20.04.5, WSL2
- [x] bash version, `echo "$BASH_VERSION"`: 5.0.17(1)-release
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.10

## Additional context
It is the following line that takes 30 seconds, where `cur` is `""`. The output of the `compgen -c -- ''` command contains 10k+ lines on my system.
https://github.com/scop/bash-completion/blob/ac68847b33a1cd030ccbb1accaa5015ae4a9a797/completions/ssh#L356

## Debug trace
[bash_completion_log.txt](https://github.com/scop/bash-completion/files/11207775/bash_completion_log.txt)

szsam:
I suggest that we do not try to complete when `cur` is an empty string.

szsam:
I'm using Ubuntu on WSL (Windows Subsystem for Linux). That might be the reason why the `compgen` is so slow.

akinomyoga:
Similar codes can be found in more completions within the codebase.

```bash
./completions/_svn:42:                COMPREPLY=($(compgen -c -- "$cur"))
./completions/cpio:27:            COMPREPLY=($(compgen -c -- "$cur"))
./completions/dpkg:107:            COMPREPLY=($(compgen -c -- "$cur"))
./completions/fio:78:            COMPREPLY=($(compgen -c -- "$cur"))
./completions/fio:91:            COMPREPLY=($(compgen -c -- "$cur"))
./completions/firefox:36:            COMPREPLY=($(compgen -c -- "$cur"))
./completions/man:33:            COMPREPLY=($(compgen -c -- "$cur"))
./completions/mr:57:                COMPREPLY=($(compgen -c -- "${cur}"))
./completions/mussh:42:            COMPREPLY+=($(compgen -c -- "$cur"))
./completions/mypy:28:            COMPREPLY=($(compgen -c -- "${cur:-py}"))
./completions/perlcritic:37:            COMPREPLY=($(compgen -c -- "$cur"))
[...]

PR #978, opened 2023-05-12
feat: no empty command completion if `no_empty_cmd_completion` is on

scop:
On some systems, generating the entire list of available commands takes a long time. `no_empty_cmd_completion` is semantically close enough to piggyback on it here.

Closes https://github.com/scop/bash-completion/issues/926

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2708,7 +2708,7 @@ _complete_as_root()
 _comp_compgen_commands()
 {
     [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return
-    _comp_compgen -- -c
+    _comp_compgen -- -c -o plusdirs
     # for e.g. spaces in paths to and in command names
     ((${#COMPREPLY[@]} == 0)) || compopt -o filenames
 }
--- a/bash_completion
+++ b/bash_completion
@@ -2706,7 +2706,7 @@ _complete_as_root()
 # @since 2.12
 _comp_compgen_commands()
 {
-    [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return
+    [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return 1
     # -o filenames for e.g. spaces in paths to and in command names
     _comp_compgen -- -c -o plusdirs && compopt -o filenames
 }
--- a/bash_completion
+++ b/bash_completion
@@ -2702,6 +2702,15 @@ _complete_as_root()
     [[ $EUID -eq 0 || ${root_command-} ]]
 }
 
+# Complete on available commands, subject to `no_empty_cmd_completion`.
+#
+# @since 2.12
+_comp_compgen_commands()
+{
+    [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return
+    _comp_compgen -- -c
+}
+
 # @since 2.12
 _comp_longopt()
 {
--- a/bash_completion
+++ b/bash_completion
@@ -2709,6 +2709,8 @@ _comp_compgen_commands()
 {
     [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return
     _comp_compgen -- -c
+    # for e.g. spaces in paths to and in command names
+    ((${#COMPREPLY[@]} == 0)) || compopt -o filenames
 }
 
 # @since 2.12
--- a/bash_completion
+++ b/bash_completion
@@ -2707,9 +2707,8 @@ _complete_as_root()
 _comp_compgen_commands()
 {
     [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return
-    _comp_compgen -- -c -o plusdirs
-    # for e.g. spaces in paths to and in command names
-    ((${#COMPREPLY[@]} == 0)) || compopt -o filenames
+    # -o filenames for e.g. spaces in paths to and in command names
+    _comp_compgen -- -c -o plusdirs && compopt -o filenames
 }
 
 # @since 2.12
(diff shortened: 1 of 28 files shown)


test/t/unit/test_unit_compgen_commands.py
scop, line 1: Tests do not work but timeout for some reason, need to get that fixed up.

bash_completion
```diff
@@ -2557,6 +2557,15 @@ _complete_as_root()
     [[ $EUID -eq 0 || ${root_command-} ]]
 }
 
+# Complete on available commands, subject to `no_empty_cmd_completion`.
+#
+# @since 2.12
+_comp_compgen_commands()
+{
+    [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return
+    _comp_compgen -- -c
```
scop, line 2566: This does not escape for example spaces in command names in `$PATH` properly. We apparently try that by setting `compopt -o filenames` in some completions (but not all), but that doesn't help.

It was similarly broken before this change too, but I'd like to either fix that while we work on this, or make it consistent if we do/don't try the filenames compopt.

scop: I'm not sure how I tested before, but `compopt -o filenames` does seem to help.

akinomyoga (commented):
There are some other cases where we might suppress command name generations on empty completions.

```bash
./bash_completion:2497:        _comp_compgen -- -d -c
./completions/_su:23:            _comp_compgen -- -d -c
./completions/gdb:23:            _comp_compgen -- -d -c
```

Currently, `_comp_compgen_commands` only generates the command names, but maybe it could also generate directories by `plusdirs` by default.

scop:
Good points. Added `-o plusdirs` and addressed the noted remaining ones.

The note in gdb completion regarding bash builtins etc being include is valid for quite a few other things as well, that's something we may want to address sometime.

bash_completion
```diff
@@ -2614,6 +2612,17 @@ _complete_as_root()
...
 
+# Complete on available commands, subject to `no_empty_cmd_completion`.
+#
+# @since 2.12
+_comp_compgen_commands()
+{
+    [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return
+    _comp_compgen -- -c -o plusdirs
+    # for e.g. spaces in paths to and in command names
+    ((${#COMPREPLY[@]} == 0)) || compopt -o filenames
```
akinomyoga, line 2623:
```suggestion
    # -o filenames for e.g. spaces in paths to and in command names
    _comp_compgen -- -c -o plusdirs && compopt -o filenames
```

For the CI error.

akinomyoga: I now realized that I haven't added `@return` in the doc comment of `_comp_compgen`, but `_comp_compgen -- [options]` is intended to succeed only when at least one completion is generated, and this behavior is already used in the codebase.

It is mentioned in [api-and-naming.md in #993](https://github.com/scop/bash-completion/pull/993/files#diff-0cb3af15071817f620202dd47344251e6a8bff22af238511159fb03ef2e886f0R161-R163). I added https://github.com/scop/bash-completion/pull/993/commits/ed8e81e246d20c8ecf7f943dfe7f8e41060504ac in #993.

bash_completion
```diff
@@ -2614,6 +2612,17 @@ _complete_as_root()
     [[ $EUID -eq 0 || ${root_command-} ]]
 }
 
+# Complete on available commands, subject to `no_empty_cmd_completion`.
+#
```
akinomyoga, line 2616: I think we can define the behavior of `@return`, though there are choices.

bash_completion
```diff
@@ -2614,6 +2612,17 @@ _complete_as_root()
...
+# @since 2.12
+_comp_compgen_commands()
+{
+    [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return
```
akinomyoga, line 2620: If we make it the same as `_comp_compgen -- [...]` (i.e., 0 only when at least one completion is generated), we can return `1` for `no_empty_cmd_completion`.

```suggestion
    [[ ! ${cur-} ]] && shopt -q no_empty_cmd_completion && return 1
```

scop: SGTM in principle, although returning 1 feels a bit "wrong" in the sense that it's not an error, it's just that the completions are inhibited because of config. There probably isn't such a use case in-tree at least yet, but a pattern like `_comp_compgen_command || do_something_else_to_generate_completions` doesn't seem completely unrealistic, but we might not want to enter the "or" branch when the nonzero is due to config. In that sense, returning 0 here could be worth considering.

But I guess it could be argued the other way in favor of a nonzero. And that'd be kind of nice for consistency, too. No strong feelings on this one here.

akinomyoga: I have already let `_comp_split` and `_comp_compgen -- [...]` use exit status 1 when no completions are generated although it's not an error. Usage error is notified by status 2. A typical use case for `return 1` that I assume is `_comp_compgen -v command && process "${arr[@]}"`. But if there is another use case, there's no problem in returning 0 here.

I guess whether to run the "or" branch depends on the actual action in `do_something_elese_to_generate_completions`. For example, if `do_something_elese_to_generate_completions` has a small computational cost, I think it can be run even if `_comp_compgen_command` is canceled because the purpose of `no_empty_cmd_completion` is not to block the completion.

akinomyoga (approved):

bash_completion
akinomyoga:
```suggestion
# @return True (0) if one or more completions are generated, or otherwise False
# (1).  Note that it returns 1 even when the completion generation is canceled
# by `shopt -s no_empty_cmd_completion`.
#
```

PR #981, opened 2023-05-15
test(dcop): skip if basic invocation errors out

scop:
It's highly unlikely that a DCOP server would be running in containers.

### DIFF
--- a/test/t/test_dcop.py
+++ b/test/t/test_dcop.py
@@ -2,6 +2,8 @@ import pytest
 
 
 class TestDcop:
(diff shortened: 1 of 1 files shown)


akinomyoga (approved):
Does that mean, even if the command `dcop` exists in a container, it doesn't function as expected?

scop:
Yes. This was uncovered by the work on #979 (just like the pyston one) -- dcop completion is a symlink so we never actually ended up running it in CI.
```
$ docker run --rm -it ghcr.io/scop/bash-completion/test:fedoradev dcop
ERROR: Couldn't attach to DCOP server!
```

PR #985, opened 2023-05-21
fix: add misc fixes before `_comp_compgen_split`

akinomyoga:
Before applying `_comp_compgen_split` mentioned in https://github.com/scop/bash-completion/issues/958#issuecomment-1537214890, I would like to add the following fixes that are overlapping with it.


scop (approved):
Thanks!

PR #986, opened 2023-05-21
doc: add rules for quoting

### LINKED ISSUE #932: Decide and document quoting rule
> Hmm, I suppose this means we should be quoting all arguments containing variables to any commands or functions we invoke? This PR does not address them all throughout the codebase, but we can take care of them in followup PR's.
>
> I'm starting to feel that we should write up some rules about all this stuff in our docs somewhere, perhaps styleguide.md. And it would be great if there was some tooling that would help by flagging or fixing undesired uses. I wonder if shellcheck or shfmt would accept some of that stuff.
>
> _Originally posted by @scop in https://github.com/scop/bash-completion/pull/931#discussion_r1174403080_

akinomyoga:
Resolve #932 

Note: `shfmt` tries to remove the quoting of `[[ "$*" ]]`, which is needed to avoid the behavior change in bash 4.2 => 4.3. There is a similar problem with `[[ "${arr[*]}" ]]`. Actually, the current codebase doesn't have cases of `[[ "$*" ]]` or `[[ "${arr[*]}" ]]` where `shfmt` tries to remove the necessary quoting. But this could conflict in the future.

### DIFF
--- a/doc/styleguide.md
+++ b/doc/styleguide.md
@@ -122,3 +122,62 @@ it.
 ## Function and variable names
 
 See [API and naming](api-and-naming.md).
+
+## Quoting of words
+
+To avoid unexpected word splitting and pathname expansions, an argument of
+commands needs to be properly quoted when it contains shell expansions such as
+`$var`, `$(cmd)`, and `$((expr))`.
+
+When one intentionally wants word splitting and pathname expansions, one should
+consider using the utility functions provided by bash-completion.  To safely
+split a string without being affected by non-standard `IFS` and pathname
+expansions, use the shell function `_comp_split`.  To safely obtain filenames
+by pathname expansions without being affected by `failglob`, etc., use the
+shell function `_comp_expand_glob`.  Note that `_comp_expand_glob` should be
+always used for the pathname patterns even if the pattern does not contain
+shell expansions.
+
+In the following contexts, the quoting to suppress word splitting and pathname
+expansions are not needed.
+
+- The right-hand sides of variable assignments ... `v=WORD` (e.g. `v=$var`)
+- The arguments of conditional commands ... `[[ WORD ]]` (e.g. `[[ $var ]]`)
+- The argument specified to `case` statement ... `case WORD in foo) ;; esac`
+  (e.g. `case $var in foo) ;; esac`)
+
+In bash-completion, we do not quote them by default.  However, there are
+exceptions where the quoting is still needed for other reasons.
+
+- When the word *directly* contains shell special characters (space, tab,
+  newline, or a character from ``;|&()<>\\$`'"#!~{``), these characters need to
+  be quoted.  The "*directly*" means that the special characters produced by
+  shell expansions are excluded here.  For example, when one wants to include a
+  whitespace as a part of the value of the word, the right-hand side can be
+  quoted as `v="a b"`.
+- An empty word (i.e., the word whose value is an empty string) is specified by
+  `""`.  The right-hand side of an assignment technically can be an empty
+  string as `var=`, but we still use `var=""` there because `shellcheck`
+  suggests that e.g. `var= cmd` is confusing with `var=cmd`.
+- `$*` and `${array[*]}` need to be always quoted because they can be affected
+  by the word splitting in bash <= 4.2 even in the above contexts.
+- In the following contexts, double-quoting of shell expansions is needed
+  unless the result of expansions is intentionally treated as glob patterns or
+  regular expressions.
+  - The right-hand sides of `==`, `!=`, and `=~` in the conditional commands
+    ... `[[ word == "$var" ]]`
+  - The case patterns ... `case word in "$var") ;; esac`
+
+Note: Here strings `cat <<<$var` are also supposed to be safe against word
+splitting and pathname expansions without quoting, but bash <= 4.3 has a bug
+[1], so they need to be quoted for as long as we support bash 4.3.
+
+- [koalaman/shellcheck#1009 (comment)](https://github.com/koalaman/shellcheck/issues/1009#issuecomment-488395630)
+
+There are also preferences on the type of quoting, which are though not too
+strict.  We prefer to use double quotes over single quotes by default.  When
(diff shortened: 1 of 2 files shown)


doc/styleguide.md
```diff
@@ -122,3 +122,43 @@ it.
...
+- When the word contains shell special characters ``/[ \t\n;|&()<>\\$`'"#!~{]/``,
+  these characters need to be quoted
+- `$*` and `${array[*]}` need to be always quoted because they can be affected
+  by the word splitting in bash <= 4.2 even in the above contexts.
+- In the following contexts, double-quoting of shell expansions is needed
+  unless the result of expansions is intentionally treated as glob patterns or
+  regular expressions.
+  - The right-hand sides of `==`, `!=`, and `=~` in the conditional commands
+    ... `[[ word == "$var" ]]`
+  - The case patterns ... `case word in $var) ;; esac`
```
scop, line 158: For consistency with the preceding example?
```suggestion
  - The case patterns ... `case word in "$var") ;; esac`
```

doc/styleguide.md
```diff
@@ -122,3 +122,43 @@ it.
...
+
+Note: Here strings `cat <<<$var` are also supposed to be safe against word
+splitting and pathname expansions without quoting, but bash <= 4.3 has a bug
+[1], so they need to be quoted as far as we support bash 4.3.
```
scop, line 162:
```suggestion
[1], so they need to be quoted for as long as we support bash 4.3.
```

doc/styleguide.md
```diff
@@ -122,3 +122,43 @@ it.
...
+In the following contexts, the quoting to suppress word splitting and pathname
+expansions are not needed.
+
+- The right-hand sides of variable assignments ... `v=$var`
+- The arguments of conditional commands ... `[[ $var ]]`
+- The argument specified to `case` statement ... `case $var in foo) ;; esac`
+
+However, there are exceptions where the quoting is still needed.
+
+- When the word contains shell special characters ``/[ \t\n;|&()<>\\$`'"#!~{]/``,
```
scop, line 149: What is "the word" here, could we rephrase/clarify this some, perhaps through an example?

scop (approved):
Good stuff, pre-approved with some trivial changes/comments to address.

doc/styleguide.md
akinomyoga: Hmm. The "word" here means the range of characters in the shell code that the user wants to treat as a single shell WORD. WORD is defined in [POSIX XCU 2.3 Token Recognition](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_03) and in [XCU 2.10.2 Shell Grammar Rules](https://pubs.opengroup.org/onlinepubs/9699919799/). So, if the text range contains unquoted shell special characters, that is actually no longer a shell word. In that sense, the word cannot contain these characters without quoting.

akinomyoga:
> > `shfmt` tries to remove the quoting of `[[ "$*" ]]`
> 
> Is this something that should be reported as a bug against shfmt?

I think it should be reported to shfmt if it hasn't been reported so far, though I'm not sure we could claim that is a bug. The behavior of bash <= 4.2 may be considered as a bug of Bash, and it's possible that `shfmt` intentionally excludes workarounds of bash bugs. Anyway, it's worth reporting.

doc/styleguide.md
scop: Sorry, I was probably unclear, let me try to elaborate. The bullet point examples above contain uses of `$var` (which are not "word"s AFAIU), but this bullet talks about exceptions, and I read them being exceptions to the above points. But because the above points do not contain any "word" uses -> confusion ensues exactly to which are these exceptions for.

So reading this I'm left wondering if quoting is necessary e.g. (using the slash as an example) for literal `foo/bar` so we'd need to do `v="foo/bar"`, or uses of variables whose values might contain the slash (in which case we might not know, and it would arguably to be better to just always quote).

akinomyoga: Thank you for your clarification! I still planned to update the corresponding part, but I was thinking about how I should fix it. Now I understand the problems.

In my mind, I was talking about "the words that *contain* shell expansions" in the entire section as in the section title, and `$var` was the simplest case of such a word. However, as you pointed out, the examples seem to just talk about the shell expansions so are confusing. Then, I guess adding examples would be fine to avoid the confusion.

> So reading this I'm left wondering if quoting is necessary e.g. (using the slash as an example) for literal `foo/bar` so we'd need to do `v="foo/bar"`,

Sorry for confusing you, but the slash is not a part of the set of special characters. The enclosing `[` and `]` are also not a part of the special characters. I have a habit of indicating a regular expression by `/.../`. (Actually, I initially wrote `` \t\n;|&()<>\\$`'"#!~{`` without `/[` and `]/`, but the pre-commit hook (probably PyMarkdown) has complained that a whitespace character cannot be included at the beginning of the code span.) I'll try to fix it.

akinomyoga: ca0fde09

- I changed the section title to `Quoting of words` to extend the target to the general words.
- I made it explicit that we avoid unnecessary quoting.
- I explained the exception of shell special characters in more detail.
- I added the case of an empty word as an exception.
- I also added our preferences in the type of quoting as suggested in [#932 (comment)](https://github.com/scop/bash-completion/issues/932#issuecomment-1556879298).

scop: Looks great to me :+1:

akinomyoga:
> Would you like to have a go at reporting the shfmt issue discussed above? I can do it as well, but I guess you know the details better by heart.

I'll think about it. Before reporting, I'd like to think about whether it is technically reasonable to implement it in the current codebase of `mvdan/sh`. I guess we need to add some codes in [this case branch](https://github.com/mvdan/sh/blob/9e323b8552f73daa5fd63f7dff0dfbc3f7bc3af8/syntax/simplify.go#L115), but if the implementation would be too complicated, the suggestion might be just left untouched just like https://github.com/mvdan/sh/issues/996.

PR #987, opened 2023-05-22
docs(api-and-naming): use `REPLY` for return variable instead of `ret`

akinomyoga:
https://github.com/scop/bash-completion/discussions/537#discussioncomment-5961630

### DIFF
--- a/doc/api-and-naming.md
+++ b/doc/api-and-naming.md
@@ -7,17 +7,25 @@ inject them to the `COMPREPLY` array variable, as required for completions to
 work.
 
 Most other functions make use of "output" variables, i.e. assign values to
-them. The most common one of these is named `ret`. Consult the commentary
-before each function in the source to find out the specific names.
-`local`izing output variables before invoking a function that populates them
-is the caller's responsibility.
-Note that if calling multiple functions that assign output to the same variable
-during one completion function run, each result should be copied to another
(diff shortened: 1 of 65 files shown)


doc/api-and-naming.md
```diff
@@ -7,7 +7,7 @@ inject them to the `COMPREPLY` array variable, as required for completions to
 work.
 
 Most other functions make use of "output" variables, i.e. assign values to
-them. The most common one of these is named `ret`. Consult the commentary
+them. The most common one of these is named `REPLY`. Consult the commentary
```
scop, line 10: Could we add a few words here why we're using `REPLY` for this? Uppercase is unconventional, and would be good to note the specific choice of word here as well.

Looks good to me, pre-approved and good to merge with this addressed.

scop (approved):

doc/api-and-naming.md
akinomyoga: Thanks for the suggestion. I separated the sentence of `REPLY` into a new paragraph and added a description (update: [`774454d0..2993392a4`](https://github.com/scop/bash-completion/compare/774454d0e0b9e59c4f0b4a1dbf90e88715fa8668..2993392a402d642effd7a7c27161d4aceb7ed847)).

PR #988, opened 2023-05-23
refactor: more util/xfunc primary command convention adjustments

scop:
Remaining found cases as mentioned in https://github.com/scop/bash-completion/pull/965#issuecomment-1558479426


akinomyoga (approved):
If the `pathcmd` modification of `PATH` is widely used with `_comp_xfunc`, maybe it could be implemented as an option to `_comp_xfunc`? But I also see use cases with other completion generations, so it could be better implemented as an option to `_comp_compgen` just like `-C dir`. Anyway, such a refactoring can be done in a later PR. We can easily search the related part by `grep pathcmd`.

PR #993, opened 2023-05-26
doc(api-and-naming): add rules for xfuncs and generators

akinomyoga:
Document the naming rules and implementation advices for xfuncs and `_comp_compgen_NAME`

https://github.com/scop/bash-completion/pull/954#discussion_r1186805078
https://github.com/scop/bash-completion/pull/964
https://github.com/scop/bash-completion/discussions/962#discussioncomment-5832685

Note: `_comp_compegn [-i CMD|-x CMD]` are not yet impl;emented.

### DIFF
--- a/doc/api-and-naming.md
+++ b/doc/api-and-naming.md
@@ -41,15 +41,17 @@ deprecated in.
 Due to its nature, bash-completion adds a number of functions and variables in
 the shell's environment.
 
-|                                     | `bash_completion`   | `completions/*`                                                            |
-|:------------------------------------|:--------------------|:---------------------------------------------------------------------------|
-| public configuration variables      | `BASH_COMPLETION_*` | `BASH_COMPLETION_CMD_${Command^^}_${Config^^}`                             |
-| private non-local variables         | `_comp__*`          | `_comp_cmd_${Command}__${Data}`                                            |
-| private non-local mutable variables | `_comp__*_mut_*`    | `_comp_cmd_${Command}__mut_${Data}`                                        |
-| exporter function local variables   | `_*` (not `_comp*`) | `_*` (not `_comp*`)                                                        |
-| public/exported functions           | `_comp_*`           | `_comp_cmd_${Command}` (functions for `complete -F`)                       |
-|                                     |                     | `_comp_xfunc_${Command}_${Utility}` (functions for use with `_comp_xfunc`) |
-| private/internal functions          | `_comp__*`          | `_comp_cmd_${Command}__${Utility}` (utility functions)                     |
+|                                     | `bash_completion`       | `completions/*`                                                                       |
+|:------------------------------------|:------------------------|:--------------------------------------------------------------------------------------|
+| public configuration variables      | `BASH_COMPLETION_*`     | `BASH_COMPLETION_CMD_${Command^^}_${Config^^}`                                        |
+| private non-local variables         | `_comp__*`              | `_comp_cmd_${Command}__${Data}`                                                       |
+| private non-local mutable variables | `_comp__*_mut_*`        | `_comp_cmd_${Command}__mut_${Data}`                                                   |
+| exporter function local variables   | `_*` (not `_comp*`)     | `_*` (not `_comp*`)                                                                   |
+| public/exported functions           | `_comp_*`               | `_comp_cmd_${Command}` (functions for `complete -F`)                                  |
+|                                     |                         | `_comp_xfunc_${Command}_${Utility}` (functions for use with `_comp_xfunc`)            |
+|                                     | `_comp_compgen_${Name}` | `_comp_xfunc_${Command}_compgen_${Name}` (generators for use with `_comp_compgen -x`) |
+| private/internal functions          | `_comp__*`              | `_comp_cmd_${Command}__${Utility}` (utility functions)                                |
+|                                     |                         | `_comp_cmd_${Command}__compgen_${Name}` (generators for use with `_comp_compgen -i`)  |
 
 `${Command}` refers to a command name (with characters not allowed in POSIX
 function or variable names replaced by an underscore), `${Config}` the name of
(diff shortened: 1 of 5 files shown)


doc/api-and-naming.md
```diff
@@ -110,3 +112,64 @@ distinctive and clash free enough.
...
+Exported functions (xfunc) are the functions defined in completion files for
+specific commands but exposed to other completion files.  The xfuncs have the
+name `_comp_xfunc_CMD_NAME` where `CMD` is the name of the associated command,
+and `NAME` is the name of the utility.  The other functions defined in specific
+completion files are considered private and should not be called outside the
+file.
+
+The xfuncs can be called by `_comp_xfunc CMD NAME ARGS` from external files.
+The xfuncs are supposed to be directly called as `_comp_xfunc_CMD_NAME ARGS`
+from the same file where they are defined.
```
scop, line 127:
```suggestion
from the same file where they are defined, or if they wrap a `_comp_cmd_NAME__*`
function, that one should be called directly instead.
```

doc/api-and-naming.md
```diff
@@ -110,3 +112,64 @@ distinctive and clash free enough.
...
+
+- The `_comp_compgen -- COMPGEN_ARGS` returns whether there is at least one
+  completions.  This is useful when one wants to reuse the array content with
+  `"${tmp[@]}"` avoiding `nounset` error.
+- Some use other rules for the exit status. E.g., `help` and `usage` return
+  whether there were options *before* filtering by cur. This is used for
+  `_comp_compgen_help || _comp_compgen_usage`.
+
+Whether to clear the target variable on the runtime error (when `-a` is not
+specified in `OPTS`) is implementation-defuned. On the other hand, the
```
scop, line 168:
```suggestion
specified in `OPTS`) is implementation-defined. On the other hand, the
```

scop (commented):
Looks mostly good to me, just a few small comments.

Kind of a head scratcher is that per this definition, I guess the majority of functions would actually be generators, and we'd be up for another round of renamings to bring them up to spec :thinking:

Another thing is choosing between options 2 and 3 for internal generators per https://github.com/scop/bash-completion/discussions/962#discussioncomment-5841276

akinomyoga:
> Kind of a head scratcher is that per this definition, I guess the majority of functions would actually be generators, and we'd be up for another round of renamings to bring them up to spec 🤔

For the utility functions defined in `bash_completion`, I was thinking about adjusting them at the same time as the renaming. E.g., directly renaming `_filedir => _comp_compgen_filedir` instead of `_filedir => _comp_filedir => _comp_compgen_filedir`. For other functions defined in specific completion files, as far as the function is used only in a particular way (e.g., the target is fixed to be `COMPREPLY`, etc.), I was thinking that they don't necessarily need to be updated to be generators soon.

> Another thing is choosing between options 2 and 3 for internal generators per [#962 (reply in thread)](https://github.com/scop/bash-completion/discussions/962#discussioncomment-5841276)

I left a comment at https://github.com/scop/bash-completion/discussions/962#discussioncomment-6023519.

akinomyoga:
9cd10f79 I added an implementation of `-i cmd` and `-x cmd`.

doc/api-and-naming.md
```diff
@@ -110,3 +112,65 @@ distinctive and clash free enough.
...
+A generator function should replace the existing content of the variable by
+default.  When the appending behavior is favored, the caller should specify it
+through `_comp_compgen -a NAME`.  The generator function do not need to process
+it because internal `_comp_compgen` calls automatically reflects the option
+`-a` specified to the outer calls of `_comp_compgen`.
+
+The exit status is implementation-defined.
+
+- The `_comp_compgen -- COMPGEN_ARGS` returns whether there is at least one
+  completions.  This is useful when one wants to reuse the array content with
```
scop, line 162:
```suggestion
  completion.  This is useful when one wants to reuse the array content with
```

doc/api-and-naming.md
```diff
@@ -110,3 +112,65 @@ distinctive and clash free enough.
...
+  `"${tmp[@]}"` avoiding `nounset` error.
+- Some use other rules for the exit status. E.g., `help` and `usage` return
+  whether there were options *before* filtering by cur. This is used for
+  `_comp_compgen_help || _comp_compgen_usage`.
+
+Whether to clear the target variable on the runtime error (when `-a` is not
```
scop, line 168: Above, it was noted that
> One should not directly modify or reference the target variable.

This seems to be in conflict with that. If we don't modify or reference it, we certainly shouldn't clear it either?

Also,
```suggestion
Whether to clear the target variable on runtime error (when `-a` is not
```

doc/api-and-naming.md
```diff
@@ -110,3 +112,65 @@ distinctive and clash free enough.
...
+specified in `OPTS`) is implementation-defined. On the other hand, the
+generator function should not leave any side effects in the target variable on
```
scop, line 170: Above mentioned conflict similarly here?

doc/api-and-naming.md
```diff
@@ -110,3 +112,65 @@ distinctive and clash free enough.
...
+the usage error.
```
scop, line 171:
```suggestion
usage error.
```

scop (approved):
Looks good to me, pre-approved after addressing the apparent conflict in the docs with regards to referencing/modifying target vars in generators.

doc/api-and-naming.md
akinomyoga: Thanks!

doc/api-and-naming.md
akinomyoga: I was assuming the case that the target variable is indirectly modified by premature internal calls of `_comp_compgen_NAME` inside the generators. I'll clarify that.

akinomyoga: 27cd492124fad2fb18c491d43044769d51b5dfaf

doc/api-and-naming.md
akinomyoga: 27cd492124fad2fb18c491d43044769d51b5dfaf

scop:
Cheers :+1:

akinomyoga:
Thanks!

PR #999, opened 2023-06-07
fix(lvm): update call to sub-command

aaronkrogers:
This fixes an error breaking completion with lvm when called with the `lvm <subcommand>` format. It builds the internal function call to the new function name format and passes in the command line arguments.

Before:
```
lvm lvcreate -bash: _lvcreate: command not found
```
After:
```
lvm lvcreate -
--activate                --ignoreactivationskip    --raidintegrity
--addtag                  --ignoremonitoring        --raidintegrityblocksize
--alloc                   --journal                 --raidintegritymode
--autobackup              --lockopt                 --readahead
```
```
lvcreate -
--activate                --ignoreactivationskip    --raidintegrity
--addtag                  --ignoremonitoring        --raidintegrityblocksize
--alloc                   --journal                 --raidintegritymode
--autobackup              --lockopt                 --readahead
```

### DIFF
--- a/completions/lvm
+++ b/completions/lvm
@@ -895,7 +895,7 @@ _comp_cmd_lvm()
                 vgdisplay | vgexport | vgextend | vgimport | vgmerge | vgmknodes | vgreduce | \
                 vgremove | vgrename | vgs | vgscan | vgsplit | lvchange | lvcreate | lvdisplay | \
                 lvextend | lvreduce | lvremove | lvrename | lvresize | lvscan)
-                "_${words[1]}"
+                "_comp_cmd_${words[1]}" "${words[@]:1}"
                 ;;
         esac
     fi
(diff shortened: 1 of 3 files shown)


completions/lvm
```diff
@@ -895,7 +895,7 @@ _comp_cmd_lvm()
                 vgdisplay | vgexport | vgextend | vgimport | vgmerge | vgmknodes | vgreduce | \
                 vgremove | vgrename | vgs | vgscan | vgsplit | lvchange | lvcreate | lvdisplay | \
                 lvextend | lvreduce | lvremove | lvrename | lvresize | lvscan)
-                "_${words[1]}"
+                "_comp_cmd_${words[1]}" "${words[@]:1}"
```
akinomyoga, line 898: I believe we need to pass `"${words[1]}" "${@:2}"` as arguments. The completion function is supposed to receive the command name (`$1`), the current word (`$2`), and the previous word if any (`$3`). Also, we need to shift the content of `COMP_WORDS`, etc.

Actually, we already have a utility to process all of these handlings. I guess we can call that utility:

```suggestion
                _comp_command_offset 1
```

akinomyoga (commented):
Thank you for the report!

completions/lvm
aaronkrogers: Thank you for the suggestion! That looks much better and works great too

scop:
Thanks! Added a test case in 42b8a00ab34569d05b56f025fc5e24bb794027b1.

Issue #103, opened 2016-12-24
allow user to complete filenames after 'set --'


jidanni:
User wishes to have filenames completed when hit hits TAB at
```
$ set -- filena<TAB>
```
however he is thwarted.
The only choices are
```
$ set -- <TAB>
allexport             functrace             interactive-comments  noglob                physical              vi
braceexpand           hashall               keyword               nolog                 pipefail              xtrace
emacs                 histexpand            monitor               notify                posix
errexit               history               noclobber             nounset               privileged
errtrace              ignoreeof             noexec                onecmd                verbose
```
Pray tell what good are they after `--`?

Why can't you let me complete filenames anymore?

bash-completion:
  Installed: 1:2.1-4.3

jidanni:
BTW
```$ help set```
shows that the above offered choices are only appropriate after `-o` too!

jidanni:
In fact there should be no need to even need me to add `--` to turn on filename completion!

jidanni:
Workaround while you are fixing the bug:
```$ \set bla bla```

jidanni:
Workaround. Put in .bashrc:
`complete -F _filedir_xspec -A setopt set`

PR #1005, opened 2023-06-13
fix(eog): add completion for AVIF and WEBP files

Rogach:
I'm not sure what to write here - it is what it says on the tin, I've just added two more file extensions.

### DIFF
--- a/completions/eog
+++ b/completions/eog
@@ -19,7 +19,7 @@ _comp_cmd_eog()
         return
     fi
 
-    _comp_compgen_filedir '@(ani|?(w)bmp|gif|ico|j2[ck]|jp[cefgx2]|jpeg|jpg2|pcx|p[gp]m|pn[gm]|ras|svg?(z)|tga|tif?(f)|x[bp]m)'
+    _comp_compgen_filedir '@(ani|?(w)bmp|gif|ico|j2[ck]|jp[cefgx2]|jpeg|jpg2|pcx|p[gp]m|pn[gm]|ras|svg?(z)|tga|tif?(f)|x[bp]m|avif|webp)'
 } &&
     complete -F _comp_cmd_eog eog
 


akinomyoga (approved):
I tried to find a page that lists the supported file formats, but it seems to be hard to list the supported file formats. The page [1] doesn't seem to list many file formats including the suggested `.avif` and `.webp`. Meanwhile, the Wikipedia article  [2] seems to list those file formats, but it doesn't seem to be based on official information.

- [1] [Supported image formats](https://help.gnome.org/users/eog/stable/formats-view.html.en)
- [2] [Eye of GNOME - Wikipedia](https://en.wikipedia.org/wiki/Eye_of_GNOME)

The source repository [3] doesn't seem to contain the file formats either (based on the code search). Instead, it seems to call an external function `gdk_pixbuf_get_formats` to get the list of formats [4].

- [3] [GNOME / Eye of GNOME · GitLab](https://gitlab.gnome.org/GNOME/eog)
- [4] [src/eog-pixbuf-util.c · master · GNOME / Eye of GNOME · GitLab](https://gitlab.gnome.org/GNOME/eog/-/blob/master/src/eog-pixbuf-util.c#L41-60)

The function `gdk_pixbuf_get_formats` seems to be from the GdkPixbuf library. The GdkPixbuf library seems to recommend creating an external loader module for each file format, so it is hard to pre-determine which formats are supported by the system. At least, the file formats `avif` and `webp` seem to be supported by external modules.

- [5] [[feature request] add support for webp (#103) · Issues · GNOME / gdk-pixbuf · GitLab](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/issues/103#note_535385)
- [6] [Add an AVIF loader (!76) · Merge requests · GNOME / gdk-pixbuf · GitLab](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/merge_requests/76#note_845043)
- [7] [Add a gdk-pixbuf module by linkmauve · Pull Request #182 · AOMediaCodec/libavif](https://github.com/AOMediaCodec/libavif/pull/182)
- [8] [aruiz/webp-pixbuf-loader: WebP Image format GdkPixbuf loader](https://github.com/aruiz/webp-pixbuf-loader)

----

**Edit**: There seem to be several ways to dynamically get the file formats supported in the present system [9,10].

- [9] [c - What image formats are supported by Gdk-Pixbuf (Gtk-Image?) by Default? - Stack Overflow](https://stackoverflow.com/questions/7704670/what-image-formats-are-supported-by-gdk-pixbuf-gtk-image-by-default)
- [10] [gdk-pixbuf-query-loaders(1) - Linux man page](https://linux.die.net/man/1/gdk-pixbuf-query-loaders)

scop:
> I'm not sure what to write here - it is what it says on the tin, I've just added two more file extensions.

FWIW, I usually just leave it empty or say "SSIA.".

> There seem to be several ways to dynamically get the file formats supported in the present system

That would be interesting. However at least on Ubuntu 20, `gdk-pixbuf-query-loaders` is not installed in `$PATH` (it's buried in `/usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0`), and doesn't seem that the output format is documented (not that that would prevent us from happily scraping stuff everywhere though ;)). The Python recipe looks cleaner, but requires modules that are not necessarily present (e.g. `gi`), and I'm not sure if even Python can be assumed present. But I guess we could make something work from these, if someone's interested.

scop:
Resolved conflicts, sorted the regex elements, fixed up commit message, merged, thanks :+1:

PR #1006, opened 2023-06-13
Add ppm extension for eog

seiferteric:
Currently eog completion for netpbm format (https://en.wikipedia.org/wiki/Netpbm) only has pgm and ppm, but there is also ppm and is supported by eog.


scop (approved):

scop:
Thanks!

PR #1010, opened 2023-06-21
feat(eog): associate with `*.heic` and `*.jxl`

calestyo:
eog supports these via add-on gdk-pixbuf loaders.

### DIFF
--- a/completions/eog
+++ b/completions/eog
@@ -19,7 +19,7 @@ _comp_cmd_eog()
         return
     fi
 
-    _comp_compgen_filedir '@(ani|avif|?(w)bmp|gif|ico|j2[ck]|jp[cefgx2]|jpeg|jpg2|pcx|p[bgp]m|pn[gm]|ras|svg?(z)|tga|tif?(f)|webp|x[bp]m)'
+    _comp_compgen_filedir '@(ani|avif|?(w)bmp|gif|heic|ico|j2[ck]|jp[cefgx2]|jpeg|jpg2|jxl|pcx|p[bgp]m|pn[gm]|ras|svg?(z)|tga|tif?(f)|webp|x[bp]m)'
 } &&
     complete -F _comp_cmd_eog eog
 


completions/eog
```diff
@@ -19,7 +19,7 @@ _comp_cmd_eog()
         return
     fi
 
-    _comp_compgen_filedir '@(ani|avif|?(w)bmp|gif|ico|j2[ck]|jp[cefgx2]|jpeg|jpg2|pcx|p[bgp]m|pn[gm]|ras|svg?(z)|tga|tif?(f)|webp|x[bp]m)'
+    _comp_compgen_filedir '@(ani|avif|?(w)bmp|gif|heic|ico|j2[ck]|jp[cefgx2]|jpeg|jpg2|jxl|pcx|p[bgp]m|pn[gm]|ras|svg?(z)|tga|tif?(f)|webp|x[bp]m)'
```
scop, line 22: Should some more besides just `heic` be added for HEIC?

- https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format
- https://www.iana.org/assignments/media-types/image/heic

calestyo: Hmm good question.

What I am using is `heif-gdk-pixbuf`, which is based on [libheif](https://github.com/strukturag/libheif).<br>
It's not really clear to me, whether the different extension names are actually all different formats (okay some are HVEC based and some AV1) and whether libheif supports all of them.

OTOH, one could argue that these are simply well known extensions for image formats, `eog` is an image viewer and even if `heif-gdk-pixbuf` wouldn't support all formats, other loaders may and so it's still worth for bash-completion to recognise them.

If you wish, you can just extend my commit with the other extensions... or I could update it if you prefer.

akinomyoga: https://github.com/strukturag/libheif/blob/4a2f42f01e475258e67371db99108636d32f561c/gdk-pixbuf/pixbufloader-heif.c#L224-L229

akinomyoga: > It's not really clear to me, whether the different extension names are actually all different formats (okay some are HVEC based and some AV1) and whether libheif supports all of them.

Could you investigate them?

calestyo: Well I guess it's the one from the code you just referenced.

But as said before... do we even really want to include only those which are supported by `heif-gdk-pixbuf`? That could change any time, or there could be other GDK pixbuf loaders.

akinomyoga: > Well I guess

Could you confirm that?

> it's the one from the code you just referenced.

The code tells which extensions are at least supported, but it doesn't tell whether an extension not listed there is compatible with the listed one and could be loaded. If the file format is determined by only the filename extension, we should just list the extensions listed by the pixbuf implementation, but if the file format can be detected by other means (e.g. by checking several bytes at the beginning of the file), we can also list compatible extensions.

> But as said before... do we even really want to include only those which are supported by `heif-gdk-pixbuf`? That could change any time, or there could be other GDK pixbuf loaders.

We shouldn't include any extensions that are not known to be supported. Suggesting the file in the completion implies to the users that the file is supported by `eog`.

calestyo:
> Could you confirm that?

I'm afraid not really.<>
I even only found sample files for `heif` and `heic` (for that I have plenty, or at least my smartphone claims it would bei `heic`).
```
$ file sample*
sample1.heic: ISO Media, HEIF Image
sample1.heif: ISO Media, HEIF Image HEVC Main or Main Still Picture Profile
```
eog would be able to display both, but whether the 2nd is really `.heif`  - who knows

> We shouldn't include any extensions that are not known to be supported. Suggesting the file in the completion implies to the users that the file is supported by eog.

Guess that's more a matter of taste, isn't it? Take e.g. JPEG2000 ... that has about 10 different substandards and I think only the "main" ones were ever really used/supported, yet they may all share the same extension.

I don't think it's feasible for bash-completion to keep up with what exactly is supported and what not. It makes sense to not complete `*.wav` of course... but if it's a well known image file extension, why not just complete it.

Even `.heic` won't work, if the extra pixbuf loader isn't installed.

scop (approved):
Let's go with the basics that we have here for now. Thanks!

PR #1013, opened 2023-07-20
feat(hash): new completion

scop:


completions/hash
```diff
@@ -0,0 +1,26 @@
...
+    # shellcheck disable=SC2254
+    case $prev in
+        -${noargopts}p)
+            _comp_compgen_filedir
+            return
+            ;;
+    esac
+
+    if [[ $cur == -* ]]; then
+        _comp_compgen_help
```
akinomyoga, line 18: bash < 4.4 doesn't support `hash --help`, so we need to use `help hash`.

```suggestion
        _comp_compgen_help -c help "$1"
```

scop: Thanks! Better that way in the first place.

akinomyoga (approved):
Thanks!

PR #1015, opened 2023-07-29
add executable completion to exec

s1037989:


completions/carton
```diff
@@ -72,6 +72,11 @@ _comp_cmd_carton()
     fi
 
     case $command in
+        exec)
+            COMPREPLY=($(for i in $(compgen -f -d -- ${cur}); do if test -d "$i"; then echo "$i/"; elif test -x "$i"; then echo "$i "; fi; done));
+            test -d "${COMPREPLY[0]}" && compopt -o nospace
```
akinomyoga, line 77: Thank you for your contribution. I guess the behavior you expect could be achieved by just the following. Could you check it?

```suggestion
            _comp_compgen_filedir
```

s1037989: Success!

akinomyoga: Thanks for updating.

akinomyoga:
Do you know good documentation about `carton exec`? What I can find is an example in [README](https://github.com/perl-carton/carton#synopsis), in which what can be actually specified to the arguments of `carton exec` is not so clear. Is it an arbitrary command name available in the current system plus arguments? If so, it is probably better to use `_comp_command_offset` to invoke the completion for the specific command.

```bash
_comp_command_offset "$((i + 1))"
```

s1037989:
The purpose of `carton exec` is to run the perl scripts in your project folder, such as `carton exec -- script/app`, but, indeed, one might call `perl` first as `carton exec -- perl script/app`.

I tested with `_comp_command_offset` as you suggested, but the disadvantage to that approach is that without the `--` it tab completes every possible command (a positive), but with the `--` (needed to prevent `carton exec` from parsing all of the command line args from the script) it only completes the files in the project folder, however, it completes all files, not just executables.

I think it would be ideal for `carton exec` and `carton exec --` to be treated the same, and then if the next word is `perl` handle it like the perl tab completion, otherwise the next word should be treated like an executable and after that treat it like any other unknown command... probably tab complete any file at that point?

akinomyoga:
Thank you for the information!

https://github.com/perl-carton/carton/issues/228 seems to be related to `carton exec --`. I'm still unsure if I correctly understand it. `carton` seems to treat `carton exec cmd -v` as if `carton -v exec -- cmd`, but

- How does it treat `carton exec cmd arg`? Is it equivalent to `carton exec -- cmd arg`?
- How does it treat `carton exec cmd -v arg`? Is it something like `carton -v exec -- cmd arg` or `carton -v arg exec -- cmd`?
- How does it treat `carton exec -v cmd`? Is it equivalent to `carton -v exec -- cmd`?

Perhaps we could skip all the options (such as `--` and `-v`) to identify the position of an executable name.

akinomyoga:
I've pushed an experimental implementation of skipping options (798a7984ea8b38236d095a3564090f11d6282386) to branch `patch-1`. I haven't tested it because I am not familiar with `carton`. Could you test it?

s1037989:
I tried your patch and the behavior seems to be as before.

> How does it treat `carton exec cmd arg`? Is it equivalent to `carton exec -- cmd arg`?
> How does it treat `carton exec cmd -v arg`? Is it something like `carton -v exec -- cmd arg` or `carton -v arg exec -- cmd`?
> How does it treat `carton exec -v cmd`? Is it equivalent to `carton -v exec -- cmd`?

Well, this is interesting...

```
$ carton exec perl -v
carton v1.0.35
$ carton exec -v perl
carton v1.0.35
$ carton -v exec perl
carton v1.0.35
$ carton exec perl -E 'say 1'
1
$ carton exec perl -E 'say scalar localtime'
Tue Aug  1 20:46:22 2023
$ carton exec -- perl -v
This is perl 5, version 38, subversion 0 (v5.38.0) built for x86_64-linux
$ carton exec -v -- perl
carton v1.0.35
$ carton -v exec -- perl
carton v1.0.35
$ carton exec perl -v -E 'say scalar localtime'
carton v1.0.35
```

So I guess carton is only taking its known arguments?  But `carton --help` isn't really documenting all of its arguments.  In fact, the only three arguments it documents are `install`, `exec`, and `bundle`.

From [Carton/CLI.pm](https://metacpan.org/release/MIYAGAWA/Carton-v1.0.35/source/lib/Carton/CLI.pm#L41):
```
-v|--version, -h|--help, --verbose
bundle
fatpack
install: -p|--path string, --cpanfile string, --without string, --deployment, --cached
show
list: --distfile
tree
check: --cpanfile
update
run
exec
```

And each of those subcommands, in addition to the specific arguments, takes additional arguments of more-or-less arbitrary nature.

Of course, the best way to ensure that carton doesn't steal your exec's args is to use `--`.  And *after* the `--`, what is ideal is to treat what comes next as a normal bash default:

- without any characters on the command line (meaning nothing after `--`) tab completion does nothing
- after one character, tab completion starts work to complete the command
- subsequent words are tab completions of the command / "first word" (after `--`) as arguments to that command
- and anything else that's standard bash behavior

IMO, anything after `carton exec --` should be treated as if you were doing a normal shell command -- you're only doing `carton exec --` to set a specific environment unique to what comes next; i.e. you don't want to set it for the entire shell:

```
$ perl -E 'say $]'
5.034000
$ perl -MMojolicious::Plugin::Pager -E 'say $Mojolicious::Plugin::Pager::VERSION'
Can't locate Mojolicious/Plugin/Pager.pm
$ carton exec -- perl -MMojolicious::Plugin::Pager -E 'say $Mojolicious::Plugin::Pager::VERSION'
0.06
$ perl -MMojolicious::Plugin::Pager -E 'say $Mojolicious::Plugin::Pager::VERSION'
Can't locate Mojolicious/Plugin/Pager.pm
```

akinomyoga:
> I tried your patch and the behavior seems to be as before.

Thank you for testing. Does that mean the following problem still persists?

> https://github.com/scop/bash-completion/pull/1015#issuecomment-1659918378
> with the `--` (needed to prevent `carton exec` from parsing all of the command line args from the script) it only completes the files in the project folder, however, it completes all files, not just executables.

Also, the following behavior is supposed to be implemented in the version I pushed.

> https://github.com/scop/bash-completion/pull/1015#issuecomment-1659918378
> I think it would be ideal for `carton exec` and `carton exec --` to be treated the same, and then if the next word is `perl` handle it like the perl tab completion, otherwise the next word should be treated like an executable

All of the above seem to work in an expected way at least in my environment. Could you check the behavior again making sure that you use the version I pushed? If there is any problem, could you describe it in detail? Thank you for your help.

----

Also, thank you for the interesting test results. Those are useful for understanding the nature of Carton's command line parser.

s1037989:
Ok, I must have not done a full load of the changed files, because all of my testing at this point seems to indicate that tab completion for carton is working exactly as expected for all of the scenarios that I can see!

LGTM!  Thank you for the excellent patch to get this working better for carton!

I'm glad that the test results I provided were also helpful to you!

akinomyoga:
rebased

akinomyoga (approved):

scop:
Thanks!

PR #1018, opened 2023-08-01
fix(python): complete filenames for script arguments

akinomyoga:
Ref. https://github.com/ohmybash/oh-my-bash/issues/462

When `[-c command | -m module | file]` is already specified before `cword`, the current implementation of the `python` completer tries to complete filenames as arguments to the command/module/script. However, this doesn't work after an option of the form `-?`. The following is a case reported in https://github.com/ohmybash/oh-my-bash/issues/462:

```bash
python some_script.py -p [TAB]
```

The current implementation seems to try to detect the situation by checking the previous and the second previous arguments, but we should explicitly test if `[-c command | -m module | file]` is already specified.

### Note: Removing an old case pattern that excludes `python` and `-?`
I think this is actually related to the following line that we once discussed in #952 (see replies [1,2]):

https://github.com/scop/bash-completion/blob/8b4db1c8b6581d89dd9df8031f847c717778f742/completions/python#L77-L81

- [1] https://github.com/scop/bash-completion/pull/952#issuecomment-1535620829
- [2] https://github.com/scop/bash-completion/pull/952#discussion_r1186281373

After looking at this code again, I now believe this piece of code tried to detect whether the current word is an argument just after `[-c command | -m module | file]` or not. The `case` pattern and the if statement exclude the case where

1) the previous argument is the command name (`?(*/)?(micro)python*([0-9.])|?(*/)py@(py|ston)*([0-9.])`),
2) the previous argument is an option (`-?`), or
3) the second previous and previous arguments form a pair of an option plus an option argument (`${words[cword - 2]} != -[QWX]`).

The remaining situation is likely to be only the case where the previous argument is a command/module/script of `[-c command | -m module | file]`. Since we now explicitly check `[-c command | -m module | file]`, we should remove this heuristic detection of the first argument.

### DIFF
--- a/test/t/test_python.py
+++ b/test/t/test_python.py
@@ -45,3 +45,42 @@ class TestPython:
     )
     def test_bb(self, completion):
         assert "-bb" in completion
+
+    @pytest.mark.complete("python foo ", cwd="python")
+    def test_script_arg(self, completion):
+        assert "bar.txt" in completion
+
+    @pytest.mark.complete("python -- foo ", cwd="python")
+    def test_script_arg_with_double_hyphen(self, completion):
+        assert "bar.txt" in completion
+
+    @pytest.mark.complete("python -m foo bar -p ", cwd="python")
+    def test_module_arg(self, completion):
+        assert "bar.txt" in completion
+
+    @pytest.mark.complete("python foo bar -p ", cwd="python")
+    def test_script_arg_after_option(self, completion):
+        assert "bar.txt" in completion
+
+    @pytest.mark.complete("python -- foo bar -p ", cwd="python")
+    def test_script_arg_after_option_with_double_hyphen(self, completion):
+        assert "bar.txt" in completion
+
+    @pytest.mark.complete("python -m foo bar -p ", cwd="python")
+    def test_module_arg_after_option(self, completion):
+        assert "bar.txt" in completion
+
+    @pytest.mark.complete("python -mfoo bar -p ", cwd="python")
+    def test_module_arg_after_option_with_connected_m_arg(self, completion):
+        assert "bar.txt" in completion
+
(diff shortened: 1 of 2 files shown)


scop (changes requested):
Makes sense. This is a bit of a corner case maybe, could we add some test cases to cover/guard the intended behavior?

akinomyoga:
I added small adjustments and tests.

test/t/test_python.py
```diff
@@ -45,3 +45,23 @@ def test_9(self, completion):
...
+    def test_script_args_3(self, completion):
+        assert "bar.txt" in completion
+
+    @pytest.mark.complete("python -- ", cwd="python")
+    def test_script_args_4(self, completion):
+        assert "bar.txt" not in completion
+
+    @pytest.mark.complete("python -W -mfoo ", cwd="python")
+    def test_script_args_5(self, completion):
+        assert "bar.txt" not in completion
```
scop, line 67: To clarify, this is because `-mfoo` gets treated as an arg to `-W`, and ignored as invalid -> this is effectively the same as calling just `python ` (as far as the python executable goes)? Perhaps reflecting that in the test function name would be in order, e.g. like this, or in its docstring? Pre-approved with something like that.

```suggestion
    @pytest.mark.complete("python -W -mfoo ", cwd="python")
    def test_script_invalid_w_arg(self, completion):
        assert "bar.txt" not in completion
```

scop (approved):

test/t/test_python.py
akinomyoga: I've updated the function names and added a docstring to the last one.

scop (approved):

scop:
Thanks!

PR #1019, opened 2023-08-01
feat(eog): add missing extension .heif

akinomyoga:
The support for `.heic` has been added in #1010, but the gdk pixbuf loader, which supports `.heic`, [explicitly supports also `.heif` and `.avif`](https://github.com/strukturag/libheif/blob/4a2f42f01e475258e67371db99108636d32f561c/gdk-pixbuf/pixbufloader-heif.c#L224-L229). The extension `.avif` is already included in the list, but `.heif` is currently missing.

### DIFF
--- a/completions/eog
+++ b/completions/eog
@@ -19,7 +19,7 @@ _comp_cmd_eog()
         return
     fi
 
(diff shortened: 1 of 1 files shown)


scop:
Thanks!

PR #1020, opened 2023-08-04
feat(pipenv): add fallback 3rd party completion loader

scop:
https://pipenv.pypa.io/en/latest/shell/#shell-completion

### DIFF
--- a/completions/_pipenv
+++ b/completions/_pipenv
@@ -8,7 +8,7 @@ eval -- "$(
     # shellcheck disable=SC2154
     ucname="${cmdname^^}"
     ucname=${ucname//-/_}
-    eval "v=_${ucname}_COMPLETE=bash_source; export \$v"
+    export "_${ucname}_COMPLETE=bash_source"
     "$1" 2>/dev/null
 )"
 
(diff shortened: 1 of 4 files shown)


completions/_pipenv
```diff
@@ -0,0 +1,15 @@
...
+# This pattern is used by programs built with https://click.palletsprojects.com
+#
+# This serves as a fallback in case the completion is not installed otherwise.
+
+eval -- "$(
+    # shellcheck disable=SC2154
+    ucname="${cmdname^^}"
+    ucname=${ucname//-/_}
+    eval "v=_${ucname}_COMPLETE=bash_source; export \$v"
+    "$1" 2>/dev/null
```
scop, line 12: I'm not too happy with how this turned out with the nested evals and all, but I didn't spend too much thinking about it either. Maybe there's a cleaner way?

akinomyoga: It's not so different, but we can directly use `export`. (At least, it will correctly fail without unexpected command execution when the command name contains a string by an attacker, though I'm not sure if the completion loader can be invoked for such an unexpected command names)

```suggestion
    # shellcheck disable=SC2154
    ucname="${cmdname^^}"
    ucname=${ucname//-/_}
    export "_${ucname}_COMPLETE=bash_source"
    "$1" 2>/dev/null
```

scop: Yeah, I like that better :+1:

akinomyoga (approved):

PR #1021, opened 2023-08-04
pytest: complete new --import-mode value

calestyo:


scop:
Thanks!

Looks like `pytest --import-mode=something-garbage` would output the valid choices, but the format is very clearly not intended for scraping, so perhaps better to steer clear of it.

BTW, pytest would also come with argcomplete completions, I considered dropping ours and switching to it a while back. But the test class and test case name completions we have here seemed too valuable to be thrown away to me.

calestyo:
Thanks for merging.

I'd anyway not recommend to hold one's breath for any new import mechanism ;-)

I know about the argcomplete support... but I kinda dislike enabling (IIRC it always requires some env var to be set)... and I, personally, hate "spoiling" the environment with countless vars. ^^

scop:
Yeah, but generally speaking this stuff is not only about whether we'll see new import mechanisms, but also there being older versions for which we may offer invalid options.

FYI we've added support for loading quite a few "known" 3rd party completions recently with no need for users to do anything about it, pytest could be one of them. But at least for now, as said, I've decided against it for it. `git grep "3rd party"` and symlinks pointing to those files for more info.

calestyo:
> FYI we've added support for loading quite a few "known" 3rd party completions recently with no need for users to do anything about it

That's really nice to hear.

In general I think, that such "framework-functionality" really makes a difference.

What I'm for example still struggling the most with is: easily re-using already existing completions at different places.
1. For example for aliases.
    In simple cases like if I make aliases for `ls` it's still easy, by doing something like:
    ```
    _longopt_alias_wrapper_ls() { shift; _longopt ls "$@"; }; complete -F _longopt_alias_wrapper_ls ls1 l ll lll
    ```
    but for more complex cases like `systemctl` that doesn't seem to work.

2. And similarly if one has wrapper commands (like `faketime` or `script -c command`) or the like, so that commands after that are completed.

PR #1023, opened 2023-08-05
test(pids,pgids): check current p(g)id in completions

scop:


test/t/unit/test_unit_pids.py
```diff
@@ -26,4 +28,6 @@ def test_ints(self, bash):
                 want_output=True,
             ).split()
         assert completion
+        if hasattr(os, "getpid"):
+            assert str(os.getpid()) in completion
```
scop, line 32: I'm not sure what "stubbed" means in https://docs.python.org/3/library/os.html#os.getpid wrt the return value, but that seems to affect only platforms not that relevant (?) to us.

akinomyoga (approved):

PR #1027, opened 2023-08-06
fix(dict): nounset error when host/port not set

scop:


akinomyoga:
I pushed some fixes.

- 687cf994 -- To check whether any completions are generated, it is not sufficient to check `completion` is not empty. Since we pass `-o default` to the completion setting, even when no completions are generated, the default completions are performed as a fallback, so the test always succeeds. To properly test it, we need to exclude the fallback completions.
- b83b0f5a, b31fe1e3 -- The current implementation requires the normal IFS so that `$host $port` are correctly expanded. This fails when a custom IFS is set.

scop:
Ah, I was using the database completion just to see that the unset host/port won't cause an error, not to test database completion beyond that at all. I don't mind improving it further, but the modified tests fail now in the non-networked Debian 10 container.

We could skip or xfail these tests (see test_ether_wake.py for an example), but even with networking enabled this could be a flaky test depending on dict.org availability and/or whatever the local `dict` is configured to access if a host is not set.

akinomyoga:
Thank you for the advice. I've added a `xfail` for the network test.

> We could skip or xfail these tests (see test_ether_wake.py for an example), but even with networking enabled this could be a flaky test depending on dict.org availability and/or whatever the local `dict` is configured to access if a host is not set.

I'm not familiar with `dict`, but is it possible to specify a local database? Can we specify some file path or do we need to launch a `dictd` server? It seems to be the latter.

scop:
Actually I'm not familiar with `dict` either. Previously I thought a serverless mode existed, but I didn't come across one when looking around yesterday. Let's go with this now, and revisit if we hear about problems.

PR #1026, opened 2023-08-06
feat(xrandr): comma separated `--setmonitor` third argument

scop:
Requires/stacks on top of #1025, drafting until that one is merged.

PR #1028, opened 2023-08-06
fix(_comp_compgen_*): avoid conflicts with "-v var"

akinomyoga:
To avoid conflicts between local variables and variable names specified by `_comp_compgen -v var`, the local variables basically need to have the name `_foo`.

### Handling of conflicts with internal generator calls

When a generator needs to call other generators to store results in a local variable, the local variable name cannot have the name `_foo` because it is rejected by the internal generator calls. In such a case, we first declare the variable without the underscore (e.g. `foo`), call the internal generator to store the result to `foo`, move the results to an underscored variable (`_foo`), and remove the local variable `foo` by `_comp_unlocal foo`. The current version of the PR uses this approach.

However, maybe this is too complicated. Another possibility is to add a new option `_comp_compgen -U foo` to internally perform `_comp_unlocal foo` immediately before storing the results.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -838,7 +838,7 @@ _comp__get_cword_at_cursor()
     fi
 
     local IFS=$' \t\n'
-    local "$2" "$3" "$4" && _comp_upvars -a${#words[@]} "$2" ${words[@]+"${words[@]}"} \
+    local "$2" "$3" "$4" && _comp_upvars -a"${#words[@]}" "$2" ${words[@]+"${words[@]}"} \
         -v "$3" "$cword" -v "$4" "${cur:0:index}"
 }
 
--- a/bash_completion
+++ b/bash_completion
@@ -422,6 +422,20 @@ _comp_compgen__error_fallback()
 #           The array name should not start with an underscores "_", which is
 #           internally used.  The array name should not be either "IFS" or
 #           "OPT{IND,ARG,ERR}".
+#   -U var  Unlocalize VAR before performing the assignments.  This option can
+#           be specified multiple times to register multiple variables.  This
+#           option is supposed to be used in implementing a generator (G1) when
+#           G1 defines a local variable name that does not start with `_`.  In
+#           such a case, when the target variable specified to G1 by `-v VAR1`
+#           conflicts with the local variable, the assignment to the target
+#           variable fails to propagate outside G1.  To avoid such a situation,
+#           G1 can call `_comp_compgen` with `-U VAR` to unlocalize `VAR`
+#           before accessing the target variable.  For a builtin compgen call
+#           (i.e., _comp_compgen [options] -- options), VAR is unlocalized
+#           after calling the builtin `compgen` but before assigning results to
+#           the target array.  For a generator call (i.e., _comp_compgen
+#           [options] G2 ...), VAR is unlocalized before calling the child
+#           generator function `_comp_compgen_G2`.
 #   -c cur  Set a word used as a prefix to filter the completions.  The default
 #           is ${cur-}.
 #   -R      The same as -c ''.  Use raw outputs without filtering.
@@ -512,6 +526,7 @@ _comp_compgen()
     local _dir=""
     local _ifs=$' \t\n' _has_ifs=""
     local _icmd="" _xcmd=""
+    local -a _upvars=()
 
     local _old_nocasematch=""
     if shopt -q nocasematch; then
@@ -519,7 +534,7 @@ _comp_compgen()
         shopt -u nocasematch
     fi
     local OPTIND=1 OPTARG="" OPTERR=0 _opt
-    while getopts ':av:Rc:C:lF:i:x:' _opt "$@"; do
+    while getopts ':av:U:Rc:C:lF:i:x:' _opt "$@"; do
         case $_opt in
             a) _append=set ;;
             v)
@@ -528,6 +543,16 @@ _comp_compgen()
                     return 2
                 fi
                 _var=$OPTARG
+                ;;
+            U)
+                if [[ $OPTARG == @(*[^_a-zA-Z0-9]*|[0-9]*|'') ]]; then
+                    printf 'bash_completion: %s: -U: invalid variable name `%s'\''\n' "$FUNCNAME" "$OPTARG" >&2
+                    return 2
+                elif [[ $OPTARG == @(_*|IFS|OPTIND|OPTARG|OPTERR) ]]; then
+                    printf 'bash_completion: %s: -U: unnecessary to mark `%s'\'' as upvar\n' "$FUNCNAME" "$OPTARG" >&2
+                    return 2
+                fi
+                _upvars+=("$OPTARG")
                 ;;
             c) _cur=$OPTARG ;;
             R) _cur="" ;;
@@ -587,6 +612,8 @@ _comp_compgen()
             printf 'bash_completion: %s: unrecognized generator `%s'\'' (function %s not found)\n' "$FUNCNAME" "$1" "${_generator[0]}" >&2
             return 2
         fi
+
+        ((${#_upvars[@]})) && _comp_unlocal "${_upvars[@]}"
 
         if [[ $_dir ]]; then
             local _original_pwd=$PWD
@@ -653,6 +680,7 @@ _comp_compgen()
         return
     }
 
+    ((${#_upvars[@]})) && _comp_unlocal "${_upvars[@]}"
     _comp_split -l ${_append:+-a} "$_var" "$_result"
 }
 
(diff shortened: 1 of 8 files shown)


akinomyoga:
> Another possibility is to add a new option `_comp_compgen -U foo` to internally perform `_comp_unlocal foo` immediately before storing the results.

This is an example implementation for this direction.

```diff
diff --git a/bash_completion b/bash_completion
index 98c67eb0..0799824e 100644
--- a/bash_completion
+++ b/bash_completion
@@ -512,6 +512,7 @@ _comp_compgen()
     local _dir=""
     local _ifs=$' \t\n' _has_ifs=""
     local _icmd="" _xcmd=""
+    local -a _upvars=()

     local _old_nocasematch=""
     if shopt -q nocasematch; then
@@ -519,7 +520,7 @@ _comp_compgen()
         shopt -u nocasematch
     fi
     local OPTIND=1 OPTARG="" OPTERR=0 _opt
-    while getopts ':av:Rc:C:lF:i:x:' _opt "$@"; do
+    while getopts ':av:U:Rc:C:lF:i:x:' _opt "$@"; do
         case $_opt in
             a) _append=set ;;
             v)
@@ -529,6 +530,16 @@ _comp_compgen()
                 fi
                 _var=$OPTARG
                 ;;
+            U)
+                if [[ $OPTARG == @(*[^_a-zA-Z0-9]*|[0-9]*|'') ]]; then
+                    printf 'bash_completion: %s: -U: invalid variable name `%s'\''\n' "$FUNCNAME" "$OPTARG" >&2
+                    return 2
+                else if [[ $OPTARG == @(_*|IFS|OPTIND|OPTARG|OPTERR) ]]; then
+                    printf 'bash_completion: %s: -U: `%s'\'' unneccessary to mark upvar\n' "$FUNCNAME" "$OPTARG" >&2
+                    return 2
+                fi
+                _upvars+=("$OPTARG")
+                ;;
             c) _cur=$OPTARG ;;
             R) _cur="" ;;
             C)
@@ -573,6 +584,9 @@ _comp_compgen()
         if [[ $_has_ifs ]]; then
             printf 'bash_completion: %s: `-l'\'' and `-F sep'\'' are not supported for generators\n' "$FUNCNAME" >&2
             return 2
+        elif ((${#_upvars[@]})); then
+            printf 'bash_completion: %s: `-U var'\'' is not supported for generator calls\n' "$FUNCNAME" >&2
+            return 2
         fi

         local -a _generator
@@ -653,6 +664,7 @@ _comp_compgen()
         return
     }

+    ((${#_upvars[@]})) && _comp_unlocal "${_upvars[@]}"
     _comp_split -l ${_append:+-a} "$_var" "$_result"
 }
```

akinomyoga:
I decided to use the approach of the previous reply https://github.com/scop/bash-completion/pull/1028#issuecomment-1667010520.

akinomyoga:
Rebased & resolved conflicts.

akinomyoga:
Added fixes mentioned in #1026.

completions/cvs
```diff
@@ -56,7 +56,7 @@ _comp_cmd_cvs()
     local cur prev words cword comp_args
     _comp_initialize -n : -- "$@" || return
 
-    local count mode="" i cvsroot="" has_cvsroot="" cvsroots pwd
+    local count mode="" i cvsroot="" has_cvsroot="" pwd
```
scop, line 59: This hunk cherry-picked/done in #1040

completions/ssh
```diff
@@ -8,7 +8,7 @@ _comp_cmd_ssh__compgen_queries()
         key-plain key-sig protocol-version compression sig ciphers macs
         kexalgorithms pubkeyacceptedkeytypes hostkeyalgorithms
         hostbasedkeytypes hostbasedacceptedkeytypes)
-    _comp_compgen -ac "${cur,,}" -- -W '"${ret[@]}" help"'
+    _comp_compgen -c "${cur,,}" -U ret -- -W '"${ret[@]}" help"'
```
scop, line 11: This hunk cherry-picked/done in #1040

completions/ssh
```diff
@@ -37,7 +37,7 @@ _comp_cmd_ssh__compgen_ciphers()
     [[ ${ret-} ]] || ret=(3des-cbc aes128-cbc aes192-cbc aes256-cbc
         aes128-ctr aes192-ctr aes256-ctr arcfour128 arcfour256 arcfour
         blowfish-cbc cast128-cbc)
-    _comp_compgen -a -- -W '"${ret[@]}"'
+    _comp_compgen -U ret -- -W '"${ret[@]}"'
```
scop, line 40: Drop of `-a` cherry-picked/done in #1040

completions/ssh
```diff
@@ -46,7 +46,7 @@ _comp_cmd_ssh__compgen_macs()
     _comp_compgen -v ret -i ssh query "$1" mac
     [[ ${ret-} ]] || ret=(hmac-md5 hmac-sha1 umac-64@openssh.com
         hmac-ripemd160 hmac-sha1-96 hmac-md5-96)
-    _comp_compgen -a -- -W '"${ret[@]}"'
+    _comp_compgen -U ret -- -W '"${ret[@]}"'
```
scop, line 49: Drop of `-a` cherry-picked/done in #1040

completions/openssl
```diff
@@ -2,27 +2,27 @@
 
 _comp_cmd_openssl__compgen_sections()
 {
-    local config f
+    local config _i _file
```
scop, line 5: `local` of `i` cherry-picked/done in #1040

completions/cvs
```diff
@@ -43,9 +43,9 @@ _comp_xfunc_cvs_compgen_roots()
     local -a cvsroots=()
     [[ -v CVSROOT ]] && cvsroots=("$CVSROOT")
     [[ -r ~/.cvspass ]] && cvsroots+=($(awk '{ print $2 }' ~/.cvspass))
-    [[ -r CVS/Root ]] && mapfile -tO ${#cvsroots[@]} cvsroots <CVS/Root
+    [[ -r CVS/Root ]] && mapfile -tO "${#cvsroots[@]}" cvsroots <CVS/Root
```
scop, line 46: I wonder/forget what's the reason for this change, the length can never be anything but an integer, right?

bash_completion
```diff
@@ -2267,11 +2290,11 @@ _known_hosts()
...
-    local options=""
-    [[ ${1-} == -a || ${2-} == -a ]] && options=-a
-    [[ ${1-} == -c || ${2-} == -c ]] && options+=" -c"
-    # shellcheck disable=SC2086
-    _known_hosts_real ${options-} -- "$cur"
+    local -a options=()
+    [[ ${1-} == -a || ${2-} == -a ]] && options+=(-a)
+    [[ ${1-} == -c || ${2-} == -c ]] && options+=(-c)
+    local IFS=$' \t\n' # Workaround for connected ${v+"$@"} in bash < 4.4
+    _known_hosts_real ${options[@]+"${options[@]}"} -- "$cur"
```
scop, line 2297: I wonder what's the site where we'd need the said shellcheck disables?

scop (commented):
Slowly getting started with this, got some trivial bits merged out of the way.

completions/cvs
akinomyoga: This is based on the [quoting rule](https://github.com/scop/bash-completion/blob/master/doc/styleguide.md#quoting-of-words). These are the background discussions:

- https://github.com/scop/bash-completion/pull/931#discussion_r1174403080
- https://github.com/scop/bash-completion/issues/932
- https://github.com/scop/bash-completion/pull/986

More explicitly, `IFS` can contain an integer. Even if such an `IFS` is unusual, there is no reason to leave it unquoted when just quoting can avoid it. For this reason, the unquoted arguments of commands have always been a red flag for me, and it is annoying that I stumble upon the line every time I glance at the code.

bash_completion
akinomyoga: Every line. If we do not make `options` array and another function uses an array `options`, `shellcheck` will issue a warning for every line that uses `options` above.

akinomyoga:
Rebased.

bash_completion
scop: I can see the warnings, but what I'm wondering is what is that _another function_ that uses such an array in this context? I see `_comp_compgen_usage` and `_comp_compgen_help` declare it as an array, but neither is being invoked anywhere before `_known_hosts` that I can tell.

Or maybe is this just a naive check without considering the actual call graph, i.e. if the variable *actually* ends up ever being used as array and non-array. Maybe the "use as an array" is triggered by just the (undeclared) use as such in `_comp_compgen_help__parse` without deeper consideration.

(No objections to the change, use as an array looks cleaner to me anyway, but just want to understand what I'm looking at.)

scop (approved):
I think it will take me quite some time to become comfortable with this, but I do see the point and think I even understand some parts of this stuff :) So let's go with this, thanks!

bash_completion
akinomyoga: > Or maybe is this just a naive check without considering the actual call graph,

This is the case. That seems to be the `shellcheck` behavior.

akinomyoga:
Thanks

PR #1029, opened 2023-08-06
refactor: adjust API/naming of generators

akinomyoga:
Changes on top of #1028 / Should be considered after merging #1028

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2244,7 +2244,7 @@ _comp_compgen_dvd_devices()
 # @since 2.12
 _comp_compgen_terms()
 {
-    _comp_compgen -a split -- "$({
+    _comp_compgen_split -- "$({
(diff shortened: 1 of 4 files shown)


bash_completion
```diff
@@ -2237,7 +2237,7 @@ _comp_compgen_dvd_devices()
 # @since 2.12
 _comp_compgen_terms()
 {
-    _comp_compgen -a split -- "$({
+    _comp_compgen_split -- "$({
```
scop, line 2240: This will affect also the API of the old `_terms`. I suppose it would be prudent to follow what's been done e.g. with `_usb_ids` and `_pci_ids` etc.

akinomyoga:
Rebased.

bash_completion
akinomyoga: added 19ef4ae5 and squashed into a03da8ac.

scop (approved):
LGTM, thanks!

PR #1030, opened 2023-08-06
fix(_known_hosts_real): fix nounset/failglob with _comp_split, etc

akinomyoga:


bash_completion
```diff
@@ -2442,27 +2439,29 @@ _known_hosts_real()
             [[ -r $i && ! -d $i ]] && kh+=("$i")
         done
         for i in /etc/ssh2/knownhosts ~/.ssh2/hostkeys; do
-            [[ -d $i ]] && khd+=("$i"/*pub)
+            [[ -d $i ]] || continue
+            _comp_expand_glob tmpkh '"$i"/*pub'
+            ((${#tmpkh[@]})) && khd+=("${khd[@]}")
```
scop, line 2444: I suppose
```suggestion
            ((${#tmpkh[@]})) && khd+=("${tmpkh[@]}")
```

bash_completion
```diff
@@ -2442,27 +2439,29 @@ _known_hosts_real()
             [[ -r $i && ! -d $i ]] && kh+=("$i")
         done
         for i in /etc/ssh2/knownhosts ~/.ssh2/hostkeys; do
-            [[ -d $i ]] && khd+=("$i"/*pub)
+            [[ -d $i ]] || continue
+            _comp_expand_glob tmpkh '"$i"/*pub'
```
scop, line 2443: Aside, not introduced here, but I wonder if this is really supposed to be `.../*pub` rather than `.../*.pub`. Could fix it in this MR while at it (or not, your call), unless we can think of reasons for it to really be without the period.

scop (approved):
Pre-approved with a couple of comments addressed if you agree with them.

bash_completion
akinomyoga: Ah, nice catch. Thank you.

bash_completion
akinomyoga: Thank you for pointing this out. I also think `.pub` is more reasonable. I'll include the fix.

akinomyoga: 6d7c0258abd2c060439008634e6c79233b6dda16 and then squashed into 672215e010d5f78085caf61cecda620b4c67f4ae

bash_completion
akinomyoga: 6d7c0258abd2c060439008634e6c79233b6dda16 and then squashed into 672215e010d5f78085caf61cecda620b4c67f4ae

PR #1032, opened 2023-08-11
Rename and refactor several functions and a variable in `bash_completion`

akinomyoga:
The remaining functions are

- the functions that are supposed to be passed to `complete -F`
- `_known_hosts_real`, which needs to be updated to be a generator. In rewriting to a generator, the names of local variables need to be discussed as in PR #1028.


akinomyoga:
commit e633b626 (`{ => _comp_compgen}_shells`) was moved to PR #1035 1f8e77c757e8c427bd601c2857030dbc7d3337f3.

bash_completion.d/000_bash_completion_compat.bash
```diff
@@ -29,6 +29,11 @@ _comp_deprecate_func 2.12 _modules _comp_compgen_kernel_modules
 _comp_deprecate_func 2.12 _installed_modules _comp_compgen_inserted_kernel_modules
 _comp_deprecate_func 2.12 _usergroup _comp_compgen_usergroup
 _comp_deprecate_func 2.12 _complete_as_root _comp_as_root
+_comp_deprecate_func 2.12 _included_ssh_config_files _comp__included_ssh_config_files
```
scop, line 32: This is being renamed/made to a private function, so exposing the deprecated name is not necessary.
```suggestion
```

completions/mussh
```diff
@@ -39,7 +39,7 @@ _comp_cmd_mussh()
             ;;
         -c)
             compopt -o filenames
-            _comp_compgen -a commands
+            _comp_compgen_commands
```
scop, line 42: Unrelated in this PR, but I guess we can let it pass :)

bash_completion
```diff
@@ -2930,8 +2929,8 @@ _minimal()
 # https://lists.gnu.org/archive/html/bug-bash/2012-01/msg00045.html
 complete -F _minimal ''
 
-# TODO:API: rename per conventions
-__load_completion()
+# @since 2.12
+_comp_load()
```
scop, line 2933: I'm wondering if this is something we want/need to expose as a public function? External completions shouldn't need to use it I think.

akinomyoga: Thank you.

Hmm, `ble.sh` actually wants to use it. `_completion_loader` (which is also used by `git-completion.bash` from the Git project) sets `_minimal` up when none is found, but `ble.sh` wants to fall back into its own completion if none is found. Another option is to make `_completion_loader` accept an option that suppresses the fallback to `_minimal`.

----

I now noticed that I overlooked `_completion_loader` in renaming functions. I think `_completion_loader` should be renamed to `_comp_load` instead of `__load_completion`. I'll fix it.

akinomyoga: > Another option is to make `_completion_loader` accept an option that suppresses the fallback to `_minimal`.

I suggested this, but `_completion_loader` is used as a completion function for `complete -F`, so I realized it causes a problem when a user inputs a command name starting with `-` in the command line. So we should prepare two separate functions for `complete -F` and for a feature to load the completion from a script.

Also, as far as I search on GitHub, `__load_completion` is directly used by some users (regardless of whether it was originally intended by us).

----

The function `_completion_loader` currently seems to be intended for two different roles. One is as a function that is specified to `complete -F`. For example, returning status `124` on successful loading is a part of this use case. The other is as a function that can be called by a user to load a completion for a specific command. Here, I suggest splitting `_completion_loader` into two: 1) `_comp_xxx "$cmd"` that can be used to load a completion for the specified command, and 2) `_comp_yyy` that can be specified to `complete -F`.

Then, I would also like to have an option that does not define the fallback completion `_minimal` when the specific completion is not found. I use it for ble.sh. There are also other scripts that use `__load_completion` (though I'm not sure if they recognize `_completion_loader`). There are two ways:

- **Opt-out `_minimal`** -- `_comp_xxx -- "$cmd"` defines the fallback setting `_minimal` by default, and `_comp_xxx -R "$cmd"` does not. 
- **Opt-in `_minimal`** -- `_comp_xxx -- "$cmd"` does not define `_minimal`, and `_comp_xxx -D "$cmd"` defines `_minimal`.

The opt-in behavior seems natural to me as defining `_minimal` is an additional behavior on top of just loading a specific completion. Actually, the opt-in behavior is mostly the same as the current implementation of `__load_completion`, so my current suggestion is to merge the feature `_comp_xxx` (feature 1 of `_completion_loader`) into `__load_completion` as an option `-D` (defining the default setting `_minimal`) and to make it a public interface under the name `_comp_load`.

The feature `_comp_yyy` is named as a function `_comp_complete_load` in accordance with the naming convention suggested in #1037. I can adjust the naming.

I've pushed those adjustments in commit 77497bc0.

scop (approved):
LGTM, thanks!

akinomyoga:
Thank you.

PR #1033, opened 2023-08-11
refactor: rename `{ => _comp}{_get_first_arg,_count_args}`

akinomyoga:


completions/xdg-mime
```diff
@@ -25,9 +25,9 @@ _comp_cmd_xdg_mime()
     local cur prev words cword comp_args
     _comp_initialize -- "$@" || return
 
-    local args
-    _count_args
-
+    local ret
+    _comp_count_args
+    local args=$ret
```
scop, line 30: Would it not be better to just use `ret` below, or was it considered too generic a name from the location it is set, or...?

scop (changes requested):
Looks mostly good to me, but I'd be happier if we had some tests for the changed behavior e.g. regarding `-` and `--`, as well as the last commit (optarg check fixes).

completions/xdg-mime
akinomyoga: In the original code, the variable `args` has been used also by lines 42, 46, and 53 in addition to line 31. There are no other function calls that affect the variable `ret`, so we can technically replace all of these references to `args` with `ret`. However, it is not so obvious that `ret` is not affected by other codes in between, and it might be broken in the future. For this reason, I basically give a distinct variable name when the variable is referenced by the codes far from the position where it is obtained.

akinomyoga:
There were no tests for `_get_first_arg` at all, so I added basic tests (87fee4e52eebcd20c7ce1c61997769a73a1d8e9a). Then, I added tests in commits 37f458daf10bfd341cced531f545268215296228, 443d0a2c74e898d90fda10c51d41e576cf58bc60, and 36965850.

bash_completion
```diff
@@ -2157,39 +2157,48 @@ _comp_realcommand()
 }
 
 # This function returns the first argument, excluding options
-# @param $1 chars  Characters out of $COMP_WORDBREAKS which should
-#     NOT be considered word breaks. See _comp__reassemble_words.
-# TODO:API: rename per conventions
-_get_first_arg()
+# @var[out] ret First argument if any, or otherwise an empty string
```
scop, line 2160:
```suggestion
# @var[out] ret First argument before current being completed if any, or otherwise an empty string
```

test/t/unit/test_unit_get_first_arg.py
```diff
@@ -0,0 +1,72 @@
...
+        assert_bash_exec(
+            bash,
+            '_comp__test_unit() { local -a "words=$1"; local cword=$2 ret=; shift 2; _comp_get_first_arg "$@" && printf "%s\\n" "$ret"; return 0; }',
+        )
+
+    def test_1(self, bash, functions):
+        assert_bash_exec(bash, "_comp__test_unit '()' 0")
+
+    def test_2(self, bash, functions):
+        """a b| should set args to 1"""
```
scop, line 19: Most docstrings throughout this file seem to be off, as we're returning the first arg (if any), not the number of args.

scop (approved):
Thanks!

Left some final remarks, pre-approved with them addressed.

test/t/unit/test_unit_get_first_arg.py
akinomyoga: Thank you. You are right. Those tests are just converted from the ones for `_count_args`, but I forgot to update the docstrings.

PR #1036, opened 2023-08-13
refactor: refactor `_comp_{get_first_word,count_args}`

akinomyoga:
Waiting for #1033 and #1034

----

There are many handwritten for-loops in the codebase that can be replaced by these functions with small adjustments. This is the first step:
- 0f14cc030ebdb54f3d120d8cadc08651dad00624 ad4472fb9dc66a77414e6a9f8a1ea17319511764 Change the interface to accept detailed specs through options rather than through positional parameters.
- 0384bd51df3f68a3e065aeead4e921462fff6655 Add an interface to get a position of the first argument (instead of its content)
- 521d2bb5330de4a721ae20d37d330190aa4beef0 29398efab4dd06fceb7c94c5329c5002b14cb8b4 3127703f12f3e82ddf3f950a801ab430024aca22 Change behaviors (considering `<>&` and skipping `cword`/`words` re-assembling that seems redundant). These need discussions.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2157,17 +2157,36 @@ _comp_realcommand()
 }
 
 # This function returns the first argument, excluding options
+#
+# Options:
+#     -a GLOB  Pattern of options that take an option argument
+#
 # @var[out] ret First argument before current being completed if any, or
 # otherwise an empty string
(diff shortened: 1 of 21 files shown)


scop (approved):
Nice :+1:

PR #1048, opened 2023-09-18
feat: add fallback 3rd-party completion loaders

akinomyoga:
Ref. https://github.com/ohmybash/oh-my-bash/pull/477


scop (approved):
Nice, pre-approved with one suggestion.

PR #1049, opened 2023-09-23
fix: add misc fixes

akinomyoga:
These are changes that I'd like to apply before submitting the coming PRs.

### DIFF
--- a/completions/apt-get
+++ b/completions/apt-get
@@ -22,7 +22,7 @@ _comp_cmd_apt_get()
     _comp_initialize -n ':=' -- "$@" || return
 
     local special="" i
-    for ((i = 1; i < ${#words[@]} - 1; i++)); do
+    for ((i = 1; i < cword; i++)); do
         if [[ ${words[i]} == @(install|remove|auto?(-)remove|purge|source|build-dep|download|changelog) ]]; then
             special=${words[i]}
             break
--- a/completions/carton
+++ b/completions/carton
@@ -36,11 +36,7 @@ _comp_cmd_carton()
     fi
 
     case $prev in
-        --version | -v)
-            return
-            ;;
-        --help | -h)
-            [[ $has_command ]] || _comp_cmd_carton__commands "$1"
+        --version | --help | -[vh])
             return
             ;;
         --cpanfile)
--- a/bash_completion
+++ b/bash_completion
@@ -719,6 +719,7 @@ _comp_compgen_set()
 #   -P arg  The same as the compgen option -P.
 #   -o arg  The same as the compgen option -o.
 # @param $1 String to split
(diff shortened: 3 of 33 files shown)


completions/carton
```diff
@@ -75,6 +75,7 @@ _comp_cmd_carton()
         exec)
             # skip all the options --, -v, etc. and identify the command name
             # position.
+            local i
```
scop, line 78: This seems unnecessary, `i` is localized on line 21 above, and I think the intent is to continue iterating here where the earlier loop left off.

completions/apt-get
```diff
@@ -22,7 +22,7 @@ _comp_cmd_apt_get()
     _comp_initialize -n ':=' -- "$@" || return
 
     local special="" i
-    for ((i = 1; i < ${#words[@]} - 1; i++)); do
+    for ((i = 1; i < cword; i++)); do
```
scop, line 25: Just a note, the class of these changes makes us more likely complete a "special" when one was already specified after the completion point. It's not necessarily a bad thing, and in some cases we ended up doing that anyway through some possible weird accidents, and in most cases completing something that's valid for a "special" before it isn't the right thing to do anyway. So all good as far as I'm concerned.

scop (approved):
LGTM, pre-approved with the carton comment addressed.

completions/apt-get
akinomyoga: I also thought about that possibility (that this type of loop is intended for checking also a "special" after the completion point), but in that case, I would expect `i < ${#words[@]}` instead of `i < ${#words[@]} - 1`. Maybe I miss something. Just out of curiosity, do you know or remember the reason for `- 1`? My guess was just that it was written under the assumption that the completion point is always in the last word.

The only exception is `apt-cache` where the case of a "special" after the completion point seems to be explicitly handled. This was processed in a separate commit a57f1b415610af604f275622212573f0231c96de.

completions/carton
akinomyoga: Ah, thank you for pointing out. I missed it. I removed the first loop in an (unsubmitted) commit [`b92f4ce`](https://github.com/scop/bash-completion/commit/b92f4ce1838a47df1c421ba1f2286f09e229caa2#diff-91682bca968cc4516eec6a97d940a3cb4af23ede5aa518e28813e6635bfbfec9), but I seemed to have missed that `i` of the first loop is reused here. I'll drop the change. I'll also update the unsubmitted commit.

akinomyoga:
I added another commit cd063f9a40e4dc85e0a9de7ceeac019cdac57449 for `carton`.

completions/carton
akinomyoga: I've dropped the change from the leakvar commit.

PR #1051, opened 2023-09-24
refactor(completions/*): apply API renaming

akinomyoga:


completions/apt-get
```diff
@@ -1,20 +1,22 @@
...
     fi
 }
 
-_comp_deprecate_func 2.12 _comp_cmd_apt_get_installed_packages \
-    _comp_xfunc_apt_get_installed_packages
+# @deprecated 2.12
+_comp_cmd_apt_get_installed_packages()
+{
+    _comp_compgen -ax apt-get installed_packages "$@"
+}
```
scop, line 19: I think this did not exist before 2.12, so perhaps we can remove the function altogether?
```suggestion
```

scop (approved):
Thanks, pre-approved with a few comments.

completions/apt-get
akinomyoga: Thank you. OK, it was introduced as a function `_apt_get_installed_packages` in commit ef1b97befb02a4f710c498a37df5c0a8f29b8f5b on 2020-11-07 (which is later than 2.11 release on 2020-07-25).

akinomyoga:
rebased

PR #1053, opened 2023-09-24
fix: add misc fixes (misc-2)

akinomyoga:


completions/_umount.linux
```diff
@@ -109,8 +109,9 @@ _comp_cmd_umount()
                 jfs minix msdos ncpfs nfs nfs4 ntfs ntfs-3g proc qnx4 ramfs
                 reiserfs romfs squashfs smbfs sysv tmpfs ubifs udf ufs umsdos
                 usbfs vfat xfs'
-            _comp_compgen -a fstypes
-            [[ $split ]] && COMPREPLY=(${COMPREPLY[@]/#/$prev,})
+            _comp_compgen -a _fstypes
```
scop, line 112: This raises an eyebrow, I suppose we should keep it as `fstypes`?

akinomyoga: Ah, this line should not be changed. I don't remember why I edited this line. Fixed it in commit 41457649ccb8c8191ed8ab65b5457a2d185e19ed.

scop (approved):

PR #1059, opened 2023-10-04
feat(_comp_variable_assignments): complete only well-known `LC_*`-variables

calestyo:
POSIX.1-2017 standardises:
LANG, LC_ALL, LC_COLLATE, LC_CTYPE, LC_MESSAGES, LC_MONETARY, LC_NUMERIC and LC_TIME

The GNU C Library version 2.38 standardises:
LC_ADDRESS, LC_IDENTIFICATION, LC_MEASUREMENT, LC_NAME, LC_PAPER and LC_TELEPHONE

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1285,7 +1285,10 @@ _comp_variable_assignments()
         TERM)
             _comp_compgen_terms
             ;;
-        LANG | LC_*)
+        LANG | LC_ALL | LC_COLLATE | LC_CTYPE | LC_MESSAGES | LC_MONETARY | \
+        LC_NUMERIC | LC_TIME | \
+        LC_ADDRESS | LC_IDENTIFICATION | LC_MEASUREMENT | LC_NAME | \
+        LC_PAPER | LC_TELEPHONE)
             _comp_compgen_split -- "$(locale -a 2>/dev/null)"
             ;;
         LANGUAGE)


akinomyoga:
> The GNU C Library version 2.38 standardises:

They are not standard. They are GNU extensions. However, bash-completion doesn't assume GNU systems.

Also, is there any standard that LC_* not listed in POSIX and GNU extensions are prohibited? Otherwise, I don't see a reason for this patch.

calestyo:
> They are not standard. They are GNU extensions. However, bash-completion doesn't assume GNU systems.

Well but right now, where `LC_*` is used, they would also be completed.

Also, doesn't it make sense to apply:
https://github.com/scop/bash-completion/blob/15b74b1050333f425877a7cbd99af2998b95c476/CONTRIBUTING.md?plain=1#L184-L185

only to GNU features that bash-completion would need to **depend** on?

Here, no dependency would be introduced,... merely a completion offer, that non-GNU platforms would get, despite their platform possibly not honouring these variable names as special.

> Also, is there any standard that LC_* not listed in POSIX and GNU extensions are prohibited? Otherwise, I don't see a reason for this patch.

No, but just because their is no such standard, it makes IMO sense to have the patch:<br>
Anyone is free to use `LC_MY_SUPER_CONFIG_VAR`… but it doesn't make sense to provide locale name completions for such cases.

akinomyoga:
> Also, doesn't it make sense to apply:
> 
> https://github.com/scop/bash-completion/blob/15b74b1050333f425877a7cbd99af2998b95c476/CONTRIBUTING.md?plain=1#L184-L185
> 
> only to GNU features that bash-completion would need to **depend** on?

Yes, that is the reason we supported `LC_*`. The support for `LC_*` does not *depend* on GNU extensions, so we can support them for the general systems that might support extra locale variables.

Your reason that the support for GNU extensions should be added in bash-completion appears to be that bash-completion can be permissive for GNU extensions. This is true. Then, we can make it permissive for any `LC_*`. However, you seem to try to be restrictive for any extensions other than GNU ones. This patch tries to make it permissive for GNU extensions and restrictive for any other extensions, which is not consistent.

For consistency, there are two options: Be restrictive for any extensions (i.e., only list the locale variables defined by POSIX), or be permissive for any extensions as in the current `master`.

> Anyone is free to use `LC_MY_SUPER_CONFIG_VAR`…

Yes.

> but it doesn't make sense to provide locale name completions for such cases.

Why?

calestyo:
> The support for LC_* does not depend on GNU extensions, so we can support them for the general systems that might support extra locale variables.

Well I don't think "my code" does depend on GNU extensions either, it just supports only well-known names, those standardised by POSIX and those added by GNU. If other libc implementations would add further names, we could and should of course add them, too.<br>
I've look, and at least couldn't find and such for musl libc and a quick search didn't reveal anything for FreeBSD.

> However, you seem to try to be restrictive for any extensions other than GNU ones.

That would absolutely not be my intention. If any non-GNU `LC_*` extensions were known we should add them - well at least if there is some "reasonable" userbase (meaning that I compile my own fork of some libc and add `LC_CALESTYO`, then this woulnd't be well-known enough to be added ;-) ).

> Why?

Well, why don't we provide locale name completions for any other variables for which nothing else is done? `LC_CONFIG` might be the location of a config variable of some project named `lc` or maybe an acronym of a longer name. For that it would surely be better to get the file completion.


Anyway...  I just meant this as a minor improvement to complete only those var names with locale strings, for which we know that they're actually used as such. Didn't expect this was so controversial. :-D

Feel free to simply close :-)

akinomyoga:
> If other libc implementations would add further names, we could and should of course add them, too.

Who pays that maintenance cost for the hardcoded list? I don't think it's worth paying the maintenance cost just to replace `LC_*` and hardcode the list of known names in the wild.

> I've look, and at least couldn't find and such for musl libc and a quick search didn't reveal anything for FreeBSD.

Have you checked macOS? Have you checked NetBSD, OpenBSD, Solaris, Minix, AIX, etc.? There are even more operating systems where Bash runs. Could you make a detailed survey and summarize the support for the locale variables in all the major systems? Could you make a report on the locale variables every year (or more frequently)? That is essentially "the maintenance cost" mentioned above.

> Well, why don't we provide locale name completions for any other variables for which nothing else is done?

Because these are the most probable completions. If there is no other prior information, a variable of the name `LC_*` has a higher probability of expecting locale names than expecting filenames. A variable of other variables has a higher probability of expecting filenames than locale names. Such probabilities are subjective in nature in the Bayesian sense, but I'd be surprised if your *belief* is that a variable of the name `LC_*` (if there is no additional information) expects filenames more probably than locale names.

calestyo:
> Who pays that maintenance cost for the hardcoded list? I don't think it's worth paying the maintenance cost just to replace LC_* and hardcode the list of known names in the wild.

I think the list would be rather quite stable... POSIX hasn't changed in... forever? ^^ ... GNU hasn't changed since I can think back.<br>
So rather, I'd expect that other libc implementations would come along and add something new. But even there, from the main ones I know, I couldn't find any which did so.


Anyway... as said... just reject it, fine by me :-D

akinomyoga:
> I think the list would be rather quite stable... POSIX hasn't changed in... forever? ^^ ... GNU hasn't changed since I can think back.

The major maintenance cost is not the one to update the list. The maintenance cost arises when we confirm that it doesn't change constantly. No matter how stable you think the list is or no matter how you expect other libc implementations follow GNU, unless POSIX declares that the list will never be changed forever, unless GNU declares it will never change the list, and unless all the other libc implementations declares it will completely follow POSIX and GNU forever, we need to check if the list is changed or unchanged.

PR #1064, opened 2023-10-28
Added RequiredRSASize option for ssh client.

inntran:
This is actually a bug, because the "RequiredRSASize" exists in recent OpenSSH source.

The reason I make this change are as follows:
1. The RequiredRSASize option was added in openssh/openssh-portable@54b333d .
2. RHEL 9.1 changed minimum RSA key size to 2048 by default, broke connections to old SSH servers. Reference: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/9.1_release_notes/new-features#BZ-2066882
3. Fedora Linux sync'd the change from RHEL 9.1. Reference: https://packages.fedoraproject.org/pkgs/openssh/openssh-server/fedora-37.html
4. I was connecting to a legacy device over SSH and tried very hard to remember this option.

I have ran the supplied test for ssh, got the following output:
```
====== 12 passed, 1 xfailed in 35.62s ======
```

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -73,11 +73,11 @@ _comp_xfunc_ssh_options()
         PasswordAuthentication PermitLocalCommand PKCS11Provider Port
         PreferredAuthentications ProxyCommand ProxyJump ProxyUseFdpass
         PubkeyAcceptedAlgorithms PubkeyAuthentication RekeyLimit RemoteCommand
-        RemoteForward RequestTTY RevokedHostKeys SecurityKeyProvider SendEnv
-        ServerAliveCountMax ServerAliveInterval SetEnv StreamLocalBindMask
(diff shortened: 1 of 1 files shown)


inntran:
To get the latest available OpenSSH client options:
```
curl --silent https://raw.githubusercontent.com/openssh/openssh-portable/master/ssh_config.5 | awk '$1==".It" && $2=="Cm" && $3!="Host" && $3 ~ /^[A-Z]/ {print $3}' | xargs | fold -s -w 72
```
I didn't include this because this might break user's expectations on old version of OpenSSH client.
Or we could include this instruction to help downstream distributions to patch their own.

scop (approved):
Thanks! Merging with commit message tweaked to our standards.

PR #1065, opened 2023-10-31
Add completion for BSD chflags(1) command

Low-power:


completions/chflags
```diff
@@ -0,0 +1,31 @@
...
+        _comp_compgen -- -W '-f -h -v -R $opts'
+    else
+        local ret
+        # The first argument is a list of flags; the rest are filedir.
+        _comp_count_args
+        if ((ret == 1)); then
+            _comp_compgen -- -W 'simmutable nosimmutable sappend nosappend
+                                 archived noarchived opaque noopaque nodump
+                                 dump uimmutable nouimmutable uappend
+                                 nouappend hidden nohidden'
```
akinomyoga, line 23: The first argument seems to be a comma-separated list. Could we use `_comp_delimited` to support the completion of the comma-separated list? Also, does this list depend on the operating system? `chflags` of OpenBSD and NetBSD seem to define different sets of flags. Which implementations did you reference in preparing the list?

- [chflags(1) - OpenBSD manual pages](https://man.openbsd.org/chflags.1)
- [chflags(1)](https://man.freebsd.org/cgi/man.cgi?chflags(1))
- [chflags(1) - NetBSD Manual Pages](https://man.netbsd.org/NetBSD-9.0/chflags.1)

completions/chflags
```diff
@@ -0,0 +1,31 @@
+# chflags(1) completion                                      -*- shell-script -*-
+
+_comp_cmd_chflags()
+{
+    local cur prev words cword was_split comp_args
```
akinomyoga, line 5:
```suggestion
    local cur prev words cword comp_args
```

`was_split` is only needed when the flag `-s` is specified to `_comp_initialize`.

Low-power:
> Could we use _comp_delimited to support the completion of the comma-separated list?

Yes, that would be more useful.

> Which implementations did you reference in preparing the list?

FreeBSD and Darwin.

> `was_split` is only needed when the flag `-s` is specified to `_comp_initialize`.

Yep, I forgot to remove this declaration.

completions/chflags
akinomyoga: > > Which implementations did you reference in preparing the list?
> 
> FreeBSD and Darwin.

Could you switch the list of flags based on the current system? E.g., using `[[ $OSTYPE == openbsd* ]]` and `[[ $OSTYPE == netbsd* ]]`.

akinomyoga:
> > Could we use _comp_delimited to support the completion of the comma-separated list?
> 
> Yes, that would be more useful.
> 
> > `was_split` is only needed when the flag `-s` is specified to `_comp_initialize`.
> 
> Yep, I forgot to remove this declaration.

Thank you for those updates.

completions/chflags
```diff
@@ -0,0 +1,31 @@
...
+
+    if [[ $cur == -* ]]; then
+        # Complete -options
+        local w opts=""
+        for w in "${words[@]}"; do
+            [[ $w == -R ]] && opts="-H -L -P" && break
+        done
+        _comp_compgen -- -W '-f -h -v -R $opts'
+    else
+        local ret
```
scop, line 16: Per #987
```suggestion
        local REPLY
```

completions/chflags
scop, line 1: Let's add at least a basic test case for this; a starting point would be `cd test; ./generate chflags`. See other test cases in `test/t` for inspiration for more.

scop: We should add an entry to `completions/Makefile.am` for this.

completions/chflags
```diff
@@ -0,0 +1,31 @@
...
+    local cur prev words cword comp_args
+    _comp_initialize -- "$@" || return
+
+    if [[ $cur == -* ]]; then
+        # Complete -options
+        local w opts=""
+        for w in "${words[@]}"; do
+            [[ $w == -R ]] && opts="-H -L -P" && break
+        done
+        _comp_compgen -- -W '-f -h -v -R $opts'
```
scop, line 14: FreeBSD (15.0) has `-x`.
```suggestion
        _comp_compgen -- -W '-f -h -v -x -R $opts'
```

completions/chflags
```diff
@@ -0,0 +1,31 @@
...
+    else
+        local ret
+        # The first argument is a list of flags; the rest are filedir.
+        _comp_count_args
+        if ((ret == 1)); then
```
scop, line 19:
```suggestion
        if ((REPLY == 1)); then
```

scop (changes requested):

akinomyoga:
Somehow the CI has failed to start. We recently updated the version of release-please-action in #1075. @Low-power Could you rebase it again on top of the latest `master`?

https://github.com/scop/bash-completion/actions/runs/7713997204

completions/chflags
akinomyoga: @Low-power Thanks, I confirmed this is fixed in your last push.

completions/chflags
akinomyoga: This is also fixed.

completions/chflags
akinomyoga: If `-x` is not available in other BSDs, I think we should reference `$OSTYPE` for `-x` the same as for the first word.

akinomyoga: 44f29dae2c11f6e228d7e947e9aa0e6d9c1fb106

akinomyoga:
I rebased. I also added links to the references in commit 5bd938bee9fe9a150397a04f2431b1d635b0c633.

scop (approved):

scop:
Thanks!

Issue #1067, opened 2023-11-18
Request for way to disable "xspec" extension-based completion


duelafn:
Hello, I often work with files with missing or incorrect file extensions and the builtin extension-based completions for things like unzip make this difficult. I am aware that I can undo individual completions by ading, say, `complete -o default -o bashdefault unzip` to my .bashrc (and I'm currently doing this for the most common offenders), but really I'd rather be able to fully disable extension-based filtering for all commands without having to continually update my list of exceptions.

I'm hoping for some kind of environment variable that that I can set even after loading bash-completion that will make, at least, the "_filedir_xspec" function fall back to normal file matching (maybe something like `BASH_COMPLETION_DISABLE_EXTENSION_FILTERS`).

Thanks for some great software and thanks for considering my feature request!

akinomyoga:
It should be reasonable that emptying `_comp_xspecs` disables all the xspec completions.

```bash
declare -gA _comp_xspecs=()
```

However, it doesn't work currently because emptying `_comp_xspecs` leaving the completion settings `-F _filedir_xspecs` disables any filenames.

akinomyoga:
@duelafn I realized that the above setting actually works with the current implementation (except for `7z`, `file-roller`, and `jar`, for which `_filedir_xspec` are directly called internally). When the `_comp_xspecs` entry does not exist, the completion loader does not set up `_filedir_xspec` in the first place. If you are using the release version, you need to empty `_xspecs` instead of `_comp_xspecs`. In summary, you can add the following setting immediately after loading `bash_completion`:

```bash
declare -gA _comp_xspecs=() _xspecs=()
```

duelafn:
Thanks for the update - I was just experimenting and came to the same conclusion (that clearing the array works currently and that clearing _xspecs works for the ancient release included in Debian). Fantastic!

PR #1068, opened 2023-11-18
feat(_filedir_xspec): allow unsetting _comp_xspec for plain completion

akinomyoga:
~~Resolve https://github.com/scop/bash-completion/issues/1067~~

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2938,7 +2938,7 @@ _filedir_xspec()
     _comp_quote_compgen "$cur"
     local quoted=$ret
 
-    local xspec=${_comp_xspecs[${1##*/}]-${_xspecs[${1##*/}]-}}
+    local xspec=${_comp_xspecs[${1##*/}]-${_xspecs[${1##*/}]-'*'}}
     local -a toks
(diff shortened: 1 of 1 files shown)


akinomyoga:
@duelafn Could you check if this version works as you expected? You can clone the repository `akinomyoga/oh-my-bash` and check out the branch `empty-xspec`. Then, you can set `declare -gA _comp_xspec=()` after loading `bash-completion` to disable all the xspec-based completions.

akinomyoga:
Since just `_comp_xspecs=() _xspecs=()` turns out to work in #1067, this PR is not needed. However, it would be still useful to fall back to the normal filename completions when the corresponding `_comp_xspec` empty is unspecified in case the `_comp_xspec` entry is removed after attempting a completion.

akinomyoga:
I realized that the pattern specified to `_comp_xspecs` is reverted, so what we should specify here is `'!*'` instead of `'*'`, but then it is equivalent to the empty string. This PR has the opposite effect, and if modified to use `'!*'`, this PR wouldn't have any effect. Let me close the PR.

Issue #1062, opened 2023-10-16
use of `$FUNCNAME` when `set -u` is in place and `FUNCNAME` was unset


calestyo:
## Describe the bug

Throughout the code `$FUNCNAME` is used, which causes an unbound error, if `set -u` is in place and `FUNCNAME` was unset.

Interestingly, which I don't quite understand why, one can `unset FUNCNAME` even outside a function (where I thought it wasn't supposed to exist) and it will have the effect of it being no longer special.

## Expected behavior

I guess `${FUNCNAME-}` or so should be used.

## Additional context

If it helps, I could provide a PR, but I have no idea which value we should use (especially also in the `FUNCNAME[i]` cases)... simply `unknown`?<br>
The empty string? Though that would look ugly in many error messages (because of the `"bash_completion: : message`)... so maybe use something `${FUNCNAME:+"$FUNCNAME": }` and print the `: ` via that (and only if there is a funcname?

Thanks,
Chris.

calestyo:
Oh, I haven't really checked whether any of `bash`’s other "special" variables that can be unset are used in a fashion that would cause errors when `set -u` is in use.

Maybe that should be done.

akinomyoga:
You shouldn't unset the Bash special variables in an interactive *Bash* session even though it is possible. At least, bash-completion cannot be responsible for every kind of broken environment. There are many different ways to break the environment by removing or disabling the Bash built-in features.
- For example, one can overwrite builtin commands by functions, aliases, etc. To protect ourselves, one might use the `builtin` builtin, but a function or an alias can also overwrite the `builtin` builtin. One might then think about using the POSIX mode to access the builtins directly, but it again fails when the builtins are disabled by `enable -n`. If the `enable` builtin itself is disabled, then there is no way to recover the builtins.
- For another example, one can mark random variables `readonly` at the top level of the shell context (e.g. `readonly i`), and then any local variables of the same name cannot be defined.
- Unsetting `FUNCNAME` is a breakage of the fundamental environment at the same level. There are many similar variables including `BASH`, `COMP_WORDBREAKS`, `LINENO`, `OSTYPE`, `PATH`, `PWD`, etc. It might be worked around by checking the existence of all the variables, but it complicates the codebase unnecessarily.

I have a doubt about complicating the codebase just to protect us from such unrealistic cases.

[...]

PR #1069, opened 2023-11-18
refactor(make): Use awk to extract target names

akinomyoga:
This is the last piece of `#TODO:API`.

This is also the refactoring mentioned in https://github.com/scop/bash-completion/pull/546#issuecomment-871197885. As mentioned there, we might separate the awk script to another file in `helpers/`.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -3207,6 +3207,22 @@ _comp_xfunc()
     "$xfunc_name" "${@:3}"
 }
 
+# Call a POSIX-compatible awk.  Solaris awk is not POSIX-compliant, but Solaris
+# provides a POSIX-compatible version through /usr/xpg4/bin/awk.  We switch the
+# implementation to /usr/xpg4/bin/awk in Solaris if any.
+# @since 2.12
+if [[ $OSTYPE == *solaris* && -x /usr/xpg4/bin/awk ]]; then
+    _comp_awk()
+    {
+        /usr/xpg4/bin/awk "$@"
+    }
+else
+    _comp_awk()
+    {
+        command awk "$@"
+    }
+fi
+
 # source compat completion directory definitions
 _comp__init_compat_dirs=()
 if [[ ${BASH_COMPLETION_COMPAT_DIR-} ]]; then
--- a/completions/make
+++ b/completions/make
@@ -15,7 +15,7 @@ _comp_cmd_make__extract_targets()
     [[ $mode == -d && $prefix == */* ]] &&
         prefix_replace=${prefix##*/}
 
-    awk -f "${BASH_SOURCE[0]%/*}/../helpers/make-extract-targets.awk"
+    _comp_awk -f "${BASH_SOURCE[0]%/*}/../helpers/make-extract-targets.awk"
 }
 
 # Truncate the non-unique filepaths in COMPREPLY to only generate unique
--- a/completions/make
+++ b/completions/make
@@ -1,91 +1,92 @@
 # bash completion for GNU make                             -*- shell-script -*-
 
-# TODO: rename per API conventions, rework to use vars rather than outputting(?)
-_make_target_extract_script()
+# Extract the valid target names starting with PREFIX from the output of
+# `make -npq'
+# @param mode    If this is `-d', the directory names already specified in
+#                PREFIX are omitted in the output
+# @param prefix  Prefix of the target names
+_comp_cmd_make__extract_targets()
 {
-    local mode="$1"
-    shift
+    local mode=$1
+    local -x prefix=$2
 
-    local prefix="$1"
-    local prefix_pat=$(command sed 's/[][\,.*^$(){}?+|/]/\\&/g' <<<"$prefix")
-    local basename=${prefix##*/}
-    local dirname_len=$((${#prefix} - ${#basename}))
-    # Avoid expressions like '\(.\{0\}\)', FreeBSD sed doesn't like them:
-    # > sed: 1: ...: RE error: empty (sub)expression
-    local dirname_re=
-    ((dirname_len > 0)) && dirname_re="\(.\{${dirname_len}\}\)"
+    # display mode, only output current path component to the next slash
+    local -x prefix_replace=$prefix
+    [[ $mode == -d && $prefix == */* ]] &&
+        prefix_replace=${prefix##*/}
 
-    if [[ ! $dirname_re ]]; then
-        local output="\1"
-    elif [[ $mode == -d ]]; then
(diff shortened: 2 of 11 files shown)


completions/make
```diff
@@ -1,91 +1,92 @@
...
+        prefix_replace=${prefix##*/}
+
+    awk '
+      BEGIN {
+        prefix = ENVIRON["prefix"];
+        prefix_replace = ENVIRON["prefix_replace"];
+        is_target_block = 0;
+        target = "";
+      }
+      function starts_with(str, prefix) {
```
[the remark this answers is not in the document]
akinomyoga: Solaris awk lacks many features. It doesn't have a user-defined function as you guessed. It doesn't support `ENVIRON` It doesn't support `-v var=value` command-line arguments. It doesn't seem to support the condition of the form `!expr` in the block beginning.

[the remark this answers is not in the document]
akinomyoga: One way is to use `/usr/xpg4/bin/awk`. Solaris intensionally preserves the ancient feature set of core utilities, but at the same time they provide POSIX utilities in `/usr/xpg4/bin`. Another way is to use `awk` with a limited set of features, which I'm not even sure if it is possible.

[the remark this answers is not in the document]
scop: Maybe trying to limp along with the simplest of awk would lose the readability benefits of using awk in the first place.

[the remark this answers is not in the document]
akinomyoga: I separated the AWK script in commit db3a82d46c67fb6320967158e69c0022a251b3e3.

For Solaris awk, I found also other cases where the feature that Solaris awk does not support is used. In commit 6a9b7fe41f76fe863cd07f36d48b0c5e48f6c7e4, I added a utility `_comp_awk` (which calls `/usr/xpg4/bin/awk "$@"` in Solaris or otherwise `awk "$@"`) and replaced a few calls of `awk` that need POSIX features. The use of `ENVIRON` in `portinstall` was recently introduced by me in #1053.

In commit 36a2e375c, I have modified the condition `$NF=2` to `NF >= 2`. Solaris awk doesn't like the assignment `$NF=2` in the block condition, but to begin with, I doubt this condition. See the commit message for the details.

completions/make
```diff
@@ -1,91 +1,92 @@
...
+      /^# +Variables/, /^# +Files/ { next; }      # skip until files section
+      /^# +Not a target/, /^$/     { next; }      # skip not target blocks
+
+      # The stuff above here describes lines that are not
+      #  explicit targets or not targets other than special ones
+      # The stuff below here decides whether an explicit target
+      #  should be output.
+
+      starts_with($0, prefix) { is_target_block = 1; }
+      !is_target_block { next; }
```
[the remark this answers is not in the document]
akinomyoga: I changed the condition to `is_target_block == 0` in the separated AWK file.

bash_completion
```diff
@@ -3207,6 +3207,22 @@ _comp_xfunc()
     "$xfunc_name" "${@:3}"
 }
 
+# Call a POSIX-compatible awk.  Solaris awk is not POSIX-compliant, but Solaris
+# provides a POSIX-compatible version through /usr/xpg4/bin/awk.  We switch the
+# implementation to /usr/xpg4/bin/awk in Solaris if any.
+# @since 2.12
+if [[ ${OSTYPE-} == solaris* && -x /usr/xpg4/bin/awk ]]; then
```
scop, line 3214: I have no actual information on this, but the other cases in the tree we have test for `*solaris*`. No other use accounts for `$OSTYPE` being unset either, I suppose we can follow suit here.
```suggestion
if [[ $OSTYPE == *solaris* && -x /usr/xpg4/bin/awk ]]; then
```

bash_completion
```diff
@@ -3207,6 +3207,22 @@ _comp_xfunc()
...
+    _comp_awk()
+    {
+        /usr/xpg4/bin/awk "$@"
+    }
+else
+    _comp_awk()
+    {
+        awk "$@"
```
scop, line 3222: Now that we're doing this, maybe...?
```suggestion
        command awk "$@"
```

helpers/make-extract-targets.awk
scop, line 1: We may want to run this file through `gawk --pretty-print` sometime (and if so, make that happen in pre-commit, too). That may run into odds with our `.editorconfig` though, and anyway no need to do it in this PR.

scop (approved):

bash_completion
akinomyoga: Thanks for the suggestion. There are some comments as follows, but I'll apply the suggested changes.

- For the unset guard, I remembered the recent comment https://github.com/scop/bash-completion/issues/1062#issuecomment-1765557130. Although this case was easily worked around by `${var-}`, we can still assume that `OSTYPE` is preserved.
- For the value of `OSTYPE` in Solaris, as far as I see the change history of `support/config.guess` (and `support/mkmachtype` at the early stage) in the upstream Bash repository, it seems to have always started with `solaris`. The related commits in the Bash repository are 5e6f45d9b, 50fc6e33f, 5c5c75236, 1e8fce0f8, 7117c2d22, bb70624e9, d166f0488, ccc6cda31, 726f6388. However, I'm not sure if the style will change in the future. Also, I'm not sure if there is a custom version of Bash that has other values.

bash_completion
akinomyoga: There are many occurrences of `awk` in the codebase. Can I apply the change to all the occurrences in a separate PR? Or do we use `command` only in this case?

helpers/make-extract-targets.awk
akinomyoga: I checked the output of `gawk --pretty-print`. I see what you say; The indentation is performed by TAB characters, which conflicts with `indent_style = space` in the `.editorconfig`.

bash_completion
akinomyoga: Or another way is to replace all the other occurrences of `awk` with `_comp_awk`.

akinomyoga:
amend [`1a221053..10d1ec2a`](https://github.com/scop/bash-completion/compare/1a221053b52bc00765e9f6196fb0239d7c418df6..10d1ec2aed0d2c348b215dc3704aafbadd139d74): I added more cases where XPG awk is needed in Solaris.

bash_completion
akinomyoga: Got it.

akinomyoga: e18667a9a17d7db0b480426311ad0fe911c26dc3..302dc52b9b02dc9a8fa9b1833012b3c4f96b6d7b

PR #1072, opened 2023-11-27
refactor: prefix `command` to the `awk` command

akinomyoga:
As discussed in https://github.com/scop/bash-completion/pull/1069#discussion_r1402672706.

In this PR, I change `awk` to `command awk` where Solaris awk works (65ea69942).

However, another possibility is that we entirely switch to `_comp_awk` so that we do not have to care about the feature set of Solaris awk. If you like this direction, I can adjust the PR.

There are additional changes:

- There are a few cases of directly calling `sed` without prefixing `command`, which were recently added.  I also fix these `sed` cases in this PR (2e382861c).
- In addition, there are several compatibility issues related to the implementations of `awk` (b207dfc38, 7b975fce9, a1c5c30dc) and `sed` (0d798df58, 17e6d83d8, 7bea087aa).
- I also apply `shfmt` and `shellcheck` to `test/runLint` similarly to #1071 for `test/fallback/update-fallback-links` (bae5ee80a).

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1702,7 +1702,7 @@ _comp_compgen_available_interfaces()
         else
             ifconfig -a || ip -c=never link show || ip link show
         fi
-    } 2>/dev/null | awk \
+    } 2>/dev/null | _comp_awk \
         '/^[^ \t]/ { if ($1 ~ /^[0-9]+:/) { print $2 } else { print $1 } }')" &&
         _comp_compgen -U generated set "${generated[@]}"
 }
@@ -1943,7 +1943,7 @@ _comp_compgen_services()
         systemctl list-units --full --all ||
             systemctl list-unit-files
     } 2>/dev/null |
-        awk '$1 ~ /\.service$/ { sub("\\.service$", "", $1); print $1 }')
+        _comp_awk '$1 ~ /\.service$/ { sub("\\.service$", "", $1); print $1 }')
     _comp_split -la services "$_generated"
 
     if [[ -x /sbin/upstart-udev-bridge ]]; then
@@ -2016,7 +2016,7 @@ _comp_compgen_kernel_modules()
 _comp_compgen_inserted_kernel_modules()
 {
     _comp_compgen -c "${1:-$cur}" split -- "$(PATH="$PATH:/sbin" lsmod |
-        awk '{if (NR != 1) print $1}')"
+        _comp_awk '{if (NR != 1) print $1}')"
 }
 
 # This function completes on user or user:group format; as for chown and cpio.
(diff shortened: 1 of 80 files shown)


bash_completion
```diff
@@ -1702,7 +1702,7 @@ _comp_compgen_available_interfaces()
         else
             ifconfig -a || ip -c=never link show || ip link show
         fi
-    } 2>/dev/null | awk \
+    } 2>/dev/null | command awk \
```
scop, line 1705: In https://github.com/scop/bash-completion/pull/1069#discussion_r1406662760 I tried to say that I'd prefer use of `_comp_awk` throughout (not `command awk`). Now that we have the function, I think it makes perfect sense to make use of it everywhere. Unless of course there's a reason we should not do that, but I'm not aware of any.

completions/chronyc
```diff
@@ -4,7 +4,7 @@ _comp_cmd_chronyc__command_args()
 {
     local -a args
     _comp_split args "$("$1" help 2>/dev/null |
-        awk '/^'"$prev"'\s[^ []/ { gsub("\\|", " ", $2); print $2 }')"
+        command awk '/^'"$prev"'[ 	][^ []/ { gsub("\\|", " ", $2); print $2 }')"
```
scop, line 7: I wonder if there's a better way to express this, the literal tab isn't quite obvious just by reading.

scop (approved):
Left a couple of comments, pre-approved with them addressed.

completions/chronyc
akinomyoga: I'll replace it with `\t`. Although POSIX ERE does not define `\t`, POSIX awk specifies the extension to the ERE for escape sequences of control characters including `\t`.

bash_completion
akinomyoga: I used `_comp_awk` in the alternative commit 8372459deba23b86b6eb687665df90c11bda38d9.

Now we do not need Solaris-awk-specific adjustments, so I dropped a1c5c30dc72621d3b36329154e3d1a9babef08c9. Instead, I added commit 9ec2e9eeb0d7af3bd65ba402fb58d53488690146 for the adjustment in the other way.

completions/chronyc
akinomyoga: Fixed it in https://github.com/scop/bash-completion/pull/1072/commits/1989ba9ad3de5cf33da3c9623fa87be5194fac08#diff-24150d0a0f171f44ed1383b6c2dff5a662fdee266f192390f67c3e26c22a4c42

akinomyoga:
As mentioned in https://github.com/scop/bash-completion/pull/1072#discussion_r1407642224, there is a new commit 9ec2e9eeb0d7af3bd65ba402fb58d53488690146.

scop:
Thanks!

PR #1074, opened 2023-11-28
Work around `shopt -s patsub_replacement` newly supported in Bash 5.2

akinomyoga:
The main fixes are the last three commits: caeedee5, cfdcc937, and c1883818. See the commit messages.

The other commits are relatively trivial fixes overlapping with the above ones: 87c535a0, 59ad04d1, 85203cb1, a15f6308, 21d28ad0.


test/runLint
```diff
@@ -61,3 +64,9 @@ gitgrep "$cmdstart"'unset [^-]' 'Explicitly specify "unset -v/-f"'
 
 gitgrep "$cmdstart"'((set|shopt)\s+[+-][a-z]+\s+posix\b|(local\s+)?POSIXLY_CORRECT\b)' \
     'fiddling with posix mode breaks keybindings with some bash versions'
+
+gitgrep '\$\{([^{}\n]|\{.*\})+/([^{}\n]|\{.*\})+/([^{}"\n]|\{.*\})*\$.*\}' \
+    '$rep of ${var/pat/$rep} needs to be double-quoted for shopt -s patsub_replacement (bash >= 5.2)'
```
scop, line 69: The commit message of 5b959a7e55fc743b59b55249d8529113294ea022 details a case why we should be rather using `_comp_compgen` than double quoting. Should we update this to suggest that, too?

Looks good to me otherwise, pre-approved and good to merge with this addressed (or not addressed if that's the right thing to do).

scop (approved):

test/runLint
akinomyoga: This becomes long, so in commit c35af29e96dc62eb85b0306397ddee72c5441623, I added a new subsection in `doc/styleguide.md` under the quoting section and added pointers to the related lint items.

akinomyoga:
[`32f0088d..2a11209b`](https://github.com/scop/bash-completion/compare/32f0088dc85cf4c90aff18a524bf91a2fd436053..2a11209b867c22256916e5d37c18596c7ed8df59): I added a change for `ret => REPLY` (#987).

PR #1079, opened 2023-12-22
add y4m and heic/heif/avif extensions to feh

chemag:
* y4m supported in imlib2 since [1]
* heif Supported in feh through imlib2 since [2]

[1] https://git.enlightenment.org/old/legacy-imlib2/commit/be721b0335845f4d127f6c9041aa98a1d5e7ee22
[2] https://git.enlightenment.org/old/legacy-imlib2/commit/4b4f72c7ffd476df6e450d84f7acb078808cb931

### DIFF
--- a/completions/feh
+++ b/completions/feh
@@ -111,7 +111,7 @@ _comp_cmd_feh()
 
     # FIXME: It is hard to determine correct supported extensions.
     # feh can handle any format that imagemagick can plus some others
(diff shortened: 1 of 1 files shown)


scop:
Looks good to me otherwise, but I wonder if we should be adding some more HEIF extensions, ref https://git.enlightenment.org/old/legacy-imlib2/src/commit/3106fbd0d2403b5c98f8dec41cf7696d13dac250/src/lib/loaders.c#L35

chemag:
> Looks good to me otherwise, but I wonder if we should be adding some more HEIF extensions, ref https://git.enlightenment.org/old/legacy-imlib2/src/commit/3106fbd0d2403b5c98f8dec41cf7696d13dac250/src/lib/loaders.c#L35

Happy to add "avci" and "avcs". Would you mind if we merge this one, and then I send another pull request?

chemag:
On a second thought, avci is not supported be libheif, which means it's not supported by imlib2, which means it's not supported by feh. 

Sent a new pull request: https://github.com/scop/bash-completion/pull/1081

PR #1081, opened 2023-12-29
feat(feh): deassociate with avci/avcs

chemag:
feh supports several HEIF sub-types through imlib2, which gets the support from libheif. The latter does not support avci/avcs yet.

### DIFF
--- a/completions/feh
+++ b/completions/feh
@@ -111,7 +111,7 @@ _comp_cmd_feh()
 
     # FIXME: It is hard to determine correct supported extensions.
     # feh can handle any format that imagemagick can plus some others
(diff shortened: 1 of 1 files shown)


scop:
Oh. Is this something that should be reported against imlib2 too in that case, per https://github.com/scop/bash-completion/pull/1079#issuecomment-1868511393 ?

chemag:
> Oh. Is this something that should be reported against imlib2 too in that case, per [#1079 (comment)](https://github.com/scop/bash-completion/pull/1079#issuecomment-1868511393) ?

Yeah, agree. 

https://github.com/strukturag/libheif/blob/master/libheif/heif.h#L355

scop:
Thanks. I don't have an account needed for reporting this against imlib2 nor do I feel like getting one for this purpose, leaving that part for someone else.

PR #1084, opened 2024-01-01
fix: fix misc problems before applying `_comp_split` changes

akinomyoga:
These are a set of trivial fixes that I would like to apply before submitting the _comp_split/_comp_compgen_split/_comp_expand_glob changes.


scop:
Thanks!

PR #1090, opened 2024-01-08
fix: prefer using iproute2 instead of ifconfig

### LINKED ISSUE #1089: Wrong network interface name if the name is long
## Describe the bug

network interface name get truncated with ifconfig

## To reproduce

```
$ ip link show up
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
3: nameislong: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc fq_codel state DOWN mode DEFAULT group default qlen 1000

$ ifconfig 
lo        Link encap:Local Loopback
eth0      Link encap:Ethernet
nameislongLink encap:(hwtype unknown)
```

https://github.com/scop/bash-completion/blob/2b7fc195aeb404b41bc3e5fa8c1e98f10f1ee26a/bash_completion#L1723-L1738

Then the `_comp_compgen_available_interfaces -a` will return `nameislongLink` instead of `nameislong`

Reproduce:

```sh
sudo tunctl -t nameislong
sudo ip addr add 10.23.45.0/24 dev nameislong
sudo ip link set dev  nameislong up

# TAB
curl --interface nameis
```

- [x] Operating system name/distribution and version: `NixOS 24.05.20240102.bd645e8`
- [x] bash version, `echo "$BASH_VERSION"`: `5.2.21(1)-release`
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: `2.11`

## Additional context

It's bug of upstream inetutils, but maybe we can simply use iproute2 as default choose

`ip -c=never link show up || ip link show up || ifconfig`

https://github.com/scop/bash-completion/blob/2b7fc195aeb404b41bc3e5fa8c1e98f10f1ee26a/bash_completion#L1731-L1733

hellodword:
resolves #1089

Same with this https://github.com/scop/bash-completion/blob/2b7fc195aeb404b41bc3e5fa8c1e98f10f1ee26a/completions/wol#L18

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1628,7 +1628,7 @@ _comp_compgen_mac_addresses()
     # - ip link: link/ether
     _comp_compgen -v addresses split -- "$(
         {
-            LC_ALL=C ifconfig -a || ip -c=never link show || ip link show
+            ip -c=never link show || ip link show || LC_ALL=C ifconfig -a
         } 2>/dev/null | command sed -ne \
             "s/.*[[:space:]]HWaddr[[:space:]]\{1,\}\($_re\)[[:space:]].*/\1/p" -ne \
             "s/.*[[:space:]]HWaddr[[:space:]]\{1,\}\($_re\)[[:space:]]*$/\1/p" -ne \
@@ -1699,7 +1699,7 @@ _comp_compgen_ip_addresses()
     local PATH=$PATH:/sbin
     local addrs
     _comp_compgen -v addrs split -- "$({
-        LC_ALL=C ifconfig -a || ip -c=never addr show || ip addr show
+        ip -c=never addr show || ip addr show || LC_ALL=C ifconfig -a
     } 2>/dev/null |
         command sed -e 's/[[:space:]]addr:/ /' -ne \
             "s|.*inet${_n}[[:space:]]\{1,\}\([^[:space:]/]*\).*|\1|p")" ||
@@ -1733,9 +1733,12 @@ _comp_compgen_available_interfaces()
         if [[ ${1-} == -w ]]; then
             iwconfig
         elif [[ ${1-} == -a ]]; then
-            ifconfig || ip -c=never link show up || ip link show up
+            # Note: we prefer ip (iproute2) to ifconfig (inetutils) since long
+            # interface names will be truncated by ifconfig [1].
+            # [1]: https://github.com/scop/bash-completion/issues/1089
+            ip -c=never link show up || ip link show up || ifconfig
         else
-            ifconfig -a || ip -c=never link show || ip link show
+            ip -c=never link show || ip link show || ifconfig -a
         fi
     } 2>/dev/null | _comp_awk \
         '/^[^ \t]/ { if ($1 ~ /^[0-9]+:/) { print $2 } else { print $1 } }')" &&
--- a/test/fixtures/shared/bin/ip
+++ b/test/fixtures/shared/bin/ip
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+# Dummy "ip addr show" and "ip link show up" emulator
+
+for arg in "$@"; do
+case "$arg" in
+link)
+cat <<EOF
+1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
+    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
+2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default 
+    link/ether 33:33:33:33:33:33 brd ff:ff:ff:ff:ff:ff link-netnsid 0
+EOF
+exit 0
+  ;;
+addr)
+cat <<EOF
+1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
+    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
+    inet 127.0.0.1/8 scope host lo
+       valid_lft forever preferred_lft forever
+    inet6 ::1/128 scope host 
+       valid_lft forever preferred_lft forever
+2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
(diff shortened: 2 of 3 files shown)


bash_completion
```diff
@@ -1620,7 +1620,7 @@ _comp_compgen_mac_addresses()
     # - ip link: link/ether
     _comp_compgen -v addresses split -- "$(
         {
-            LC_ALL=C ifconfig -a || ip -c=never link show || ip link show
+            LC_ALL=C ip -c=never link show || ip link show || ifconfig -a
```
scop, line 1623: This makes LC_ALL=C apply to the `ip` command and removes it from the `ifconfig`. I assume that is not on purpose; if it is, I think it deserves a comment.

scop (commented):
Could we add a comment, preferably in the commit message, about why are we making this change?

bash_completion
```diff
@@ -1618,9 +1618,10 @@ _comp_compgen_mac_addresses()
     # - ifconfig on Linux: HWaddr or ether
     # - ifconfig on FreeBSD: ether
     # - ip link: link/ether
+    # - long interface name will be truncated by ifconfig (#1089), prefer iproute2
```
akinomyoga, line 1621: Usually, the long lines in the code that do not fit in 79 columns should be wrapped.

We haven't included references to the GitHub issues with just a number like \#1089. GitHub is not a unique and permanent platform for this project. It should better be a full link (`https://...`) or at least `GitHub\#1089` or `GitHub: #1089`. Existing cases are

https://github.com/scop/bash-completion/blob/b631277c23fb221fb2761f05ce2eb8457364bb96/bash_completion#L1063-L1065

https://github.com/scop/bash-completion/blob/b631277c23fb221fb2761f05ce2eb8457364bb96/bash_completion#L2562-L2563

https://github.com/scop/bash-completion/blob/b631277c23fb221fb2761f05ce2eb8457364bb96/bash_completion#L3050-L3051

akinomyoga (commented):
The commit message starts with `chore:` (which is used for something outside the code, such as library and CI setups), but I think this is more like `fix:` because this fixes an actual problem #1089.

Also, a test for `wol` is failing.

bash_completion
scop: Also, this comment about interface name truncation is misplaced in the sense that the function is about generating mac addresses and I suppose it has nothing to do with interface names. So perhaps we should not do this change in `_comp_compgen_mac_addresses` at all -- or if we do, some commentary why we are doing it in the commit message would be in order.

bash_completion
```diff
@@ -1694,7 +1695,7 @@ _comp_compgen_ip_addresses()
     local PATH=$PATH:/sbin
     local addrs
     _comp_compgen -v addrs split -- "$({
-        LC_ALL=C ifconfig -a || ip -c=never addr show || ip addr show
+        ip -c=never addr show || ip addr show || LC_ALL=C ifconfig -a
```
scop, line 1698: Similarly as above -- `_comp_compgen_ip_addresses` is not affected by interface name truncation I think, so let's not make this change or add a comment in the commit message why we're doing it.

scop (changes requested):

bash_completion
akinomyoga: I've moved the comment inside `_comp_compgen_available_interfaces`.

bash_completion
akinomyoga: I added reasoning in the commit message.

akinomyoga:
I rebased it and adjusted the code comment and the commit message.

akinomyoga:
The test for `wol` is failing because the test requires `33:33:33:33:33:33`, which is missing with the iproute2 implementation. The `ifconfig` command generates `33:33:33:33:33:33` as an `HWaddr` configured in `test/fixtures/shared/bin/ifconfig` while `ip` doesn't read `HWaddr` in `test/fixtures/shared/bin/ifconfig`.

akinomyoga (approved):
Thank you. I added another commit to solve another issue in the test framework. I think now this is ready.

bash_completion
```diff
@@ -1731,9 +1731,12 @@ _comp_compgen_available_interfaces()
         if [[ ${1-} == -w ]]; then
             iwconfig
         elif [[ ${1-} == -a ]]; then
-            ifconfig || ip -c=never link show up || ip link show up
+            # Note: we prefer ip (iproute2) to inetutils (ifconfig) since long
```
yedayak, line 1734: nit:
```suggestion
            # Note: we prefer ip (iproute2) to ifconfig (inetutils) since long
```

akinomyoga: Ah, right. Thanks

yedayak:
I rebased on origin/main

yedayak (approved):

yedayak:
The issue of truncation seems to only affect the `inetutils` version of ifconfig, but the `net-tools` version doesn't seem to work currently with long interfaces, presumably because it needs some changes in the awk command. I still think this should be merged, but ideally we will make a follow-up PR that fixes ifconfig where it's possible.

akinomyoga:
> ~The issue of truncation seems to only affect the `inetutils` version of ifconfig, but the `net-tools` version doesn't seem to work currently with long interfaces, presumably because it needs some changes in the awk command. I still think this should be merged, but ideally we will make a follow-up PR that fixes ifconfig where it's possible.~ My testing was flawed (I tested with an interface with numbers in the name). The issue only seems to happen to the inetutils version.

Thank you. So does that mean the issue of truncation will be totally solved by this PR? Or does it mean there is another issue with the truncation with `inetutils`? In the latter case, is the part "I still think this should be merged, but ideally we will make a follow-up PR that fixes ifconfig where it's possible." (which was deleted) would be still valid?

yedayak:
Yeah this should totally fix the problem, I think we should merge this. Since the other ifconfig doesn't have an issue there is no need to fix it in a new PR

akinomyoga:
Thank you for your clarification!

scop:
Unfortunately this has fallen through the cracks based on my old "changes requested" review that has not been reset (I look through open PR's from the list view but do not generally open all of them, including those in that state). And because of that, it also missed the 2.15.0 release :/

Sorry about that, merging now.

PR #1101, opened 2024-02-08
fix(gdb): fix regression that fails to generate command names

akinomyoga:
The current version fails to generate any command names for the first word after the command name "gdb".  This is a regression introduced in https://github.com/scop/bash-completion/pull/1086 and pointed out in the comment: https://github.com/scop/bash-completion/commit/73c5292b7a39a2f3d1ae01057e0afd3aa24c8a32#r138406647.

### DIFF
--- a/completions/gdb
+++ b/completions/gdb
@@ -26,7 +26,7 @@ _comp_cmd_gdb()
             # functions and aliases. Thus we need to retrieve the program
             # names manually.
             local path_array
-            _comp_compgen -Rv path_array split -X '' -- "$PATH"
+            _comp_compgen -Rv path_array split -F : -X '' -- "$PATH"
             _comp_compgen_split -o plusdirs -- "$(
                 find ${path_array[@]+"${path_array[@]}"} . -mindepth 1 \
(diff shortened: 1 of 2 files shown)


scop (approved):
Thanks!

akinomyoga:
> but it doesn't seem to work in the `alpine` container. I'll later take a look.

https://github.com/scop/bash-completion/blob/ca8e2405ae49c927f9c9d312aa0efdde3d231bf2/completions/gdb#L30-L34

The `find` command in the `alpine` container doesn't have the operator `-printf`. According to POSIX [1], `-mindepth` is not POSIX, `-maxdepth` is not POSIX, `-not` is not POSIX, `-executable` is not POSIX, and `-printf` is not POSIX. Only `-type` out of the six operators used in this implementation is POSIX-compatible.

- [1] [XCU 4 - find](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/find.html)

It didn't cause an error before because it's never been tested.

akinomyoga:
> I suppose we could use `require_cmd=True` and `gdb` as the test command.

Ah, yeah. That is clever.

PR #1104, opened 2024-02-08
refactor(rtx): rename to mise

scop:
To follow upstream tool rename, https://mise.jdx.dev/rtx.html

### DIFF
--- a/completions/.gitignore
+++ b/completions/.gitignore
@@ -184,6 +184,7 @@
 /_metalctl
 /micropython
 /_minikube
+/_mise
 /mkisofs
 /_mmctl
 /mogrify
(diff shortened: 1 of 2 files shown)


akinomyoga:
When did that change happen? I think we can keep the old name for some period.

The reason for the change seems to be the confusion with the series of graphics cards, but it doesn't seem to be the conflicts on the command names.

scop:
I don't know exactly when it happened, but recently. Why I thought I'd rather rename than keep the old is that the rtx project was fairly new too, and now that it no longer exists by that name, it could be somewhat more likely that another utility by that name appears and our completion loader is likely not doing the right thing with it.

But I have no strong opinion on this, can also leave the old name in if you feel it's better tha way.

akinomyoga:
> that the rtx project was fairly new too,

Ah, I got it. Thank you for explaining the reason. Then, I'm fine with the current version, though the test seems to fail.

akinomyoga (approved):

PR #1107, opened 2024-02-12
feat(ip): Add completion for monitor subcommand

yedayak:
Currently it stops after completing the first option, although you should be able to choose multiple options.
I'll be happy for some pointers on how I can make it keep completing after the first option, but I think this is enough on its own.

### DIFF
--- a/completions/ip
+++ b/completions/ip
@@ -357,7 +357,14 @@ _comp_cmd_ip()
                 all) ;;
                 *)
                     ((cword == subcword)) &&
-                        _comp_compgen -- -W 'help all'
+                        _comp_compgen_split -- "help all $(
+                            {
+                                "$1" -c=never monitor help || "$1" monitor help
+                            } 2>&1 | command sed -e \
+                                '/OBJECTS := /,/[^|]$/!d' -e \
+                                's/OBJECTS := *//' -e \
+                                's/|//g'
+                        )"
                     ;;
             esac
             ;;
--- a/test/t/test_ip.py
+++ b/test/t/test_ip.py
@@ -14,6 +14,15 @@ class TestIp:
     def test_r_r(self, completion):
         assert completion
 
+    @pytest.mark.complete(
+        "ip monitor ",
+        require_cmd=True,
+        skipif="ip monitor help 2>/dev/null; (( $? != 255 ))",
+    )
+    def test_monitor(self, completion):
+        assert "neigh" in completion
+        assert "all" in completion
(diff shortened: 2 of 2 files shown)


completions/ip
```diff
@@ -357,7 +357,14 @@ _comp_cmd_ip()
                 all) ;;
                 *)
                     ((cword == subcword)) &&
-                        _comp_compgen -- -W 'help all'
+                        _comp_compgen_split -- "help all $(
+                            {
+                                $1 -c=never monitor help || $1 monitor help
+                            } 2>&1 | command sed -e \
+                                '/OBJECTS := /,/|.*$/!d' -e \
```
akinomyoga, line 364: I think `/|.*$/` matches any line containing `|`, so this condition fails when the number of lines of OBJECTS becomes three or more. For example, with this mock text:

```bash
$ printf '%s\n' 'OBJECTS := address | link |' '    mroute | neigh |' '    nexthop | stats'
OBJECTS := address | link |
    mroute | neigh |
    nexthop | stats
```

The range only selects the first two:

```bash
$ printf '%s\n' 'OBJECTS := address | link |' '    mroute | neigh |' '    nexthop | stats' | sed -ne '/OBJECTS := /,/|.*$/p'
OBJECTS := address | link |
    mroute | neigh |
$
```

completions/ip
```diff
@@ -357,7 +357,14 @@ _comp_cmd_ip()
                 all) ;;
                 *)
                     ((cword == subcword)) &&
-                        _comp_compgen -- -W 'help all'
+                        _comp_compgen_split -- "help all $(
+                            {
+                                $1 -c=never monitor help || $1 monitor help
```
akinomyoga, line 362:
```suggestion
                                "$1" -c=never monitor help || "$1" monitor help
```

It's better to quote them, though there are still many unquoted cases in the codebase.

akinomyoga (changes requested):
Thank you for your contribution. The change in this PR is generally nice, but I think some minor points should be updated.

completions/ip
yedayak: Fixed

completions/ip
yedayak: Nicely spotted, I changed it to match until a line that don't end with with `|`.

akinomyoga: Thanks!

akinomyoga (approved):

scop:
Thanks!

PR #1108, opened 2024-02-15
fix(ip): Remove non-existent route completion

yedayak:
The ip route usage used to mention monitor as a subcommand, but this was removed in https://github.com/iproute2/iproute2/commit/c0c44bfedda05e52572cc16c9d7777e1077d86cf

Remove it from the completions


scop:
Thanks!

PR #1110, opened 2024-02-19
feat(dot): support filename extension .gv

### LINKED ISSUE #1109: Filename autocompletion for dot program does not accept .gv extension
## Describe the bug

Dot is a program in the graphviz package. Its source files used to use `.dot` as a file extension. However, according to [wikipedia](https://en.wikipedia.org/wiki/DOT_(graph_description_language)), `.gv` is now the preferred file extension.

Bash autocompletion only accepts files with the legacy `.dot` extension. It will not autocomplete files with the `.gv` extension.

## To reproduce

1. Create a blank file called `test.gv`. For example, it can be created by running `: >test.gv`.
2. Type `dot test`.
3. Hit tab. Watch how the filename is not completed.

## Expected behavior

The `test.gv` filename from the example should be expanded when tab is pressed.

- [ ] Distro: Fedora Linux 39
- [ ] bash version: 5.2.26(1)-release
- [ ] bash-completion version: 2.11

## Debug trace

[trace.txt](https://github.com/scop/bash-completion/files/14323798/trace.txt)

akinomyoga:
Fixes #1109

### DIFF
--- a/completions/dot
+++ b/completions/dot
@@ -39,7 +39,7 @@ _comp_cmd_dot()
         return
     fi
 
-    _comp_compgen_filedir dot
+    _comp_compgen_filedir '@(gv|dot)'
 } &&
     complete -F _comp_cmd_dot dot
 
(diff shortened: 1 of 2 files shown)


scop:
Thanks!

PR #1115, opened 2024-02-22
test(perldoc): Test if real perldoc command exists

yedayak:
On some systems the perldoc command exists, but only as a stub that prompts to install the "real" perl-doc pacakge. Skip the test that requires the command if only the stub exists (exits with exit code 1).

### DIFF
--- a/test/t/test_perldoc.py
+++ b/test/t/test_perldoc.py
@@ -9,7 +9,11 @@ class TestPerldoc:
         assert "fixtures/" not in completion  # Our fixtures/ dir
         assert not [x for x in completion if "File::File::" in x]
 
-    @pytest.mark.complete("perldoc -", require_cmd=True)
+    @pytest.mark.complete(
+        "perldoc -",
+        require_cmd=True,
+        skipif="! perldoc -V &>/dev/null",
+    )
     def test_2(self, completion):
         assert completion
 


test/t/test_perldoc.py
```diff
@@ -9,7 +9,11 @@ def test_1(self, completion):
         assert "fixtures/" not in completion  # Our fixtures/ dir
         assert not [x for x in completion if "File::File::" in x]
 
-    @pytest.mark.complete("perldoc -", require_cmd=True)
+    @pytest.mark.complete(
+        "perldoc -",
+        require_cmd=True,
+        skipif="perldoc 2>/dev/null; (( $? == 1 ))",
```
scop, line 15: This feels a bit specific to me, especially given we're making an assumption about a perldoc that is not the real thing and could be basically whatever yet making an assumption that it'll exit with some particular status. I think I'd rather go with:
```suggestion
        skipif="! perldoc -V &>/dev/null",
```

scop (changes requested):

test/t/test_perldoc.py
yedayak: Makes sense, changed to that.

akinomyoga (approved):

scop (approved):

scop:
Thanks!

PR #1117, opened 2024-02-24
fix(cryptsetup): add file completions to header option

aaronkrogers:
An encrypted device's header should be stored in a file. This change enables filename completion for cryptsetup's --header option.

### DIFF
--- a/completions/cryptsetup
+++ b/completions/cryptsetup
@@ -22,7 +22,7 @@ _comp_cmd_cryptsetup()
             --skip | --iter-time | --timeout | --tries | -${noargopts}[chslSbopitT])
             return
             ;;
-        --key-file | --master-key-file | --header-backup-file | -${noargopts}d)
+        --key-file | --master-key-file | --header | --header-backup-file | -${noargopts}d)
             _comp_compgen_filedir
             return
             ;;


completions/cryptsetup
```diff
@@ -22,7 +22,7 @@ _comp_cmd_cryptsetup()
             --skip | --iter-time | --timeout | --tries | -${noargopts}[chslSbopitT])
             return
             ;;
-        --key-file | --master-key-file | --header-backup-file | -${noargopts}d)
+        --key-file | --master-key-file | --header | --header-backup-file | -${noargopts}d)
```
akinomyoga, line 25: `cryptsetup --help` says the option argument of `--header` is "Device or file". I'm not familiar with `cryptsetup`, but doesn't that mean it is better to use `_comp_cmd_cryptsetup__device` for the completion of the option argument of `--header`?

aaronkrogers: Isn't _comp_cmd_cryptsetup__device functionally equivalent to _comp_compgen_filedir defaulting to "/dev/"? Since the detached header might be stored in either a file or a device, using _comp_compgen_filedir means users will not be unexpectedly completing /dev/ when trying to select a file in their current directory.

akinomyoga: > Isn't _comp_cmd_cryptsetup__device functionally equivalent to _comp_compgen_filedir defaulting to "/dev/"?

Yes, it's equivalent. So I thought it could be used for "Device or file".

> using _comp_compgen_filedir means users will not be unexpectedly completing /dev/ when trying to select a file in their current directory.

So, do you intentionally avoid the completion of devices on the empty word? Thank you for the clarification. I was unsure about whether that is the intention, or just you haven't thought about it about it.

In this case, do you think it is useful to complete both devices and filenames?

```bash
_comp_compgen_filedir
[[ /dev/ == "$cur"?* ]] &&
  _comp_compgen -c /dev/ filedir
```

aaronkrogers: Personally, I prefer to type out /dev/\<tab\> when attempting to complete devices. I think it fits better with other bash-completion scripts that begin at pwd unless some other path is specified. `_comp_compgen_filedir` handles this usage fine on its own without the need for `_comp_compgen -c "/dev/" filedir`

```
aaron@sphen keys $ cryptsetup --header 
root.header  root.key     

aaron@sphen keys $ cryptsetup --header /dev/
Display all 241 possibilities? (y or n)

aaron@sphen keys $ cryptsetup --header /dev/nv
nvme0%     nvme0n1#   nvme0n1p1# nvme0n1p2# nvme0n1p3# nvme0n1p4# nvram%
```

I'm honestly not a fan of the _comp_cmd_cryptsetup__device function here since completing from /dev on empty isn't very useful. Trying to complete on do so gives the following:
```
aaron@sphen ~ $ cryptsetup --header
Display all 241 possibilities? (y or n)
```

So we can display all 241 devices or hit no and give it more information to work with. If we really wanted a file, then we have to back out the autocompleted "/dev/" and start over. It doesn't save time over providing "/dev/" on the command line or with more information to begin with, i.e. "/dev/sd\<tab\>".

For the same reason, I'm not for completing both devices and pwd files on empty as that would still give 200+ files to sift through by default which over-complicates things if I'm trying to find completions on a handful of local files.

scop: I think the suggested implementation is ok, let's go with it. Fixed the commit message syntax, merging with that. Thanks!

Issue #1111, opened 2024-02-19
env: Complete command as well


hseg:
## Describe the feature/solution

`env`'s completion currently completes long flags and filenames.
Could we instead have `env` complete like `sudo`, where it first completes
`env`'s flags with their parameters, then some sequence of words containing `=`
(perhaps using `_comp_variable_assignments`?), then a command (using
`_comp_command_offset`)?

## Maintenance (please complete the following information)

- [ ] This is a request for new completion
- [ ] Link to upstream project query about shipping the completion:
Given the presence of completions for `coreutils` binaries in repo, I suppose
this has already been addressed?

scop:
Makes sense, https://github.com/scop/bash-completion/pull/1118

hseg:
Given #1118 has been merged (and a quick check confirms it works pretty much as expected), this can be closed, no?

scop:
Ah yes, I should have used "Closes" instead of "Refs" in the commit message to make that happen automatically. Thanks!

PR #1118, opened 2024-02-26
feat(env): complete commands and variable assignments

scop:
Refs https://github.com/scop/bash-completion/issues/1111

### DIFF
--- a/completions/env
+++ b/completions/env
@@ -1,0 +1,49 @@
+# bash completion for env(1)                              -*- shell-script -*-
+
+_comp_cmd_env()
+{
+    local cur prev words cword was_split comp_args
+    _comp_initialize -s -- "$@" || return
+
+    local i noargopts='!(-*|*[uCS]*)'
+    for ((i = 1; i <= cword; i++)); do
+        if [[ ${words[i]} != -* && ${words[i]} != *=* ]]; then
+            _comp_command_offset $i
+            return
+        fi
+        # shellcheck disable=SC2254
+        [[ ${words[i]} == @(--@(unset|chdir|split-string)|-${noargopts}[uCS]) ]] &&
+            ((i++))
+    done
+
+    # shellcheck disable=SC2254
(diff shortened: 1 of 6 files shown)


scop:
Test coverage is likely something to expand before we merge this.

Would be nice to also address `--*-signal=SIG` too, but that's not a blocker I think.

completions/env
```diff
@@ -0,0 +1,49 @@
+# bash completion for env(1)                              -*- shell-script -*-
+
+_comp_cmd_env()
+{
+    local cur prev words cword was_split comp_args
+    _comp_initialize -s -- "$@" || return
+
+    local i noargopts='!(-*|*[uCS]*)'
+    for ((i = 1; i <= cword; i++)); do
+        if [[ ${words[i]} != -* ]]; then
```
akinomyoga, line 10: In addition to options, I think the the assignments of the form `var=...` should also be skipped. Otherwise, the assignments are treated as the command name in `_comp_command_offset`. Also, the later `_comp_variable_assignments` is ineffective.

akinomyoga (commented):
I think we can add tests.

completions/env
scop: Good point, done now. The later `_comp_variable_assignments` is not entirely ineffective, it will still be helpful with completing known values for variables where supported, e.g. TZ=, LANG=, and the fallback to filedir.

Added one test, but you're right, more should be added. Marking as a draft for that will get to that soonish I hope (but feel free to chime in and push more here in the meantime if you like).

scop: Added some more tests, lifting draft.

scop (commented):
Nice! Good to go as far as I'm concerned (GH won't let me approve this because I'm the original submitter).

akinomyoga (approved):

akinomyoga:
Thanks!

There are actually similar cases for `type -P`, but I think we should consider them in another PR.

PR #1130, opened 2024-03-09
Complete ip link types

yedayak:
Split this into a few commits, adding a function and using it in two places.

### DIFF
--- a/completions/ip
+++ b/completions/ip
@@ -177,11 +177,13 @@ _comp_cmd_ip()
                         _comp_compgen_available_interfaces
                         _comp_compgen -a -- -W 'dev scope to label dynamic
                             permanent tentative deprecated dadfailed temporary
-                            primary secondary up'
+                            primary secondary up type'
                     elif [[ $prev == dev ]]; then
                         _comp_compgen_available_interfaces
                     elif [[ $prev == scope ]]; then
                         _comp_cmd_ip__iproute2_etc rt_scopes
+                    elif [[ $prev == type ]]; then
+                        _comp_cmd_ip__link_types "$1"
                     fi
                     ;;
                 *)
--- a/completions/ip
+++ b/completions/ip
@@ -100,19 +100,12 @@ _comp_cmd_ip()
                     # TODO
                     ;;
                 delete)
-                    case $((cword - subcword)) in
-                        1)
-                            _comp_compgen_available_interfaces
-                            ;;
-                        2)
-                            _comp_compgen -- -W 'type'
-                            ;;
-                        3)
-                            [[ $prev == type ]] &&
-                                _comp_compgen -- -W 'vlan veth vcan dummy ifb
-                                    macvlan can'
-                            ;;
-                    esac
+                    if [[ $prev == type ]]; then
+                        _comp_cmd_ip__link_types "$1"
+                    else
+                        _comp_compgen_available_interfaces
+                        _comp_compgen -a -- -W 'type'
+                    fi
                     ;;
                 set)
                     if ((cword - subcword == 1)); then
--- a/completions/ip
+++ b/completions/ip
@@ -15,6 +15,19 @@ _comp_cmd_ip__netns()
     )"
 }
 
+_comp_cmd_ip__link_types()
+{
+    _comp_compgen_split -- "$(
+        {
+            ${1-ip} -c=never link help || ${1-ip} link help
+        } 2>&1 | command sed -e \
+            '/TYPE := /,/}/!d' -e \
+            's/.*{//' -e \
+            's/}.*//' -e \
+            's/|/ /g'
+    )"
+}
+
 _comp_cmd_ip()
 {
     local cur prev words cword comp_args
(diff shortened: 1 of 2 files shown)


completions/ip
```diff
@@ -15,6 +15,19 @@ _comp_cmd_ip__netns()
...
+_comp_cmd_ip__link_types()
+{
+    _comp_compgen_split -- "$(
+        {
+            ${1-ip} -c=never link help || ${1-ip} link help
+        } 2>&1 | command sed -e \
+            '/TYPE := /,/}/!d' -e \
+            's/.*{//' -e \
+            's/}.*//' -e \
+            's/|//g'
```
akinomyoga, line 27:
```suggestion
            's/|/ /g'
```

Currently it seems to be separated by `<space>|<space>`, but it's safer to assume `|`. Or you could do `s/ | / /g`.

completions/ip
```diff
@@ -87,19 +100,12 @@ _comp_cmd_ip()
...
-                            ;;
-                        3)
-                            [[ $prev == type ]] &&
-                                _comp_compgen -- -W 'vlan veth vcan dummy ifb
-                                    macvlan can'
-                            ;;
-                    esac
+                    _comp_compgen_available_interfaces
+                    if [[ $prev == type ]]; then
+                        _comp_cmd_ip__link_types "$1"
```
akinomyoga, line 105: Could you explain those changes in behavior?

- Why was the check for the word position removed? Hmm, I tried to find a description in `man ip` but cannot find the description on the arguments to `delete`.
- The completions generated by `_comp_compgen_available_interfaces` seem to be overwritten by `_comp_cmd_ip__link_types`, but then why `_comp_compgen_available_interfaces` is called outside the `if` statement? How is it different from calling `_comp_compgen_available_interfaces` inside the `false` branch of the `if` statement?

akinomyoga: Ah, OK. For the first point, I noticed the explanation in the commit message. Thanks.

completions/ip

completions/ip
yedayak: You're correct, I didn't notice it. I moved `_comp_compgen_available_interfaces` to the else branch, since it only needs to be called than.

yedayak:
I'm not sure why the alpine CI fails? I tried running it on latest alpine myself and it seems to work fine

completions/ip
akinomyoga: Thanks for updating!

akinomyoga:
That typically happens, so I recently try to do debugging on the CI.

- [test: [debug] reduce matrix · akinomyoga/bash-completion@1160cee](https://github.com/akinomyoga/bash-completion/actions/runs/8225287017/job/22490130322)

With this change,

```diff
diff --git a/completions/ip b/completions/ip
index ac004be75..384de08b9 100644
--- a/completions/ip
+++ b/completions/ip
@@ -17,6 +17,27 @@ _comp_cmd_ip__netns()

 _comp_cmd_ip__link_types()
 {
+    {
+        echo '==== test1 ===='
+        ${1-ip} -c=never link help
+        echo '==== test2 ===='
+        ${1-ip} link help
+        echo '==== test3 ===='
+        {
+            ${1-ip} -c=never link help || ${1-ip} link help
+        } 2>/dev/null |
+            command sed '/TYPE := /,/}/!d'
+        echo '==== test4 ===='
+        {
+            ${1-ip} -c=never link help || ${1-ip} link help
+        } 2>/dev/null |
+            command sed '/TYPE := /,/}/!d;s/.*{//;s/}.*//'
+        echo '==== test5 ===='
+        {
+            ${1-ip} -c=never link help || ${1-ip} link help
+        } 2>/dev/null |
+            command sed '/TYPE := /,/}/!d;s/.*{//;s/}.*//;s/|/ /g'
+    } >&2
     _comp_compgen_split -- "$(
         {
             ${1-ip} -c=never link help || ${1-ip} link help
```

The result is this:

```
==== test1 ====
BusyBox v1.36.1 (2023-11-07 18:53:09 UTC) multi-call binary.

Usage: ip [OPTIONS] address|route|link|tunnel|neigh|rule [ARGS]

OPTIONS := -f[amily] inet|inet6|link | -o[neline]

ip addr add|del IFADDR dev IFACE | show|flush [dev IFACE] [to PREFIX]
ip route list|flush|add|del|change|append|replace|test ROUTE
ip link set IFACE [up|down] [arp on|off] [multicast on|off]
[promisc on|off] [mtu NUM] [name NAME] [qlen NUM] [address MAC]
[master IFACE | nomaster] [netns PID]
ip tunnel add|change|del|show [NAME]
[mode ipip|gre|sit] [remote ADDR] [local ADDR] [ttl TTL]
ip neigh show|flush [to PREFIX] [dev DEV] [nud STATE]
ip rule [list] | add|del SELECTOR ACTION
==== test2 ====
ip: invalid argument 'help' to 'ip'
==== test3 ====
==== test4 ====
==== test5 ====
```

yedayak:
Thanks @akinomyoga, the issue was `ip link help` doesn't work in busybox ip. I added a skip to check for that and rebased, hopefully should work now

akinomyoga (approved):

scop:
Thanks!

PR #1131, opened 2024-03-10
fix(available_interfaces): fix regression of unwanted trailing colons

### LINKED ISSUE #1129: _comp_compgen_available_interfaces returns trailing colons
## Describe the bug

SSIA.

## To reproduce

```
iftop -i <TAB>
...        lo:               ...
```

## Expected behavior

Available interfaces shown with no trailing colon.

- [x] bash version, `echo "$BASH_VERSION"`: 5.1.16(1)-release
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: current master

## Additional context

Regression in b60353508eb470515f1063a1e1a75bdf1fda730f

akinomyoga:
Fixes #1129

~~I noticed some files are not added in `test/t/unit/Makefile.am`. I added them in the second commit.~~ (**edit**: moved to #1132)

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1737,7 +1737,7 @@ _comp_compgen_available_interfaces()
         fi
     } 2>/dev/null | _comp_awk \
         '/^[^ \t]/ { if ($1 ~ /^[0-9]+:/) { print $2 } else { print $1 } }')" &&
-        _comp_compgen -U generated set "${generated[@]}"
+        _comp_compgen -U generated set "${generated[@]%%[[:punct:]]*}"
 }
 
 # Echo number of CPUs, falling back to 1 on failure.
--- a/test/t/unit/Makefile.am
+++ b/test/t/unit/Makefile.am
@@ -2,6 +2,7 @@ EXTRA_DIST = \
 	test_unit_abspath.py \
 	test_unit_command_offset.py \
 	test_unit_compgen.py \
+	test_unit_compgen_available_interfaces.py \
 	test_unit_compgen_commands.py \
 	test_unit_count_args.py \
 	test_unit_delimited.py \
--- a/test/t/unit/test_unit_compgen_available_interfaces.py
+++ b/test/t/unit/test_unit_compgen_available_interfaces.py
@@ -1,0 +1,25 @@
+import pytest
+
+from conftest import assert_bash_exec
+
+
+@pytest.mark.bashcomp(cmd=None)
+class TestUtilCompgenAvailableInterfaces:
+    @pytest.fixture
+    def functions(self, bash):
+        assert_bash_exec(
+            bash,
+            "_comp__test_dump() { ((${#arr[@]})) && printf '<%s>' \"${arr[@]}\"; echo; }",
+        )
+        assert_bash_exec(
+            bash,
+            '_comp__test_compgen() { local -a arr=(00); _comp_compgen -v arr "$@"; _comp__test_dump; }',
+        )
+
+    def test_1_trailing_colons(self, bash, functions):
+        output = assert_bash_exec(
+            bash,
+            "_comp__test_compgen available_interfaces",
+            want_output=True,
+        )
+        assert ":" not in output.strip()


akinomyoga:
Tests in `test_unit_load_completion.py` fail. Somehow, the symbolic link and the target are reversed in the tar ball.

```bash
$ LANG=C find test/fixtures/_comp_load -name cmd1 -ls
998988373      0 lrwxrwxrwx   1 murase   murase         19 Dec 25 05:23 test/fixtures/_comp_load/bin/cmd1 -> ../prefix1/bin/cmd1
1015156558      4 -rwxr-xr-x   1 murase   murase         10 Dec 25 05:23 test/fixtures/_comp_load/prefix1/bin/cmd1
1048877481      4 -rw-r--r--   1 murase   murase         56 Dec 25 05:23 test/fixtures/_comp_load/prefix1/share/bash-completion/completions/cmd1
1065355955      4 -rw-r--r--   1 murase   murase         57 Dec 25 05:23 test/fixtures/_comp_load/userdir1/completions/cmd1
$ LANG=C tar tvf bash-completion-2.12.0.tar.xz --no-anchored cmd1
-rwxr-xr-x murase/murase    10 2023-12-25 05:23 bash-completion-2.12.0/test/fixtures/_comp_load/bin/cmd1
hrwxr-xr-x murase/murase     0 2023-12-25 05:23 bash-completion-2.12.0/test/fixtures/_comp_load/prefix1/bin/cmd1 link to bash-completion-2.12.0/test/fixtures/_comp_load/bin/cmd1
-rw-r--r-- murase/murase    56 2023-12-25 05:23 bash-completion-2.12.0/test/fixtures/_comp_load/prefix1/share/bash-completion/completions/cmd1
-rw-r--r-- murase/murase    57 2023-12-25 05:23 bash-completion-2.12.0/test/fixtures/_comp_load/userdir1/completions/cmd1
```

Automake seems to specify the option `-h` (resolves symbolic links) to `tar` to create a tar ball.

```bash
tar --format=posix -chf - bash-completion-2.12.0
```

akinomyoga:
I decided to separate a PR for `test_unit_load_completion.py` so removed the related commits from this PR.

bash-completion:
Thanks!

bash_completion
```diff
@@ -1737,7 +1737,7 @@ _comp_compgen_available_interfaces()
         fi
     } 2>/dev/null | _comp_awk \
         '/^[^ \t]/ { if ($1 ~ /^[0-9]+:/) { print $2 } else { print $1 } }')" &&
-        _comp_compgen -U generated set "${generated[@]}"
+        _comp_compgen -U generated set "${generated[@]%%[[:punct:]]*}"
```
bash-completion, line 1740: On a second thought, this is not correct, it will break display of things like `br-something` and `lo:0` (aliases only come out from `ifconfig` that way it seems).

scop, line 1740: On a second thought, this is not correct, it will break display of things like br-something and lo:0 (aliases only come out from ifconfig that way it seems).

akinomyoga: Hmm, thank you for pointing this out. @yedayak I think you can make a proper fix for that. Could you take a look?

yedayak: I just checked some stuff, and it seems you can create interfaces with most punctuation in them, for example `a_b-c` and an interface named `veth2@veth3`. I didn't manage to create interfaces with `:` in them, but as @scop mentioned ifconfig can output them like that. I think the best solution will be to remove the last colon always, and maybe to use `ip -json` when available?

yedayak: Although maybe interfaces with `@` in them are a edge case we shouldn't care about, in which case we should strip everything from there until the end

akinomyoga: Does that mean you suggest something like `"${generated[@]%:}"`?

scop: Re how to create interfaces with colons in them, e.g.
```
ip addr add 127.0.0.42 dev lo label lo:42
ifconfig lo:43 127.0.0.43 up
```

These show up in ifconfig output like
```
lo:42: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.42  netmask 255.255.255.255
        loop  txqueuelen 1000  (Local Loopback)

lo:43: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.43  netmask 255.0.0.0
        loop  txqueuelen 1000  (Local Loopback)
```

They do not show up in `ip link` output for me. They would show up in `ip addr` output, but not in a way we'd currently parse anyway.

There's also
```
ip link set lo alias lo:45
```
...but that does not show up in ifconfig output. It does in ip link output, but again not in a way we'd currently parse.

scop: I would suggest just stripping the trailing colon as well (did also in https://github.com/scop/bash-completion/issues/1133).

JSON output is nice in theory, but opens the can of worms that with what would we parse the JSON with? `jq`, `yq`, ...? Probably quite a bit of code and even more fallbacks involved, not sure if it's worth it.

yedayak: > Does that mean you suggest something like `"${generated[@]%:}"`?

Pretty much, although I'm not sure I understand how it works. only thing is it still completes with `@` if ifconfig isn't available. I think `@` and following characters should be stripped, at least when `ip` is being used. Maybe making a case for that is worth it?

> JSON output is nice in theory, but opens the can of worms that with what would we parse the JSON with? jq, yq, ...? Probably quite a bit of code and even more fallbacks involved, not sure if it's worth it.

Agreed about that.

PR #1134, opened 2024-03-12
fix(ip): Complete more variations of subcommands

yedayak:


completions/ip
```diff
@@ -275,7 +275,7 @@ _comp_cmd_ip()
             esac
             ;;
 
-        neigh)
+        neigh*)
```
akinomyoga, line 278: This is going to match also wrong words, which will produce errors, such as `neighxxxx` or `neighzzzz`. This might be considered harmless, but the other subcommand matching seems to enumerate typical abbreviations:

https://github.com/scop/bash-completion/blob/445e29470b396f789605bf8c7e307371ccd81ff9/completions/ip#L142

It doesn't cover every case such as `ad`, `add`, `addre`, and `addres`, but it seems fine. It seems inconsistent that only `neigh` accepts an arbitrary string starting with `neigh` while the others enumerate typical abbreviations.

If we take this approach, doesn't this need to be `n*` instead of `neigh*`?

One option is to match `n | neigh | neighbor | neighbour)`. Another option is to change the matching of every other subcommand in the form `a*)`, etc. I feel the former is better.

scop (approved):
Thanks!

PR #1136, opened 2024-03-14
feat(curl): Complete protocols for --proto-default

yedayak:


completions/curl
```diff
@@ -127,6 +127,9 @@ _comp_cmd_curl()
             _comp_compgen_known_hosts -- "$cur"
             return
             ;;
+        --proto-default)
+            _comp_compgen -- -W "$("$1" --version 2>/dev/null | command sed -e '/Protocols/!d' -e 's/Protocols://g')"
```
scop, line 131: See https://github.com/scop/bash-completion/blob/222230eabd45ce30b39b65a143b457a38ac1861e/bash_completion#L747-L749

scop (changes requested):

completions/curl
yedayak: Thanks, fixed

PR #1140, opened 2024-03-21
fix(ip): Quote network namespace names

yedayak:
Network namespaces can have spaces in them, which lead to incorrect completions in that case. Fix it by calling _comp_quote on all namespace names.

There are probably better ways to do this than overwriting `$COMPREPLY`, but I couldn't find one.

### DIFF
--- a/completions/ip
+++ b/completions/ip
@@ -8,11 +8,20 @@ _comp_cmd_ip__iproute2_etc()
 
 _comp_cmd_ip__netns()
 {
-    _comp_compgen_split -- "$(
+    local unquoted
+    _comp_split -l unquoted "$(
         {
             ${1-ip} -c=never netns list 2>/dev/null || ${1-ip} netns list
-        } | _comp_awk '{print $1}'
+        } | command sed -e 's/ (.*//'
     )"
+    # namespace names can have spaces, so we quote all of them if needed
+    local ns quoted=()
+    for ns in "${unquoted[@]}"; do
+        local namespace
+        printf -v namespace '%q' "$ns"
+        quoted+=("$namespace")
+    done
+    ((${#quoted[@]})) && _comp_compgen -- -W '"${quoted[@]}"'
 }
 
 _comp_cmd_ip__link_types()


completions/ip
```diff
@@ -8,11 +8,21 @@ _comp_cmd_ip__iproute2_etc()
 
 _comp_cmd_ip__netns()
 {
-    _comp_compgen_split -- "$(
+    _comp_compgen_split -l -- "$(
```
akinomyoga, line 11:
```suggestion
    local unquoted
    _comp_split -l unquoted "$(
```

completions/ip
```diff
@@ -8,11 +8,21 @@ _comp_cmd_ip__iproute2_etc()
...
         {
             ${1-ip} -c=never netns list 2>/dev/null || ${1-ip} netns list
-        } | _comp_awk '{print $1}'
+        } | command sed -e 's/ (.*//g'
     )"
+    # namespace names can have spaces, so we quote all of them if needed
+    local unquoted=("${COMPREPLY[@]}")
+    COMPREPLY=()
```
akinomyoga, line 18:
```suggestion
    local i
```

`local i` is missing.

akinomyoga (commented):
You can first split the string using `_comp_split` and then finally filter them by `_comp_compgen`.

completions/ip
```diff
@@ -8,11 +8,21 @@ _comp_cmd_ip__iproute2_etc()
...
+    for i in "${!unquoted[@]}"; do
+        local REPLY
+        printf -v REPLY %q "${unquoted[i]}"
+        _comp_quote "$REPLY"
+        _comp_compgen -a -- -W "$REPLY"
```
akinomyoga, line 23: When using `compgen -W`, one can (and needs to) quote the words inside the option argument as `'"$REPLY"'` because the option argument to `-W` is going to be processed by shell expansions.

```suggestion
        printf -v REPLY %q "${unquoted[i]}"
        _comp_compgen -a -- -W '"$REPLY"'
```

----

In addition, calling `_comp_compgen` for each completion would be slow. We can first collect the results in another array `quoted` and finally perform `((${#quoted[@]})) && _comp_compgen -- -W '"${quoted[@]}"'`.

completions/ip

completions/ip

yedayak:
I moved to using `_comp_quote_compgen` which seems more appropriate. This also allowed me to remove the %q printf

akinomyoga:
Thank you for updating!

> I moved to using `_comp_quote_compgen` which seems more appropriate.

Do you have an explanation for it? I feel `_comp_quote_compgen` is not the one because it is intended for `cur` (i.e., the string used to filter the generated completions) but not for the generated completions. It does some extra modifications that seem inappropriate to apply to the generated completions.

yedayak:
> Thank you for updating!
> 
> > I moved to using `_comp_quote_compgen` which seems more appropriate.
> 
> Do you have an explanation for it? I feel `_comp_quote_compgen` is not the one because it is intended for `cur` (i.e., the string used to filter the generated completions) but not for the generated completions. It does some extra modifications that seem inappropriate to apply to the generated completions.

From what I tested it seems `_comp_quote` adds single quotes unconditionally, while `_comp_quote_compgen` quotes spaces and such with backslashes. Also, using `_comp_quote` just doesn't seem to actually complete, it completes `'` but that's it (this is with `_comp_quote`): 
```
$ ip netns list
blah bah
rgergerg
```
```
$ ip netns delete # Completes '
```
```
$ ip netns delete '<TAB>
'blah bah'  'rgergerg'
$ ip netns delete 'b<TAB>  # Completes nothing
```

akinomyoga:
> From what I tested it seems `_comp_quote` adds single quotes unconditionally, while `_comp_quote_compgen` quotes spaces and such with backslashes. Also, using `_comp_quote` just doesn't seem to actually complete, it completes `'` but that's it (this is with `_comp_quote`):

What I suggested in https://github.com/scop/bash-completion/pull/1140#discussion_r1535012572 is to use `printf %q` (but not to use `_comp_quote`). Do you have an explanation for the reason that `_comp_quote_compgen` is preferred over `printf %q`? In particular, [this part](https://github.com/scop/bash-completion/blob/4f9ad009487452cb5896cd129a7c87ac4196cb21/bash_completion#L1076-L1078) and [this part](https://github.com/scop/bash-completion/blob/4f9ad009487452cb5896cd129a7c87ac4196cb21/bash_completion#L1082-L1090) are the differences from `printf %q`, but how are they used in the present case?

yedayak:
> > From what I tested it seems `_comp_quote` adds single quotes unconditionally, while `_comp_quote_compgen` quotes spaces and such with backslashes. Also, using `_comp_quote` just doesn't seem to actually complete, it completes `'` but that's it (this is with `_comp_quote`):
> 
> What I suggested in [#1140 (comment)](https://github.com/scop/bash-completion/pull/1140#discussion_r1535012572) is to use `printf %q` (but not `_comp_quote`). Do you have an explanation for the reason that `_comp_quote_compgen` is preferred over `printf %q`? In particular, [this part](https://github.com/scop/bash-completion/blob/4f9ad009487452cb5896cd129a7c87ac4196cb21/bash_completion#L1076-L1078) and [this part](https://github.com/scop/bash-completion/blob/4f9ad009487452cb5896cd129a7c87ac4196cb21/bash_completion#L1082-L1090) are the differences from `printf %q`, but how are they used in the present case?

OK, I didn't get that. Just using `printf %q` seems to work fine, pushed a version with that

completions/ip
```diff
@@ -8,11 +8,21 @@ _comp_cmd_ip__iproute2_etc()
...
     )"
+    # namespace names can have spaces, so we quote all of them if needed
+    local i quoted=()
+    for i in "${!unquoted[@]}"; do
+        local namespace
+        printf -v namespace '%q' "${unquoted[i]}"
+        quoted+=("$namespace")
+    done
+    ((${#quoted[@]})) && _comp_compgen -- -W '"${quoted[@]}"'
+
```
akinomyoga, line 25: extra line?

completions/ip
```diff
@@ -8,11 +8,21 @@ _comp_cmd_ip__iproute2_etc()
 
 _comp_cmd_ip__netns()
 {
-    _comp_compgen_split -- "$(
+    local unquoted
+    _comp_split -l unquoted "$(
         {
             ${1-ip} -c=never netns list 2>/dev/null || ${1-ip} netns list
-        } | _comp_awk '{print $1}'
+        } | command sed -e 's/ (.*//g'
```
akinomyoga, line 15:
```suggestion
        } | command sed -e 's/ (.*//'
```

I think it matches only up to once because  `.*` anyway matches to the end of the line.

completions/ip
```diff
@@ -8,11 +8,21 @@ _comp_cmd_ip__iproute2_etc()
...
     )"
+    # namespace names can have spaces, so we quote all of them if needed
+    local i quoted=()
+    for i in "${!unquoted[@]}"; do
```
akinomyoga, line 19: Should we directly iterate over the values instead of the keys (because we do not use the key value)?

akinomyoga (approved):
There are minor comments. Otherwise, I approve the PR.

completions/ip
yedayak: Removed

completions/ip
yedayak: Yeah, it works without, removed

completions/ip
yedayak: Makes sense, changed it

akinomyoga (approved):

PR #1141, opened 2024-03-24
ip: Various ip net namespace fixes and feats

yedayak:
`ip netns delete` also supports the -all flag, I might add that in a follow up PR

### DIFF
--- a/completions/ip
+++ b/completions/ip
@@ -430,8 +430,26 @@ _comp_cmd_ip()
                 add | identify | list-id)
                     # TODO
                     ;;
-                delete | exec | pids | set)
+                delete | pids | set)
                     [[ $prev == "$subcmd" ]] && _comp_cmd_ip__netns "$1"
+                    ;;
+                exec)
+                    local all_offset=0 i
+                    for ((i = 1; i <= cword; i++)); do
+                        case ${words[i]} in
+                            -a | -all)
+                                all_offset=1
+                                break
+                                ;;
+                        esac
+                    done
+                    if [[ $prev == "$subcmd" && $all_offset != 1 ]]; then
+                        _comp_cmd_ip__netns "$1"
+                    else
+                        local offset
+                        offset="$((subcword + 2 - all_offset))"
+                        _comp_command_offset "$offset"
+                    fi
                     ;;
                 *)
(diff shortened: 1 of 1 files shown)


completions/ip
```diff
@@ -384,12 +384,30 @@ _comp_cmd_ip()
                 add | identify | list-id)
                     # TODO
                     ;;
-                delete | exec | pids | set)
+                delete | pids | set)
                     [[ $prev == "$subcmd" ]] && _comp_cmd_ip__netns "$1"
                     ;;
+                exec)
+                    local is_all=0
```
scop, line 391: The typical way to handle boolean flags like this in the codebase is to initialize to empty (when false is desired as the default), then set it to `set` on true, and check with `if [[ $is_all ]]`. grepping for `has_` yields some examples.

Ah, but reading a bit further, for this case, I see the value is actually being used as an integer, so renaming it to something more descriptive would be in order.

completions/ip
```diff
@@ -384,12 +384,30 @@ _comp_cmd_ip()
...
+                    for ((i = 1; i <= cword; i++)); do
```
scop, line 392: I believe `i` needs to be made `local` to avoid a leak. It is made so further down below, but in a hunk of code where it is not being used.

completions/ip
```diff
@@ -384,12 +384,30 @@ _comp_cmd_ip()
...
+                            -a | -all)
+                                is_all=1
+                                break
+                                ;;
+                        esac
+                    done
+                    if [[ $prev == "$subcmd" && $is_all != 1 ]]; then
+                        _comp_cmd_ip__netns "$1"
+                    else
+                        local offset i
```
scop, line 403: `i` should likely be localized earlier, see above.
```suggestion
                        local offset
```

scop (changes requested):

completions/ip
yedayak: Yeah that's a mistake, fixed

completions/ip
yedayak: Changed the name to all_offset, does that seem fine?

scop: Yes :+1:

scop (approved):
Thanks!

PR #1142, opened 2024-03-24
feat(ip): Complete help for unknown subcommands

yedayak:


completions/ip
```diff
@@ -417,10 +417,14 @@ _comp_cmd_ip()
                     ;;
                 *)
                     ((cword == subcword)) &&
-                        _comp_compgen -- -W 'state policy monitor'
+                        _comp_compgen -- -W 'state policy monitor help'
```
akinomyoga, line 420: The other cases seem to have `help` at the beginning of the list.

completions/ip
```diff
@@ -405,7 +405,7 @@ _comp_cmd_ip()
                     fi
                     ;;
                 *)
-                    _comp_compgen -- -W 'show'
+                    _comp_compgen -- -W 'show help'
```
akinomyoga, line 408: The other cases seem to check `((cword == subword))`, but this branch doesn't check that. Do you think there is a reason? For example, is there a use case of `ip netconf <something>... show` or `ip netconf <something>... help`?

completions/ip
yedayak: Changed it, also in netconf

completions/ip
yedayak: No, that was an oversight. `ip netconf <something>... help` doesn't do anything useful. Fixed it

akinomyoga (approved):
Thanks!

PR #1143, opened 2024-03-25
feat(xmllint,xmlwf): also suggest *.rss files

ittner:
Add suggestion for .rss files to xmllint; this suffix is pretty common for RSS feeds stored offline.

### DIFF
--- a/completions/xmllint
+++ b/completions/xmllint
@@ -46,7 +46,7 @@ _comp_cmd_xmllint()
         return
     fi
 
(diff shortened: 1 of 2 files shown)


scop (commented):
Looks good to me. I think we should also add it for `xmlwf` though, would you mind doing this as part of this PR?

ittner:
Ok! I just amended the commit to add xmlwf too.

Also renamed the issue, but the branch name does still lists xmllint only to allow the forced-push to GH.

scop (approved):

scop:
Thanks!

PR #1144, opened 2024-03-26
fix(ip): make addrlabel flush, list not complete

yedayak:


completions/ip
```diff
@@ -189,13 +189,16 @@ _comp_cmd_ip()
...
             case $subcmd in
-                list | add | del | flush)
+                add | del)
                     if [[ $prev == dev ]]; then
                         _comp_compgen_available_interfaces
+                    elif [[ $prev == prefix || $prev == label ]]; then
+                        : # TODO - Is there a way to complete these?
                     else
-                        : # TODO
+                        _comp_compgen -- -W 'dev prefix label'
```
scop, line 198: This does not quite seem to fit the commit message, it adds some completions to add/del. Could we split this to two commits so that the message of each describes what they do, so they end up in the release notes appropriately?

scop (changes requested):

completions/ip
yedayak: Split into two commits

scop (approved):
Thanks!

PR #1145, opened 2024-03-27
feat(ip): Complete neigh show and flush

yedayak:


completions/ip
```diff
@@ -289,7 +302,20 @@ _comp_cmd_ip()
                     # TODO
                     ;;
                 show | flush)
-                    # TODO
+                    case "$prev" in
+                        nud)
+                            _comp_cmd_ip__neigh_states "$1"
+                            ;;
+                        dev)
+                            _comp_compgen_available_interfaces "$1"
```
scop, line 310: This passes the command being completed to `_comp_compgen_available_interfaces`, but that's not the correct arg it takes. It would be kind of nice if it invoked the same `ip` we are invoking here, but as it stands, that's not available.

```suggestion
                            _comp_compgen_available_interfaces
```

completions/ip
```diff
@@ -289,7 +302,20 @@ _comp_cmd_ip()
...
+                            ;;
+                        proxy | to | vrf) # TODO - Maybe we can complete vrf here?
+                            :
+                            ;;
+                        *)
+                            _comp_compgen -- -W 'proxy to nud vrf dev'
```
scop, line 316: Should we handle `nomaster` here, and also in switch case conditions above?

scop (changes requested):

completions/ip

completions/ip
scop: I think we are not doing it for any other similar cases in `ip` either, so hardcoding it would be fine with me.

yedayak: OK, done. forced push

scop: LGTM, one conflict needs to be resolved, good to go otherwise.

yedayak: O fixed the conflict

scop (approved):
Thanks!

PR #1146, opened 2024-03-30
feat(ip): Complete stats subcommand

yedayak:


completions/ip
```diff
@@ -282,6 +282,41 @@ _comp_cmd_ip()
...
             ;;
+        stats)
+            case "$subcmd" in
+                show)
+                    if [[ $prev == dev ]]; then
+                        _comp_compgen_available_interfaces
+                    elif [[ $prev == group ]]; then
+                        _comp_compgen -- -W "$(
+                            # stats command was added after color, should always have it
+                            $1 -c=never stats help 2>&1 |
```
scop, line 293: Path to the `ip` command being invoked can theoretically contain spaces, so `"$1"` would be in order here. But I see we're failing that already in other places in the file, want to submit a separate followup PR to fix them all?

yedayak: Sure, fixed it here, and will send a PR for the rest

scop (approved):
LGTM, thanks!

One thing I notice is that `ip <tab>` does _not_ offer `stats` as a subcommand, but that's hardly surprising because it's not included in `ip help` output. This is with Ubuntu 23.10.

yedayak:
> LGTM, thanks!
> 
> One thing I notice is that `ip <tab>` does _not_ offer `stats` as a subcommand, but that's hardly surprising because it's not included in `ip help` output. This is with Ubuntu 23.10.

Yeah, I fixed that :) 
https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/commit/?id=32bb7f8f99d27edd66071dbda4a58b3f27c911fd

PR #1152, opened 2024-04-03
feat: ssh-add improvements

scop:


akinomyoga (approved):
I thought `_comp_complete_user_at_host` might better be a new generator, but the current one may be actually the best solution because it relies on `words` by `_comp_initialize -n :` which is different from `words` in the caller's context. I think it is useful to leave a code comment about that.

PR #1159, opened 2024-04-05
fix: use `--` to pass an arbitrary string to builtins

akinomyoga:
These are fixes for the builtins receiving arbitrary strings as strings (but not as options). This generalizes the `type -P` fix in #1118.


scop (approved):
Thanks!

PR #1169, opened 2024-04-18
feat(ip): Complete neigh add, del, change, replace

yedayak:


completions/ip
```diff
@@ -357,7 +357,23 @@ _comp_cmd_ip()
...
+                        lladdr)
+                            _comp_compgen_mac_addresses
+                            ;;
+                        nud)
+                            _comp_cmd_ip__neigh_states "$1"
+                            ;;
+                        dev)
+                            _comp_compgen_available_interfaces
+                            ;;
+                        proxy)
```
scop, line 370: `protocol` takes an arg, should we try and complete it or explicitly "non-complete" it?

scop (changes requested):

completions/ip
yedayak: Made it complete using `/etc/iproute2/rt_protos`

scop (approved):
Thanks!

PR #1170, opened 2024-04-18
feat(ip): Complete 'route get' options

yedayak:


completions/ip
```diff
@@ -249,7 +249,21 @@ _comp_cmd_ip()
...
+                            ;;
+                        uid)
+                            _comp_compgen_uids
+                            ;;
+                        oif | iif)
+                            _comp_compgen_available_interfaces -a
+                            ;;
+                        *)
+                            # TODO: 'iif' only works if also 'from' is specified
+                            _comp_compgen -- -W 'dport from fibmatch iif ipproto mark oif sport to tos uid vrf'
```
scop, line 264: `to` seems to be missing from the help message, but I see it is being handled at https://github.com/iproute2/iproute2/blob/89210b9ec1c445ae963d181b5816d12a0cdafbb6/ip/iproute.c#L2077

That options parsing hunk of code at the above link has some other args not included in the help output or here, seemingly at least `notify`, `connected`, and `as`.

`dev` we could support too, but I suggest not including it in this list here but just in the `oif | iif | dev` case above.

scop (changes requested):

completions/ip
yedayak: Oh yeah I didn't see thos options. added, also the `dev` you talked about.

scop: `as` is still missing, was it left out on purpose?

yedayak: > `as` is still missing, was it left out on purpose?

fixed

PR #1172, opened 2024-04-24
fix(ip): Complete link change as well as set

yedayak:
This was missing from the help and man page, but worked for a while.


scop (approved):
Because upstream has just made it available in the docs as well in https://github.com/iproute2/iproute2/commit/11543416d9bcdf58c385c635295b074dcdb6bc5a, let's add it here too. If they hadn't, I feel we should not have included it in our suggestions either, but only the `set | change` part.

Thanks!

PR #1173, opened 2024-04-27
feat(ip): Complete addr add/change/replace options

yedayak:


completions/ip
```diff
@@ -186,7 +186,7 @@ _comp_cmd_ip()
                     elif [[ $prev == scope ]]; then
                         _comp_cmd_ip__iproute2_etc rt_scopes
                     else
-                        : # TODO
+                        _comp_compgen -- -W 'dev scope home nodad mngtmpaddr noprefixroute autojoin'
```
scop, line 189: Should we include all the words in `IFADDR`  as well? Ref https://github.com/iproute2/iproute2/blob/89210b9ec1c445ae963d181b5816d12a0cdafbb6/ip/ipaddress.c#L63-L66

scop (changes requested):

completions/ip
yedayak: Yeah we should, will do that soon

yedayak: Added them and pushed, should be ready

completions/ip
```diff
@@ -181,13 +181,24 @@ _comp_cmd_ip()
...
+                            ;;
+                        broadcast | anycast | peer | metric)
+                            :
+                            ;;
+                        valid_lft | preferred_lft)
+                            _comp_compgen -- -W 'forever'
+                            ;;
+                        *)
+                            _comp_compgen -- -W 'anycast broadcast dev home label scope nodad metric mngtmpaddr
+			        noprefixroute valid_lft peer preferred_lft'
```
scop, line 199: Ubuntu 24.04 has `autojoin` in `CONFFLAG`

scop (changes requested):
LGTM, just one small remark

completions/ip
yedayak: Added

scop (approved):
Thanks!

PR #1184, opened 2024-05-01
feat(ssh-keygen): -O arg improvements

scop:


completions/ssh-keygen
```diff
@@ -60,17 +60,39 @@ _comp_cmd_ssh_keygen()
...
-                    permit-X11-forwarding no-touch-required source-address=
-                    verify-required
-
-                    lines= start-line= checkpoint= memory= start= generator=
-
-                    application= challenge= device= no-touch-required resident
-                    user= write-attestation='
+                local -a opts=()
+                case ${words[*]} in
+                    *\ -*M\ *)
```
akinomyoga, line 65: I think we want to exclude a space in the second `*`, i.e., `!(* *)`. But then, maybe we want to introduce `noargopts` with a modification to exclude a space. For example,

```bash
local noargopts='!(-*|*[ aCIJjMNPSVWzbEFRDwfGKsTmnO]*)'
```

completions/ssh-keygen
```diff
@@ -60,17 +60,39 @@ _comp_cmd_ssh_keygen()
...
-                    permit-port-forwarding permit-pty permit-user-rc
-                    permit-X11-forwarding no-touch-required source-address=
-                    verify-required
-
-                    lines= start-line= checkpoint= memory= start= generator=
-
-                    application= challenge= device= no-touch-required resident
-                    user= write-attestation='
+                local -a opts=()
+                case ${words[*]} in
```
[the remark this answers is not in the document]
scop: Applied initially as is, but then again we have quite a few `${words[*]}` elsewhere, so pulled up in 472856b9a7a52e891cc7a76a6a3e36ef07f4dec8

completions/ssh-keygen
akinomyoga: Thank you!

akinomyoga (approved):

PR #1199, opened 2024-05-22
Ignore comments in ssh_config Include parsing

remil1000:
When parsing _Include_ directive in _ssh_config_ files, comments on the same line as the directive are considered as files to be recursively processed. Best case the recursive `max_depth=16` terminates this early, but on old versions it may lead to random reads or an infinite recursion, hanging the shell process

Example before/after:

A minimal ssh client config file
```
$ cat config 
Include foo bar .ssh/baz # do not include
```

The current sed expression
```
$ sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' config 
foo bar .ssh/baz # do not include
```

Proposed change to sed expression (ignore anything coming after a `#`)
```
$ sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\([^#]*\).*$/\1/p' config
foo bar .ssh/baz 
```


bash_completion
scop, line 1: Let's amend our test suite to cover for this case, see test/t/unit/test_unit_known_hosts.py

scop (changes requested):
Thanks!

In addition to the requested test suite update, please search for "conventional commits" in CONTRIBUTING.md for info how to format the commit message and modify it accordingly (+ force-push with it).

bash_completion
remil1000: - test suite amended (hopefully I understood it well)
- commit message updated also

also I realized we may be in murky waters at this inline/trailing comment support has only been gradually added to sshd_config then ssh_config around SSH version 8.5 (sshd_config - https://bugzilla.mindrot.org/attachment.cgi?id=3460 and ssh_config https://bugzilla.mindrot.org/show_bug.cgi?id=2320)

So for example code
```
_comp_split hosts "$(command sed -ne 's/^[[:blank:]]*[Hh][Oo][Ss][Tt][[:blank:]=]\{1,\}\(.*\)$/\1/p' "${config[@]}")"; then
```
in block https://github.com/scop/bash-completion/blob/main/bash_completion#L2646-L2652 may need changes also to handle those inline comments

test/t/unit/test_unit_known_hosts.py
```diff
@@ -127,6 +126,8 @@ def test_included_configs(self, bash, hosts):
         expected.extend("asterisk_1 asterisk_2".split())
         # fixtures/_known_hosts/.ssh/config_question_mark
         expected.append("question_mark")
+        # fixtures/_known_hosts/.ssh/config_comment
+        expected.append("with_comment")
```
akinomyoga, line 130: The test seems to require the hostname found by the `Include` directive with inline comments, but the test doesn't seem to require that the words in the comment are excluded.

test/t/unit/test_unit_known_hosts.py
```diff
@@ -1,7 +1,6 @@
 from itertools import chain
 
 import pytest
-
```
akinomyoga, line 4: Is this change intentional?

bash_completion
```diff
@@ -2447,7 +2447,8 @@ _comp__included_ssh_config_files()
     # https://github.com/openssh/openssh-portable/blob/5ec5504f1d328d5bfa64280cd617c3efec4f78f3/readconf.c#L2240
     local max_depth=16
     while ((${#included[@]} > 0 && depth++ < max_depth)); do
-        _comp_split include_files "$(command sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]\(.*\)$/\1/p' "${included[@]}")" || return
+        _comp_split include_files "$(command sed -n -e 's/[[:blank:]]*#.*//' -e 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]*\(..*\)/\1/p' "${included[@]}")" || return
```
akinomyoga, line 2450: Should we remove the trailing/preceding whitespaces? Those are stripped by `_comp_split` anyway. Then, the sed expression can be simplified:

```suggestion
        _comp_split include_files "$(command sed -ne 's/^[[:blank:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:blank:]]*\([^#]*\).*$/\1/p' "${included[@]}")" || return
```

akinomyoga (commented):
Thanks for updating.

akinomyoga:
> So for example code
> 
> ```
> _comp_split hosts "$(command sed -ne 's/^[[:blank:]]*[Hh][Oo][Ss][Tt][[:blank:]=]\{1,\}\(.*\)$/\1/p' "${config[@]}")"; then
> ```
> 
> in block https://github.com/scop/bash-completion/blob/main/bash_completion#L2646-L2652 may need changes also to handle those inline comments

I agree. We want to apply the changes consistently for similar cases.

test/t/unit/test_unit_known_hosts.py
remil1000: the original fix is to discard inline comments after an `Include` directive

if the included file would contain `Host with_comment # to be ignored` then this patch would also require to update the `Host` directive parsing sed expression (which I already tried) only to find out it breaks https://github.com/scop/bash-completion/blob/main/test/fixtures/_known_hosts/config#L9 `Host = hus%%eth0 !negated  #not-a-comment` as until now fixtures considered inline comments as actual content to be parsed

This is precisely why I only patched the `Include` parsing expression and mentioned this future `Host` issue, because tests would fail one after the other

test/t/unit/test_unit_known_hosts.py
remil1000: sorry, this one I missed while git diff'ing - fixed

bash_completion
remil1000: this is mainly because I discovered OpenSSH maintainers, in their parser, are actually proceeding like this, simply stopping processing as soon as they hit a `#` sign and trimming
```
	/* Remove comment and trailing whitespace */
	if ((cp = strchr(orig, '#')) != NULL)
		*cp = '\0';
	rtrim(orig);
```
https://github.com/openbsd/src/blob/master/usr.bin/ssh/readconf.c#L3229-L3232

I can revert to your proposition, which, incidentally is very similar to my first proposal in the PR

akinomyoga: Does the result change with this implementation detail?

test/t/unit/test_unit_known_hosts.py
akinomyoga: I think the fixture for `Host` should also be updated.

----

By the way, I started this "conversation" (in the GitHub term) to discuss what should be tested. You've modified the test, but it doesn't seem to test the expectation by the original fix for `Include`. Could you test that a word that matches a filename in the comment is excluded? You can add another file in the fixture and put a hostname that should not be collected there. Then, you can include the filename in the inline comment on a `Include` line.

bash_completion
akinomyoga: I didn't get the response, but I applied the suggestion since that is the minimal change to the current main, and it is also simpler.

test/t/unit/test_unit_known_hosts.py
akinomyoga: I've again checked this conversation. In https://github.com/scop/bash-completion/pull/1199#discussion_r1625448295, I requested to add a regression test for the main purpose of this PR; i.e., we should test that "inside_comment" of `Include with_comment # inside_comment` is excluded from include targets.

However, the reply https://github.com/scop/bash-completion/pull/1199#discussion_r1626233739 talked about something completely different; it discussed a similar fix for Host to exclude the inline comments. To this, I replied (in the former half of https://github.com/scop/bash-completion/pull/1199#discussion_r1626965206) that `Host` should also be updated similarly, but I now think it doesn't need to be done within this PR.

In the latter half of https://github.com/scop/bash-completion/pull/1199#discussion_r1626965206, I pointed out that the test is not updated to include the regression test, but there is no further action from the original author of the PR.

akinomyoga: 7aa420a2871cd17436fefd9fa8b468c7235bee47 for a regression test.

akinomyoga:
I rebased it.

bash_completion
akinomyoga: I think the original purpose of this conversation is already satisfied as a test has already been updated. For the possible change to the parsing of the `Host` directive, we can process that in a separate PR.

akinomyoga:
I squashed commits.

A similar change for the Host directive discussed in the following places haven't been processed, but I think they can be processed in a separate PR:

> https://github.com/scop/bash-completion/pull/1199#discussion_r1625438365 by @remil1000 
>
> [...] also I realized we may be in murky waters at this inline/trailing comment support has only been gradually added to sshd_config then ssh_config around SSH version 8.5 (sshd_config - https://bugzilla.mindrot.org/attachment.cgi?id=3460 and ssh_config https://bugzilla.mindrot.org/show_bug.cgi?id=2320)
> 
> So for example code
> 
> ```
> _comp_split hosts "$(command sed -ne 's/^[[:blank:]]*[Hh][Oo][Ss][Tt][[:blank:]=]\{1,\}\(.*\)$/\1/p' "${config[@]}")"; then
> ```
> 
> in block https://github.com/scop/bash-completion/blob/main/bash_completion#L2646-L2652 may need changes also to handle those inline comments

> https://github.com/scop/bash-completion/pull/1199#discussion_r1626233739 by @remil1000 
>
> [...] if the included file would contain `Host with_comment # to be ignored` then this patch would also require to update the `Host` directive parsing sed expression (which I already tried) only to find out it breaks https://github.com/scop/bash-completion/blob/main/test/fixtures/_known_hosts/config#L9 `Host = hus%%eth0 !negated #not-a-comment` as until now fixtures considered inline comments as actual content to be parsed
> 
> This is precisely why I only patched the `Include` parsing expression and mentioned this future `Host` issue, because tests would fail one after the other

PR #1200, opened 2024-05-23
fix(cargo): avoid local: can only be used in a function warning

nekopsykose:
when loading bash-completion currently, the following is printed

```
$ bash -l
bash: local: can only be used in a function
bash-5.2$ 
```

tracing this with `bash -x` points to the _cargo completion. it seems that this is not actually sourced by _comp_load and isn't in a function

elide this by inlining the variable use

### DIFF
--- a/completions/_cargo
+++ b/completions/_cargo
@@ -2,8 +2,7 @@
 #
 # This serves as a fallback in case the completion is not installed otherwise.
 
-# shellcheck disable=SC2168 # "local" is ok, assume sourced by _comp_load
-local rustup="${1%cargo}rustup" # use rustup from same dir
-eval -- "$("$rustup" completions bash cargo 2>/dev/null)"
+# use rustup from same dir
+eval -- "$("${1%cargo}rustup" completions bash cargo 2>/dev/null)"
 
 # ex: filetype=sh


akinomyoga:
> tracing this with `bash -x` points to the _cargo completion. it seems that this is not actually sourced by _comp_load

Can you tell the reason why it is not sourced by `_comp_load`? For example, what is the corresponding line in the output of the following command?

```bash
PS4='+[${FUNCNAME[*]}][${BASH_SOURCE[*]}]' bash -xl
```

nekopsykose:
it was actually a configuration error- the _cargo ended up in `compatdir`, where it gets loaded differently. if you move `_cargo` to /etc/bash_completion.d (default compatdir) where it's eagerly loaded you can reproduce. without being in compat it loads fine and this doesn't happen

..so i think this isn't actually an issue. but this is the only thing that has this error, and all the other completions can be (eagerly) loaded without warnings. do you think it's still worth changing?

akinomyoga:
> and all the other completions can be (eagerly) loaded without warnings

Even if the error message would not be shown, I think there are errors. For example, [`_golangci-lint`](https://github.com/scop/bash-completion/blob/main/completions/_golangci-lint) uses `"$1"` to call the command, which is only set by `_comp_load`. If it is directly loaded through `bash_completion.d`, it internally calls `"" completion bash`. This is supposed to fail, but the error message is suppressed because of the redirection `2>/dev/null`. This behavior is common for all our "3rd party completion loaders".

> do you think it's still worth changing?

I don't think so.

nekopsykose:
> I don't think so.  

sure thing :)

Issue #479, opened 2020-12-29
ifup, ifdown, _ip_addresses: check NETWORK=none instead/in addition to in_container?


scop:
>     * `TestIfDown`, `TestIfup` - it's using `_configured_interfaces` (https://github.com/scop/bash-completion/blob/fa21b5cecc0ef5885f9eb41abbefcee2b59b258c/bash_completion#L1013) and gets nothing because NixOS is not supported by this function. NixOS doesn't provide `ifup` or `ifdown` commands.

For these we currently check if we're running in a container and xfail if so, but I suppose we should additionally make them do so with `NETWORK=none`.

_Originally posted by @scop in https://github.com/NixOS/nixpkgs/issues/107768#issuecomment-752013329_

Ditto `_ip_addresses`

PR #1208, opened 2024-06-03
test(ifdown,iftest): Skip if we don't have network

yedayak:
Without network we likely don't have any interfaces configured either

Fixes: https://github.com/scop/bash-completion/issues/479


akinomyoga (approved):

scop:
Thanks!

Issue #453, opened 2020-10-16
OSTYPE check too strict in `watch`


celynwalters:
This is on the same lines as #122

I found that for [`watch`](https://github.com/scop/bash-completion/blob/master/completions/watch#L3) completions are not used on Mac OSX because of a check to the `OSTYPE` environment variable:

```bash
[[ $OSTYPE == *linux* ]] || return 1
```

Yes, `watch` isn't normally available on OSX, but it can be simply installed via `brew install watch`.
I couldn't find any others like this by doing a recursive `grep`.
Perhaps the line should be more like [`killall`](https://github.com/scop/bash-completion/blob/master/completions/killall#L3)?

```bash
[[ $OSTYPE == *@(linux|freebsd|darwin)* ]] || return 1
```

I don't know for which other OSes it can extend to, though.
On OSX my `OSTYPE` variable is `darwin18.7.0`.
I have this repo as a submodule in my dotfiles, so I'm always up to date with master.

scop:
The point of exclusions like the current case at hand is not whether the command in question is usually available or not, but rather that the command by that name is usually something completely different than what the completion is written for. For example in the case of `watch`, compare https://linux.die.net/man/1/watch and https://www.unix.com/man-page/FreeBSD/8/watch/ -- so enabling this completion for watch on FreeBSD is not something we should do.

This is of course not a very good general pattern -- better would be to somehow detect whether the command we're completing is the expected one. There are some examples of this in the tree.

Now, I don't know what is the expected `watch` command on macs, but if you can confirm that there's no command called `watch` installed on macs by default, I think that alone would be enough to enable this completion for the `darwin`s.

Then again, perhaps we should look into some detection here instead. One possible (but still rather feeble) example would be to invoke `watch --help &>/dev/null` and if it doesn't indicate an error, install the completion, otherwise not. The `watch` on e.g. Ubuntu doesn't error out, but the FreeBSD does. How does yours behave?

celynwalters:
Wow, what a detailed answer!

> The point of exclusions like the current case at hand is not whether the command in question is usually available or not, but rather that the command by that name is usually something completely different than what the completion is written for. For example in the case of `watch`, compare https://linux.die.net/man/1/watch and https://www.unix.com/man-page/FreeBSD/8/watch/ -- so enabling this completion for watch on FreeBSD is not something we should do.

Ah, I was just using it as an example. I don't know anything about FreeBSD.

[...]

PR #1210, opened 2024-06-06
pkgutil: don't install on platforms where it cannot be used

eli-schwartz:
In commit 0e3a17df7eac79ff386891945a951a23134373b0, a short-circuit was added that causes the completion to return 1 if the host OS is not solaris.

On Gentoo, this triggers a warning:

```
 * QA Notice: Problems with installed bash completions were found:
 *
 * 	pkgutil: does not define any completions (failed to source?).
 *
 * For more details on installing bash-completions, please see:
 * https://wiki.gentoo.org/wiki/Bash/Installing_completion_files
```

The purpose of the check is to find broken completions that need to be fixed by their respective upstreams, and in this case the completion is "intentionally broken". It seems straightforward to respect that at an earlier level, which both avoids the QA check and ensures that unneeded files don't get installed on any distro.

And... it also means that macOS users can ship their own pkgutil completion without worrying about file clashes with the bash-completion project. The entire reason for the short-circuit was because there's a *different* pkgutil there.

Fixes: 0e3a17df7eac79ff386891945a951a23134373b0

### DIFF
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -120,7 +120,6 @@ bashcomp_DATA = 2to3 \
 		firefox \
 		flake8 \
 		_flamegraph \
-		freebsd-update \
 		freeciv \
 		freeciv-server \
 		function \
@@ -215,8 +214,6 @@ bashcomp_DATA = 2to3 \
 		_keyring \
 		kill \
 		killall \
-		kldload \
-		kldunload \
 		koji \
 		_kontena \
 		ktutil \
@@ -324,7 +321,6 @@ bashcomp_DATA = 2to3 \
 		_pipenv \
 		pkg-config \
 		pkg-get \
-		pkg_delete \
 		pkgadd \
 		pkgrm \
 		pkgtool \
@@ -333,9 +329,6 @@ bashcomp_DATA = 2to3 \
 		pm-is-supported \
 		pm-powersave \
 		pngfix \
-		portinstall \
-		portsnap \
-		portupgrade \
 		postcat \
 		postconf \
 		postfix \
@@ -506,6 +499,21 @@ bashcomp_DATA = 2to3 \
 if SOLARIS
 bashcomp_DATA += pkgutil
 endif
+
+if BSD
+bashcomp_DATA += pkg_delete
+endif
+
+if FREEBSD
+bashcomp_DATA += \
+		freebsd-update \
+		kldload \
+		kldunload \
+		portinstall \
+		portsnap \
+		portupgrade
+endif
+
 
 EXTRA_DIST = $(bashcomp_DATA)
 
@@ -770,8 +778,6 @@ CLEANFILES = \
 	ping4 \
 	ping6 \
 	_pip3 \
-	pkg_deinstall \
-	pkg_info \
 	pkgconf \
 	pkill \
 	plzip \
@@ -1159,8 +1165,10 @@ symlinks: $(DATA)
 		_black _blackd _flask _httpx
 	$(ss) pkg-config \
 		pkgconf
+if BSD
 	$(ss) pkg_delete \
 		pkg_deinstall pkg_info
+endif
 	$(ss) pgrep \
 		pkill
 	$(ss) pm-hibernate \
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -328,7 +328,6 @@ bashcomp_DATA = 2to3 \
 		pkgadd \
 		pkgrm \
 		pkgtool \
-		pkgutil \
 		plague-client \
 		pm-hibernate \
 		pm-is-supported \
@@ -502,6 +501,11 @@ bashcomp_DATA = 2to3 \
 		yum-arch \
 		zopfli \
 		zopflipng
+
+
+if SOLARIS
+bashcomp_DATA += pkgutil
+endif
 
 EXTRA_DIST = $(bashcomp_DATA)
 
--- a/configure.ac
+++ b/configure.ac
@@ -14,12 +14,22 @@ fi
 
 AC_CANONICAL_HOST
 
+build_bsd=
+build_freebsd=
 build_solaris=
 case ${host_os} in
+    *bsd*)
+        build_bsd=yes
+        case ${host_os} in
+            *freebsd*)
+                build_freebsd=yes ;;
+        esac ;;
     *solaris*)
         build_solaris=yes ;;
 esac
 
+AM_CONDITIONAL([BSD], [test "${build_bsd}" = yes])
+AM_CONDITIONAL([FREEBSD], [test "${build_freebsd}" = yes])
 AM_CONDITIONAL([SOLARIS], [test "${build_solaris}" = yes])
 
 AC_CONFIG_FILES([
--- a/configure.ac
+++ b/configure.ac
@@ -11,6 +11,17 @@ AC_ARG_WITH([pytest],[  --with-pytest=executable],[PYTEST="$withval"])
 if test "x$PYTEST" = x; then
     AC_CHECK_PROGS([PYTEST],[pytest pytest-3],[pytest])
 fi
+
+AC_CANONICAL_HOST
+
+build_solaris=
+case ${host_os} in
+    *solaris*)
+        build_solaris=yes ;;
+esac
+
+AM_CONDITIONAL([SOLARIS], [test "${build_solaris}" = yes])
+
 AC_CONFIG_FILES([
 Makefile
 completions/Makefile


configure.ac
```diff
@@ -11,6 +11,14 @@ AC_ARG_WITH([pytest],[  --with-pytest=executable],[PYTEST="$withval"])
 if test "x$PYTEST" = x; then
     AC_CHECK_PROGS([PYTEST],[pytest pytest-3],[pytest])
 fi
+
+case ${host_os} in
```
akinomyoga, line 15: AC_CANONICAL_HOST?

configure.ac
```diff
@@ -11,6 +11,14 @@ AC_ARG_WITH([pytest],[  --with-pytest=executable],[PYTEST="$withval"])
...
+    *solaris*)
+        build_solaris=yes ;;
+esac
+
+AM_CONDITIONAL([SOLARIS], [test "${build_solaris}" = yes])
```
akinomyoga, line 20:
```suggestion
AM_CONDITIONAL([SOLARIS], [test "x${build_solaris}" = xyes])
```

The environment might have the environment variable `build_solaris` starting with `-`.

eli-schwartz: This was barely relevant at all, since requiring a leading x only affects shells that aren't POSIX conformant, but AFAIK the last bug in a valid /bin/sh was fixed in 2008 and the last bug in an invalid /bin/sh (which is to say, zsh) was fixed in 2015.

akinomyoga: The configure script (autoconf) traditionally tried to cover non-POSIX conformant shells as well. Prefixing `x` seems consistent with the autoconf convention. It might not be relevant in practical cases now, but is there a reason to intentionally drop non-POSIX shells, which autoconf has been supported? This is a small fix. Is there a reason to stick with the style which is inconsistent with the autoconf convention?

configure.ac
akinomyoga: Thanks

configure.ac
akinomyoga: I mean I wouldn't try to argue that this would become a practical issue, but I should probably say it would be more like a style issue.

eli-schwartz: I fixed it by explicitly setting it equal to the blank string before the case statement. :D Environment variables are now entirely ignored.

akinomyoga: Thanks. That works.

If it were me, I'd still follow the `x` style in the Autoconf configurations even when the variable value is ensured to be safe, but that should be a preference.

> which means you cannot confuse configure by running `build_solaris=yes ./configure`

Yeah, I assumed you intentionally left the possibility of that usage.

eli-schwartz: https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Multiple-Cases.html

> Here is an example that uses the shell variable `fstype` to keep track of whether the remaining cases need to be checked. Note that since the value of `fstype` is under our control, we don’t have to use the longer ‘`test "x$fstype" = xno`’.

Seems like a suggestion that it's something beneficial to avoid... (for readability reasons it feels better to avoid IMHO)

eli-schwartz: > Yeah, I assumed you intentionally left the possibility of that usage.

In general one would do this via `./configure --host=XXX`.

akinomyoga: > https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Multiple-Cases.html
> 
> > Here is an example that uses the shell variable `fstype` to keep track of whether the remaining cases need to be checked. Note that since the value of `fstype` is under our control, we don’t have to use the longer ‘`test "x$fstype" = xno`’.
> 
> Seems like a suggestion that it's something beneficial to avoid...

Oh, I see. I didn't know that recommendation. Thank you.

> > Yeah, I assumed you intentionally left the possibility of that usage.
> 
> In general one would do this via `./configure --host=XXX`.

Yes.

completions/Makefile.am
```diff
@@ -503,6 +502,11 @@ bashcomp_DATA = 2to3 \
 		zopfli \
 		zopflipng
 
+
+if SOLARIS
+bashcomp_DATA += pkgutil
+endif
```
scop, line 508: Looks like the following utilities would benefit from a similar treatment: freebsd-update, killall, kldload, kldunload, portsnap, route, and watch.

We don't necessarily have to address them here, but I would not object to doing so at all. But if not, we should make a followup PR for them once this is in.

Aside, I don't know if we should make the SOLARIS etc defines more specific to avoid clashes with some possibly existing ones or if it's no concern at all, e.g. BASHCOMP_OS_SOLARIS, BASHCOMP_OS_LINUX, etc.

scop (commented):
Looks good to me on surface, just one remark and will defer to @akinomyoga to comment and approve if you have further things to note.

completions/Makefile.am
eli-schwartz: The defines should be fine -- AM_CONDITIONAL defines live in their own namespace and this is generated as SOLARIS_TRUE / SOLARIS_FALSE in ac_subst_vars. In general, you are supposed to be allowed to choose your own arbitrary project-specific names here.

Good point on the other freebsd programs: https://gitweb.gentoo.org/repo/gentoo.git/tree/app-shells/bash-completion/bash-completion-2.14.0.ebuild?id=8aa148634191c504242b4c4ab82f2ef622aede22

Gentoo is already opportunistically deleting a bunch of freebsd stuff for similar reasons.

akinomyoga: - Can we also switch the files for `completions/_umount{,.linux}` and `completions/_mount{,.linux}` on the installation stage?
- Also, `route`, `watch`, and `killall` have similar `OSTYPE` checks.
- `ifup` seems to depend on a condition based on the output of `uname -s`.
- `_modules` also seem to depend on the system arrangement, but I'm not sure if we should skip the installation because the condition can be dynamic. The required directory might be added later.

akinomyoga:
Would it be possible to have a configure option to generate a host-independent version, which includes all completions and is ready to use in any operating system just by sourcing `bash_completion`<sup>*</sup>? It doesn't need to be done within this PR, but I think it should be useful to leave a way to generate the installation that includes everything as we've been doing until now.

*: I think `/etc/profile.d/bash_completion.sh` wouldn't work because it contains a hardcoded path to the install location, but still we can directly source the file `bash_completion`.

completions/Makefile.am
eli-schwartz: "can" we is a good question.

> We don't necessarily have to address them here, but I would not object to doing so at all. But if not, we should make a followup PR for them once this is in.

I already dug deeper and handled everything that was simple, clean and obvious to me. Some of those files, I'm not at all sure (yet) how they should be handled, and the rest have OS conditionals which I don't quite understand but feel a bit wrong (e.g. there are comments on other tickets about how the checks are being used as a shortcut for "probably not the same command with the same options").

The current changes I made, I'm highly confident they are always correct, so I would rather merge them as is and follow up on the rest.

eli-schwartz:
You mean it includes all completions specific to FreeBSD, in addition to all completions specific to Solaris, in addition to all completions specific to Linux?

I am not certain what scenario that would be used on, since I'm unaware of any package format that both supports all these operating systems *and* shares a binary cache between them all. And I doubt that it's being installed to a USB drive and run on multiple operating systems. Unlike e.g. Windows' [PortableApps](https://portableapps.com/) ecosystem, you would end up with too much software that has to be installed multiple times anyway, once per OS.

akinomyoga:
> You mean it includes all completions specific to FreeBSD, in addition to all completions specific to Solaris, in addition to all completions specific to Linux?

Yes.

> I am not certain what scenario that would be used on, since I'm unaware of any package format that both supports all these operating systems _and_ shares a binary cache between them all.

That's not related to the case I assume.

> And I doubt that it's being installed to a USB drive and run on multiple operating systems. [...]

This is also irrelevant.

I'm not talking about packages for distributions or a USB drive including binary files. `bash-completion` is essentially a set of shell scripts and doesn't include any binary files, so it is essentially possible to just copy the files to another host and source `bash_completion` from e.g. `~/.bashrc`.

For example, one would want to do this when one uses a shared system, where users don't have an administrative privilege or users may want to use specific versions of shell settings without affecting other users. You might say one should rebuild `bash-completion` for each host, but it wasn't needed so far. One could install `bash-completion` to somewhere once `--prefix="$HOME/tmp"` and could copy the results. I don't see the necessity of requiring extra steps to generate `bash-completion` for each host the user logs in.

In fact, there was no need to rebuild `bash-completion` for each operating system. We cannot say it is really necessary to prepare a copy of `bash-completion` for each operating system. I agree that It is cleaner for the packages for specific operating systems to exclude unnecessary completions. However, I don't think the current way of installing everything causes a real problem. In case a distribution wants to override the completions prepared by bash-completion (such as `pkgutil` in macOS mentioned in the original explanation of the PR), we can prefix `_` to lower the precedence (e.g. rename `pkgutil` to `_pkgutil`).

akinomyoga:
The CI is failing. Could you modify the format of the commit message (as described in the following item I quoted from CONTRIBUTING.md)?

https://github.com/scop/bash-completion/blob/8304d3342a07e0f1a402ff0a6752e513c2a0541a/CONTRIBUTING.md?plain=1#L166-L177

completions/Makefile.am
akinomyoga: Thanks for the reply. I understand that you won't work on it within this PR.

> the rest have OS conditionals which I don't quite understand but feel a bit wrong (e.g. there are comments on other tickets about how the checks are being used as a shortcut for "probably not the same command with the same options").

Which one are you talking about? I found #453, but the explanation about the check there appears understandable to me. What do you feel a bit wrong with? Are there other tickets related to this?

The conditions for `ifup` and `_modules` would be hard to check for cross-compiling, so we shouldn't try to switch their installation probably.

The other tests using `OSTYPE` seem consistent with the checks of the completions for which you adjusted the installation. You may think it is unneeded because it sets up completions in Gentoo Linux as expected, but it won't set up completions in non-Linux systems. I currently believe it would be consistent to switch the installations of the completions for Linux in a similar way as you did for non-Linux completions.

eli-schwartz:
> The CI is failing. Could you modify the format of the commit message (as described in the following item I quoted from CONTRIBUTING.md)?

I'm unfamiliar with this "convention" -- as far as I'm aware it's not actually the conventional way to do anything, it seems to violate a number of best practices. That being said, a bit of digging around tells me this style was invented by the JavaScript world which... explains a whole lot about JavaScript.

The explanation provided for how it is necessary to generate changelogs makes sense to me -- I had been wondering why the changelog doesn't effectively describe user-facing changes but dives into the weeds of developer commit messages. Personally, I would suggest migrating to [Towncrier](https://towncrier.readthedocs.io/)

Unfortunately I lack the bandwidth at the moment to figure out how this unusual commit message style works or how to integrate it into the intended semantics of a changelog. And since if I understand correctly this is a merge blocker, that also means I don't have the bandwidth to implement the rest of your feedback. Unfortunately I can't commit to much at the moment, other than to keep this review in mind before opening further PRs to enhance bash-completion.

eli-schwartz:
Suggestion for the future: currently CONTRIBUTING.md is very hard to read because there's a lot of disorganized information leading to cognitive overload in people that just want to submit a PR.

akinomyoga:
Then, can I take over the PR?

scop:
To me, following projects' conventions is a basic thing to do when contributing to them, no matter if I agree with them or not. There's no point in arguing "how" conventional or to whom the commit message style this project want is or its subjective merits: it is what this project uses, and it uses tooling that does useful things when the style is properly followed, and there's documentation about it. (I do think the name of the commit message style is somewhat unfortunate though, but :shrug:)

CONTRIBUTING.md is hard to read, agreed. Some structuring would help, also with ability to link to relevant sections in it. README.md's FAQ has the same problem (at least with linkability). PR's highly appreciated!

https://github.com/scop/bash-completion/pull/1214 resurrects what was in this PR, with the commit message formatting issues addressed. I intend to merge it after CI passes.

eli-schwartz:
> To me, following projects' conventions is a basic thing to do when contributing to them, no matter if I agree with them or not.

Not exactly. It's a basic thing to do when you intend to get your stuff merged. Sometimes you realize you lack the resources to follow said conventions, in which case you change your mind about contributing, or ideally choose to delay contributing until you can get up to speed on the conventions. In both cases you are no longer in a state of currently contributing to the project, which means that the question of following the conventions falls off the map (but may become relevant later on).

> There's no point in arguing "how" conventional or to whom the commit message style this project want is or its subjective merits in this PR: it is what this project uses, and it uses tooling that does useful things when the style is properly followed, and there's documentation about it.

There is certainly a point. It's a topic that came up as a side point, as a result of the PR. Its value as a ***subjective*** critique is equally good (or equally bad!) regardless of whether it appears as a comment on another PR, or as a distinct ticket. And it came up organically. :)

eli-schwartz:
Thanks, both, for tweaking the PR into mergeable condition. :)

@akinomyoga, apologies for neglecting to give the "go-ahead" earlier.

scop:
I'll refrain from further arguing the semantic and subjective differences here, just speaking for myself that the likelihood of something persisting as a TODO item from a closed PR having a different topic than the one at hand for this project is much, much lower than if it was a "proper" discussion or a dedicated issue.

PR #1211, opened 2024-06-07
feat(sops): add 3rd party fallback completion loader

scop:
https://github.com/getsops/sops

### DIFF
--- a/completions/_sops
+++ b/completions/_sops
@@ -1,0 +1,24 @@
+# 3rd party completion loader for commands using           -*- shell-script -*-
+# version 1 of the https://cli.urfave.org library.
+#
+# This serves as a fallback in case the completion is not installed otherwise.
+
+# https://github.com/urfave/cli/blob/v1-maint/docs/v1/manual.md#bash-completion
+# https://github.com/urfave/cli/blob/v1-maint/autocomplete/bash_autocomplete
+_comp_cmd__urfave_cli_v1()
+{
+    local cur prev words cword comp_args
+    _comp_initialize -- "$@" || return
+
+    local compcmd=("${words[@]:0:cword}")
+    if [[ $cur == -* ]]; then
+        compcmd+=("$cur")
+    fi
+    compcmd+=(--generate-bash-completion)
+
+    _comp_compgen_split -- "$("${compcmd[@]}" 2>/dev/null)"
+} &&
+    complete -o bashdefault -o default -o nospace \
+        -F _comp_cmd__urfave_cli_v1 "$1"
+
+# ex: filetype=sh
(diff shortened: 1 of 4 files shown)


completions/_sops
```diff
@@ -0,0 +1,24 @@
+# 3rd party completion loader for commands using           -*- shell-script -*-
+# version 1 of the https://cli.urfave.org library.
+#
+# This serves as a fallback in case the completion is not installed otherwise.
+
+# https://github.com/urfave/cli/blob/v1-maint/docs/v1/manual.md#bash-completion
+# https://github.com/urfave/cli/blob/v1-maint/autocomplete/bash_autocomplete
+_comp_cmd__urfave_cli_v1()
```
scop, line 8: Not sure if this is according to the naming guideline, but at least it is "in spirit".

completions/_sops
```diff
@@ -0,0 +1,24 @@
+# 3rd party completion loader for commands using           -*- shell-script -*-
+# version 1 of the https://cli.urfave.org library.
```
scop, line 2: Not sure if there are substantial differences at the moment in the v1 vs v2 completions at the moment, but I initially looked into using the upstream main branch one as a template for `sops`  and didn't get it to work immediately, so thought we might want to add one for each version. That'll mean some more maintenance when tools upgrade their urfave/cli.

- v1: https://github.com/urfave/cli/blob/v1-maint/autocomplete/bash_autocomplete
- v2: https://github.com/urfave/cli/blob/v2-maint/autocomplete/bash_autocomplete
- to-be v3: https://github.com/urfave/cli/blob/main/autocomplete/bash_autocomplete

completions/_sops
```diff
@@ -0,0 +1,24 @@
...
+    local compcmd=("${words[@]:0:cword}")
+    if [[ $cur == -* ]]; then
+        compcmd+=("$cur")
+    fi
+    compcmd+=(--generate-bash-completion)
+
+    _comp_compgen_split -- "$("${compcmd[@]}" 2>/dev/null)"
+} &&
+    complete -o bashdefault -o default -o nospace \
+        -F _comp_cmd__urfave_cli_v1 "$1"
```
scop, line 22: Followed upstream in the use of various `-o`. Notably `nospace` does not seem to work that well at least with `sops`, but I don't know if we want to "fix" it and deviate from upstream on this.

scop:
Added upstream FYI discussion at https://github.com/urfave/cli/discussions/1922

akinomyoga (approved):
I'm not familiar with the upstream framework, but as they seem to have left 👍 , I think this should be fine.

completions/_sops
akinomyoga: I think this is OK.

PR #1212, opened 2024-06-10
fix(_comp_expand_glob): set LC_COLLATE for the sorting order

### LINKED ISSUE #1207: 000_bash_completion_compat.bash sourced last in some locales
## Describe the bug

Scripts in /etc/bash_completion.d/ often rely on the compatibility stuff like the `have` function but some locales order numbers last, so the compat stuff isn't available.

## To reproduce

1. install something that relies on `have` being available, such as pmount 0.9.23-7.1 in Debian
2. `LANG=cs_CZ.UTF-8 bash`

## Expected behavior

no errors, as opposed to `bash: have: příkaz nenalezen` (which means "command not found")

- [X] Operating system name/distribution and version: Debian testing (as of 2024-06-03)
- [X] bash version, `echo "$BASH_VERSION"`: 5.2.21(1)-release
- [X] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.14.0

## Additional context

This glob: https://github.com/scop/bash-completion/blob/35293af0aaa2d37cee5549d3687004a045003578/bash_completion#L3323 needs to be evaluated with `LC_COLLATE=C`.

akinomyoga:
Fixes #1207. Now `_comp_expand_glob` adjusts `LC_COLLATE` (2d7b9b136) and the files in compatibility directories are generated by `_comp_expand_glob` (0beaa288b).

In addition to the fix, I modified `_comp_expand_glob` to take account of the new `GLOBSORT` added in Bash 5.3 (1322589f1).

This is just refactoring, but I also moved the codes to list the files in a function since the number of temporary global variables is increasing (f6c99d7). Only the list of the files to source is written to the global variable `_comp__init_startup_configs`, and the files are sourced outside the function.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -320,7 +320,7 @@ _comp_upvars()
 # interference from user's shell options/settings or environment variables.
 # @param $1 array_name  Array name
 #   The array name should not start with an underscore "_", which is internally
-#   used.  The array name should not be "GLOBIGNORE".
+#   used.  The array name should not be "GLOBIGNORE" or "GLOBSORT".
 # @param $2 pattern     Pattern string to be evaluated.
 #   This pattern string will be evaluated using "eval", so brace expansions,
 #   parameter expansions, command substitutions, and other expansions will be
@@ -335,7 +335,7 @@ _comp_expand_glob()
         printf 'bash-completion: %s: unexpected number of arguments\n' "$FUNCNAME" >&2
         printf 'usage: %s ARRAY_NAME PATTERN\n' "$FUNCNAME" >&2
         return 2
-    elif [[ $1 == @(GLOBIGNORE|_*|*[^_a-zA-Z0-9]*|[0-9]*|'') ]]; then
+    elif [[ $1 == @(GLOBIGNORE|GLOBSORT|_*|*[^_a-zA-Z0-9]*|[0-9]*|'') ]]; then
         printf 'bash-completion: %s: invalid array name "%s"\n' "$FUNCNAME" "$1" >&2
         return 2
     fi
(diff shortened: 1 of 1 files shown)


bash_completion
```diff
@@ -3298,45 +3303,65 @@ else
...
-    # run-in-place-from-git-clone setups.  Notably we do it after the system
-    # location here, in order to prefer in-tree variables and functions.
-    if [[ ${BASH_SOURCE%/*} == */share/bash-completion ]]; then
-        _comp__init_compat_dir=${BASH_SOURCE%/share/bash-completion/*}/etc/bash_completion.d
-    elif [[ $BASH_SOURCE == */* ]]; then
-        _comp__init_compat_dir="${BASH_SOURCE%/*}/bash_completion.d"
+# List custom/extra completion files to source on the startup
+## @param $1 path Path to "bash_completion"
+## @var[out] _comp__init_startup_configs
+_comp__init_collect_startup_configs()
```
scop, line 3309: I suppose we could `unset -f` this as it is only being used at load time?

Pre-approved with that.

scop (approved):

bash_completion
akinomyoga: Yes, and it already unsets itself at the end.

https://github.com/scop/bash-completion/pull/1212/files#diff-a4757074ff650000804fd3eaabe9b0a9e02e33040ca5b8afd4c0275fc5f3e136R3357

Is it better to do it at the beginning of the function?

scop: Ah! Missed that.

I don't know if it matters much where the `unset` happens, if something is only to be called once anyway. Perhaps at the end sounds like the best option for me. We are not consistent with that currently (could address in a followup MR if we care):
- `_comp_dequote__initialize`: beginning
- `_comp__init_set_up_service_completions`, `_comp_cmd_tar`, `_comp_cmd_vncviewer__bootstrap`: end
- `_comp_cmd_makepkg__bootstrap`: middle

akinomyoga: I'll submit a PR.

As for `_comp_cmd_makepkg__bootstrap`, I guess it wanted to put `$func "$@"` at the end of the function to reflect the exit status.

akinomyoga: #1215

PR #1213, opened 2024-06-16
test: fix tests for failglob

akinomyoga:
This PR fixes the problems of tests that would be triggered by setting `failglob` in `test/config/bashrc`. (However, this PR doesn't include the `failglob` setting itself.)

Commit aa8e4afe8dfeeae3ee959b83e486780f4bffe04d is an extra commit to update the names of test classes and files.


scop (approved):
Thanks!

PR #1215, opened 2024-06-17
refactor: normalize unset positions of call-once functions

akinomyoga:
As mentioned in https://github.com/scop/bash-completion/pull/1212#discussion_r1642366683.


scop (approved):
Thanks!

PR #1224, opened 2024-06-23
feat(feh): deassociate with xcf, pdf, dvi, txt

Maelan:
See commit message.

### DIFF
--- a/completions/feh
+++ b/completions/feh
@@ -111,7 +111,7 @@ _comp_cmd_feh()
 
     # FIXME: It is hard to determine correct supported extensions.
     # feh can handle any format that imagemagick can plus some others
-    _comp_compgen_filedir 'xpm|tif?(f)|png|p[npgba]m|iff|?(i)lbm|jp?(e)g|jfi?(f)|gif|bmp|arg?(b)|tga|xcf|ani|ico|?(e)ps|pdf|dvi|txt|svg?(z)|cdr|[ot]tf|ff?(.gz|.bz2)|webp|y4m|hei[cf]?(s)|avif?(s)'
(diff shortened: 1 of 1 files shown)


akinomyoga:
```
- opening *.xcf (Gimp vectorial image) is mentioned in feh’s manpage
  but I have been unable to open any such file;
```

What is the problem in this case? According to the man page, it seems to use ImageMagick for `.xcf` when it is enabled by `--conversion-timeout`. There are many possibilities:

- ImageMagick stopped supporting `.xcf` after the corresponding part of the man page of `feh` was written. In this case, we can drop `.xcf` from the list.
- ImageMagick needs a special configuration to open `.xcf`. In this case, some users may set it up, so I'm not sure if we should simply drop it.
- ImageMagick works, but `feh` cannot use the result of ImageMagick for some reason. In this case, maybe we should ask the upstream `feh` about whether this is intentional.

```
- opening *.ps, *.pdf, *.dvi (document files) works through ImageMagick,
  but only when the document contains a single page;
  + PDF and DVI are primarily used for possibly-multipage textual
    documents, PDF viewers are widely available and they are much better
    for this purpose than an image viewer that first converts documents
    to bitmaps; it’s unlikely that a user wants to use feh for these.
```

Is it? The images that will be embedded in PDFs are usually generated as single-page PDFs. I always create plots in academic papers as single-page PDFs. Although I'm not a user of `feh`, I'm not sure if we can easily conclude that no users use `feh` to preview images in the form of single-page PDFs.

scop:
I have not tested the behavior, but I'm inclined to agree with @akinomyoga here, unless some more feedback confirming that deassociating these would be the right thing to do.

scop:
Closing due to lack of feedback.

PR #1228, opened 2024-06-26
feat(curl): Complete "%" for --trace

yedayak:
"%" makes trace go to stderr

### DIFF
--- a/completions/curl
+++ b/completions/curl
@@ -11,8 +11,8 @@ _comp_cmd_curl()
         --abstract-unix-socket | --alt-svc | --config | --cookie | \
             --cookie-jar | --dump-header | --egd-file | --etag-compare | \
             --etag-save | --hsts | --key | --libcurl | --netrc-file | \
-            --output | --proxy-key | --random-file | --trace | --trace-ascii | \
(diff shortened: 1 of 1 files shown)


akinomyoga (approved):

completions/curl
```diff
@@ -153,6 +153,11 @@ _comp_cmd_curl()
             _comp_compgen -- -W 'SRP'
             return
             ;;
+        --trace)
+            _comp_compgen -- -W '%'
```
scop, line 157: If we do this, should we complete `-` as well (for sending the trace to stdout)?

Also, `--trace-ascii` appears to take `%` and `-` as well even though in my curl man page only the latter is documented.

scop (changes requested):

completions/curl
yedayak: > If we do this, should we complete `-` as well (for sending the trace to stdout)?

Yeah, added 

> Also, `--trace-ascii` appears to take `%` and `-` as well even though in my curl man page only the latter is documented.

Added for that too, it seems `%` was added to the man here: https://github.com/curl/curl/commit/1159dc359da767c36ed724f246cd0a3bd4b1f7ce

PR #1231, opened 2024-07-02
fix(curl): Complete all and category for --help

yedayak:


completions/curl
```diff
@@ -111,6 +111,7 @@ _comp_cmd_curl()
                     [[ $x != -* ]] || return
                 done
                 _comp_compgen -- -W '"${categories[@]}"'
+                _comp_compgen -a -- -W 'category all'
```
akinomyoga, line 114:
```suggestion
                _comp_compgen -- -W '"${categories[@]}" category all'
```

yedayak: Done

akinomyoga (approved):

PR #1236, opened 2024-07-22
feat(ip): complete ip ntable show

yedayak:


completions/ip
```diff
@@ -456,7 +456,21 @@ _comp_cmd_ip()
...
-                    # TODO
+                    case $prev in
+                        dev)
+                            _comp_compgen_available_interfaces
+                            ;;
+                        name)
+                            _comp_compgen_split -- "$(
+                                {
+                                    ip -c=never ntable show 2>/dev/null || ip ntable show
+                                } | sed -e '/inet.*/!d' -e 's/inet6\{0,1\} //g'
```
akinomyoga, line 467:
```suggestion
                                } | sed -ne 's/^inet6\{0,1\} //p'
```

`/inet.*/` and `/inet6\{0,1\} /` match `inet ` in the middle of a line (such as `    dev clarinet ...`). If it's supposed to match at the beginning of line, the flag `g` is unnecessary.

yedayak: The joining of the sed commands is clever, changed according to the suggestions. Also added `command` before the sed as it is elsewhere

akinomyoga: Thanks!

completions/ip
```diff
@@ -456,7 +456,21 @@ _comp_cmd_ip()
...
-                    # TODO
+                    case $prev in
+                        dev)
+                            _comp_compgen_available_interfaces
+                            ;;
+                        name)
+                            _comp_compgen_split -- "$(
+                                {
+                                    ip -c=never ntable show 2>/dev/null || ip ntable show
+                                } | command sed -ne 's/^inet6\{0,1\} //p'
```
akinomyoga, line 467: In my system, this seems to suggest duplicate names:

```bash
$ ip ntable show | command sed -ne 's/^inet6\{0,1\} //p'
arp_cache
arp_cache
arp_cache
arp_cache
arp_cache
arp_cache
arp_cache
ndisc_cache
ndisc_cache
ndisc_cache
ndisc_cache
ndisc_cache
ndisc_cache
ndisc_cache
ndisc_cache
```

Could you add `sort -u`?

yedayak: Sure, although it seems `_comp_comgen` doesn't care about having repeating options, and just completes the unique ones. I understand that usually it's best practice to use less commands if possible to reduce exec/forking. Do you see an advantage of making it unique here?

akinomyoga: Hmm, OK. You are right. I thought there was a problem with duplicate entries, but now I cannot seem to find any problems with having duplicate entries. Maybe it was only in a specific case (e.g. like [cd](https://github.com/scop/bash-completion/blob/e209afb2d0c96ee6bfe83da8b2742a8f8e88d372/completions/cd#L48)). The current version is good.

akinomyoga: As for the exec/fork cost, another solution would be to replace `sed` with `awk`, i.e.,
```suggestion
                                {
                                    ip -c=never ntable show 2>/dev/null || ip ntable show
                                } | command awk 'sub(/^inet6\{0,1\} /, "") && !uniq[$0]++'
```

akinomyoga (approved):

completions/ip
scop: IIRC bash's `COMPREPLY` duplicate filtering is quite slow. If that's true, a fork might be worth considering if there a re a lot of duplicates expected.

scop: Anyway I'll merge as is, feel free to submit followup PR's if it looks some would be in order. The proposed awk thing looks nice, but I'm not sure about the portability (in the sense that I don't have any factual reservations, I just don't know).

PR #1237, opened 2024-08-04
Complete stdin "-" in file locations, add more options that allow it

yedayak:
Every option that reads files using `@` allow also `-` for stdin

### DIFF
--- a/completions/curl
+++ b/completions/curl
@@ -54,6 +54,7 @@ _comp_cmd_curl()
             --proxy-header | -${noargopts}[dH])
             if [[ $cur == \@* ]]; then
                 _comp_compgen -c "${cur:1}" filedir
+                _comp_compgen -a -c "${cur:1}" -- -W '-'
                 if [[ ${#COMPREPLY[@]} -eq 1 && -d ${COMPREPLY[0]} ]]; then
                     COMPREPLY[0]+=/
                     compopt -o nospace
--- a/completions/curl
+++ b/completions/curl
@@ -30,8 +30,8 @@ _comp_cmd_curl()
             --sasl-authzid | --service-name | --socks5-gssapi-service | \
             --speed-limit | --speed-time | --telnet-option | --tftp-blksize | \
             --time-cond | --tls13-ciphers | --tlspassword | --tlsuser | \
-            --url | --user | --user-agent | --version | --write-out | \
-            -${noargopts}[CFmQreYytzuAVw])
+            --url | --user | --user-agent | --version | \
+            -${noargopts}[CFmQreYytzuAV])
             return
             ;;
         --cacert | --cert | --proxy-cacert | --proxy-cert | -${noargopts}E)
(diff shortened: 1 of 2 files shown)


completions/curl
```diff
@@ -50,10 +50,11 @@ _comp_cmd_curl()
             _comp_compgen_filedir crl
             return
             ;;
-        --data | --data-ascii | --data-binary | --data-urlencode | --header | \
-            --proxy-header | -${noargopts}[dH])
+        --data | --data-ascii | --data-binary | --data-urlencode | --json | \
+            --header | --proxy-header | --write-out | -${noargopts}[dHw])
             if [[ $cur == \@* ]]; then
                 _comp_compgen -c "${cur:1}" filedir
+                _comp_compgen -a -c "${cur:1}" -- -W '-'
```
scop, line 57: No strong opinions, but I suppose we could simplify/optimize this to the following unless we see downsides to doing that? WDYT, and @akinomyoga too? 
```suggestion
                [[ $cur == @?(-) ]] && COMPREPLY+=(-)
```

akinomyoga: Hmm, although it is functionally equivalent, I prefer using `_comp_compgen` which is more idiomatic.

It takes time for me to understand that `[[ $cur == @?(-) ]] && COMPREPLY+=(-)` is functionally equivalent to `_comp_compgen -ac "${cur:1}" -- -W '='` and thus doesn't have a bug nor an extra intentional effect.

akinomyoga: There seem to be both types of existing cases:

```bash
./completions/7z:59:            _comp_compgen -c "${cur:4}" -- -P"${cur:0:4}" -W '-'
./completions/curl:144:            _comp_compgen -- -W '-'
./completions/gpgv:35:                _comp_compgen -- -W '-'
./completions/iperf:55:            _comp_compgen -- -W 'C'
./completions/pack200:35:            _comp_compgen -- -W '-'
./completions/pdftotext:32:            _comp_compgen -- -W '-'
./completions/puppet:160:                    _comp_compgen -- -W '-'
./completions/tshark:97:            _comp_compgen -- -W 'n'
./completions/unpack200:19:            _comp_compgen -- -W '-'
```

```bash
./completions/cppcheck:38:            [[ ! $cur || $cur == - ]] && COMPREPLY+=(-)
./completions/mr:46:                    COMPREPLY+=(-)
./completions/nslookup:59:        [[ $REPLY -eq 1 && $cur == @(|-) ]] && COMPREPLY+=(-)
./completions/patch:24:            [[ ! $cur || $cur == - ]] && COMPREPLY=(-)
./completions/tshark:67:            [[ $cur == @(|-) ]] && COMPREPLY+=(-)
./completions/wget:78:            _comp_compgen_filedir && [[ $cur == - || ! $cur ]] && COMPREPLY+=(-)
```

scop: For me it's the opposite, I'm far from being "natural" with the stuff we've added over the few last releases and have to look up stuff a lot, almost certainly because I haven't been writing all that much code in this project lately, whereas straight bash is much easier for me. But I'm fine with those who are doing most of the actual work these days getting to decide the flavor. Would not hurt to be consistent with it though through the whole tree.

Pre-approved either way you guys decide.

yedayak: I don't really mind either, I think the second option is easier to read without context, but the first one makes more sense here since it follows the same pattern as the line before it.

akinomyoga: I again thought about it, but I think it is still better to generate completions through `_comp_compgen` even in simple cases because it would be easier to apply changes later in a consistent way (by modifying the behavior of `_comp_compgen` or by grep'ping the calls of `_comp_compgen`).

PR #1244, opened 2024-09-05
test(scp): Add unit tests for getting remote files

yedayak:
Since there doesn't seem to be any unit tests for xfuncs, I'm not sure if the naming here makes sense, or if I should make some more directories, for example a `unit/xfunc` or `fixtures/xfunc`?

I did this since there seems to be a lot of issues with this function, as seen in https://github.com/scop/bash-completion/pull/910 and https://github.com/scop/bash-completion/pull/765#issuecomment-1173097965
This should be a first good step to make fixing the issues easier IMO.

### DIFF
--- a/test/fixtures/mount/bin/showmount
+++ b/test/fixtures/mount/bin/showmount
@@ -1,6 +1,6 @@
-#!/bin/sh
+#!/bin/bash
 
-if [ "$1" = -e ] && [ "$2" = mocksrv ]; then
+if [[ $1 == -e && $2 == "mocksrv" ]]; then
     echo "Header line"
     echo "/test/path"
     echo "/test/path2"
--- a/test/fixtures/scp/bin/ssh
+++ b/test/fixtures/scp/bin/ssh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -eu
+args=("$@")
+while true; do
+    arg="${args[0]-}"
+    case "$arg" in
+        -o)
+            args=("${args[@]:2}")
+            ;;
+        local)
+            args=("${args[@]:1}")
+            ;;
+        *)
+            break
+            ;;
+    esac
+done
+#shellcheck disable=SC2068
+${args[@]}
--- a/test/t/test_scp.py
+++ b/test/t/test_scp.py
@@ -2,7 +2,7 @@ from itertools import chain
 
 import pytest
 
-from conftest import assert_bash_exec, assert_complete
+from conftest import assert_bash_exec, assert_complete, bash_env_saved
 
 LIVE_HOST = "bash_completion"
 
@@ -23,7 +23,7 @@ class TestScp:
                     )
                 ),
                 # Local filenames
-                ["config", "known_hosts", r"spaced\ \ conf"],
+                ["bin/", "config", "known_hosts", r"spaced\ \ conf"],
             )
         )
         assert completion == expected
@@ -43,7 +43,7 @@ class TestScp:
                     )
                 ),
                 # Local filenames
-                ["config", "known_hosts", r"spaced\ \ conf"],
+                ["bin/", "config", "known_hosts", r"spaced\ \ conf"],
             )
         )
         assert completion == expected
@@ -101,3 +101,43 @@ class TestScp:
         completion = assert_complete(bash, "scp remote_host:spaces")
         assert_bash_exec(bash, "unset -f ssh")
         assert completion == r"\\\ in\\\ filename.txt"
+
+    def test_xfunc_remote_files(self, bash):
+        with bash_env_saved(bash) as bash_env:
+            bash_env.save_variable("COMPREPLY")
+            bash_env.write_variable(
+                "PATH",
+                "$PWD/scp/bin:$PATH",
+                quote=False,
+            )
+            bash_env.write_variable("cur", "local:shared/default/")
+            completions_regular_escape = (
+                assert_bash_exec(
+                    bash,
+                    r'_comp_compgen -x scp remote_files; printf "%s\n" "${COMPREPLY[@]}"',
+                    want_output=True,
+                )
+                .strip()
+                .splitlines()
+            )
+            completions_less_escape = (
+                assert_bash_exec(
+                    bash,
+                    r'_comp_compgen -x scp remote_files -l; printf "%s\n" "${COMPREPLY[@]}"',
+                    want_output=True,
+                )
+                .strip()
+                .splitlines()
+            )
+        assert completions_regular_escape == [
+            "shared/default/bar ",
+            r"shared/default/bar\\\ bar.d/",
+            "shared/default/foo ",
+            "shared/default/foo.d/",
+        ]
+        assert completions_less_escape == [
+            "shared/default/bar ",
+            r"shared/default/bar\ bar.d/",
+            "shared/default/foo ",
+            "shared/default/foo.d/",
+        ]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
     hooks:
       - id: shfmt
         types: [text]
-        files: ^(bash_completion(\.d/[^/]+\.bash)?|completions/.+|test/(config/bashrc|fallback/update-fallback-links|runLint|update-test-cmd-list)|.+\.sh(\.in)?)$
+        files: ^(bash_completion(\.d/[^/]+\.bash)?|completions/.+|test/(config/bashrc|fixtures/.+/bin/.+|fallback/update-fallback-links|runLint|update-test-cmd-list)|.+\.sh(\.in)?)$
         exclude: ^completions/(\.gitignore|Makefile.*)$
 
   - repo: https://github.com/shellcheck-py/shellcheck-py
@@ -22,7 +22,7 @@ repos:
       - id: shellcheck
         args: [-f, gcc]
         types: [text]
-        files: ^(bash_completion(\.d/[^/]+\.bash)?|completions/.+|test/(config/bashrc|fallback/update-fallback-links|runLint|update-test-cmd-list)|.+\.sh(\.in)?)$
+        files: ^(bash_completion(\.d/[^/]+\.bash)?|completions/.+|test/(config/bashrc|fixtures/.+/bin/.+|fallback/update-fallback-links|runLint|update-test-cmd-list)|.+\.sh(\.in)?)$
         exclude: ^completions/(\.gitignore|Makefile.*)$
         require_serial: false  # We disable SC1090 anyway, so parallel is ok
 


akinomyoga:
I think this should be tested within `test/t/test_scp.py`. The `test/t/unit/test_unit_*.py` are the tests for the core components in `bash_completion`. Except for that, LGTM.

test/t/unit/test_unit_xfunc_scp_remote_files.py
```diff
@@ -0,0 +1,33 @@
...
+                "$PWD/_comp_compgen_xfunc_scp_remote_files/bin:$PATH",
+                quote=False,
+            )
+            yield
+
+    def test_1(self, bash, ssh_path):
+        completions = (
+            assert_bash_exec(
+                bash,
+                r'cur=local:shared/default/ _comp_compgen -ax scp remote_files; printf "%s\n" "${COMPREPLY[@]}"; unset cur',
```
akinomyoga, line 22: It's actually better to use `bash_env` also for `cur`. That also includes the check that `_comp_xfunc_scp_cmpgen_remote_files` doesn't break the existing `cur`.

yedayak: I changed it to use `bash_env_saved` also in the test itself, is that what you meant?

yedayak:
> I think this should be tested within `test/t/test_scp.py`. The `test/t/unit/test_unit_*.py` are the tests for the core components in `bash_completion`. Except for that, LGTM.

Moved to `test/t/test_scp.py`.
I had to add `@pytest.mark.bashcomp(ignore_env=r"^\+COMPREPLY=")` to the entire class, I'm not sure if there is a better way to only apply it to the necessary functions.

test/t/unit/test_unit_xfunc_scp_remote_files.py
akinomyoga: Thanks.


> I changed it to use `bash_env_saved` also in the test itself, is that what you meant?

Yes.

akinomyoga:
> I'm not sure if there is a better way to only apply it to the necessary functions.

You can call `bash_env.save_variable("COMPREPLY")`.

test/t/test_scp.py
```diff
@@ -95,3 +96,32 @@ def test_remote_path_with_nullglob(self, completion):
...
+
+    @pytest.fixture(scope="class")
+    def ssh_mock(self, request, bash):
+        with bash_env_saved(bash) as bash_env:
+            bash_env.write_variable(
+                "PATH",
+                "$PWD/scp/bin:$PATH",
+                quote=False,
+            )
+            yield
```
akinomyoga, line 108: After another look, I feel this doesn't need to be a pytest fixture. This fixture is used by only one test, so this temporary value of `PATH` can be just set inside `test_xfunc_remote_files`.

yedayak:
> > I'm not sure if there is a better way to only apply it to the necessary functions.
> 
> You can call `bash_env.save_variable("COMPREPLY")`.

This doesn't seem to work, the diff doesn't look at the saved variables, it just uses `get_env()` which is basically just `_comp__test_get_env`.

akinomyoga:
> > > I'm not sure if there is a better way to only apply it to the necessary functions.
> >
> > You can call `bash_env.save_variable("COMPREPLY")`.
> 
> This doesn't seem to work,

How have you checked that it doesn't seem to work? I now tried that, but it works.

> the diff doesn't look at the saved variables, it just uses `get_env()` which is basically just `_comp__test_get_env`.

The saved variables are restored by `bash_env`. `get_env()` is called **after** the `with` statement calls `bash_env_saved.__exit__()` and the saved variables are restored.

test/t/test_scp.py
akinomyoga: I was actually wondering whether the lifetime of the generator fixture `ssh_mock` (and thus the `with` statement inside the generator) is well-defined and matches our expectations. It now seems to work, but this can be broken when we add multiple tests relying on the same fixture or when another condition is changed. Can you explain why this works, in particular, how the ordering of our `get_env()` check and `bash_env_saved.__exit__` is stably sequenced?

yedayak:
> > > > I'm not sure if there is a better way to only apply it to the necessary functions.
> > > 
> > > 
> > > You can call `bash_env.save_variable("COMPREPLY")`.
> > 
> > 
> > This doesn't seem to work,
> 
> How have you checked that it doesn't seem to work? I now tried that, but it works.
> 
> > the diff doesn't look at the saved variables, it just uses `get_env()` which is basically just `_comp__test_get_env`.
> 
> The saved variables are restored by `bash_env`. `get_env()` is called **after** the `with` statement calls `bash_env_saved.__exit__()` and the saved variables are restored.

Ah I managed it, I was trying to save the variable after it was changed, which in hindsight obviously doesn't make sense :)

test/t/test_scp.py
yedayak: >After another look, I feel this doesn't need to be a pytest fixture. This fixture is used by only one test, so this temporary value of PATH can be just set inside test_xfunc_remote_files.

I get that, I was thinking of adding a a test for the new `-l` (less escaping) option here, which will need basically the same steps. Maybe I can just do both in the same test instead?

> I was actually wondering whether the lifetime of the generator fixture `ssh_mock` (and thus the `with` statement inside the generator) is well-defined and matches our expectations. It now seems to work, but this can be broken when we add multiple tests relying on the same fixture or when another condition is changed. Can you explain why this works, in particular, how the ordering of our `get_env()` check and `bash_env_saved.__exit__` is stably sequenced?

Everything that happens after the yield (the finalization) happens in the [reverse order of calling the fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended). Since `ssh_mock` depends on the `bash` fixture, `bash` gets called first when running, and last in the finalization, so `get_env(0` will happen after `bash_env_saved.__exit__` is called.

akinomyoga: > Maybe I can just do both in the same test instead?

I think that is better because other tests in the same class may also be affected by this temporary variable change.

> Everything that happens after the yield (the finalization) happens in the [reverse order of calling the fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended).

Thank you! It's good to know that!

> Since `ssh_mock` depends on the `bash` fixture,

I also missed this point...

yedayak: OK, for now I removed the `ssh_mock`, if we will need it we can add it later

test/fixtures/scp/bin/ssh
akinomyoga, line 1: This is just a hypothetical case, but this implementation may produce an error message when e.g. `local mycmd -o -o -o -o` is passed as arguments. This tries to perform `shift 1` once and `shift 2` four times, but the number of arguments is less than 9. Then, ash/dash/ksh/zsh produces an error message.

Since this implementation assumes the hostname be specified as `local`, can we just remove the arguments until the first occurrence of `local`?

```bash
until [ "${1-local}" = local ]; do shift 1; done
[ "${1-}" = local ] && shift 1
```

yedayak: > This is just a hypothetical case, but this implementation may produce an error message when e.g. `local mycmd -o -o -o -o` is passed as arguments. This tries to perform `shift 1` once and `shift 2` four times, but the number of arguments is less than 9. Then, ash/dash/ksh/zsh produces an error message.

This invocation will make an error also with the real `ssh` command, do I don't think it really matters here
> Since this implementation assumes the hostname be specified as `local`, can we just remove the arguments until the first occurrence of `local`?
> 
> ```shell
> until [ "${1-local}" = local ]; do shift 1; done
> [ "${1-}" = local ] && shift 1
> ```
This doesn't work if the `-o "<Options>"` is after `local`, e.g. `ssh local -o blah ls`, which does work when running the real `ssh`.

test/fixtures/scp/bin/ssh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+for arg in "$@"; do
+    case "$arg" in
+	-o)
+		shift 2;;
+	local)
+		shift 1;;
```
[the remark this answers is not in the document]
yedayak: Did this, the reason this passed CI is because shfmt doesn't run on fixtures, added a commit to make both shfmt and shellcheck run on fixtures/*/bin/*

test/fixtures/scp/bin/ssh
akinomyoga: > > This is just a hypothetical case, but this implementation may produce an error message when e.g. `local mycmd -o -o -o -o` is passed as arguments. This tries to perform `shift 1` once and `shift 2` four times, but the number of arguments is less than 9. Then, ash/dash/ksh/zsh produces an error message.
> 
> This invocation will make an error also with the real `ssh` command, do I don't think it really matters here

Have you tried? It works in my environment (where chatoyancy is a hostname):

```bash
$ ssh chatoyancy echo -o -o -o -o
-o -o -o -o
```

Anyway, what I wanted to point out is that the implementation doesn't seem logical. The current implementation removes the arguments "from the beginning" while determining the number of removed arguments by looking at "arbitrary positions" in the argument list. For example, when `local` is the last word of the argument list, this removes a word at the beginning while keeping the word `local`. 

> 
> > Since this implementation assumes the hostname be specified as `local`, can we just remove the arguments until the first occurrence of `local`?
> > ```shell
> > until [ "${1-local}" = local ]; do shift 1; done
> > [ "${1-}" = local ] && shift 1
> > ```
> 
> This doesn't work if the `-o "<Options>"` is after `local`, e.g. `ssh local -o blah ls`, which does work when running the real `ssh`.

Ah, you are right. So it should eliminate the words before the first command name.

test/fixtures/scp/bin/ssh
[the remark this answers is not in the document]
akinomyoga: Thanks! I'm unsure if all `fixtures/*/bin/*` would be written in shell language in the future, but it should be fine. We can adjust it when we add a new executable file not written in shell language.

test/fixtures/scp/bin/ssh
yedayak: > > > This is just a hypothetical case, but this implementation may produce an error message when e.g. `local mycmd -o -o -o -o` is passed as arguments. This tries to perform `shift 1` once and `shift 2` four times, but the number of arguments is less than 9. Then, ash/dash/ksh/zsh produces an error message.
> > 
> > 
> > This invocation will make an error also with the real `ssh` command, do I don't think it really matters here
> 
> Have you tried? It works in my environment (where chatoyancy is a hostname):
> 
> ```shell
> $ ssh chatoyancy echo -o -o -o -o
> -o -o -o -o
> ```

Yeah my testing was wrong, you're correct.
 
> Anyway, what I wanted to point out is that the implementation doesn't seem logical. The current implementation removes the arguments "from the beginning" while determining the number of removed arguments by looking at "arbitrary positions" in the argument list. For example, when `local` is the last word of the argument list, this removes a word at the beginning while keeping the word `local`.

I changed the script to just operate on an array, instead of trying to change `$@`. This should handle these edge cases much better
> 
> > > Since this implementation assumes the hostname be specified as `local`, can we just remove the arguments until the first occurrence of `local`?
> > > ```shell
> > > until [ "${1-local}" = local ]; do shift 1; done
> > > [ "${1-}" = local ] && shift 1
> > > ```
> > 
> > 
> > This doesn't work if the `-o "<Options>"` is after `local`, e.g. `ssh local -o blah ls`, which does work when running the real `ssh`.
> 
> Ah, you are right. So it should eliminate the words before the first command name.

akinomyoga: Thanks!

test/fixtures/mount/bin/showmount
```diff
@@ -1,6 +1,6 @@
-#!/bin/sh
+#!/bin/bash
 
-if [ "$1" = -e ] && [ "$2" = mocksrv ]; then
+if [[ $1 == -e ]] && [[ $2 == "mocksrv" ]]; then
```
akinomyoga, line 3:
```suggestion
if [[ $1 == -e && $2 == "mocksrv" ]]; then
```

akinomyoga (approved):
The rest seems good. The mock `ssh` is of course not a complete implementation of the ssh argument parsing,  but we can at least explain its limited behavior.

akinomyoga (approved):

PR #1247, opened 2024-09-16
feat(slabtop): new completion

scop:


akinomyoga (approved):

yedayak:
Do you think it's better to have this here and not in procps-ng?

yedayak (approved):

scop:
Yes, in principle I'd like to see these upstream, but as we already have a bunch and upstream has none, I think it's fine to continue that for now. Could ping upstream sometime though if they'd be interested in taking them over though.

PR #1248, opened 2024-09-18
Deprecate {ins,rm}mod and mod{info,probe}

evelikov:
I'm aiming to have these in the upstream project and having the ones in bash-completion means that distributions will get clashing files. 

Rename these with the fallback prefix and (try) to fix all the build magic ;-)

### DIFF
--- a/completions/_modinfo
+++ b/completions/_modinfo
@@ -1,5 +1,8 @@
 # Linux modinfo(8) completion                              -*- shell-script -*-
 
+# Use of this file is deprecated.
+# Upstream completion is expected to be available in kmod >= 35, use that instead.
+
 _comp_cmd_modinfo()
 {
     local cur prev words cword was_split comp_args
--- a/test/fallback/completions/rmmmod
+++ b/test/fallback/completions/rmmmod
@@ -1,0 +1,1 @@
+../../../completions/_rmmod
--- a/completions/.gitignore
+++ b/completions/.gitignore
@@ -139,7 +139,7 @@
 /_incus
 /_infracost
 /inotifywatch
-/insmod.static
+/_insmod.static
 /iperf3
 /_istioctl
 /javac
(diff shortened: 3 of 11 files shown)


completions/_modinfo
```diff
@@ -1,5 +1,8 @@
 # Linux modinfo(8) completion                              -*- shell-script -*-
 
+# Use of this file is deprecated.
+# Upstream completion is available in slackpkg >= 15.0.4, use that instead.
```
yedayak, line 4: Seems like a copy paste error, should mention kmod?

scop: Thanks! LGTM in principle, but should we split this PR to two, just in case it'll take more time for modinfo/probe to be done upstream?

scop: Basically what I mean is that we now have "use one in kmod >= 35" notices in both commit messages and contents of modinfo and modprobe here, but the respective completions don't even exist yet in upstream master branch, nor do I see a PR for them upstream.

So until some of that actually happens upstream, we are giving potentially confusing information about an upstream version that does not exist yet (well this applies to the >= 34 notes as well as it stands), and are taking a leap of faith that these will actually end up in kmod 35 even if there does not seem to be publicly visible progress with them yet.

This is not  that big of a deal and I think we can trust you to get these done according to the plan upstream, and file PR's to update the notices in the deprecated files we ship in case the release numbers don't end up the way they are here. But it would be nice if you could confirm this.

evelikov: Is there anything else I can do to unblock this? Do you have any rough ETA or an issue/milestone tracker for the next bash-completions release?

Thanks in advance o/

scop: Let's change the comment wordings to "... is expected to be available in ..." and I think it's good enough.

No tracker or real ETA, but I'd say we could roll a new release "soon".

evelikov: Love the suggestion - updated modinfo/modprobe to use it.

Ftr Debian experimental has picked kmod from git, so they had to manually remove to upstream bash completions. So a bash-completion release would be amazing, thank you :bow:

test/fallback/completions/rmmmod
scop, line 1: JFYI #1264 1264

PR #1249, opened 2024-09-24
fix(dpkg): Add missing short option -R

elboulangero:
Completion for '--raw-extract' is already supported, however the equivalent short option '-R' was missing. Now fixed.

Cf. dpkg-deb help message:

```
dpkg-deb --help
[...]
  -R|--raw-extract <deb> <directory>
[...]
```

### DIFF
--- a/completions/dpkg
+++ b/completions/dpkg
@@ -72,11 +72,11 @@ _comp_cmd_dpkg()
         done
     fi
 
-    local noargopts='!(-*|*[ciAIfexXbsplWSrVLPD]*)'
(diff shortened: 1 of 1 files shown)


akinomyoga (changes requested):
Thanks for your contribution. The value of `noargopts` on line 75 should also be updated.

elboulangero:
Indeed, thanks for your review! I updated the PR.

akinomyoga (approved):

yedayak (approved):

PR #1252, opened 2024-10-14
feat(ngrok): add 3rd party fallback completion loader

akinomyoga:
https://ngrok.com/docs/agent/cli/#ngrok-completion

### DIFF
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -765,6 +765,7 @@ CLEANFILES = \
 	neomutt \
 	_nerdctl \
 	_nfpm \
+	_ngrok \
 	_nomad \
 	_npm \
 	_oc \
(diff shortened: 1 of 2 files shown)


completions/Makefile.am
```diff
@@ -1000,7 +1001,7 @@ symlinks: $(DATA)
 	$(ss) freeciv-server \
 		civserver
 	$(ss) _gaiacli \
-		_gaiad _npm _tendermint
+		_gaiad _npm _tendermint _ngrok
```
scop, line 1004: I've attempted to keep these sorted alpabetically, pre-approved with this change
```suggestion
		_gaiad _ngrok _npm _tendermint
```

akinomyoga: Ah, right! I'll fix it. Thank you for your review.

akinomyoga: 2de6e1b1226c4d06fc90392dfbbb547cf523ad74, and then squashed into 3ad11012c03cee677a2080674b0056d39441676b.

PR #1256, opened 2024-10-30
feat(openssl): Stop having subcommands hardcoded

yedayak:
Instead we get the command list by parsing openssl help, which also adds some subcommands that were missing, and removes some non-existant ones (at least on versions I have).
The parsing only happens when it's necessary, when the subcommand wasn't chosen.

### DIFF
--- a/completions/openssl
+++ b/completions/openssl
@@ -42,22 +42,11 @@ _comp_cmd_openssl()
     local cur prev words cword comp_args
     _comp_initialize -- "$@" || return
 
-    local commands command formats
-
-    commands='asn1parse ca ciphers crl crl2pkcs7 dgst dh dhparam dsa dsaparam
-        ec ecparam enc engine errstr gendh gendsa genrsa nseq ocsp passwd
-        pkcs12 pkcs7 pkcs8 prime rand req rsa rsautl s_client s_server s_time
-        sess_id smime speed spkac storeutl verify version x509 md2 md4 md5 rmd160
-        sha sha1 aes-128-cbc aes-128-ecb aes-192-cbc aes-192-ecb aes-256-cbc
-        aes-256-ecb base64 bf bf-cbc bf-cfb bf-ecb bf-ofb camellia-128-cbc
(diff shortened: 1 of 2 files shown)


completions/openssl
```diff
@@ -42,22 +42,11 @@ _comp_cmd_openssl()
...
-        camellia-256-ecb cast cast-cbc cast5-cbc cast5-cfb cast5-ecb cast5-ofb
-        des des-cbc des-cfb des-ecb des-ede des-ede-cbc des-ede-cfb des-ede-ofb
-        des-ede3 des-ede3-cbc des-ede3-cfb des-ede3-ofb des-ofb des3 desx rc2
-        rc2-40-cbc rc2-64-cbc rc2-cbc rc2-cfb rc2-ecb rc2-ofb rc4 rc4-40
-        sha224 sha256 sha384 sha512 genpkey pkey pkeyparam pkeyutl'
+    local command formats
 
     if ((cword == 1)); then
+        local commands
+        commands="$($1 help 2>&1 | grep -v -e commands -e 'help:')"
```
scop, line 49:
```suggestion
        commands="$("$1" help 2>&1 | command grep -vF -e commands -e help:)"
```
...but I think it's also more customary to use sed for this, e.g. `sed -e '/commands\|help:/d'`

test/t/test_openssl.py
```diff
@@ -2,7 +2,7 @@
 
 
 class TestOpenssl:
-    @pytest.mark.complete("openssl ")
+    @pytest.mark.complete("openssl ", require_cmd=True)
     def test_1(self, completion):
         assert completion
```
scop, line 7: Let's assert that we have some known/expected completions here, to gain some confidence that the parsing actually works.

scop (changes requested):

completions/openssl
yedayak: Done, used sed

test/t/test_openssl.py
yedayak: Added some sub commands from each section in the help

scop (approved):
Thanks!

Issue #1272, opened 2024-11-26
Github errors parsing changelog in CHANGELOG.md and release 2.12.0


yedayak:
## Describe the bug
When looking at release [2.12.0](https://github.com/scop/bash-completion/releases/tag/2.12.0), there are errors rendering a few of the lines:
```
 Extra open brace or missing close brace 
```
![image](https://github.com/user-attachments/assets/e60be0b4-7549-48ec-8103-ac33b70c0bd8)

The issue is that github interprets anything between two dollar signs (`$`) [as a LaTeX expression](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions#writing-inline-expressions), parsed with MathJax. The 4 lines that create the error all have a single opening brace between two dollar signs, for example:
```
CHANGELOG.md:* **dict:** work around bash-4.3 ${v+"$@"} with custom IFS ([cc21298](https://github.com/scop/bash-completion/commit/cc212983a2b2523e5b62e9735c6dff5f011bb80a))
```
Just `${$` (unquoted) is enough to trigger this.
I will create a PR to quote these in the changelog. I don't know if editing the release itself will change the date on it?
Do you think it's worth creating some kind of lint to catch stuff like that or filing an issue with release please?

## To reproduce
Look at the bug fixes in [CHANGELOG.md](https://github.com/scop/bash-completion/blob/3b1586b1470ff066daa9b1f2e620d4ef86303eae/CHANGELOG.md#2120-2024-02-21) in the 2.12.0 section
## Expected behavior
No errors when rendering
## Versions (please complete the following information)
N/A

akinomyoga:
> I don't know if editing the release itself will change the date on it?

I don't think it will. I'm not familiar with it, but `release-please` seems to just insert a section on the release time.

> Do you think it's worth creating some kind of lint to catch stuff like that or filing an issue with release please?

We may need to add a rule for the commit message so that `$` is only allowed in ``` `...` ```.

We might file an issue in the upstream `release-please`, but I'm not sure how it should behave in the general case. For the `bash-completion` project, we still expect ``` `...` ``` in the commit messages to be treated as Markdown. Likewise, some projects may want to include LaTeX formulae in the commit messages, so we cannot simply escape all the special characters of Markdown. Maybe they can allow configuring the set of characters that should be escaped?

PR #1273, opened 2024-11-26
chore: Quote changelog lines with `$.*{.*$`

yedayak:
See https://github.com/scop/bash-completion/issues/1272


akinomyoga (approved):
There are still other occurrences of raw `$`, which might better be quoted too for consistency, but it is not a big problem.

The CI error is caused by the change in #1090, which slipped off from the check due to some setup changes.

akinomyoga:
I guess the CI error should be fixed first in a separate PR.

yedayak:
> I guess the CI error should be fixed first in a separate PR.

Yeah, made https://github.com/scop/bash-completion/pull/1274

akinomyoga:
Can you rebase it to rerun the CI?

yedayak:
> Can you rebase it to rerun the CI?

Done

scop:
No problem with merging this, but as a sustainable solution, I'd rather have release-please (or our use of it) fixed to prevent issues like this, if possible.

PR #1279, opened 2024-11-28
test: Remove /usr/bin/which installation

yedayak:
This was added when fedoradev still had it, in
537c9a0e9d59c699927a2eeb04d87c05a3638490, but that was dropped in 53feba31e9b49340605304258e18ff06baacaa22.

### DIFF
--- a/test/docker/centos7/Dockerfile
+++ b/test/docker/centos7/Dockerfile
@@ -9,8 +9,6 @@ RUN set -x \
         /usr/bin/autoconf \
         /usr/bin/automake \
         /usr/bin/make \
-        # /usr/bin/which: https://bugzilla.redhat.com/show_bug.cgi?id=1443357 \
-        /usr/bin/which \
         /usr/bin/xvfb-run \
         python36-pexpect
 


akinomyoga (approved):

test/docker/centos7/Dockerfile
```diff
@@ -9,8 +9,6 @@ RUN set -x \
         /usr/bin/autoconf \
         /usr/bin/automake \
         /usr/bin/make \
-        # /usr/bin/which: https://bugzilla.redhat.com/show_bug.cgi?id=1443357 \
-        /usr/bin/which \
```
scop, line 13: This does not seem correct to me.

The workaround was removed from the Fedora build because Fedora fixed their xvfb package (by adding the dependency on which, or refactoring it so it no longer needs which).

But the centos:7 xvfb-run still needs which:
```shellsession
$ docker run --rm ghcr.io/scop/bash-completion/test:centos7 grep -wF which /usr/bin/xvfb-run
if ! which xauth >/dev/null; then
```
...but it gets installed only as a side effect of some packages to be tested that we happen to install, i.e. it remains unfixed in the centos xvfb-run providing package and thus we should keep it to avoid things breaking if we no longer installed these for some reason:
```shellsession
$ docker run --rm ghcr.io/scop/bash-completion/test:centos7 rpm -e which
error: Failed dependencies:
	which is needed by (installed) gstreamer-tools-0.10.36-7.el7.x86_64
	which is needed by (installed) fuse-2.9.2-11.el7.x86_64
	which is needed by (installed) arts-8:1.5.10-26.el7.x86_64
	which is needed by (installed) xdg-utils-1.1.0-0.17.20120809git.el7.noarch
	which is needed by (installed) checksec-2.2.2-1.el7.noarch
	/usr/bin/which is needed by (installed) munin-node-2.0.75-1.el7.noarch
```

scop (changes requested):

test/docker/centos7/Dockerfile
yedayak: Yeah, it seems I didn't understand the original issue correctly. Thanks for the explanation!

PR #1280, opened 2024-11-28
Make pytest fail when tests marked as xfail pass

yedayak:
There are currently six tests that "xpass" (at least in a recent job I checked). All of them pass correctly locally for me as well.
This removes the specific xfail markers for the passing tests and sets `xfail_strict` to true so CI will fail if any other ones will succeed by accident.

More info about xfails vs xpasses: https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail

This is split into 3 commits for different parts of this, the first one just includes some cleanup of test_mkdir.py as well as xfail removal.

### DIFF
--- a/test/setup.cfg
+++ b/test/setup.cfg
@@ -3,6 +3,7 @@ minversion = 3.6
 markers =
     bashcomp
     complete
+xfail_strict=true
 
 [mypy]
 python_version = 3.6
--- a/test/t/test_alias.py
+++ b/test/t/test_alias.py
@@ -10,7 +10,6 @@ class TestAlias:
     def test_1(self, completion):
         assert completion == "bar foo".split()
 
-    @pytest.mark.xfail  # TODO: Would like this completion to work
     @pytest.mark.complete("alias foo=")
     def test_2(self, completion):
         assert completion == "foo='bar'"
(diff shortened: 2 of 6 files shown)


yedayak:
It seems the ifup and ifdown tests don't fail in centos7, since the `lo` interface is configured there. Will have to think what the best way to make this work is.
Also, I found that ifup doesn't complete interfaces in fedoradev since interface configuration is in `/etc/NetworkManager/system-connections/` now by default, will file another issue for that

test/setup.cfg
```diff
@@ -3,6 +3,7 @@ minversion = 3.6
 markers =
     bashcomp
     complete
+xfail_strict=true
```
scop, line 6: If we want this, I think we also want/need to go all things that can be xfails or xpasses, and mark them as such, _specifically on per distro basis_ (i.e. semantically for example like "expected to fail on CentOS 7").

We could also consider changing some xfails to skips, but to me the downside of that would be that we'd get no indication if something that used to fail has started to pass due to something (which xpass gives us). There might be some skips that we'd want to convert to xfails for this reason.

scop (changes requested):

PR #1285, opened 2024-12-01
test(realcommand): Don't use pytest.skipif

yedayak:
This function doesn't actually exist. Also added an error message, this really should never happen.

### DIFF
--- a/test/t/unit/test_unit_realcommand.py
+++ b/test/t/unit/test_unit_realcommand.py
@@ -43,11 +43,11 @@ class TestUnitRealCommand:
 
     def test_basename_nonexistent(self, bash, functions):
         filename = "non-existent-file-for-bash-completion-tests"
-        skipif = "! type -P %s" % filename
+        file_exists_check = "! type -P %s" % filename
         try:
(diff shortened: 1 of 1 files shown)


akinomyoga:
Also, the variable name `skipif` seems the opposite. If the command `! type -P non-existent-...` succeeds, the test is performed. When `skipif` is **not** satisfied, the test is skipped.

test/t/unit/test_unit_realcommand.py
```diff
@@ -47,7 +47,7 @@ def test_basename_nonexistent(self, bash, functions):
         try:
             assert_bash_exec(bash, skipif, want_output=None)
         except AssertionError:
-            pytest.skipif(skipif)
+            pytest.skip(f"Found command that shouldn't exist: '{skipif}'")
```
akinomyoga, line 50:
```suggestion
            pytest.skip(f"Found command that shouldn't exist: '{filename}'")
```

yedayak: done

scop (approved):

akinomyoga (approved):

Issue #1282, opened 2024-11-29
pkg_delete file missing from release 2.15.0 release tarball


rhelmot:
## Describe the bug

`completions/pkg_delete` is missing from the tarball. This causes build errors when built on BSD systems.

## To reproduce

Build and install bash-completion on e.g. FreeBSD from the released tarball

## Expected behavior

Installation completes successfully

## Versions (please complete the following information)

- [x] Operating system name/distribution and version: FreeBSD 14.0
- [x] bash version, `echo "$BASH_VERSION"`: 5.2.21(0)-release
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.15.0

## Additional context

I originally encountered this issue cross compiling for FreeBSD from a Linux host via nix, but it reproduces just fine when building native for FreeBSD without nix.

## Debug trace

```
[+] ~/bash-completion-2.15.0% ./configure                                                                                                                                                          audrey@chrysanthemum [09:02:15 PM]
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a race-free mkdir -p... mkdir -p
checking for gawk... no
checking for mawk... no
checking for nawk... nawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking how to create a pax tar archive... gnutar
checking whether ln -s works... yes
checking for a sed that does not truncate output... /usr/bin/sed
checking for pytest... no
checking for pytest-3... no
checking build system type... x86_64-unknown-freebsd14.0
checking host system type... x86_64-unknown-freebsd14.0
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: creating completions/Makefile
config.status: creating doc/Makefile
config.status: creating helpers/Makefile
config.status: creating test/Makefile
config.status: creating test/fallback/Makefile
config.status: creating test/fallback/completions/Makefile
config.status: creating test/t/Makefile
config.status: creating test/t/unit/Makefile
[+] ~/bash-completion-2.15.0% make                                                                                                                                                                 audrey@chrysanthemum [09:02:20 PM]
Making all in completions
make[1]: don't know how to make pkg_delete. Stop

make[1]: stopped in /home/audrey/bash-completion-2.15.0/completions
*** Error code 1

Stop.
make: stopped in /home/audrey/bash-completion-2.15.0
```

rhelmot:
[...]

PR #1287, opened 2024-12-03
build: Include OS specific completions in tarball

yedayak:
This should fix https://github.com/scop/bash-completion/issues/1282, although still doesn't deal with the issue of including tests for them or not.

I will love if people who actually know Makefile/autotools will look at this :)
I'm not really sure about the variables naming, open for suggestions.

### DIFF
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -799,6 +799,8 @@ CLEANFILES = \
 	ping4 \
 	ping6 \
 	_pip3 \
+	pkg_deinstall \
+	pkg_info \
 	pkgconf \
 	pkill \
 	plzip \
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -1,5 +1,4 @@
-bashcompdir = $(datadir)/$(PACKAGE)/completions
-bashcomp_DATA = 2to3 \
+cross_platform = 2to3 \
 		7z \
 		a2x \
 		abook \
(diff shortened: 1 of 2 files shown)


scop (approved):
IDK if I qualify as someone actually knowing these tools that much, but makes sense to me, assuming it works (did not test).

completions/Makefile.am
```diff
@@ -501,27 +500,36 @@ bashcomp_DATA = 2to3 \
...
+# We include all the files in the tarball
+EXTRA_DIST = $(bashcomp_DATA) \
+	     $(solaris_extra) \
+	     $(bsd_extra) \
+	     $(freebsd_extra)
 
-EXTRA_DIST = $(bashcomp_DATA)
+bashcompdir = $(datadir)/$(PACKAGE)/completions
+bashcomp_DATA = $(cross_platform)
+# We add only the completions that are relevant for the current platform we're installing on.
```
akinomyoga, line 523:
```suggestion
# We add only the completions that are relevant for the current platform we're
# installing on.
```

akinomyoga (approved):

completions/Makefile.am
yedayak: Changed

yedayak:
> IDK if I qualify as someone actually knowing these tools that much, but makes sense to me, assuming it works (did not test).

Tests I did:
- creating tarball and seeing it has all the completions
- Installing from that tarball and seeing it only installs linux completions (on a linux host (WSL))
- Modifying the configure.ac to make linux go to the bsd branch, and seeing it installs `pkg_delete`

I also modified the configure.ac in the PR, just to make it more consistent with how the variables are used.

completions/Makefile.am
```diff
@@ -501,27 +500,37 @@ bashcomp_DATA = 2to3 \
...
 		portinstall \
 		portsnap \
 		portupgrade
-endif
 
+# We include all the files in the tarball
+EXTRA_DIST = $(bashcomp_DATA) \
+	     $(solaris_extra) \
+	     $(bsd_extra) \
+	     $(freebsd_extra)
```
akinomyoga, line 519: It doesn't cause a problem, but this causes duplicates because `*_extra` is included in `bashcomp_DATA` depending on the platform.

```suggestion
EXTRA_DIST = $(cross_platform) \
	     $(solaris_extra) \
	     $(bsd_extra) \
	     $(freebsd_extra)
```

akinomyoga, line 519: It doesn't cause a problem, but this causes duplicates because `*_extra` is included in `bashcomp_DATA` depending on the platform.

```suggestion
EXTRA_DIST = $(cross_platform) \
	     $(solaris_extra) \
	     $(bsd_extra) \
	     $(freebsd_extra)
```

but there seems to be another way: If we use `dist_bashcomp_DATA` instead of `bashcomp_DATA`, the files in `dist_bashcomp_DATA` are automatically included in `dist`. In that case, we can only specify `*_extra` to `EXTRA_DIST`, but there are anyway duplicates with that approach.

akinomyoga (approved):

PR #1288, opened 2024-12-07
Docs: Update OSX docs

hahnandrew:
Parity change for what's on the [homebrew site](https://formulae.brew.sh/formula/bash-completion@2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ installed system wide. To do this:
 ### macOS (OS X)
 Homebrew will install to $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh.
 We can source it by adding the following to our login file `~/.bash_profile`.
-Note that by default MacOS does not automatically source `~/.bashrc`. 
+Note that by default MacOS does not automatically source `~/.bashrc` or `/etc/bashrc`. 
 
 ```shell
 if [[ -s $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh ]]; then
--- a/README.md
+++ b/README.md
@@ -61,9 +61,31 @@ installed system wide. To do this:
    programmable completion for other purposes.
 
 ### macOS (OS X)
+
+If you're using macOS (formerly OS X), `/etc/bashrc` is apparently not sourced
+at all, and `~/.bashrc` is not sourced from `~/.bash_profile` by default
+(because `~/.bash_profile` is not created by default).  In this case, the
+standard way is to configure `~/.bash_profile` to source `~/.bashrc` and write
+interactive settings in `~/.bashrc`.  You can source `~/.bashrc` in
+`~/.bash_profile` in the following way:
+
+```shell
+# ~/.bash_profile
+
+if [[ -f ~/.bashrc ]]; then
+  source ~/.bashrc
+fi
+```
+
+Then, you can source `bash-completion` in your `~/.bashrc`.  It should be noted
+that `bash-completion` should not be sourced in `~/.bash_profile` because
+`~/.bash_profile` is only sourced in interactive login shell sessions.  If you
+start nested Bash sessions, the interactive settings in `~/.bash_profile` will
+disappear.  It is strongly recommended to source `~/.bashrc` from
+`~/.bash_profile` and write interactive settings in `~/.bashrc`.
+
 Homebrew will install to $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh.
-We can source it by adding the following to our login file `~/.bash_profile`.
-Note that by default MacOS does not automatically source `~/.bashrc` or `/etc/bashrc`. 
+We can source it by adding the following to our startup file `~/.bashrc`.
 
 ```shell
 if [[ -s $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh ]]; then
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ start nested Bash sessions, the interactive settings in `~/.bash_profile` will
 disappear.  It is strongly recommended to source `~/.bashrc` from
 `~/.bash_profile` and write interactive settings in `~/.bashrc`.
 
-Homebrew will install to $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh.
+Homebrew will install to `$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh`.
 We can source it by adding the following to our startup file `~/.bashrc`.
 
 ```shell
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ fi
 
 Then, you can source `bash-completion` in your `~/.bashrc`.  It should be noted
 that `bash-completion` should not be sourced in `~/.bash_profile` because
-`~/.bash_profile` is only sourced in interactive login shell sessions.  If you
+`~/.bash_profile` is only sourced in interactive _login_ shell sessions.  If you
 start nested Bash sessions, the interactive settings in `~/.bash_profile` will
 disappear.  It is strongly recommended to source `~/.bashrc` from
 `~/.bash_profile` and write interactive settings in `~/.bashrc`.
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ We can source it by adding the following to our startup file `~/.bashrc`.
 
 ```shell
 if [[ -s $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh ]]; then
-    . "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh"
+  . "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh"
 fi
 ```
 
--- a/README.md
+++ b/README.md
@@ -61,14 +61,12 @@ installed system wide. To do this:
    programmable completion for other purposes.
 
 ### macOS (OS X)
-
-If you're using macOS (formerly OS X), `/etc/bashrc` is apparently not sourced at
-all. In that case, you can put the `bash_completion` file in `/sw/etc`
-and add the following code to `~/.bash_profile`:
+Homebrew will install to $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh.
+We can source it by adding the following to our login file `~/.bash_profile`
 
 ```shell
-if [ -f /sw/etc/bash_completion ]; then
-   . /sw/etc/bash_completion
+if [[ -r "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh" && -s "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh" ]]; then
+    . "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh"
 fi
 ```
 
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Homebrew will install to $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh.
 We can source it by adding the following to our login file `~/.bash_profile`
 
 ```shell
-if [[ -r "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh" && -s "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh" ]]; then
+if [[ -s $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh ]]; then
     . "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh"
 fi
 ```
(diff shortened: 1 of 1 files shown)


akinomyoga (changes requested):
We should also update the description above the code example.

I'm not a user of macOS, but what is the original `/sw` directory? A quick search [1] suggests that it is related to [the Fink project](https://www.finkproject.org/index.php), which seems to be still active.

- [1] [Is a folder titled "SW" part of the opera… - Apple Community](https://discussions.apple.com/thread/966038?sortBy=rank)

The suggested change also assumes Homebrew, which is another framework.

Then, we shouldn't remove the settings for Fink but should describe both cases. We should also clarify that the Fink users should use `/sw`, and the Homebrew users should use `$HOMEBREW_PREFIX`.

README.md
```diff
@@ -67,9 +67,7 @@ all. In that case, you can put the `bash_completion` file in `/sw/etc`
 and add the following code to `~/.bash_profile`:
 
 ```shell
-if [ -f /sw/etc/bash_completion ]; then
-   . /sw/etc/bash_completion
-fi
+[[ -r "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh" ]] && . "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh"
```
akinomyoga, line 70: - `if ... then ... fi` should be used. `[[ ... ]] && . ...` will leave `$?` being `1` for the first prompt when the file is not found.
- If you want to specifically check the read attribute by `-r` (instead of using `-f`), you may be further specific to request a non-empty file by `-s`.

hahnandrew: Added - whoever is maintaining the Homebrew site should update accordingly

hahnandrew:
> We should also update the description above the code example.

Added

> I'm not a user of macOS, but what is the original `/sw` directory? A quick search [1] suggests that it is related to [the Fink project](https://www.finkproject.org/index.php), which seems to be still active.
> 
> * [1] [Is a folder titled "SW" part of the opera… - Apple Community](https://discussions.apple.com/thread/966038?sortBy=rank)
> 
> The suggested change also assumes Homebrew, which is another framework.
> 
> Then, we shouldn't remove the settings for Fink but should describe both cases. We should also clarify that the Fink users should use `/sw`, and the Homebrew users should use `$HOMEBREW_PREFIX`.

The context is that fink's last supported MacOS version was released 2 years ago and doesn't have support for the last two M series chips, /sw is not a common convention nor exists by default. Homebrew is MacOS's de facto package manager.

FWIW I haven't heard of fink before coming across this repository

README.md
```diff
@@ -61,14 +61,12 @@ installed system wide. To do this:
...
-If you're using macOS (formerly OS X), `/etc/bashrc` is apparently not sourced at
-all. In that case, you can put the `bash_completion` file in `/sw/etc`
-and add the following code to `~/.bash_profile`:
+Homebrew will install to $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh.
+We can source it by adding the following to our login file `~/.bash_profile`
 
 ```shell
-if [ -f /sw/etc/bash_completion ]; then
-   . /sw/etc/bash_completion
+if [[ -r "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh" && -s "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh" ]]; then
```
akinomyoga, line 68:
```suggestion
if [[ -s $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh ]]; then
```

`-s` implies `-r`.

akinomyoga:
> The context is that fink's last supported MacOS version was released 2 years ago and doesn't have support for the last two M series chips,

The *release* doesn't appear, but they explain at the top of their official page how to install it from GitHub in the recent macOS on x86_64 and M1/M2. The latest commit to [the GitHub repository](https://github.com/fink/fink) was pushed last week. This means that there are still users, even though the number might be smaller.

> /sw is not a common convention nor exists by default.

Yeah, that seems a problem because the original description didn't mention the Fink project. 

> Homebrew is MacOS's de facto package manager.
> 
> FWIW I haven't heard of fink before coming across this repository

But then, why do we need to remove the description for the Fink users?

hahnandrew:
> But then, why do we need to remove the description for the Fink users?

M4 is the current chip. Fink is niche and likely would fit better in an FAQ or wiki - for scale, Fink has 0.6% of Github stars that Homebrew has, which isn't to downplay the engineering efforts behind Fink, but just to show the relative userbases. There are tons of package managers out there, and it'd be impractical to include every single one in the landing Readme

hahnandrew:
> But then, why do we need to remove the description for the Fink users?

M4 is the current chip. Fink is niche and likely would fit better in an FAQ or wiki - for scale, Fink has 0.6% of Github stars that Homebrew has, which isn't to downplay the engineering efforts behind Fink, but to show the relative userbases. There are tons of package managers out there, and it'd be impractical to include every single one in the landing Readme

hahnandrew:
> But then, why do we need to remove the description for the Fink users?

M4 is the current chip. Fink is niche and likely would fit better in an FAQ or wiki - for scale, Fink has 0.6% of Github stars that Homebrew has, which isn't to downplay the engineering efforts behind Fink, but to show the relative userbases. There are tons of package managers out there, and it'd be impractical to include every single one in the landing Readme, and increases the surface area for maintenance

akinomyoga:
> M4 is the current chip. Fink is niche and likely would fit better in an FAQ or wiki - for scale, Fink has 0.6% of Github stars that Homebrew has, which isn't to downplay the engineering efforts behind Fink, but to show the relative userbases.

Are M1 and M2 already marked as end of support by Apple? Also, the relative size of the user base is not important. As far as there are users, we continue to support them.

> There are tons of package managers out there, and it'd be impractical to include every single one in the landing Readme, and increases the surface area for maintenance

That makes sense, though I doubt if there are really *tons* of package managers. I searched for the keyword "macos package manager list" in Google, and the first link of the search result for macOS [1] includes Fink as one of 19 package managers.

- [Top Package Managers for Mac in 2024](https://slashdot.org/software/package-managers/mac/)

I also checked who've introduced the original description about `/sw/etc`. This seems to have been introduced in 7a4383333814e48bd039b5490cecaad3df9ebae1 (2003-08-04), which is probably outdated. [The `bash-completion` package](https://github.com/fink/fink-distributions/blob/master/10.9-libcxx/stable/main/finkinfo/shells/bash-completion.info) distributed by the Fink project also has an explanation, though I'm not sure if we can reference it when we install `bash-completion` manually.

README.md
```diff
@@ -61,14 +61,12 @@ installed system wide. To do this:
    programmable completion for other purposes.
 
 ### macOS (OS X)
-
-If you're using macOS (formerly OS X), `/etc/bashrc` is apparently not sourced at
```
akinomyoga, line 65: I have again looked at the "Installation" section of README. The reason that we had a separate section for macOS seems to be because `/etc/bashrc` wasn't sourced then, but is it still the case? If yes, I think we should still want this sentence.

hahnandrew: Added

akinomyoga: `/etc/bashrc` and `~/.bashrc` are different files, but yeah, `~/.bashrc` is not configured in `~/.bash_profile` probably because they switched to Zsh and doesn't provide `~/.bash_profile` by default.

akinomyoga: I'll add a description.

hahnandrew: The convention is `/etc/bashrc` is system wide and `~/.bashrc` is per-user. Neither are sourced by default by MacOS, nor is it common to edit the systemwide login file either.

akinomyoga: I added 3d50f47179ed3aaf3aa20f85ead9eda5591b80c0. Actually, `~/.bash_profile` is not really a proper file to write interactive settings. I originally thought of discussing this in a separate issue, but the present discussion of `/etc/bashrc` and `~/.bashrc` is closely related to this, so I decided to include it in this PR.

There are additional style fixes d4c5fea5418a9729dcdb49fd1907635ad699ca03 and 0a11552a757a350013dc21bbadeff51bcbe99daf. I'll squash these small fixes later.

akinomyoga: > nor is it common to edit the systemwide login file either.

Right, but `bash-completion`'s README is not written specifically for normal users. This README also targets package maintainers of different systems.

akinomyoga: I added another commit 78ae7f6a689fed77d1c7bf35ea50f3f95669c10b to clarify that the user needs to install the `bash-completion` package using the Homebrew. To the users unfamiliar with it, the current wording appears to imply that Homebrew is provided by macOS by default, and the `bash-completion` package is also installed by default.

Actually, the reason that we had a special section for macOS was that `/etc/bashrc` and `~/.bashrc` didn't work properly. For the other environments where `/etc/bashrc` (or a similar file) and `~/.bashrc` work, we haven't described the settings for specific package managers. In that sense, it might be a bit strange that we describe the setting for Homebrew. The setting for Homebrew's `bash-completion` package seems to be anyway presented to the user on the installation [1], so it is technically unnecessary.

- [1] https://github.com/Homebrew/homebrew-core/blob/5df158e314883cda5ff0f9c258d289af64c712db/Formula/b/bash-completion%402.rb#L49-L54

Nevertheless, I think it would be useful for macOS users to provide the Homebrew setting as an example.

akinomyoga (approved):
Here's the summary of the changes. It is probably better to leave them in the commit message:

The description for macOS in the `master` branch was added in commit 7a4383333814e48bd039b5490cecaad3df9ebae1 21 years ago. The path `/sw/etc` seems to be related to the Fink project, but this does not seem to be common today. In this patch,

- We remove the (seemingly) outdated description about `/sw/etc`.
- We describe the proper setup of `~/.bash_profile` and `~/.bashrc` in macOS.
- We add an example setup with Homebrew's `bash-completion` package.

README.md
```diff
@@ -62,13 +62,36 @@ installed system wide. To do this:
...
+# ~/.bash_profile
+
+if [[ -f ~/.bashrc ]]; then
+  source ~/.bashrc
+fi
+```
+
+Then, you can source `bash-completion` in your `~/.bashrc`.  It should be noted
+that `bash-completion` should not be sourced in `~/.bash_profile` because
+`~/.bash_profile` is only sourced in interactive login shell sessions.  If you
```
scop, line 82:
```suggestion
`~/.bash_profile` is only sourced in interactive _login_ shell sessions.  If you
```

README.md
```diff
@@ -62,13 +62,36 @@ installed system wide. To do this:
...
+start nested Bash sessions, the interactive settings in `~/.bash_profile` will
+disappear.  It is strongly recommended to source `~/.bashrc` from
+`~/.bash_profile` and write interactive settings in `~/.bashrc`.
+
+For example, if you install `bash-completion` using Homebrew, it will install
+the entry point of `bash-completion` to
+`$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh`.  We can source it by
+adding the following to our startup file `~/.bashrc`:
```
scop, line 90: Doesn't homebrew do enablement of bash-completion by default out of the box?

If not, do they have instructions of their own how to do it?

If they do, I'd rather not ship specific documentation of our own (if it's redundant) but rather point to theirs, and if it's not redundant, I feel it would be good to somehow explicitly acknowledge that the package does come with docs of its own, and we recommend the user to look into that as well (or recommend against something in it, whichever way it is).

akinomyoga: Actually, this is [explained in Homebrew's `bash-completion`](https://github.com/Homebrew/homebrew-core/blob/5df158e314883cda5ff0f9c258d289af64c712db/Formula/b/bash-completion%402.rb#L49-L54). I think this string will be presented to the user on the installation time. On the other hand, they don't describe it on [its web page](https://formulae.brew.sh/formula/bash-completion).

One non-trivial thing is that the `.sh` files in `/etc/profile.d` are intended to be automatically sourced by a system-wide shell configuration. Then, one would naively expect Homebrew to configure `/etc/profile.d/*.sh` to be automatically sourced so that a user can simply source a Homebrew shell configuration file (e.g. at `$HOMEBREW_PREFIX/etc/bashrc` or `$HOMEBREW_PREFIX/etc/profile`). I don't have macOS, but my guess is that Homebrew doesn't offer such a configuration, so a user needs to manually source each file in `$HOMEBREW_PREFIX/ext/profile.d`.

README.md
```diff
@@ -62,13 +62,36 @@ installed system wide. To do this:
...
+# ~/.bash_profile
+
+if [[ -f ~/.bashrc ]]; then
+  source ~/.bashrc
+fi
+```
+
+Then, you can source `bash-completion` in your `~/.bashrc`.  It should be noted
+that `bash-completion` should not be sourced in `~/.bash_profile` because
+`~/.bash_profile` is only sourced in interactive _login_ shell sessions.  If you
```
akinomyoga, line 82:
```suggestion
`~/.bash_profile` is only loaded in interactive _login_ shell sessions.  If you
```

README.md
hahnandrew: > so a user needs to manually source each file in $HOMEBREW_PREFIX/ext/profile.d.

This is the case.

Issue #1260, opened 2024-11-07
Error occurred when completing the "umount" command


YoruStar:
## Describe the bug

When completing the `umount o` command, the following error occurred:
```
[root@localhost ~]# umount obash_completion: _comp_compgen: -C: invalid directory name `'
verlay
```

## To reproduce

The problematic version is `bash-completion-2.12`
```
$ mkdir overlay
$ mount -t tmpfs tmpfs overlay/
$ umount o[TAB]bash_completion: _comp_compgen: -C: invalid directory name `'
verlay
```

## Expected behavior
```
$ mkdir overlay
$ mount -t tmpfs tmpfs overlay/
$ umount o[TAB]verlay
```

## Versions (please complete the following information)
```
$ echo "$BASH_VERSION"
5.2.15(1)-release
$ IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}"
2.12.0
```

## Additional context

The issue was introduced after [feat(_comp_compgen): support option -C](https://github.com/scop/bash-completion/commit/6b3dfa529950f3bc51b8ead9323a82f65c5cfeb9)

## Debug trace
```
# Add relative paths to COMPREPLY
if [[ $cur && $cur != /* ]]; then
    local realcur
    [[ $cur == */ ]] && # don't let readlink drop last / from path
        realcur="$(readlink -f "$cur." 2>/dev/null)/" ||
        realcur=$(readlink -f "$cur" 2>/dev/null)
    if [[ $realcur ]]; then
        local dirrealcur="" dircur="" basecur
        if [[ $cur == */* ]]; then
            dirrealcur="${realcur%/*}/"
            dircur="${cur%/*}/"
        fi
        basecur=${cur#"$dircur"}
        local i
        for i in ${!COMPREPLY[*]}; do
            [[ ${COMPREPLY[i]} == "$realcur"* ]] &&
                _comp_compgen -aC "$dircur" -c "$basecur" -- \
                    -f -d -P "$dircur" -X "!${COMPREPLY[i]##"$dirrealcur"}"
        done
    fi
fi
```
In the above code, when executing the `_comp_compgen -aC '$dircur' -c '$basecur'` command, the value of the `dircur` variable is empty. This causes `_comp_compgen` to consider it an invalid argument, resulting in an error.
```
C)
    if [[ ! $OPTARG ]]; then
        printf 'bash_completion: %s: -C: invalid directory name `%s'\''\n' "$FUNCNAME" "$OPTARG" >&2
[...]

PR #1290, opened 2024-12-12
comp_compgen: option -C no longer checks if OPTARG is empty

YoruStar:
The `-C` option in _comp_compgen, similar to the `-P` option in bash's compgen, should support passing an empty string, for example: `compgen -P ""-f`

fix https://github.com/scop/bash-completion/issues/1260

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -594,13 +594,7 @@ _comp_compgen()
                 ;;
             c) _cur=$OPTARG ;;
             R) _cur="" ;;
-            C)
-                if [[ ! $OPTARG ]]; then
-                    printf 'bash_completion: %s: -C: invalid directory name `%s'\''\n' "$FUNCNAME" "$OPTARG" >&2
-                    return 2
-                fi
-                _dir=$OPTARG
-                ;;
+            C) _dir=$OPTARG ;;
             l) _has_ifs=set _ifs=$'\n' ;;
             F) _has_ifs=set _ifs=$OPTARG ;;
             [ix])


akinomyoga:
I believe this should be fixed at the side of `completions/_umount.linux`. It is in general suspicious to specify an invalid directory name so I don't think it is a good idea to suppress unconditionally. The directory name should be checked at the caller side. In the case of `_umount.linux`, it seems that the error still had occurred before 6b3dfa529950f3bc51b8ead9323a82f65c5cfeb9, but the message was redirected to `/dev/null` as `cd` is performed as `command cd -- "$dircur" 2>/dev/null`. The caller should ensure the validity of the path. In the present case, the caller can be updated to `_comp_compgen -aC "${dircur:--}" ...`

YoruStar:
> I believe this should be fixed at the side of `completions/_umount.linux`. It is in general suspicious to specify an invalid directory name so I don't think it is a good idea to suppress the error message unconditionally. The directory name should be checked at the caller side. In the case of `_umount.linux`, it seems that the error still had occurred before [6b3dfa5](https://github.com/scop/bash-completion/commit/6b3dfa529950f3bc51b8ead9323a82f65c5cfeb9), but the message was redirected to `/dev/null` as `cd` is performed as `command cd -- "$dircur" 2>/dev/null`. The caller should ensure the validity of the path. In the present case, the caller can be updated to `_comp_compgen -aC "${dircur:--}"`

I tried command cd -- "" and it did not fail.
```
$ command cd -- ""
$ echo $?
0
```
In the changes at https://github.com/scop/bash-completion/commit/6b3dfa529950f3bc51b8ead9323a82f65c5cfeb9, there are many modifications besides `_umount.linux`, and these may also potentially have the same issue.

Moreover, in the` _comp_compgen__call_generator` function, there is already a check to determine if the `$_dir` variable is empty, making the logic for checking if `$OPTARG` is empty in the `-C` option somewhat redundant.

Of course, aside from the umount command, I haven't found similar issues elsewhere, so modifying `_umount.linux` alone is also feasible.

akinomyoga:
> I tried `command cd -- ""` and it did not fail.
> 
> ```
> $ command cd -- ""
> $ echo $?
> 0
> ```

OK, thanks.

Nevertheless, I don't think it would be right just to suppress the error message when one sees an error message.

> In the changes at [6b3dfa5](https://github.com/scop/bash-completion/commit/6b3dfa529950f3bc51b8ead9323a82f65c5cfeb9), there are many modifications besides `_umount.linux`, and these may also potentially have the same issue.

Yes, they should also be fixed if they would pass an invalid directory. Before my first reply, I actually had checked it after I saw your PR. The cases for `feh` and `sbopkg` should also be fixed properly. The other cases would probably be fine, but we need to double-check them.

> Moreover, in the` _comp_compgen__call_generator` function, there is already a check to determine if the `$_dir` variable is empty,

No, it's used to check whether the `-C` option is specified or not. In the current `master`, exactly because we reject an empty directory, we can use `[[ $_dir ]]` to check whether the `-C` option is specified. If you break the assumption, it doesn't work as expected, so one needs to use another variable to store whether the `-C` option is specified.

> making the logic for checking if `$OPTARG` is empty in the `-C` option somewhat redundant.

This is false. It's the opposite. It relies on the fact that the empty directory for `-C` is rejected.

> Of course, aside from the umount command, I haven't found similar issues elsewhere, so modifying `_umount.linux` alone is also feasible.

Could you also look at `feh` and `sbopkg`? I haven't looked at them closely, but they might also need modifications. If you wouldn't have time, I'll later look at them someday.

akinomyoga:
Resolved by #1297

PR #1300, opened 2025-01-02
feat(nload): new completion

scop:
https://www.roland-riegel.de/nload/

More or less dormant project, does not look like there's much chance to get changes applied in timely manner, so submitting here.


akinomyoga (approved):

akinomyoga:
CI seems to complain about TAB in indentation.

akinomyoga (approved):

PR #1303, opened 2025-01-05
Add missing dependency on fixtures to help parallel testing

inconstante:
This might help with #430...

In my case, test/t/unit/test_unit_command_offset.py's test_cmd_quoted is the only one still failing

Original commit message follows:

-- 8< --
When running the test suite with xdist and 'pytest -n <jobs>', several tests fail. This happens because, in these tests, the definition of the testing function, usually in a method called 'functions' might happen after the execution of the tests themselves, i.e. in the methods whose names start with 'test_'.

This patch adds 'autouse=True' to these test-function-defining methods, so that they are executed before the tests.

### DIFF
--- a/test/t/unit/test_unit_count_args.py
+++ b/test/t/unit/test_unit_count_args.py
@@ -76,38 +76,38 @@ class TestUnitCountArgs(TestUnitBase):
         )
         assert output == "3"
 
-    def test_10_single_hyphen_1(self, bash):
+    def test_10_single_hyphen_1(self, bash, functions):
         """- should be counted as an argument representing stdout/stdin"""
         output = self._test(bash, "(a -b - c -d e)", 5, "a -b - c -d e", 12)
         assert output == "3"
 
-    def test_10_single_hyphen_2(self, bash):
+    def test_10_single_hyphen_2(self, bash, functions):
         """- in an option argument should be skipped"""
         output = self._test(
             bash, "(a -b - c - e)", 5, "a -b - c - e", 11, arg='-a "-b"'
         )
         assert output == "3"
 
-    def test_11_double_hyphen_1(self, bash):
+    def test_11_double_hyphen_1(self, bash, functions):
         """all the words after -- should be counted"""
         output = self._test(
             bash, "(a -b -- -c -d e)", 5, "a -b -- -c -d e", 14
         )
         assert output == "3"
 
-    def test_11_double_hyphen_2(self, bash):
+    def test_11_double_hyphen_2(self, bash, functions):
         """all the words after -- should be counted"""
         output = self._test(bash, "(a b -- -c -d e)", 5, "a b -- -c -d e", 13)
         assert output == "4"
 
-    def test_12_exclude_optarg_1(self, bash):
+    def test_12_exclude_optarg_1(self, bash, functions):
         """an option argument should be skipped even if it matches the argument pattern"""
         output = self._test(
             bash, "(a -o -x b c)", 4, "a -o -x b c", 10, arg='-a "-o" -i "-x"'
         )
         assert output == "2"
 
-    def test_12_exclude_optarg_2(self, bash):
+    def test_12_exclude_optarg_2(self, bash, functions):
         """an option argument should be skipped even if it matches the argument pattern"""
         output = self._test(
             bash,
@@ -119,7 +119,7 @@ class TestUnitCountArgs(TestUnitBase):
         )
         assert output == "2"
 
-    def test_12_exclude_optarg_3(self, bash):
+    def test_12_exclude_optarg_3(self, bash, functions):
         """an option argument should be skipped even if it matches the argument pattern"""
         output = self._test(
             bash,
@@ -131,21 +131,21 @@ class TestUnitCountArgs(TestUnitBase):
         )
         assert output == "1"
 
-    def test_13_plus_option_optarg(self, bash):
+    def test_13_plus_option_optarg(self, bash, functions):
         """When +o is specified to be an option taking an option argument, it should not be counted as an argument"""
         output = self._test(
             bash, "(a +o b c)", 3, "a +o b c", 7, arg='-a "+o"'
         )
         assert output == "1"
 
-    def test_14_no_optarg_chain_1(self, bash):
+    def test_14_no_optarg_chain_1(self, bash, functions):
         """an option argument should not take another option argument"""
         output = self._test(
             bash, "(a -o -o -o -o c)", 5, "a -o -o -o -o c", 14, arg='-a "-o"'
         )
         assert output == "1"
 
-    def test_14_no_optarg_chain_2(self, bash):
+    def test_14_no_optarg_chain_2(self, bash, functions):
         """an option argument should not take another option argument"""
         output = self._test(
             bash,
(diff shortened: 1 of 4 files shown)


test/t/unit/test_unit_count_args.py
akinomyoga, line 1: If we specify `autouse=True`, I guess `, functions` in each test (such as `def test_2(self, bash, functions):`) should be removed.

akinomyoga (commented):
Thank you for taking a look at this issue. The fix is good, but this seems to break consistency in the codebase. We specify the fixture to each test in some places, while this introduces `autouse` in a few places.

- *Option 1--Use autouse everywhere*: If we use `autouse`, shouldn't we update all the existing use of the fixtures?
- *Option 2--Use autouse under some conditions*: Or is there a criterion whether to use `autouse` or not (such as we would use `autouse` only when all tests in the class require the fixture)?
- *Option 3--Manually specify fixtures one by one*: If we do not touch existing ones, I think, for consistency, we should specify the fixture in the arguments of each test requiring the fixture.

I currently don't have a strong opinion about which approach to choose, but the change should at least be consistent with one rule. I'm interested in whether other maintainers @scop and @yedayak have some thoughts.

yedayak:
I'm not sure how this helps with #430. aren't fixtures supposed to always run before the tests that rely on them? I don't see anything in the [documentation](https://docs.pytest.org/en/stable/how-to/fixtures.html#autouse-fixtures-fixtures-you-don-t-have-to-request) saying this affects ordering. How did you test this fix?

inconstante:
@yedayak wrote:

> ... aren't fixtures supposed to always run before the tests that rely on them? I don't see anything in the [documentation](https://docs.pytest.org/en/stable/how-to/fixtures.html#autouse-fixtures-fixtures-you-don-t-have-to-request) saying this affects ordering.

The documentation you pointed to says that:

_In this example, the append_first fixture is an autouse fixture. Because it happens automatically, both tests are affected by it, **even though neither test requested it**. That doesn’t mean they can’t be requested though; just that it isn’t necessary._

That's exactly what's going on here... None of the tests I fixed requested the fixture, but they depend on it.

> I'm not sure how this helps with #430. [...] How did you test this fix?

I ran the test with xdist and `pytest -n <jobs>`. Here are more detailed steps to reproduce:

    Steps to reproduce:
    
    1. Make sure pytest-xdist is installed. On Debian systems this can be
       verified with the following command:
    
         dpkg --list python3-pytest-xdist
    
    2. Build and install bash-completion locally, for example with the
       following commands:
    
         autoreconf -f -i
         ./configure --prefix=$PWD/install/
         make install
    
    3. Run the test suite with a few parallel jobs, such as with:
    
         BASH_COMPLETION_TEST_BASH_COMPLETION=$PWD/install/share/bash-completion/bash_completion \
           pytest \
             -n 8 \
             test/t/unit/test_unit_count_args.py \
             test/t/unit/test_unit_dequote.py \
             test/t/unit/test_unit_get_first_arg.py \
             test/t/unit/test_unit_quote.py
    
Before this patch, these tests fail with messages similar to:
    
      FAILED test/t/unit/test_unit_quote.py::TestUnitQuote::test_3 - AssertionError: Error running "__tester "  a "": exit status=127, output="
    
After this patch, all these tests, which previously failed, pass.

I'll push a new version of the patch with these instructions in the commit message (as well as with the changes requested by @akinomyoga

inconstante:
@akinomyoga wrote:

> The fix is good, but this seems to break consistency in the codebase.

Indeed, thanks for pointing it out.

>     * _Option 3--Manually specify fixtures one by one_: If we do not touch existing ones, I think, for consistency, we should specify the fixture in the arguments of each test requiring the fixture.

Since I was just hoping to fix the issues, I just submitted a new version of the patch that implements option 3, thus removing the inconsistency in the codebase.

> I currently don't have a strong opinion about which approach to choose, but the change should at least be consistent under one rule. I'm interested in whether other maintainers @scop and @yedayak have some thoughts.

I also don't have a strong opinion. Anyhow, I probably prefer option 1, since it is easier to maintain. I can volunteer to change it everywhere in the codebase if that's what you, maintainers, prefer.

Please consider integrating my current change anyway, since it already helps with #430.

Thanks for you reviews, @yedayak and @akinomyoga.

akinomyoga:
> https://github.com/scop/bash-completion/pull/1303#issuecomment-2571729567 by @yedayak 
>
> I'm not sure how this helps with #430. aren't fixtures supposed to always run before the tests that rely on them?

Ah, yeah. Somehow, no one had described the issue. The issue is that some of our test cases don't specify the dependencies properly even though they actually depend on the fixture. I think it is clearer if you look at the new version of the PR changes (with Option 3): 8f0cceca397290e6a7069562dea8dacad160672d.

So I think #430 isn't the problem of the framework, but it's just bugs of specific test cases missing the necessary dependency specification.

> Since I was just hoping to fix the issues, I just submitted a new version of the patch that implements option 3, thus removing the inconsistency in the codebase.

Thank you. Yeah, that would be the minimal modifications that are consistent.

> > I currently don't have a strong opinion about which approach to choose, but the change should at least be consistent under one rule. I'm interested in whether other maintainers @scop and @yedayak have some thoughts.
> 
> I also don't have a strong opinion. Anyhow, I probably prefer option 1, since it is easier to maintain. I can volunteer to change it everywhere in the codebase if that's what you, maintainers, prefer.

If the discussion would turn out to take time, we can first merge the PR with option 3 as an easier fix, and then continue the discussion.

akinomyoga (commented):
CI complains about the codes in the commit message. Two lines don't fit within 80 columns. For the `BASH_COMPLETION_TEST_BASH_COMPLETION=`, you could replace the path with `/path/to/bash_completion` or something. For the line of the error message, you can refold the line in column 80.

inconstante:
> CI complains about the codes in the commit message. Two lines don't fit within 80 columns. For the `BASH_COMPLETION_TEST_BASH_COMPLETION=`, you could replace the path with `/path/to/bash_completion` or something. For the line of the error message, you can refold the line in column 80.

Commit message updated in 1fc9d78

inconstante:
New commit, 8c3efb6, to fix commit message title error flagged by the CI.

inconstante:
I'm getting this from the CI:

    perlcritic...............................................................Failed
    - hook id: perlcritic
    - exit code: 1
    
    Util.c: loadable library and perl binaries are mismatched (got first handshake key 0xeb00080, needed 0xf380080)

Is this a known problem? It doesn't look like it's related to my changes.

test/t/unit/test_unit_count_args.py
```diff
@@ -18,7 +18,7 @@ def functions(self, bash):
     def _test(self, *args, **kwargs):
         return self._test_unit("_comp__test_unit %s", *args, **kwargs)
 
-    def test_1(self, bash):
+    def test_1(self, bash, functions):
```
yedayak, line 21: I don't think this is needed here - this doesn't rely on _comp__test_unit

inconstante: Thanks for catching this. I just pushed a revised version.

inconstante:
@yedayak, @akinomyoga,

Is there anything I could do about the failing test?

yedayak (approved):

yedayak:
> @yedayak, @akinomyoga,
> 
> Is there anything I could do about the failing test?

I created #1307 that seems to fix the issue, but I don't think this should stop this PR from merging.

yedayak:
Can you rebase on main? pre-commit should pass now

akinomyoga (approved):

PR #1307, opened 2025-01-08
chore(perlcritic): Update perlcritic

yedayak:
pre-commit CI runs started failing on the perlcritic step, so update to the last version of the perlcritic pre-commit hook, corresponding to perlcritic v1.156

It will be preferable to make renovate update this, but I'm not sure why it doesn't do it. Any thoughts @scop @akinomyoga ?


scop:
No ideas re renovate offhand, but this seems fine, thanks.

PR #1313, opened 2025-01-12
feat(aptitude): add showsrc and source options

inconstante:
This fix has been provided to Debian (downstream) by Heinrich Schuchardt <heinrich.schuchardt@canonical.com>. It is currently a downstream patch, but it should be here.

### DIFF
--- a/completions/aptitude
+++ b/completions/aptitude
@@ -94,11 +94,11 @@ _comp_cmd_aptitude()
         _comp_compgen -- -W '"${COMPREPLY[@]}"' \
             ${exclude_flags:+-X "-[$exclude_flags]"}
     else
-        _comp_compgen -- -W 'update upgrade safe-upgrade forget-new clean
-            autoclean install reinstall remove hold unhold purge markauto
-            unmarkauto why why-not dist-upgrade full-upgrade download search
-            show forbid-version changelog keep keep-all build-dep add-user-tag
-            remove-user-tag versions'
+        _comp_compgen -- -W 'add-user-tag autoclean build-depends changelog
+            clean download extract-cache-subset forbid-version forget-new
+            full-upgrade help hold install keep keep-all markauto purge
+            reinstall remove remove-user-tag safe-upgrade search show showsrc
+            source unhold unmarkauto update versions why why-not'
     fi
 
 } &&


akinomyoga:
When was the original patch created? There seem to be even more subcommands in some manual:

* https://www.debian.org/doc/manuals/aptitude/rn01re01.en.html

  add-user-tag autoclean build-dep **build-depends** changelog clean **download** **extract-cache-subset** forbid-version forget-new full-upgrade **help** hold install keep-all markauto purge reinstall remove remove-user-tag safe-upgrade search show showsrc source unhold unmarkauto update versions why why-not

Also, `upgrade` and `dist-upgrade` seem to have been renamed to `full-upgrade`, and the old names were deprecated.

inconstante:
> When was the original patch created?

More than 2 years ago (https://salsa.debian.org/debian/bash-completion/-/merge_requests/5/diffs?commit_id=e68fde20600326e426e802233e46c940cab9478b)

> There seem to be even more subcommands in some manual:
> 
>     * https://www.debian.org/doc/manuals/aptitude/rn01re01.en.html
>       add-user-tag autoclean build-dep **build-depends** changelog clean **download** **extract-cache-subset** forbid-version forget-new full-upgrade **help** hold install keep-all markauto purge reinstall remove remove-user-tag safe-upgrade search show showsrc source unhold unmarkauto update versions why why-not
> 
> 
> Also, `upgrade` and `dist-upgrade` seem to have been renamed to `full-upgrade`, and the old names were deprecated.

You're right. Let me review this and write a more comprehensive patch.

Thanks for your review.

inconstante:
I pushed a revised version with the sub-commands listed by the manpage.

I used the following one-liner:

```
man aptitude | grep -E "^ {7}[a-z]" | tail -n 26 | head -n 24 | tr ',' '\n' | tr -d ' ' | sort | tr '\n' ' '
```

which found another sub-command: `keep`

**The list is now sorted alphabetically.**

PS: I think `download` has always been there.

akinomyoga (approved):

completions/aptitude
```diff
@@ -94,11 +94,11 @@ _comp_cmd_aptitude()
         _comp_compgen -- -W '"${COMPREPLY[@]}"' \
             ${exclude_flags:+-X "-[$exclude_flags]"}
     else
-        _comp_compgen -- -W 'update upgrade safe-upgrade forget-new clean
-            autoclean install reinstall remove hold unhold purge markauto
-            unmarkauto why why-not dist-upgrade full-upgrade download search
-            show forbid-version changelog keep keep-all build-dep add-user-tag
-            remove-user-tag versions'
+        _comp_compgen -- -W 'add-user-tag autoclean build-dep build-depends
```
scop, line 97: I don't know if it makes sense to offer both `build-dep` and `build-depends`, they seem to be aliases. We used to have the former, but maybe the latter is enough?

inconstante: They are listed together in the manpage, and there are no mentions to differences between them. Likewise for the online manual mentioned by @akinomyoga. It seems that they are aliases, indeed.

Also this: https://github.com/SamB/aptitude/blob/cb9a4e8776084d636c5ddab7aa30d1f7c064d5b5/src/cmdline/cmdline_do_action.cc#L153-L155

I don't have a preference for either, or for having only one or both. And I don't know how to choose. `aptitude help` shows `build-dep`, so there's some sense in keeping it.

scop: I think we should not have `build-dep` only, because with that, `aptitude build-depe<TAB>` will fail. Let's go with `build-depends` only -- for example `aptitude b<TAB>` will Just Work with it, without the useless prompt to choose between the two (if we had both).

akinomyoga: I haven't been so careful to check the aliases. Let's keep only `build-depends`, the longer and more descriptive one, as instructed in `CONTRIBUTING.md`.

https://github.com/scop/bash-completion/blob/0a13be1f6576ee4977552379e11157073b4e841b/CONTRIBUTING.md?plain=1#L136-L139

inconstante:
Pull request updated. Thanks for your reviews! :)

inconstante:
Hi, is there anything preventing the merging of this patch?

scop:
> Hi, is there anything preventing the merging of this patch?

Time. But done now, thanks!

Issue #1315, opened 2025-01-15
Support IPv6 when completing wget/curl


SoniEx2:
## Describe the feature/solution

We would like wget and/or curl to support IPv6 when completing --bind-address (wget) or --interface (curl), this is very useful for use with IPv6, where a machine generally has multiple IPv6 at the same time.

## Maintenance (please complete the following information)

- [ ] This is a request for new completion
- [ ] Link to upstream project query about shipping the completion:

yedayak:
Created PR #1316 for this, can it try the fix and see if it works?

PR #1316, opened 2025-01-15
curl, wget: Complete IP addresses, including ipv6 for --interface and --bind-address

yedayak:
Closes https://github.com/scop/bash-completion/issues/1315

### DIFF
--- a/completions/curl
+++ b/completions/curl
@@ -101,6 +101,7 @@ _comp_cmd_curl()
             ;;
         --interface)
             _comp_compgen_available_interfaces -a
+            _comp_compgen -a ip_addresses -a
             return
             ;;
         --help | -${noargopts}h)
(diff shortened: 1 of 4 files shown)


SoniEx2:
looks good! tho there is an issue with the wget completion where we can start the completion but not finish it. the curl works perfectly tho! thanks ^^

SoniEx2:
(ah we see, it seems that is answered in the README. unsure why it works for curl tho.)

yedayak:
> (ah we see, it seems that is answered in the README. unsure why it works for curl tho.)

I'm not sure either, bu a workaround is putting the 

> (ah we see, it seems that is answered in the README. unsure why it works for curl tho.)

Yeah I missed this about `_comp_compgen_ip_addresses`:
```
# If producing IPv6 completions, `_comp_initialize` with `-n :`.
```
Fixed, and added tests

akinomyoga (approved):

yedayak:
@SoniEx2 Does this fix the issue?

SoniEx2:
yes! thanks!

PR #1318, opened 2025-01-19
fix(ip): Don't offer neighbour, only neighbor

yedayak:
The previous state led to `ip neigh<TAB>` leading to this:
```
$ ip neighbo
neighbor   neighbour
```
choosing one to complete removes the need to choose between options that do the same thing, aligning with the CONTRIBUTING.md guidelines.

I chose `neighbor` because it's shorter, but it can be changed if someone wants to.

### DIFF
--- a/test/t/test_ip.py
+++ b/test/t/test_ip.py
@@ -32,6 +32,14 @@ class TestIp:
         assert "stale" in completion
 
     @pytest.mark.complete(
+        "ip neig",
+        require_cmd=True,
+        skipif="ip neighbor help 2>/dev/null; (( $? != 255 ))",
+    )
+    def test_neigh_one_completion(self, completion):
+        assert len(completion) == 1
+
+    @pytest.mark.complete(
         "ip monitor ",
         require_cmd=True,
         skipif="ip monitor help 2>/dev/null; (( $? != 255 ))",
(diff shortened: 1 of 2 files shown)


akinomyoga (approved):
There might be a possibility of determining it by the current locale, but I think just `neighbor` is fine as US English is common in computing and we don't want complication.

test/t/test_ip.py
```diff
@@ -31,6 +31,14 @@ def test_stats(self, completion):
     def test_neigh_state(self, completion):
         assert "stale" in completion
 
+    @pytest.mark.complete(
+        "ip neig",
+        require_cmd=True,
+        skipif="ip neighbor help 2>/dev/null; (( $? != 255 ))",
+    )
+    def test_neigh_one_completion(self, completion):
+        assert len(completion) == 1
```
scop, line 40: This is fine, but we could be more explicit, like
```suggestion
        assert completion == "hbor"
```
Anyway pre-approved, feel free to merge either way.

scop (approved):

test/t/test_ip.py
yedayak: Tried, but it doesn't work, since ip on centos7 (and presumably older OSs as well) have `neigh` in the help, so that's whats offered.

Issue #1324, opened 2025-02-02
Bash error since openssl updates


miken32:
## Describe the bug

Since I updated from 2.14 to 2.16 I've had this error when autocompleting openssl:
```
-bash: command substitution: line 1: unexpected EOF while looking for matching `''
```

This appears to be related to PR #1256.

In addition, output from `openssl ⇥⇥` looks misformatted:
```
mike ~ ❱❱❱ openssl -bash: command substitution: line 1: unexpected EOF while looking for matching `''
-bash: command substitution: line 1: unexpected EOF while looking for matching `''

(see                                                                      help:
Digest                                                                    idea              idea-cbc          idea-cfb          idea-ecb          
Message                                                                   idea-ofb          rc2               rc2-40-cbc        rc2-64-cbc        
Standard                                                                  info
aes-128-cbc       aes-128-ecb       aes-192-cbc       aes-192-ecb         kdf
aes-256-cbc       aes-256-ecb       aria-128-cbc      aria-128-cfb        list
aria-128-cfb1     aria-128-cfb8     aria-128-ctr      aria-128-ecb        mac
aria-128-ofb      aria-192-cbc      aria-192-cfb      aria-192-cfb1       nseq
aria-192-cfb8     aria-192-ctr      aria-192-ecb      aria-192-ofb        ocsp
aria-256-cbc      aria-256-cfb      aria-256-cfb1     aria-256-cfb8       passwd
aria-256-ctr      aria-256-ecb      aria-256-ofb      base64              pkcs12
asn1parse                                                                 pkcs7
bf                bf-cbc            bf-cfb            bf-ecb              pkcs8
bf-ofb            camellia-128-cbc  camellia-128-ecb  camellia-192-cbc    pkey
ca                                                                        pkeyparam
camellia-192-ecb  camellia-256-cbc  camellia-256-ecb  cast                pkeyutl
cast-cbc          cast5-cbc         cast5-cfb         cast5-ecb           prime
cast5-ofb         des               des-cbc           des-cfb             rand
ciphers                                                                   rc2-cbc           rc2-cfb           rc2-ecb           rc2-ofb           
cmp                                                                       rc4               rc4-40            seed              seed-cbc          
cms                                                                       rehash
commands                                                                  req
crl                                                                       rsa
crl2pkcs7                                                                 rsautl
des-ecb           des-ede           des-ede-cbc       des-ede-cfb         s_client
[...]

PR #1325, opened 2025-02-05
feat(openssl): openssl list -commands, not help

yedayak:
In https://github.com/scop/bash-completion/pull/1256 we started parsing the help output. Get the list instead from the `openssl list -commands` command.

This also stops completing the encryption and digest algorithms, since those are usually used with the `enc` and `dgst` subcommands.

Suggested-By: Michael Newton
Fixes: https://github.com/scop/bash-completion/issues/1324

### DIFF
--- a/completions/openssl
+++ b/completions/openssl
@@ -46,7 +46,7 @@ _comp_cmd_openssl()
 
     if ((cword == 1)); then
         local commands
-        commands="$("$1" help 2>&1 | command sed -e '/commands\|help:/d')"
+        commands="$("$1" list -commands)"
         _comp_compgen -- -W "$commands"
     else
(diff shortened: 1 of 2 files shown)


yedayak:
The `openssl list` subcommand doesn't seem to exist in the version shipping with centos7, `OpenSSL 1.0.2k-fips  26 Jan 2017`

miken32:
It doesn't exist in the version that shipped with CentOS 4, 5, or 6 either? CentOS 7 is long EOL and anyone still using openssl 1.0 likely aren't keeping their bash-completion on the bleeding edge.

But hey if you're willing to fight with sed to make it work, I won't complain!

PR #1327, opened 2025-02-08
openssl: Fix completions for MacOS, don't complete algorithms as subcommands

yedayak:
This should fix https://github.com/scop/bash-completion/issues/1324

This replaces https://github.com/scop/bash-completion/pull/1325, which doesn't work on some openssl versions

### DIFF
--- a/completions/openssl
+++ b/completions/openssl
@@ -111,6 +111,11 @@ _comp_cmd_openssl()
                 dgst | req | x509)
                     _comp_compgen -a -i openssl digests "$1"
                     ;;
+                enc)
+                    # The -list option was added in openssl 1.1.1e, so we ignore the error here.
+                    # We do get some of the ciphers "by accident", since some are detected in the help output.
+                    _comp_compgen -a split -- "$("$1" enc -list 2>/dev/null | command sed '/Supported ciphers:/d')"
+                    ;;
             esac
         else
             if [[ $command == speed ]]; then
--- a/completions/openssl
+++ b/completions/openssl
@@ -46,7 +46,8 @@ _comp_cmd_openssl()
 
     if ((cword == 1)); then
         local commands
-        commands="$("$1" help 2>&1 | command sed -e '/commands\|help:/d')"
+        # We only want the standard commands, so we delete everything starting after and including "Message Digest commands"
+        commands="$("$1" help 2>&1 | command sed -e '/Standard commands/d;/help:/d' -e '/Message Digest commands/,$d')"
         _comp_compgen -- -W "$commands"
     else
         command=${words[1]}
--- a/test/t/test_openssl.py
+++ b/test/t/test_openssl.py
@@ -5,9 +5,7 @@ class TestOpenssl:
     @pytest.mark.complete("openssl ", require_cmd=True)
     def test_1(self, completion):
         assert completion
-        assert all(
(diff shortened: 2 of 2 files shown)


completions/openssl
```diff
@@ -110,6 +111,9 @@ _comp_cmd_openssl()
                 dgst | req | x509)
                     _comp_compgen -a -i openssl digests "$1"
                     ;;
+                enc)
+                    _comp_compgen -- -W "$("$1" enc -list)"
```
akinomyoga, line 115:
```suggestion
                    _comp_compgen_split -- "$("$1" enc -list)"
```

The argument to `-W` is subject to expansion, so we prepared `_comp_compgen_split` to prevent the output of command substitutions from undergoing shell expansions. Even when we know that the present specific call of command substitution doesn't produce outputs looking like shell expansions, it is not obvious just by looking at this line. We might write `-W '"$(...)"'` instead, but shellcheck wouldn't be applied that way.

completions/openssl
```diff
@@ -46,7 +46,8 @@ _comp_cmd_openssl()
 
     if ((cword == 1)); then
         local commands
-        commands="$("$1" help 2>&1 | command sed -e '/commands\|help:/d')"
+        # We only want the standard commands, so we delete everything starting after and including "Message Digest commands"
+        commands="$("$1" help 2>&1 | command sed -E -e '/Standard commands|help:/d' -e '/Message Digest commands/,$d')"
```
akinomyoga, line 50: Maybe this is fine, but `-E` was added to POSIX recently. GNU/BSD `sed`'s are fine, but it might cause an issue with some other environments. However, `\|` in the original BRE is not POSIX. We could split the regular expression as `/Standard commands/d;/help:/d`

completions/openssl
yedayak: Done

completions/openssl
yedayak: Changed to your suggestion

akinomyoga (approved):

completions/openssl
```diff
@@ -110,6 +111,11 @@ _comp_cmd_openssl()
                 dgst | req | x509)
                     _comp_compgen -a -i openssl digests "$1"
                     ;;
+                enc)
+                    # The -list option was added in openssl 1.1.1e, so we ignore the error here.
+                    # We do get some of the ciphers "by accident", since some are detected in the help output.
+                    _comp_compgen -a split -- "$("$1" enc -list 2>/dev/null || true | sed '/Supported ciphers:/d')"
```
akinomyoga, line 117: `cmd1 || cmd2 | cmd3` would be interpreted as `cmd1 || { cmd2 | cmd3; }`, which I think is not the intention.

```suggestion
                    _comp_compgen -a split -- "$({ "$1" enc -list 2>/dev/null || true; } | sed '/Supported ciphers:/d')"
```

akinomyoga: Do we actually need to insert `true` here? In what situation, would the non-zero exit status cause an issue? I thought that `-o pipefail` might cause a signal in `sed`, but it doesn't seem to be the case. `-o errexit` doesn't seem to cause an issue either. The exit status of the command substitution would anyway be ignored.

yedayak: Yeah you're correct. Added `command`

akinomyoga: Thanks!

akinomyoga (approved):

PR #1333, opened 2025-02-14
support Android  app bundles as 'unzip' file type (XAPK/APKM/APKS)

eighthave:
XAPK/APKM/APKS are all ZIP containers for shipping Android apps.

* APKS https://developer.android.com/tools/bundletool#generate_apks
* XAPK https://openxapkfile.net/structure.html
* APKM https://fileinfo.com/extension/apkm


scop:
Thanks!

PR #1335, opened 2025-02-18
lvm: improve completions for `lvremove`

siteshwar:
It should have a completion for `/dev/mapper` path.

### DIFF
--- a/completions/lvm
+++ b/completions/lvm
@@ -33,6 +33,8 @@ _comp_cmd_lvm__logicalvolumes()
         for i in "${!COMPREPLY[@]}"; do
             [[ ${COMPREPLY[i]} == */control ]] && unset -v 'COMPREPLY[i]'
         done
+    elif [[ $cur == /d* ]]; then
+	    COMPREPLY+="/dev/mapper"
     fi
 }
 


Boomtokn (approved):

completions/lvm
```diff
@@ -33,6 +33,8 @@ _comp_cmd_lvm__logicalvolumes()
         for i in "${!COMPREPLY[@]}"; do
             [[ ${COMPREPLY[i]} == */control ]] && unset -v 'COMPREPLY[i]'
         done
+    elif [[ $cur == /d* ]]; then
+	    COMPREPLY+="/dev/mapper"
```
akinomyoga, line 37: This will prepend the string `/dev/mapper` to the first completion, which would result in a result e.g. `/dev/vg0/gfslv/dev/mapper` when a completion `/dev/vg0/gfslv` is already generated.

Also, shouldn't we complete `/dev/mapper/` instead of `/dev/mapper`?

In addition, we already have the utility of "check prefix & add to COMPREPLY". We shouldn't do that manually. For the present change, you can simply `_comp_compgen -- -W '/dev/mapper'`. However, we already call `_comp_compgen_split` on line 28, so we can simply add `/dev/mapper/` at the end of line 29.

akinomyoga (changes requested):

akinomyoga (approved):

akinomyoga:
I think we can add a test for `/dev/mapper/`.

akinomyoga:
I fixed the commit message to follow Conventional Commits and added a test.

PR #1338, opened 2025-02-24
refactor(rsync): prefix sed with command

yedayak:


akinomyoga (approved):
I think the commit message and the PR title should be `sed` instead of `cd`. Otherwise, this is good. There do not seem to be other similar cases (which may be checked by running `test/runLint`. Note: we do not care about the test codes so much).

akinomyoga:
We can consider breaking the line to fit it within 79 columns (ref [`styleguide`](https://github.com/scop/bash-completion/blob/f565bf9f2c3315a102174eb8b6c93253fc3e9353/doc/styleguide.md#line-wrapping)).

Issue #726, opened 2022-04-09
Add macOS to CI test matrix


scop:

scop:
Depends on #727 

WIP branch: https://github.com/scop/bash-completion/tree/macos-ci

PR #1339, opened 2025-02-24
Macos ci

yedayak:
See https://github.com/scop/bash-completion/issues/726

Cherry-picked from here:
- [x] https://github.com/scop/bash-completion/pull/1440
- [x] https://github.com/scop/bash-completion/pull/1441
- [x] https://github.com/scop/bash-completion/pull/1442
- [x] https://github.com/scop/bash-completion/pull/1456

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,4 +87,4 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           persist-credentials: false
-      - run: env PYTESTFLAGS="--verbose -p no:cacheprovider" test/macos-script.sh
+      - run: env PYTESTFLAGS="--verbose -p no:cacheprovider --color=yes" test/macos-script.sh
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,3 +80,11 @@ jobs:
         with:
           path: bash-completion-*.tar.xz
         if: matrix.dist == 'alpine'
+
+  distcheck-macos:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+      - run: env PYTESTFLAGS=--verbose test/macos-script.sh
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -82,7 +82,7 @@ jobs:
         if: matrix.dist == 'alpine'
 
   distcheck-macos:
-    runs-on: macos-latest
+    runs-on: macos-26
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,4 +87,4 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           persist-credentials: false
-      - run: env PYTESTFLAGS=--verbose test/macos-script.sh
+      - run: env PYTESTFLAGS="--verbose -p no:cacheprovider" test/macos-script.sh
--- a/test/macos-script.sh
+++ b/test/macos-script.sh
@@ -1,0 +1,22 @@
+#!/bin/sh -eux
+
+# Note that this script is intended to be run only in throwaway environments;
+# it may install undesirable things to system locations (if it succeeds in
+# that).
+
+brew install \
+    automake \
+    bash
+
+python3 -m pip install -r test/requirements.txt
+
+export bashcomp_bash=bash
+env
+
+autoreconf -i
+./configure
+make -j
+
+make distcheck \
+    PYTESTFLAGS="${PYTESTFLAGS---verbose --numprocesses=auto --dist=loadfile}"
+cp -p bash-completion-*.tar.* "$oldpwd/"
--- a/test/macos-script.sh
+++ b/test/macos-script.sh
@@ -15,6 +15,11 @@ python3 -m pip install -r test/requirements.txt
 
 export bashcomp_bash=bash
 env
+LC_ALL=C sort <<<$'python-dev\npython3-dev'
+LC_COLLATE=C sort <<<$'python-dev\npython3-dev'
+LC_ALL=en_US.UTF-8 sort <<<$'python-dev\npython3-dev'
+LC_COLLATE=en_US.UTF-8 sort <<<$'python-dev\npython3-dev'
+
 
 autoreconf -i
 ./configure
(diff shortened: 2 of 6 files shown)


yedayak:
The CI run failed because for some reason it didn't manage to delete the temp directory it creates in `test_unit_load.py` using shutil.rmtree in `prepare_fixture_dir`

scop:
Thanks for working on this! I think I've seen the temp dir removal issue myself as well. Would be great if we could figure out a workaround. If we can't, we should allow it to fail only on macOS.

test/t/unit/test_unit_load.py
```diff
@@ -33,7 +34,11 @@ def fixture_dir(self, request, bash):
         assert_bash_exec(
             bash, "ln -sf ../prefix1/sbin/cmd2 %s/bin/cmd2" % tmpdir
         )
-        return str(tmpdir)
+        yield str(tmpdir)
+        if sys.platform == "darwin":
+            assert_bash_exec(bash, "sudo rm -rf %s/*" % tmpdir)
```
yedayak: I don't love running sudo in the test suite, but I couldn't find a another way to delete these files

.github/workflows/ci.yaml
```diff
@@ -107,3 +107,11 @@ jobs:
...
           RELEASE_PLEASE_TAG_NAME: ${{steps.release.outputs.tag_name}}
         if: steps.release.outputs.release_created
+
+  distcheck-macos:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+      - run: env PYTESTFLAGS="--verbose -p no:cacheprovider --color=yes" test/macos-script.sh
```
scop, line 90: Just to confirm, is `--color=yes` needed here? We don't seem to add it in Linux envs, but do get color nevertheless.

test/t/unit/test_unit_dequote.py
```diff
@@ -38,7 +40,8 @@ def test_5_brace(self, bash, functions):
 
     def test_6_glob(self, bash, functions):
         output = assert_bash_exec(bash, "__tester 'a?b'", want_output=True)
-        assert output.strip() == "<a b><a$b><a&b><a'b>"
+        items = sorted(re.findall(r"<[^>]*>", output))
```
scop: It's not immediately clear to me why this is needed on macOS and not on others, could we elaborate in a comment?

...after having a look further in the PR, it seems various sorts are added elsewhere as well. That smells a bit as if the collation options would not be properly in effect. Or are they, but just sort differently in macOS nevertheless?

test/t/unit/test_unit_load.py
scop: I wonder why we need to do this manual removal in the first place. In prepare_fixture_dir we add a finalizer that is supposed to clean up the temp dir altogether. Does that not work here? Does it work in other places where we use it to set up a fixture dir?

But if it is needed, some followup questions:

Can we assume sudo is present? What if it asks for credentials?

Any idea if this is needed on all macOS setups, or just the GH one?

If it's a GH only thing, and perhaps otherwise as well, I suppose we could just not fail on error here if the remove fails and let the error message come through so people can clean up manually.

Let's add a comment with details why it doesn't work without it here.

Maybe consider doing the sudo rm in the prepare_fixture_dir finalizer on darwin, if it is needed?

test/macos-script.sh
scop, line 1: No that strong opinions, but perhaps we could take a look if it would become more messier than it's good for if we'd combine this with test/docker/entrypoint.sh with appropriate conditionals to avoid duplicating a bunch of things.

test/t/unit/test_unit_load.py
yedayak: > I wonder why we need to do this manual removal in the first place. In prepare_fixture_dir we add a finalizer that is supposed to clean up the temp dir altogether. Does that not work here? Does it work in other places where we use it to set up a fixture dir?

From what I gathered, the issue is that this is the only place where we create more files and folders inside the fixture_dir. Those extra files are created by other processes and not python, so I think that's what causes the issue. See [this note](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/removefile.3.html#NOTES:~:text=Write%20protected%20files%20owned%20by%20another%20process%20cannot%20be%20removed%20by%20removefile%28%29%2C%20regardless%20of%20the%20permissions%20on%20the%20directory%20containing%20the%20file%2E) in the `removefile()` syscall: 
>> Write protected files owned by another process cannot be removed by
>> removefile(), regardless of the permissions on the directory containing
>> the file.

I might have just not found the correct magic spell to make these files deletable without sudo

> 
> But if it is needed, some followup questions:
> 
> Can we assume sudo is present? What if it asks for credentials?

Yeah that's a problem. It seems the default behavior is asking [for a password](https://support.apple.com/en-il/guide/terminal/apd5b0b6259-a7d4-4435-947d-0dff528912ba/mac#:~:text=You%E2%80%99re%20asked%20for%20the%20password%20of%20the%20current%20user)
> 
> Any idea if this is needed on all macOS setups, or just the GH one?

I don't know, the man pages seem to imply this will happen for all macOS's, but maybe that's not the real issue. I don't have a physical one to test on.

> If it's a GH only thing, and perhaps otherwise as well, I suppose we could just not fail on error here if the remove fails and let the error message come through so people can clean up manually.

I'm not sure what a good way to let the error come through without failing the tests is
> 
> Let's add a comment with details why it doesn't work without it here.
> 
> Maybe consider doing the sudo rm in the prepare_fixture_dir finalizer on darwin, if it is needed?

`prepare_fixture_dir` doesn't currently have access to bash, so it can't run sudo. It also isn't needed for all of the uses of it. I can make it accept a `bash` instance if you think that's better though.

test/t/unit/test_unit_dequote.py
yedayak: Yeah I'm not sure what's up with that. I just checked and LC_COLLATE seems to be set to C. I'll need to look into that

test/t/unit/test_unit_load.py
akinomyoga: *From https://github.com/scop/bash-completion/pull/1531#discussion_r2653914871* by [@yedayak](https://github.com/yedayak)

> [...] a follow up might be to convert usages of `prepare_fixture_dir` to also use `tmp_path_factory`, maybe it will help with [#1339 (comment)](https://github.com/scop/bash-completion/pull/1339#discussion_r2442558142), But this isn't needed for this PR.

.github/workflows/ci.yaml
```diff
@@ -103,3 +103,11 @@ jobs:
           GH_TOKEN: ${{github.token}}
           RELEASE_PLEASE_TAG_NAME: ${{steps.release.outputs.tag_name}}
         if: steps.release.outputs.release_created
+
+  distcheck-macos:
+    runs-on: macos-latest
```
scop: macos-latest will switch to 26 in coming weeks, https://github.com/actions/runner-images/issues/14167

Hopefully that'll help rather than hurt some :)

yedayak: Tried now, didn't change. The only issue really left is the sorting of completion items being different, which I think is done by readline using `strcoll(3)` if available

PR #1348, opened 2025-03-12
feat(*sum): handle more common checksum filenames

scop:
checksums, checksums.txt, ${sumtype}sums, ${sumtype}sums.txt


completions/sha256sum
```diff
@@ -20,18 +20,20 @@ _comp_cmd_sha256sum()
...
     sumtype=${sumtype%sum}
+    local sumglob="@(*.$sumtype|@(check|${sumtype})sums?(.txt))"
 
-    local opt
+    local files opt
     for opt in "${words[@]}"; do
         if [[ $opt == -@(c|-check) ]]; then
-            _comp_compgen_filedir "$sumtype"
+            _comp_compgen -v files filedir &&
+                _comp_compgen -- -X "!$sumglob" -W '"${files[@]}"'
```
akinomyoga, line 29: This filters out directory names generated by `_comp_compgen_filedir`, but we might want to generate also directory names.

scop: Ah, true, drafting until fixed.

akinomyoga: As I mentioned in https://github.com/scop/bash-completion/pull/1358#issuecomment-2781356019 a bit, I'm currently thinking of extending `_comp_compgen_filedir` to accept options to specify the whole glob pattern (instead of just the filename-extension part).

scop: Hopefully better now & ready for another look, lifting draft.

akinomyoga (approved):

completions/sha256sum
akinomyoga: Ideally, `filedir` could be extended to accept a glob pattern, but the current version also works.

PR #1351, opened 2025-03-18
feat(rpm): rpmbuild short rebuild options

leviramsey:
While the `--rebuild` option isn't deprecated, [the man page now notes that the `-r?` options supersede it](https://www.man7.org/linux/man-pages/man8/rpmbuild.8.html).

Treats `-r?` as equivalent to `--rebuild` for determining which extensions to complete


akinomyoga (approved):
Thanks for the contribution. The commit message should be `feat` instead of `chore`. Otherwise, LGTM.

scop:
Thanks!

PR #1356, opened 2025-04-02
feat(_comp_log): new function for debug logging

dseomn:
I'm trying to write a completion script for tmux, by parsing some of its command introspection output. There are a bunch of ways I could mess up that parsing, so debug logging has been helpful so far. Maybe this will be useful to others too?

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -126,6 +126,18 @@ complete -b builtin
 
 # start of section containing completion functions called by other functions
 
+# Log a message to help with debugging.
+# If BASH_COMPLETION_DEBUG is set, it will be printed to stderr.
+# When running with `set -x`, the _comp_log call itself will be printed.
+#
+# @param $1 Message to log
+_comp_log()
+{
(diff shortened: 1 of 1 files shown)


akinomyoga:
There is no reason to include and maintain this function within the bash-completion project. This is premature abstraction. It seems that this function can be just kept in your projects.

- No codes in `bash-completion` use this function. This means that there is no chance of testing and improving the implementation. This function will soon become an unmaintained burden of this project.
- The `bash-completion` project is not the standard library of Bash interactive settings. It is primarily a set of completion settings and doesn't aim at providing the general utility used by other projects (and unused by bash-completion itself).
- Once this is added to a part of the public interface of bash-completion, we cannot change the interface for backward compatibility. So, if we would add a new public function that is not used in this project, its interface and behavior should be carefully designed based on the actual use cases. When many independent projects copy this function to implement their completion settings, we can say that its interface is validated by different use cases by different people, so we can include this function in our repository to provide an ultimate form of the logging function.

Issue #1255, opened 2024-10-25
Escaping of remote filenames containing spaces is broken for rsync 


omgold:
## Describe the bug

If I want to copy a file from remote to local containing spaces, the space will be escaped with triple backslashes, but it should be only one.

## To reproduce

If there is a named `test file.txt`  when I try to complete

`rsync localhost:test`

I get the completion

`rsync localhost:test\\\ file.txt`

which doesn't work.

## Expected behavior

It should be

`rsync localhost:test\ file.txt`

## Versions (please complete the following information)

- [ x] Operating system name/distribution and version: Arch
- [ x] bash version, `echo "$BASH_VERSION"`: 5.2.37(1)-release
- [ x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.14.0

yedayak:
Hi, what version of rsync are you using?
This should be fixed on the main branch, see [this pr](https://github.com/scop/bash-completion/pull/910).
Can you try using the main branch and see if it works for you?

omgold:
Okay tried it. It seems to be fix only partially, though.

It still misbehaves when there are 2 files present, `test file.txt` and `test\ file.txt`.

If I then try to complete

`rsync localhost:test`

at the first tab press I get

`rsync localhost:test\`

but on subsequent presses, I get no choices to pick from at all.

I'm using rsync 3.3.0

yedayak:
Can you add set -x lines to completions/rsync?
like this:
```
106                 set -x #<------ Add this
107                 _comp_cmd_rsync__vercomp "$rsync_version" "3.2.4"
108                 if (($? == 2)); then
109                     _comp_compgen -x scp remote_files
110                 else
111                     _comp_compgen -x scp remote_files -l
112                 fi
113                 set +x <-----  # and this
```

omgold:
@yedayak here you go.

[rsync-complete-debug-output.txt](https://github.com/user-attachments/files/17616488/rsync-complete-debug-output.txt)

scop:
I wonder if #1232 is the same issue as this one.

akinomyoga:
> I wonder if [#1232](https://github.com/scop/bash-completion/issues/1232) is the same issue as this one.

I'd say it is closely related, but it's a separate issue.

[...]

PR #1357, opened 2025-04-05
fix(java,rsync,scp): handle quoted space in filepaths properly

### LINKED ISSUE #1232: [2.14.0] rsync path completion fails if a folder name have a space
## Describe the bug

bash-completion was recently updated from 2.11.0 to 2.14.0 in Gentoo.
With bash-completion-2.14.0, rsync path completion fails if a folder name have a space.

## To reproduce

```
~ mkdir "a b" "a c"
~ rsync -av a [hit tab]
It adds a backslash to the command "rsync -av a\ " but hitting tab again fails to show :
a\ b/ a\ c/
```

## Versions
- [Gentoo ]
- [ 5.2.26(1)-release] bash version
- [ 2.14.0] bash-completion version

----
It isn't exclusive to Gentoo as I observe the exact same issue with Debian testing (Trixie) which features bash-completion-2.14.
On the other hand, with Debian stable (bookworm) and bash-completion-2.11, hitting tab correctly show the folders (with space or not)
Finally, downgrading from bash-completion from 2.14.0 to 2.11.0 on my gentoo system solves the issue the too.

akinomyoga:
This includes several fixes (whose changes overlap with each other). See the respective commit messages for the details.

Commit b653f4de1d43ececa91edde790b92a08ea925068 fixes #1232.


akinomyoga:
@Copilot[AI] The language of **completions/java** and **completions/ssh** are Bash.

> * The cwd value 'scp' appears to be incorrect for an rsync test. It may be a typo and should likely be 'rsync'.
> 
> ```
> @pytest.mark.complete("rsync -na spaced\ ", cwd="scp")
> ```

This is a good point. Since the `rsync` completion internally uses the xfunc function that `scp` provides, I reused the fixture directory of `scp`. This is intentional but might be confusing to the readers. I'll add a code comment.

akinomyoga:
I tried the Copilot code review, which was [released today](https://github.blog/changelog/2025-04-04-copilot-code-review-now-generally-available/), but I'm not sure how to respond to Copilot and get further feedback.

akinomyoga:
> **test/t/test_rsync.py:77**
> 
> * The cwd parameter is set to 'scp' in a rsync test; confirm if this is intentional or if it should reference 'rsync' to avoid potential mis-routing of the test context.

I already added the clarifying code comment. So, Copilot does not take account of the code comment.

> **test/t/test_scp.py:191**
> 
> * [nitpick] The expected completion string 'ced\ \ conf' is hard to interpret;

This is a common pattern of the test suite in this project, where the text following the already input part (i.e., "spa" in the present case) is tested. It is impractical to add clarifying comments to all the occurrences of similar tests. Ideally, we can document the global structures and conventions of our test suite somewhere, but we currently don't have the documentation at all. To document it, too many things are not explained.

> **test/t/test_scp.py:195**
> 
> * [nitpick] Consider asserting an explicit expected value for completion in 'test_local_path_with_spaces_2' to ensure the test precisely validates the intended behavior.

Thanks, this is simply a mistake. I fixed it.

akinomyoga:
I added a fix for #1255 and related changes. See the commit messages for the details.

JohnVillalovos:
Thanks @akinomyoga  I have tested this and now `rsync` and `scp` works with directories that have spaces in their names.

For example if I have the two directories: `/home/dir with space` and `/home/dir with target` the completion now works. Where before it would get stuck at `/home/dir with `

akinomyoga:
@JohnVillalovos Thanks for the confirmation!

bash_completion
```diff
@@ -207,7 +207,12 @@ _comp_dequote__initialize
 #                 quotations, parameter expansions are allowed.
 #
 # @var[out] REPLY  Array that contains the expanded results.  Multiple words or
-#                  no words may be generated through pathname expansions.
+#                  no words may be generated through pathname expansions.  If
+#                  $1 is not a safe word, REPLY contains the literal value of
+#                  $1.
+#
+# @return 0 if $1 is a safe word and the expansion result contains one word at
+#     least, or 1 otherwise.
```
yedayak, line 215: This change seems reasonable to me, do you think this might cause breakage in external uses of the function?

bash_completion
```diff
@@ -235,8 +240,37 @@ _comp_dequote__initialize
...
+        REPLY=${1-}
+        return 1
+    fi
+}
+
+# Try to reconstruct an incomplete word and apply _comp_dequote.
+# @param    $1  String to be expanded.  The same as _comp_dequote, but
+#               incomplete backslash, single quotation, and double quotation
+#               are allowed.
+# @var[out] REPLY  Result.  The same as _comp_dequoet.
```
yedayak, line 258:
```suggestion
# @var[out] REPLY  Result.  The same as _comp_dequote.
```

bash_completion
```diff
@@ -251,6 +251,28 @@ _comp_dequote()
     fi
 }
 
+# Try to reconstruct an incomplete word and apply _comp_dequote.
+# @param    $1  String to be expanded.  The same as _comp_dequote, but
+#               incomplete backslash, single quotation, and double quotation
+#               are allowed.
+# @var[out] REPLY  Result.  The same as _comp_dequoet.
+# @since 2.17
+_comp_dequote_incomplete()
```
yedayak, line 260: Maybe a unit test for this will be nice?

yedayak (commented):
I'm not sure I understand everything here, so take it with a grain of salt, but this looks fine to me. Left some comments. This definitely fixes the issues with local files completions, but I still seem to have some issues with remote file:
```
scp localhost:/home/yedaya/dev/bash-completion/test/fixtures/scp/local_path/backslash-a<TAB>\\\
```
This happens also with main, so not a blocker.

completions/ssh
```diff
@@ -461,6 +461,41 @@ _comp_cmd_sftp()
...
+# suffixes a space or a slash based on the file type.
+#
+# Note: With a non-empty prefix ($1 of _comp_xfunc_scp_compgen_local_files),
+# Bash will not recognize any filenames, so we need to perform the proper
+# quoting manually.  We also need to manually suffix a space or a slash based
+# on the file type because "-o nospace" is specified.  One might think of using
+# "compopt +o nospace" instead, but it would suffix a space to directory names
+# unexpectedly.
+#
+# @param $1 dirs_only - If a non-empty value is specified, only directory names
```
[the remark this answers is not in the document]
akinomyoga: Thanks. I tried to implement it in commit 8ca4fa95c17651562c4c36a000df3a065bea1ace, but it seems non-trivial how the caller should pass the flag to this function. In commit 8ca4fa9, I'm currently doing

```bash
        _comp_cmd_scp__escape_path ${_dirs_only:+'-d'} -- \
            "$_escape_replacement"
```

I thought we might assign `_dirs_only=-d` instead of `_dirs_only=set`, but even in that case, we cannot do `"$_dirs_only"` because it will produce an empty argument `""` without `-d`. Then we still need to write `${_dirs_only:+"$_dirs_only"}`.

Another option might be to make `_dirs_only` an array and use `"${_dirs_only[@]}"`. However, because of Bash-4.2 nounset bug, we anyway need to write `${_dirs_only[@]+"${_dirs_only[@]}"}`.

bash_completion
akinomyoga: Yeah, that is a good point.

After this change, the cases where it failed will continue to fail. Some of the cases where it succeeded and produced an empty result [e.g., `REPLY=()` for `_comp_dequote '$empty_variable'`] will start to fail after this change. Since the caller should have considered failing cases even with the current `main` branch, I expect that increasing another case of failing wouldn't cause a problem.

Rather, an empty `REPLY=()` would have caused a problem when the caller uses `$REPLY` after `_comp_dequote` succeeds.

```bash
some=""
if _comp_dequote '$some'; then # -> results in REPLY=()
  use "$REPLY" # <- This failed for "set -u" before the present change
fi
```

Another change is in the value of `REPLY` when `_comp_dequote` failed. Since `REPLY` was not supposed to be used when `_comp_dequote` failed (`REPLY` was actually empty, `REPLY=()`), I expect the change in `REPLY` in the failing case wouldn't cause a problem, though I admit that there might be a case that the user expects the empty `REPLY` in principle. I think this change would rather improve the compatibility in the case where the caller isn't careful enough to check the exit status:

```bash
_comp_dequote "$some2"
use "$REPLY" # <- This failed for "set -u", or if "set -u" is not set, an empty value was used.
```

bash_completion
akinomyoga: Thanks for the suggestion! I added tests in dcf662b5263986870eef4c883602a0f1e279672d.

test/t/test_scp.py
```diff
@@ -141,3 +179,48 @@ def test_xfunc_remote_files(self, bash):
...
+    # FIXME: This test currently fails.
+    # def test_local_path_mark_2(self, bash, tmpdir_mkfifo):
+    #     completion = assert_complete(
+    #         bash, "scp local_path_2-", cwd=tmpdir_mkfifo
+    #     )
+    #     assert completion == "pipe\\|"
+
+    @pytest.mark.complete("scp spa", cwd="scp")
+    def test_local_path_with_spaces_1(self, completion):
+        assert completion == "ced\\ \\ conf"
```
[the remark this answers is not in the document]
akinomyoga: Thank you. I added changes in b0d9f7db751fa987923efe94b1661bb8db16d501 90977b49a 811b67451 7225e9d27.

akinomyoga:
> but I still seem to have some issues with remote file:
> 
> ```
> scp localhost:/home/yedaya/dev/bash-completion/test/fixtures/scp/local_path/backslash-a<TAB>\\\
> ```
> 
> This happens also with main, so not a blocker.

Thank you. The following seems to fix the issue basically

```diff
diff --git a/completions/ssh b/completions/ssh
index 91bcee8e9..366de71ef 100644
--- a/completions/ssh
+++ b/completions/ssh
@@ -544,7 +544,7 @@ _comp_xfunc_scp_compgen_remote_files()
     local REPLY=$cur
     if [[ ! $_less_escaping ]]; then
         # unescape (3 backslashes to 1 for chars we escaped)
-        REPLY=$(command sed -e 's/\\\\\\\('"$_comp_cmd_scp__path_esc"'\)/\\\1/g' <<<"$REPLY")
+        REPLY=$(command sed -e 's/\\\\\\\('"$_comp_cmd_scp__path_esc"'\|$\)/\\\1/g' <<<"$REPLY")
     fi
     _comp_dequote_incomplete "$REPLY"
     local cur_val=${REPLY-}
```

but the problem is that `\|` is not supported by POSIX BRE. One might consider using `sed -E`, but the problem with `sed -E` is that it is added to POSIX only recently. If we separate the substitution as `s/.../\\\1/g;s/\\\\\\$/\\/`, the second substitution might unexpectedly replace the result of the first substitution.

akinomyoga:
> ```diff
> --- a/completions/ssh
> +++ b/completions/ssh
> @@ -546,3 +546,3 @@ _comp_xfunc_scp_compgen_remote_files()
>          # unescape (3 backslashes to 1 for chars we escaped)
> -        REPLY=$(command sed -e 's/\\\\\\\('"$_comp_cmd_scp__path_esc"'\)/\\\1/g' <<<"$REPLY")
> +        REPLY=$(command sed -e 's/\\\\\\\('"$_comp_cmd_scp__path_esc"'\|$\)/\\\1/g' <<<"$REPLY")
>      fi
> ```
> 
> but the problem is that `\|` is not supported by POSIX BRE. One might consider using `sed -E`, but the problem with `sed -E` is that it is added to POSIX only recently. If we separate the substitution as `s/.../\\\1/g;s/\\\\\\$/\\/`, the second substitution might unexpectedly replace the result of the first substitution.

6b15fa1af I decided to process it using Bash.

completions/ssh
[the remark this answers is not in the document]
akinomyoga: Probably the current version 8ca4fa95c17651562c4c36a000df3a065bea1ace is fine.

test/t/test_scp.py
```diff
@@ -141,3 +188,48 @@ def test_xfunc_remote_files(self, bash):
...
+        tmpdir, _, _ = prepare_fixture_dir(
+            request,
+            files=["local_path_2-pipe|"],
+            dirs=[],
+        )
+
+        try:
+            assert_bash_exec(bash, "mkfifo '%s/local_path_1-pipe'" % tmpdir)
+        except Exception:
+            pytest.skip(
```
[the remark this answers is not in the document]
akinomyoga: I just copied `colonpath` (`test_man.py`) and edited it, but true. There are three existing cases in the codebase:

```bash
./test/t/test_cancel.py-17-        except AssertionError:
./test/t/test_cancel.py:18:            pytest.skip("Could not add test print job")
./test/t/test_cancel.py-19-            return
--
./test/t/test_ls.py-34-        if not part_full:
./test/t/test_ls.py:35:            pytest.skip("No suitable test user found")
./test/t/test_ls.py-36-            return
--
./test/t/test_man.py-24-        else:
./test/t/test_man.py:25:            pytest.skip("Cygwin doesn't like paths with colons")
./test/t/test_man.py-26-            return
```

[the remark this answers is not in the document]
akinomyoga: Fixed in commit c851876b1b05e0ff3d85087e30f69628d3c27c01.

I also fixed existing cases in commit 6c885312a781231a34606a4a6b0ae9374ed64f2b.

akinomyoga:
> **test/t/test_rsync.py:83**
> 
> * The test for local path completion with spaces only checks that a value is returned without asserting a specific expected output. Consider adding an explicit expected value to better capture regressions.
> 
> ```
>         assert completion
> ```

I skipped testing the value because the expected value depends on the rsync version in the system, but we can test the result if it matches one of the possible values. I added the check in commit 35d367e0a591027e9956f91dc932dc5fef4ab859.

akinomyoga:
> **test/t/test_rsync.py:85**
> 
> * [nitpick] Add a comment to document the variations in expected output based on different rsync versions, so the reasoning behind the dual expectation is clear.
> 
> ```
> assert completion == r"\ conf" or completion == r"\\\ conf"
> ```

35c1af4f6

> **test/t/test_scp.py:196**
> 
> * [nitpick] Consider ensuring that the use of the pipe '|' character in the filename is intentional for testing special character handling and add an inline comment if needed.
> 
> ```
> files=["local_path_2-pipe|"],
> ```

9608d1ef5

akinomyoga:
@yedayak I added many changes and squashed them into the corresponding commits. The changes after your previous review can be checked on [`b456d8d..1fd977d`](https://github.com/scop/bash-completion/compare/b456d8d8dd2d8b3338a929b7a4244eb0a389100c..1fd977d8444e122500a7fa6583204bb28cedd220).

akinomyoga:
(rebased)

scop:
Instead of lumping unrelated but conflicting/overlapping PR's together, especially as many as there are in this one, I'd suggest filing several and basing them off each other. This is a good way to optimize for the reviewers -- ones with multiple unrelated ones tend to grow large, and entail many context switches for the reviewer. For example, I could have easily reviewed and approved a few of these now and we could get them out of the way, but I cannot find time and attention to do it all now, and I'm not sure when I will. (If others do, please go ahead!)

So the first PR would have branch-1 based off main, the second's branch branch-2 based off branch-1 (and set as base in the PR), etc etc. I believe for this to work the PR's need to be branches in this repo and not a fork, so this is an option directly available for people with contributor access.

akinomyoga:
Thank you. Yes..., I have to agree with you that this PR happened to become too large.

I usually tend to combine multiple changes in a reasonable size of a PR when I introduce many conflicting changes to avoid the complication of PR dependency and conflicts. If every tiny change is processed one by one as a PR, it would take a longer time until I forget the set of changes and their relationships. However, I admit that I failed to decide on the proper size of this PR. After some adjustments, the number of commits increased too much.

I'll split the PR.

akinomyoga:
I submitted PR #1371.

> I believe for this to work the PR's need to be branches in this repo and not a fork, so this is an option directly available for people with contributor access.

Sorry, I missed this part, though I'm afraid that I'm possibly still missing the intent. Is the idea that another collaborator can split the PR (instead of the original submitter)?

bash_completion
akinomyoga: Let's continue the discussion in PR #1483.

akinomyoga:
@scop @yedayak This is the final change of this PR. I also adjusted the PR title to reflect the last change.

akinomyoga:
Thank you very much!

PR #1297, opened 2024-12-30
fix(_umount,feh,sbopkg): check diretory name for `_comp_compgen -C`

akinomyoga:
Fixes #1260

This is an alternative fix to #1290. The fix suggested in #1290 was just to remove the error message, which I believe is not the right fix. This PR fixes it by checking the arguments at the caller side.

This includes an additional fix e9e665efd7623dd9ec6d72c25c5797c497d58f65 for `feh`. In the current `main` branch, an error message is printed on the TAB completion with the following command line:

```console
$ set -u
$ feh --font x[TAB] --fontpath
bash: words[i + 1]: unbound variable
```

PR #1358, opened 2025-04-06
fix(mdtool): fix uses of `-G 'pat'` and `-f -X '!pat'`

akinomyoga:
This PR addresses issues discussed in #1297. This PR also includes other changes. See the commit messages.

- Commit cac9bbcf2 work arounds https://github.com/scop/bash-completion/pull/1297#discussion_r2030093233
- Commit 2e0d45ef8 fixes https://github.com/scop/bash-completion/pull/1297#discussion_r1899507348


akinomyoga:
There is still an issue with `-f -X '!...'` (which haven't been discussed in #1297). They do not generate directory names even though the user might want to specify a file inside a subdirectory. This is actually related to the problem discussed in https://github.com/scop/bash-completion/pull/1348#discussion_r1992314181. I'm thinking of addressing it in another PR.

One of the reasons that some of the completion setting use `-f X'!...'` instead of calling `filedir` seems that those completions want to prefix or suffix additional string using `-S /` or `-P "$prefix"`, which are unsupported by `_comp_compgen_filedir`.

- One solution is to extend `filedir` to support these options `-S` and `-P`.
- Another solution that I'm currently thinking is to extend `_comp_compgen` to accept `-S` and `-P`. Actually, I have an old local branch that implements these options (with non-trivial but necessary adjustments of `cur`). I'll try to resurrect and rebase the branch.

CONTRIBUTING.md
```diff
@@ -132,6 +132,14 @@ Also, please bear the following coding guidelines in mind:
...
   the latter case, checks by shellcheck and shfmt will not be performed inside
   `'...'`.  Also, `_comp_compgen_split` is `IFS`-safe.
+  
+  Avoid using `_comp_compgen -- -G "pattern"` to generate completions.  The
+  result is not filtered by the current word `cur` due to the Bash design of
+  `compgen`.  Also, this cannot be used to generate filenames with a specified
+  extension because the `-G` specification only generates the matching
+  filepaths in the current directory.  It does not look into subdirectories
+  even when `$cur` implies the completion in a subdirectory.  One can instead
+  use `_comp_compgen -- -f -X '!pattern'`.
```
scop, line 142: This hunk could be in a commit of its own -- now it's "buried" in a `fix(mdtool)` one without a note. No big deal, but thought I'd mention.

test/runLint
```diff
@@ -52,6 +52,9 @@ gitgrep '(?<!command)'"$cmdstart"'(grep|ls|sed|cd)(\s|$)' \
 gitgrep '(?<!command)'"$cmdstart"'awk(\s|$)' \
     'invoke awk through "_comp_awk"'
 
+gitgrep '@\([^()|$]+\)' \
```
scop, line 55: The commit that adds this has a "unnecesasry" typo in the commit message. Could fix up before merging.

scop (approved):
Pre-approved with a couple of cosmetic notes, ok to merge whether addressed or not.

CONTRIBUTING.md
akinomyoga: Thanks! I separated the commit in the push [`df59a49..e98f1a7`](https://github.com/scop/bash-completion/compare/df59a49d00d8e953e0317720444b1fc3742acc36..e98f1a704e8ca17f34979f1f32b79d48092ecd5f).

test/runLint
akinomyoga: Thanks. I fixed this in the latest push too.

PR #1360, opened 2025-04-06
feat(_comp_compgen): support `-P prefix` with adjusted `cur`

akinomyoga:
See the commit message.


doc/api-and-naming.md
```diff
@@ -143,12 +143,12 @@ concept is later extended to also mean "eXported".
...
 The generator functions, which have names of the form `_comp_compgen_NAME`, are
 used to generate completion candidates.  A generator function is supposed to be
-called by `_comp_compgen [OPTS] NAME ARGS` where `OPTS = -aRl|-v var|-c cur|-C
-dir|-F sep` are the options to modify the behavior (see the code comment of
-`_comp_compgen` for details).  When there are no `opts`, the generator function
-is supposed to be directly called as `_comp_compgen_NAME ARGS`.  The result is
-stored in the target variable (which is `COMPREPLY` by default but can be
-specified by `-v var` in `OPTS`).
+called by `_comp_compgen [OPTS] NAME ARGS` where `OPTS = -aR|-v var|-c cur|-C
+dir|-U var|-P pref` are the options to modify the behavior (see the code
```
scop:
```suggestion
dir|-U var|-P prefix` are the options to modify the behavior (see the code
```

akinomyoga: Thanks for the suggestion!

akinomyoga: The original reason that I used the abbreviation `pref` was that the column width of the option in the doc comment was too short, but I decided to extend the column width so that we can fully spell `prefix` out: [`975712a..c755c680`](https://github.com/scop/bash-completion/compare/975712ae09a5cbaa6a3cbba60d1ec7cb3c1aab7e..c755c6803ecb52e39f83d0865a1b7d52255727e1).

completions/xgamma
```diff
@@ -32,8 +32,8 @@ _comp_cmd_xgamma()
                 # local screen numbers
                 local t screens=$(xrandr --query 2>/dev/null | command sed -ne \
```
akinomyoga:
```suggestion
                local screens=$(xrandr --query 2>/dev/null | command sed -ne \
```

akinomyoga: fixed.

PR #1362, opened 2025-04-07
perf(tar): cache the parse result of GNU tar help

akinomyoga:
In my setup, the `tar` completion has been somehow significantly slower than the other completions. It turned out that the help parsing, which is performed for every call of the tar completion, has been the bottleneck. This PR introduces a cache for the parse result, which significantly improves the performance of the tar completion.

The main change is 5a851086bf0eddb8aa09174913d76bf114b00659. There are other refactoring commits, but those changes are independent.

### DIFF
--- a/completions/tar
+++ b/completions/tar
@@ -133,6 +133,13 @@ _comp_cmd_gtar__parse_help_line()
 #     The parse results are returned in these variables.
 _comp_cmd_gtar__parse_help()
 {
+    long_arg_none=""
+    long_arg_opt=""
+    long_arg_req=""
+    short_arg_none=""
+    short_arg_opt=""
+    short_arg_req=""
+
     local str line arg
     while IFS= read -r line; do
         # Ok, this requires some comment probably.  The GNU help output prints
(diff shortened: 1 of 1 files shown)


scop (approved):
LGTM. Separate definition and initialization of variables seems to be something we don't generally do, I wonder if there's a reason for not doing them in one go here as well? Anyway, approved as is or with initialization moved to definitions.

akinomyoga:
> I wonder if there's a reason for not doing them in one go here as well?

Yeah, I also thought about that option. I updated the commit that way. In commit c1e96b7582708ef788b40f6958655d7cd15ebf6b, we initialize `_comp_cmd_gtar__parse_help__result` when loading the file `completions/tar` (rather than the first call of `_comp_cmd_gtar__parse_help`). Now the function `_comp_cmd_gtar__parse_help` just evaluates `"$_comp_cmd_gtar__parse_help__result"`, so I renamed it to `_comp_cmd_gtar__parse_help_load`.

PR #81, opened 2016-10-13
Add tmux completion

srsudar:
This is adapted from the following project:

https://github.com/srsudar/tmux-completion

I've tried to follow the contributing guidelines and similar PRs that add a file. My `bash` knowledge is sufficiently low that some of the contributing guidelines are over my head. If I'm missing any I apologize.

Do you have any documentation on the testing framework you're using so I can write tests? Or a pointer to a command with a minimal test that can serve as an example?

PR #1364, opened 2025-04-10
feat(tmux): new completion

dseomn:
From https://github.com/tmux/tmux/pull/259 it doesn't look like the tmux maintainers were interested in having a bash completion script in the upstream repo. However, I added license headers to put these new files under either bash-completion's license or tmux's, in case they want it there in the future.

I'm aware of a handful of existing tmux bash completion scripts, below. As far as I can tell, they all hard-code a decent amount of tmux's available options, commands, etc. Some are also abandoned and out of date with more recent versions of tmux. Rather than base this code off of those, I decided to implement completion using tmux's own introspection commands as much as possible. Hopefully that will reduce the ongoing maintenance work and make it stay up to date with most tmux changes automatically.

This commit has a relatively minimal set of completions, see the TODO in _comp_cmd_tmux__value(). I have code for more completions in varying states of readiness, but this commit is already pretty large. I'll make follow-up PR(s) for those.

I'm willing to maintain this script. (And I'm hoping that the design mentioned above will make that easier.)

Existing implementations that I'm aware of:

* https://github.com/Bash-it/bash-it/blob/master/completion/available/tmux.completion.bash
* https://github.com/Boruch-Baum/tmux_bash_completion
* https://github.com/imomaliev/tmux-bash-completion
* https://github.com/scop/bash-completion/pull/81
* https://github.com/srsudar/tmux-completion

### DIFF
--- a/completions/tmux
+++ b/completions/tmux
@@ -1,0 +1,318 @@
+# tmux(1) completion                                       -*- shell-script -*-
+# SPDX-License-Identifier: GPL-2.0-or-later OR ISC
+
+# Log a message to help with debugging.
+# If BASH_COMPLETION_DEBUG is set, it will be printed to stderr.
+# When running with `set -x`, the _comp_cmd_tmux__log call itself will be
+# printed.
+#
+# @param $1 Message to log
+_comp_cmd_tmux__log()
+{
+    if [[ ${BASH_COMPLETION_DEBUG-} ]]; then
+        printf 'tmux bash completion: %s\n' "$1" >&2
+    fi
+}
+
+# Run the tmux command being completed.
+#
+# @param $@ args to tmux
+_comp_cmd_tmux__run()
+{
+    local -a REPLY
+    _comp_dequote "${comp_args[0]-}" || REPLY=("${comp_args[0]-}")
+    LC_ALL=C "${REPLY[0]}" "$@"
+}
+
+# Parse usage output from tmux.
+#
+# @param $1 Usage from tmux, not including the (sub)command name.
+# @var[out] options associative array mapping options to their value types, or
+#     to the empty string if the option doesn't take a value
+# @var[out] args indexed array of positional arg types
+_comp_cmd_tmux__parse_usage()
+{
+    options=()
+    args=()
+
+    local i j
+    local words
+    _comp_split words "$1"
+    for ((i = 0; i < ${#words[@]}; i++)); do
+        case ${words[i]} in
+            "[-"*"]")
+                # One or more options that don't take arguments, either of the
+                # form `[-abc]` or `[-a|-b|-c]`
+                for ((j = 2; j < ${#words[i]} - 1; j++)); do
+                    if [[ ${words[i]:j:1} != [-\|] ]]; then
+                        options+=(["-${words[i]:j:1}"]="")
+                    fi
+                done
+                ;;
+            "[-"*)
+                # One option that does take an argument.
+                if [[ ${words[i + 1]-} != *"]" ]]; then
+                    _comp_cmd_tmux__log \
+                        "Can't parse option: '${words[*]:i:2}' in '$1'"
+                    break
+                fi
+                options+=(["${words[i]#"["}"]="${words[i + 1]%"]"}")
+                ((i++))
+                ;;
+            -*)
+                _comp_cmd_tmux__log "Can't parse option '${words[i]}' in '$1'"
+                break
+                ;;
+            *)
+                # Start of positional arguments.
+                args=("${words[@]:i}")
+                break
+                ;;
+        esac
+    done
+
+    if [[ ${BASH_COMPLETION_DEBUG-} || -o xtrace ]]; then
+        local arg
+        for arg in "${!options[@]}"; do
+            _comp_cmd_tmux__log "option: ${arg} ${options["$arg"]}"
+        done
+        for arg in "${args[@]}"; do
+            _comp_cmd_tmux__log "arg: ${arg}"
+        done
+    fi
+}
+
+# Complete a value either as the argument to an option or as a positional arg.
+#
+# @param $1 subcommand that the value is for, or 'tmux' if it's top-level
+# @param $2 type of the value, from _comp_cmd_tmux__parse_usage()
+_comp_cmd_tmux__value()
+{
+    local subcommand=$1 option_type=$2
+    _comp_cmd_tmux__log \
+        "Trying to complete '$option_type' for subcommand '$subcommand'"
+
+    # To get a list of these argument types, look at `tmux -h` and:
+    #
+    # tmux list-commands -F "#{command_list_usage}" |
+    #   sed 's/[][ ]/\n/g' |
+    #   grep -v ^- |
+    #   sort -u
+    #
+    # TODO: Complete more option types.
+    case $option_type in
+        command)
+            _comp_compgen_split -l -- "$(_comp_cmd_tmux__run \
+                list-commands -F "#{command_list_name}")"
+            ;;
+        directory | *-directory)
+            _comp_compgen_filedir -d
+            ;;
+        file | *-file | path | *-path)
+            _comp_compgen_filedir
+            ;;
+    esac
+}
+
+# Parse command line options to tmux or a subcommand.
+#
+# @param $@ args to tmux or a subcommand, starting with the (sub)command
+#     itself, ending before the current word to complete
+# @var[in] options from _comp_cmd_tmux__parse_usage()
+# @var[out] option_type if the word to complete is the value of an option, this
+#     is the type of that value, otherwise it's empty
+# @var[out] positional_start if option_type is empty, index in $@ of the first
+#     positional argument, or the last index plus 1 if the next word is the
+#     first positional argument, or -1 if the next word could be either the
+#     first positional argument or another option
+_comp_cmd_tmux__options()
+{
+    local command_args=("$@")
+    option_type=""
+    positional_start=-1
+
+    local i
+    for ((i = 1; i < ${#command_args[@]}; i++)); do
+        if [[ $option_type ]]; then
+            # arg to the previous option
+            option_type=""
+        elif [[ ${command_args[i]} == -- ]]; then
+            option_type=""
+            ((positional_start = i + 1))
+            return
+        elif [[ ${command_args[i]} == -?* ]]; then
+            # 1 or more options, possibly also with the value of an option.
+            # E.g., if `-a` and `-b` take no values and `-c` does, `-ab` would
+            # be equivalent to `-a -b` and `-acb` would be `-a` and `-c` with a
+            # value of `b`.
+            local j
+            for ((j = 1; j < ${#command_args[i]}; j++)); do
+                if [[ $option_type ]]; then
+                    # arg has both the option and its value
+                    option_type=""
+                    break
+                fi
+                option_type=${options["-${command_args[i]:j:1}"]-}
+            done
+        else
+            # first positional arg
+            ((positional_start = i))
+            return
+        fi
+    done
+}
+
+# Complete arguments to a subcommand.
+#
+# @param $@ the subcommand followed by its args, ending before the current word
+#     to complete
+_comp_cmd_tmux__subcommand()
+{
+    local subcommand_args=("$@")
+    local usage=$(_comp_cmd_tmux__run list-commands \
+        -F "#{command_list_name} #{command_list_usage}" -- "$1" 2>/dev/null)
+    if [[ ! $usage ]]; then
+        _comp_cmd_tmux__log "Unknown tmux subcommand: '$1'"
+        return
+    fi
+    local subcommand=${usage%% *} # not $1, because it could be an alias
+    _comp_cmd_tmux__log "Attempting completion for 'tmux $subcommand'"
+
+    local -A options
+    local -a args
+    _comp_cmd_tmux__parse_usage "${usage#* }"
+
+    local option_type
+    local positional_start
+    _comp_cmd_tmux__options "${subcommand_args[@]}"
+
+    if [[ $option_type ]]; then
+        _comp_cmd_tmux__value "$subcommand" "$option_type"
+        return
+    elif ((positional_start < 0)) && [[ $cur == -* ]]; then
+        _comp_compgen -- -W '"${!options[@]}"'
+        return
+    elif ((positional_start < 0)); then
+        # $cur (one past the end of subcommand_args) is the first positional
+        # arg
+        positional_start=${#subcommand_args[@]}
+    fi
+
+    if [[ $subcommand == display-menu ]]; then
+        # display-menu has a non-trivial repeating pattern of positional args
+        # that would need custom logic to support correctly, and it's probably
+        # used in config files or shell scripts more than interactively anyway.
+        _comp_cmd_tmux__log \
+            "Not completing positional args for 'tmux $subcommand'"
+        return
+    fi
+
+    _comp_cmd_tmux__log \
+        "'tmux $subcommand' first positional arg: '${subcommand_args[positional_start]-}'"
+
+    local args_index=$positional_start
+    local usage_args_index
+    local prev_arg_type=""
+    for ((\
+    usage_args_index = 0;  \
+    usage_args_index < ${#args[@]};  \
+    args_index++, usage_args_index++)); do
+        local arg_type=${args[usage_args_index]##+(\[)}
+        arg_type=${arg_type%%+(\])}
+        if [[ $arg_type == ... ]]; then
+            if ((usage_args_index == 0)); then
+                _comp_cmd_tmux__log "'tmux $subcommand' first arg is '...'"
+                return
+            elif ((usage_args_index != ${#args[@]} - 1)); then
+                _comp_cmd_tmux__log \
+                    "'tmux $subcommand' usage has '...' before last arg"
+                return
+            fi
+            # https://man.openbsd.org/style says "Usage statements should take
+            # the same form as the synopsis in manual pages."
+            # https://mandoc.bsd.lv/mdoc/intro/synopsis_util.html says "Some
+            # commands can optionally take more than one argument of the same
+            # kind. This is indicated by an ellipsis trailing the respective Ar
+            # macro." tmux seems to mostly use `...` in that way to only repeat
+            # the previous argument. In display-menu it uses `...` to repeat
+            # all arguments instead, but that subcommand is explicitly ignored
+            # above.
+            _comp_cmd_tmux__value "$subcommand" "$prev_arg_type"
+            return
+        elif [[ $arg_type == arguments ]]; then
+            if [[ $prev_arg_type == command ]] &&
+                ((usage_args_index == ${#args[@]} - 1)); then
+                # The usage ends in `command arguments`, so recurse to the new
+                # subcommand.
+                _comp_cmd_tmux__subcommand \
+                    "${subcommand_args[@]:args_index-1}"
+                return
+            else
+                _comp_cmd_tmux__log \
+                    "'tmux $subcommand' has unsupported 'arguments' in usage"
+                return
+            fi
+        elif ((args_index == ${#subcommand_args[@]})); then
+            # The usage arg is 1 past the end of $subcommand_args, so complete
+            # it.
+            _comp_cmd_tmux__value "$subcommand" "$arg_type"
+            return
+        fi
+        prev_arg_type=$arg_type
+    done
+
+    _comp_cmd_tmux__log "Too many args to 'tmux $subcommand'"
+}
+
+_comp_cmd_tmux()
+{
+    local cur prev words cword comp_args
+    _comp_initialize -- "$@" || return
+
+    local usage
+    usage=$(_comp_cmd_tmux__run -h 2>&1)
+    # Before https://github.com/tmux/tmux/pull/4455 (merged 2025-04-09), `-h`
+    # produced usage information because it was an error, so we have to trim
+    # the error message too.
+    usage=${usage#$'tmux: unknown option -- h\n'}
+    usage=${usage#usage: tmux }
+
+    local -A options
+    local -a args
+    _comp_cmd_tmux__parse_usage "$usage"
+
+    local option_type
+    local positional_start
+    _comp_cmd_tmux__options "${words[@]:0:cword}"
+
+    if [[ $option_type ]]; then
+        _comp_cmd_tmux__value tmux "$option_type"
+        return
+    elif ((positional_start < 0)) && [[ $cur == -* ]]; then
+        _comp_compgen -- -W '"${!options[@]}"'
+        return
+    elif ((positional_start < 0)); then
+        ((positional_start = cword))
+    fi
+
+    local i
+    local -a REPLY
+    local subcommand_start=$positional_start
+    for ((i = positional_start; i < cword; i++)); do
+        if _comp_dequote "${words[i]}" && [[ ${REPLY[-1]-} =~ (\\*)\;$ ]] &&
+            ((${#BASH_REMATCH[1]} % 2 == 0)); then
+            # end of current command
+            ((subcommand_start = i + 1))
+        fi
+    done
+
+    if ((cword == subcommand_start)); then
+        _comp_cmd_tmux__value tmux command
+    else
+        _comp_cmd_tmux__subcommand \
+            "${words[@]:subcommand_start:cword-subcommand_start}"
+    fi
+} &&
+    complete -F _comp_cmd_tmux tmux
+
+# ex: filetype=sh
--- a/test/t/test_tmux.py
+++ b/test/t/test_tmux.py
@@ -1,0 +1,169 @@
+# SPDX-License-Identifier: GPL-2.0-or-later OR ISC
+
+import pytest
+
+from conftest import assert_complete
+
+
+@pytest.mark.bashcomp(cmd="tmux", require_cmd=True)
+class TestTmux:
+    # Tests for _comp_cmd_tmux__parse_usage(). Most of this function is tested
+    # elsewhere since almost everything else depends on it. These are just
+    # tests for corner cases that aren't tested elsewhere.
+
+    @pytest.mark.complete("tmux wait-for -")
+    def test_parse_usage_alternative_options(self, completion):
+        """Tests the [-a|-b|-c] form of option parsing."""
+        assert "-L" in completion
+        assert "-S" in completion
+        assert "-U" in completion
+        assert "-|" not in completion
+        assert "--" not in completion
+
+    # Tests for _comp_cmd_tmux__value()
+
+    @pytest.mark.complete("tmux new-")
+    def test_value_command(self, completion):
+        assert "new-session" in completion
+        assert "new-window" in completion
+
+    @pytest.mark.complete("tmux new-window -c ", cwd="shared/default")
+    def test_value_directory(self, completion):
+        assert completion == ["bar bar.d/", "foo.d/"]
+
+    @pytest.mark.complete("tmux -f ", cwd="shared/default")
+    def test_value_file(self, completion):
+        assert completion == ["bar", "bar bar.d/", "foo", "foo.d/"]
+
+    # Tests for _comp_cmd_tmux__options()
+
+    @pytest.mark.complete("tmux -f /dev/null ")
+    def test_option_with_value(self, completion):
+        assert "new-session" in completion
+
+    @pytest.mark.complete("tmux -Nvf /dev/null ")
+    def test_option_multiple_with_value(self, completion):
+        assert "new-session" in completion
+
+    @pytest.mark.complete("tmux -f -f ")
+    def test_option_with_value_with_dash(self, completion):
+        """Tests that the second -f is a filename not an option."""
+        assert "new-session" in completion
+
+    @pytest.mark.complete("tmux -- ")
+    def test_option_explicit_end(self, completion):
+        assert "new-session" in completion
+
+    @pytest.mark.complete("tmux -f/dev/null ")
+    def test_option_with_attached_value(self, completion):
+        assert "new-session" in completion
+
+    @pytest.mark.complete("tmux -Nvf/dev/null ")
+    def test_option_multiple_with_attached_value(self, completion):
+        assert "new-session" in completion
+
+    @pytest.mark.complete("tmux -v ")
+    def test_option_without_value(self, completion):
+        assert "new-session" in completion
+
+    @pytest.mark.complete("tmux -Nv ")
+    def test_option_multiple_without_value(self, completion):
+        assert "new-session" in completion
+
+    # Tests for _comp_cmd_tmux__subcommand()
+
+    @pytest.mark.complete("tmux this-is-not-a-real-subcommand-i-hope ")
+    def test_subcommand_unknown(self, completion):
+        assert not completion
+
+    @pytest.mark.complete("tmux new-window -c ", cwd="shared/default")
+    def test_subcommand_option_value(self, completion):
+        assert completion == ["bar bar.d/", "foo.d/"]
+
+    @pytest.mark.complete("tmux new-window -")
+    def test_subcommand_options(self, completion):
+        assert "-a" in completion  # takes no value
+        assert "-c" in completion  # takes a value
+
+    @pytest.mark.complete("tmux display-menu ")
+    def test_subcommand_no_positional_arg_completion(self, completion):
+        assert not completion
+
+    @pytest.mark.complete("tmux source-file abc def ", cwd="shared/default")
+    def test_subcommand_repetition(self, completion):
+        assert completion == ["bar", "bar bar.d/", "foo", "foo.d/"]
+
+    @pytest.mark.complete(
+        "tmux bind-key C-a new-window -c ",
+        cwd="shared/default",
+    )
+    def test_subcommand_recursion(self, completion):
+        assert completion == ["bar bar.d/", "foo.d/"]
+
+    @pytest.mark.complete("tmux source-file ", cwd="shared/default")
+    def test_subcommand_positional_arg_1(self, completion):
+        assert completion == ["bar", "bar bar.d/", "foo", "foo.d/"]
+
+    @pytest.mark.complete("tmux bind-key C-a ")
+    def test_subcommand_positional_arg_2(self, completion):
+        assert "new-session" in completion
+
+    @pytest.mark.complete("tmux start-server ")
+    def test_subcommand_no_positional_args(self, completion):
+        assert not completion
+
+    @pytest.mark.complete("tmux choose-tree abc def ghi ")
+    def test_subcommand_too_many_positional_args(self, completion):
+        assert not completion
+
+    # Tests for _comp_cmd_tmux()
+
+    @pytest.mark.complete("tmux -f ", cwd="shared/default")
+    def test_tmux_option_value(self, completion):
+        assert completion == ["bar", "bar bar.d/", "foo", "foo.d/"]
+
+    @pytest.mark.complete("tmux -")
+    def test_tmux_options(self, completion):
+        assert "-f" in completion  # takes a value
+        assert "-v" in completion  # takes no value
+
+    @pytest.mark.parametrize(
+        "other_commands",
+        [
+            r"foo ';'",
+            r"foo';'",
+            r"foo '\\;'",  # backslash escaped in tmux
+            r"foo'\\;'",
+        ],
+    )
+    def test_tmux_multiple_commands(self, bash, other_commands):
+        completion = assert_complete(
+            bash,
+            f"tmux {other_commands} bind-key C-a ",
+        )
+        assert "new-session" in completion
+
+    @pytest.mark.parametrize(
+        "semicolon_arg",
+        [
+            r"foo '\;'",  # semicolon escaped in tmux
+            r"foo'\;'",
+            r"foo '\\\;'",  # escaped backslash then escaped semicolon in tmux
+            r"foo'\\\;'",
+            r"foo';'bar",  # semicolon doesn't need to be escaped in the middle
+        ],
+    )
+    def test_tmux_semicolon_within_subcommand(self, bash, semicolon_arg):
+        completion = assert_complete(
+            bash,
+            # Note that the next arg is a file, not a subcommand. If it were a
+            # subcommand, it wouldn't be possible to tell if the completion was
+            # parsing semicolon_arg correctly.
+            f"tmux source-file {semicolon_arg} ",
+            cwd="shared/default",
+        )
+        assert completion == ["bar", "bar bar.d/", "foo", "foo.d/"]
+
+    @pytest.mark.complete("tmux ")
+    def test_tmux_subcommand(self, completion):
+        assert "new-session" in completion
--- a/completions/Makefile.am
+++ b/completions/Makefile.am
@@ -446,6 +446,7 @@ cross_platform = 2to3 \
 		tcpnice \
 		timeout \
 		tipc \
+		tmux \
 		_tokio-console \
 		tox \
 		tracepath \
--- a/test/t/Makefile.am
+++ b/test/t/Makefile.am
@@ -624,6 +624,7 @@ EXTRA_DIST = \
 	test_time.py \
 	test_timeout.py \
 	test_tipc.py \
+	test_tmux.py \
 	test_totem.py \
 	test_touch.py \
 	test_tox.py \


test/t/test_tmux.py
```diff
@@ -0,0 +1,167 @@
...
+            f"tmux source-file {semicolon_arg} ",
+            cwd="shared/default",
+        )
+        assert completion == ["bar", "bar bar.d/", "foo", "foo.d/"]
+
+    @pytest.mark.xfail(reason="_comp_cmd_tmux() not called at all")
+    @pytest.mark.complete(r"tmux foo \; new-")
+    def test_tmux_subsequent_command_name_with_prefix(self, completion):
+        # TODO: figure out why _comp_cmd_tmux() isn't being called in this case
+        assert "new-session" in completion
```
dseomn, line 163: I think this is a bug in bash? https://savannah.gnu.org/support/?111224

Should I just remove this test case since the relevant code here is already covered by other test cases?

dseomn:
At a glance, the test failure looks unrelated. Let me know if I need to do anything about it?

akinomyoga (commented):
The general quality of this PR seems very good, but I'm busy this week. I'll look at this later.

test/t/test_tmux.py
dseomn: The bug in bash was acknowledged, so I removed this test.

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+
+    local option_type
+    local positional_start
+    _comp_cmd_tmux__options "${words[@]:0:cword}"
+
+    if [[ $option_type ]]; then
+        _comp_cmd_tmux__value tmux "$option_type"
+        return
+    elif ((positional_start < 0)) && [[ $cur == -* ]]; then
+        _comp_compgen -U options -- -W '"${!options[@]}"'
```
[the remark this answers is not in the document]
dseomn: Gotcha, thanks for the explanation. I was a bit confused about `-U`.

dseomn:
> The general quality of this PR seems very good, but I'm busy this week. I'll look at this later.

Thanks, and no rush.

(In addition to the changes in comments above, I also added a new test `test_parse_usage_alternative_options`.)

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+# When running with `set -x`, the _comp_cmd_tmux__log call itself will be
+# printed.
+#
+# @param $1 Message to log
+_comp_cmd_tmux__log()
+{
+    if [[ ${BASH_COMPLETION_DEBUG-} ]]; then
+        printf 'tmux bash completion: %s\n' "$1" >&2
+    fi
+}
```
scop, line 15: I think I'd like to see this as a general purpose public API function rather than a tmux specific one. We'd likely want to make use of it from some other completions, too.

I seem to remember there might have been some related discussion/issue/pull request lately, but can't find it right now. Could be I'm just mistaken though.

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+    # tmux list-commands -F "#{command_list_usage}" |
+    #   sed 's/[][ ]/\n/g' |
+    #   grep -v ^- |
+    #   sort -u
+    #
+    # TODO: Complete more option types.
+    case "$option_type" in
+        command)
+            _comp_compgen_split -l -- \
+                "$(LC_ALL=C tmux list-commands -F "#{command_list_name}")"
```
scop, line 94: Would be good to pass around the original tmux command invoked instead of hardcoding just `tmux` from `$PATH` (or as an alias or a function) -- `tmux` might not be in `$PATH` (or the like) in the first place, or the `tmux` there might not be the one that was invoked (maybe another one with a full path to it was).

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+}
+
+# Complete arguments to a subcommand.
+#
+# @param $@ the subcommand followed by its args, ending before the current word
+#     to complete
+_comp_cmd_tmux__subcommand()
+{
+    local subcommand_args=("$@")
+    local usage="$(LC_ALL=C tmux list-commands \
```
scop, line 160: Hardcoded `tmux` consideration as above.

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+    _comp_cmd_tmux__log "Too many args to 'tmux $subcommand'"
+}
+
+_comp_cmd_tmux()
+{
+    local cur prev words cword comp_args
+    _comp_initialize -- "$@" || return
+
+    local usage
+    usage="$(LC_ALL=C tmux -h 2>&1)"
```
scop, line 245:
```suggestion
    usage="$(LC_ALL=C "$1" -h 2>&1)"
```

scop (changes requested):
Looks good overall to me on a quick glance, thanks!

Left a few remarks, but I'll happily defer remainder of this review to someone else, as I don't use tmux myself (yet, anyway, I guess).

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+            ((subcommand_start = i + 1))
+        elif [[ ${words[i]} =~ (\\*)\;\'$ ]] &&
+            ((${#BASH_REMATCH[1]} % 2 == 0)); then
+            # end of current command
+            ((subcommand_start = i + 1))
+        elif [[ ${words[i]} =~ (\\*)\;\"$ ]] &&
+            (((${#BASH_REMATCH[1]} + 1) % 4 <= 1)); then
+            # end of current command
+            ((subcommand_start = i + 1))
+        fi
```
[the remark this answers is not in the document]
akinomyoga: I'm wondering about the descriptions in `man tmux`:

> ```
>        As  in  these  examples,  when running tmux from the shell extra care
>        must be taken to properly quote semicolons:
>
>              1.   Semicolons that should be interpreted as a command separa‐
>                   tor should be escaped according to the shell  conventions.
>                   For  sh(1)  this typically means quoted (such as ‘neww ';'
>                   splitw’) or escaped (such as ‘neww \\\\; splitw’).
>
>              2.   Individual semicolons or trailing semicolons  that  should
>                   be  interpreted as arguments should be escaped twice: once
>                   according to the shell conventions and a second  time  for
>                   tmux; for example:
>
>                         $ tmux neww 'foo\\;' bar
>                         $ tmux neww foo\\\\; bar
> 
>              3.   Semicolons  that are not individual tokens or trailing an‐
>                   other token should only be escaped once according to shell
>                   conventions; for example:
>
>                         $ tmux neww 'foo-;-bar'
>                         $ tmux neww foo-\\;-bar
> ```

Are the numbers of backslashes correct in the above examples? It seems to me that they should be

```bash
neww ';' splitw
neww \\\; splitw

$ tmux neww 'foo \;' bar
$ tmux neww foo \\\; bar

$ tmux neww 'foo-;-bar'
$ tmux neww foo-\;-bar
```

completions/tmux
akinomyoga: That was #1356. Sorry, I disliked it because no code has used it in the codebase.

completions/tmux
akinomyoga: To add more context, one simple way is to use `"$1"` (at the top-level function) or `"${comp_words[0]}"` (in the helper functions that are called from the top-level function). A more robust way is to use `_comp_dequote "${comp_words[0]}"` as `_comp_compgen_help__get_help_lines` is doing:

https://github.com/scop/bash-completion/blob/90162b0a6cb31fe32acfe6af5079c9c4454dc8fb/bash_completion#L1561-L1569

Note: I try to slightly change the behavior of `_comp_dequote` in https://github.com/scop/bash-completion/pull/1357/commits/3e088b7a8e13a7c352eaad8b17c766baf35a680c (PR #1357), so that one does not need to prefix `|| REPLY=${comp_words[0]-}`.

completions/tmux
scop: I agree adding it "just in case" without any uses would have been premature. But here, we do have a use for it now.

I'm not 100% convinced the debug logging is something we need to keep around after this is in though. So I guess I'm still inclined to either generalize this, or just remove the debug logging from here before merging. No that strong opinions though, can leave it as is if others feel it's better that way. If we do, we could look into generalizing later instead of now when/if a 2nd or 3rd use pops up.

completions/tmux
akinomyoga: After checking the implementation, I think maybe we can just use this function `_comp_compgen_help__get_help_lines` for broader purposes (with a small adjustment). Even with the current implementation, we can use `_comp_compgen_help__get_help_lines` for the present purpose in the following way:

```bash
_comp_compgen_help__get_help_lines -- list-commands -F "#{command_list_name}" &&
  REPLY=("${_lines[@]}") &&
  _comp_compgen -- -W '"${REPLY[@]}"'
```

Then, we may consider renaming this function to `_comp_run_cmd` and change its output variables from `_lines` to `REPLY`.

completions/tmux
akinomyoga: > No that strong opinions though, can leave it as is if others feel it's better that way. If we do, we could look into generalizing later instead of now when/if a 2nd or 3rd use pops up.

My feeling is this. For now, we can confine this function inside `completions/tmux` because the only user is the `tmux` completion. When another completion file wants to use this function, we can move the function to the main `bash_completion` to avoid duplicates. From the perspective of the actual use in bash-completion, one use case is enough. In that sense, now that we have one for `tmux`, I'm not strongly opposed to adding this function as a general utility in `bash_completion`. However, from the perspective of abstraction, I usually would like to see at least two use cases to consider the proper interface (which is not too tightly bound to a specific use case).

akinomyoga: > I'm not 100% convinced the debug logging is something we need to keep around after this is in though.

From an objective point of view, the logged information seems only useful for developing the completion settings, and it doesn't seem useful in the actual situation of completing the command lines. Any outputs in the middle of the programmable completion break the terminal layout, so the completion setting shouldn't output anything to the terminal.

However, this feature would be interesting to me. I have to admit that this is out of scope for this project (bash-completion), but my framework (ble.sh) may utilize this kind of information in the actual situation. If an error message is provided by the completion setting, I can technically show the message below the command line to help users understand what is wrong. (In that case, I would overwrite the `_comp_log` function with a `ble.sh` version to *steal* the information).

dseomn:
Thanks for the reviews! I'll try to get to these comments soon.

completions/tmux
dseomn: > I'm not 100% convinced the debug logging is something we need to keep around after this is in though.

> From an objective point of view, the logged information seems only useful for developing the completion settings, and it doesn't seem useful in the actual situation of completing the command lines.

While it is useful for development, that actually was not the main reason I added it. The issue template below links to guidance to use `set -x`, and I wanted to have the debug logs in any bug reports.

https://github.com/scop/bash-completion/blob/90162b0a6cb31fe32acfe6af5079c9c4454dc8fb/.github/ISSUE_TEMPLATE/bug_report.md?plain=1#L36-L41

> Any outputs in the middle of the programmable completion break the terminal layout, so the completion setting shouldn't output anything to the terminal.

In general, I agree. Do you think it should be silent even when `$BASH_COMPLETION_DEBUG` is set though? As long as there's a way for me to see the log while developing it and when reading bug reports, I don't care too much about the specifics.

> However, this feature would be interesting to me. I have to admit that this is out of scope for this project (bash-completion), but my framework (ble.sh) may utilize this kind of information in the actual situation. If an error message is provided by the completion setting, I can technically show the message below the command line to help users understand what is wrong. (In that case, I would overwrite the `_comp_log` function with a `ble.sh` version to _steal_ the information).

Wouldn't you want different messages for that though, if they're geared more at users understanding why there's no completion, rather than developers fixing a bug? I guess that's what log levels are for in most logging frameworks. For now I'd rather just use something to help debug this completion, but I'd be fine splitting the log messages into info/warning/whatever for users and debug for developers if you or anyone else want to make the log framework support that.

completions/tmux
dseomn: `_comp_compgen_help__get_help_lines` is private, so I shouldn't use it as is, right?

Should I wait for changes to `_comp_compgen_help__get_help_lines` and/or `_comp_dequote`, or use `"${comp_args[0]}"` / `"$1"` for now?

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+        esac
+    done
+
+    local arg
+    for arg in "${!options[@]}"; do
+        _comp_cmd_tmux__log "option: ${arg} ${options["$arg"]}"
+    done
+    for arg in "${args[@]}"; do
+        _comp_cmd_tmux__log "arg: ${arg}"
+    done
```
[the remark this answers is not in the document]
dseomn: https://github.com/scop/bash-completion/blob/main/README.md#troubleshooting (linked from the bug report template) doesn't mention `$BASH_COMPLETION_DEBUG`, so that would prevent those lines from showing up in bug reports. Do you think the README should mention BASH_COMPLETION_DEBUG?

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+                _comp_cmd_tmux__log "'tmux $subcommand' first arg is '...'"
+                return
+            elif ((usage_args_index != ${#args[@]} - 1)); then
+                _comp_cmd_tmux__log \
+                    "'tmux $subcommand' usage has '...' before last arg"
+                return
+            fi
+            # Repeat from the beginning of args.
+            usage_args_index=-1
+            ((args_index--))
```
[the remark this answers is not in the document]
dseomn: With tmux 3.5a:
```
$ tmux list-commands | grep --fixed-strings ...
display-menu (menu) [-MO] [-b border-lines] [-c target-client] [-C starting-choice] [-H selected-style] [-s style] [-S border-style] [-t target-pane][-T title] [-x position] [-y position] name key command ...
send-keys (send) [-FHKlMRX] [-c target-client] [-N repeat-count] [-t target-pane] key ...
source-file (source) [-Fnqv] [-t target-pane] path ...
```

I'm not too familiar with `display-menu`, but I think it takes repeated `name key command` triples. Completing the commands might be useful, but I'm not sure. `send-keys` can send multiple keys, though completing keyboard keys probably doesn't make sense. `source-file` takes multiple files, which is useful to complete.

completions/tmux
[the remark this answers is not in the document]
dseomn: > Instead of trying to match several possible quoting patterns, can we use `_comp_dequote "${words[i]}"` to obtain the unquoted value and match it with `(\\*)\:$`?

I looked at that, but the thing about `$(echo xxx >/dev/tty)` looked risky. Looking at it again, does that mean that it's only insecure if the user explicitly used `declare -n`, not for a normal environment variable? If using `_comp_dequote` in situations like this is the generally accepted practice, I'll switch to that.

https://github.com/scop/bash-completion/blob/90162b0a6cb31fe32acfe6af5079c9c4454dc8fb/bash_completion#L212-L233

> I'm wondering about the descriptions in `man tmux`:

Agreed that the escaping there doesn't look right, so I filed https://github.com/tmux/tmux/issues/4480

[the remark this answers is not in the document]
dseomn: > Instead of trying to match several possible quoting patterns, can we use `_comp_dequote "${words[i]}"` to obtain the unquoted value and match it with `(\\*)\:$`?

I just realized another difference with that approach. That would prevent something like `tmux foo "$(bar) ;" <Tab>` from completing the second subcommand, right? Though I think it would be worth it to simplify the code rather than supporting cases like that which are hopefully rare.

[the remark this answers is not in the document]
akinomyoga: > Looking at it again, does that mean that it's only insecure if the user explicitly used `declare -n`, not for a normal environment variable?

Yes.

> If using `_comp_dequote` in situations like this is the generally accepted practice,

Many places in the bash-completion codebase still do not use `_comp_dequote`, but I actually want to use it in more places to complete quoted words.

> Agreed that the escaping there doesn't look right, so I filed [tmux/tmux#4480](https://github.com/tmux/tmux/issues/4480)

Thanks!

> That would prevent something like `tmux foo "$(bar) ;" <Tab>` from completing the second subcommand, right?

Yes, but the original heuristic list of quoting patterns is incomplete too. E.g., it doesn't work for a different quoting pattern such as `cmd';'`, `cmd";"`, and `"$(bar)"';'`. There is an unlimited number of quoting patterns. It also fails to detect `;` for `cmd1='command;'` & `tmux "$cmd1"`. With `_comp_dequote`, we *consistently* give up completion for words with command substitutions.

completions/tmux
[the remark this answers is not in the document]
akinomyoga: Thanks. I'll look at `display-menu` later.

completions/tmux
akinomyoga: > Do you think it should be silent even when `$BASH_COMPLETION_DEBUG` is set though?

No. `BASH_COMPLETION_DEBUG` is a part of development/debugging.

> Wouldn't you want different messages for that though, if they're geared more at users understanding

Yes, only some of the existing messages seem to be useful for that purpose.

> but I'd be fine splitting the log messages into info/warning/whatever for users and debug for developers if you or anyone else want to make the log framework support that.

I'd support the info message only after a significant number of completions support logging in a systematic way. However, that is clearly not the scope of this PR. The log level can be introduced when it becomes necessary, so I wouldn't suggest introducing it now.

completions/tmux
akinomyoga: > `_comp_compgen_help__get_help_lines` is private, so I shouldn't use it as is, right?

Yes, so I wrote

> > Then, we may consider renaming this function to `_comp_run_cmd` and change its output variable from `_lines` to `REPLY`.

----

> Should I wait for changes to `_comp_compgen_help__get_help_lines` and/or `_comp_dequote`, or use `"${comp_args[0]}"` / `"$1"` for now?

You can change `_comp_compgen_help__get_help_lines` if other maintainers agree with it.

completions/tmux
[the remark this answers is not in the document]
akinomyoga: As I've written in another reply, I doubt we need these in bug reports at the cost of always running loops even when it is working.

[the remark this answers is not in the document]
akinomyoga: Or another option is to use `[[ ${BASH_COMPLETION_DEBUG-} && -o xtrace ]]`.

completions/tmux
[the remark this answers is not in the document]
dseomn: If I'm understanding the `if (*args_string(args, i++) == '\0')` part of cmd-display-menu.c that you linked to, it looks like if the name option is the empty string, it has no key or command and the next arg is another name. And that matches https://github.com/tmux/tmux/blob/b4f8340f16fc6e0108f9af5ef7d4ee093eac8a8c/key-bindings.c#L27-L70 which looks like it has display-menu args. (Those are in tmux config file format, not shell format though, I think. So everything between `{` and `}` would be one argument in the shell.) I just ran this to check my understanding, and it displayed a menu with two options and a separator line between them:

```
tmux display-menu foo f 'display-message "picked foo"' '' bar b 'display-message "picked bar"'
```

While I could try to implement that correctly for completion, I think it's probably better to just disable completion entirely for display-menu? `name` and `key` don't seem useful to complete, and completing `command` within a quoted string seems like more effort than it's worth.

Relatedly, I think there's a spectrum of tmux subcommands from ones that are primarily useful interactively (e.g., `attach-session`) to ones that are primarily useful in the config file (e.g., `if-shell`). `display-menu` seems much more on the config file (or maybe shell script) side of that to me, so I don't think it's worth spending much time on completing its args unless some users speak up asking for it.

I would like to support completion for `source-file` though, since it's simpler and I could see using that interactively.

completions/tmux
dseomn: With the addition of `if [[ ${BASH_COMPLETION_DEBUG-} || -o xtrace ]]` around the loop like you suggested below, are you ok with leaving the logging in? Bash function calls and simple `[[` are generally pretty fast, right?

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+{
+    local cur prev words cword comp_args
+    _comp_initialize -- "$@" || return
+
+    local usage
+    usage="$(LC_ALL=C tmux -h 2>&1)"
+    # Before https://github.com/tmux/tmux/pull/4455 (merged 2025-04-09), `-h`
+    # produced usage information because it was an error, so we have to trim
+    # the error message too.
+    usage="${usage#$'tmux: unknown option -- h\n'}"
```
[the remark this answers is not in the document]
dseomn: > If one double-quotes the right-hand side of an assignment, which includes `$'...'` inside a variable expansion, it would be affected by `shopt -u extquote`. (Somehow `extquote` of bash-5.2 doesn't work, but I think it is a 5.2 bug. You can confirm the breakage with `bash <= 5.1`).

I didn't know about that, thanks!

> Similarly, other occurrences of double quotations for `var="$(...)"`, `var="${...}"`, and `case "$var" in` are also unnecessary. See https://github.com/scop/bash-completion/blob/main/doc/styleguide.md#quoting-of-words. We already have mixed conventions in the existing codebase, but we prefer less quoting for better highlighting support by text editors.

Done. I left quotes in `var+=(key="$value")` and `local var="$value"` though because I wasn't sure about those. Should I change those too?

completions/tmux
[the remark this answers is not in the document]
dseomn: Done, thanks for the suggestion about `-o xtrace`.

completions/tmux
```diff
@@ -0,0 +1,297 @@
...
+                # form `[-abc]` or `[-a|-b|-c]`
+                for ((j = 2; j < ${#words[i]} - 1; j++)); do
+                    if [[ ${words[i]:j:1} != [-\|] ]]; then
+                        options+=(["-${words[i]:j:1}"]="")
+                    fi
+                done
+                ;;
+            "[-"*)
+                # One option that does take an argument.
+                if [[ ${words[i + 1]} != *"]" ]]; then
```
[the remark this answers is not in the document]
dseomn: Good catch, done.

dseomn:
Just FYI, https://github.com/tmux/tmux/commit/68ffe654990764ed5bdb7efb88e4d01b921fd3d5 will change some of the usage syntax in the next tmux release. The changes look relatively minor, and I'd like to support both the old and new syntaxes for at least a little while. And since this PR is already rather large, I'd like to go forward with this just supporting the syntax in released tmux versions, then make another PR adding support for the new changes, if that's ok.

completions/tmux
akinomyoga: I'm fine with leaving the logging for now.

completions/tmux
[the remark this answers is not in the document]
akinomyoga: Thanks for updating.

- `var+=(key="$value")` is not technically an assignment, so the quoting is mandatory.
- `local var="$value"` is an assignment (unless `local` is replaced by an alias), so they can be removed.

I prefer to remove the double quotation of the latter, `local var="$value"`, but it's not mandatory. We currently have mixed conventions in the codebase anyway, and we may or may not change them all at once in the future.

completions/tmux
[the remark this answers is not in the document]
akinomyoga: Thanks!

completions/tmux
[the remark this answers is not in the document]
akinomyoga: Thank you. I'm not sure if that `''` is documented but

> I just ran this to check my understanding, and it displayed a menu with two options and a separator line between them:
> 
> ```
> tmux display-menu foo f 'display-message "picked foo"' '' bar b 'display-message "picked bar"'
> ```

Ah, OK. I now checked `man tmux`, where the empty argument is mentioned:

<pre>
       display-menu    [-OM]   [-b   border-lines]   [-c   target-client]   [-C
               starting-choice]   [-H   selected-style]    [-s    style]    [-S
               border-style]  [-t  target-pane]  [-T  title]  [-x position] [-y
               position] name key command [argument ...]
                     (alias: menu)
               Display a menu on target-client.  target-pane gives  the  target
               for any commands run from the menu.

               A  menu  is passed as a series of arguments: first the menu item
               name, second the key shortcut (or empty for none) and third  the
               command  to run when the menu item is chosen.  The name and com‐
               mand are formats, see the “FORMATS” and “STYLES”  sections.   If
               the  name  begins  with  a hyphen (-), then the item is disabled
               (shown dim) and may not be chosen.  <b>The name may be empty for  a
               separator line, in which case both the key and command should be
               omitted.</b>
</pre>

> While I could try to implement that correctly for completion, I think it's probably better to just disable completion entirely for display-menu? `name` and `key` don't seem useful to complete, and completing `command` within a quoted string seems like more effort than it's worth.

Yeah, I understand it. You may disable it or just leave it as is.

completions/tmux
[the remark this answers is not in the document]
akinomyoga: > I think the original implementation did handle `cmd';'`, `cmd";"`, and `"$(bar)"';'`

Ah, I'm sorry, you are right. The examples should have been something like `cmd\\';'`, `cmd\\\\";"`, and `"$(bar)\\\\\\"';'`.

akinomyoga:
> And since this PR is already rather large, I'd like to go forward with this just supporting the syntax in released tmux versions,

OK.

completions/tmux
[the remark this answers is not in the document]
dseomn: > `local var="$value"` is an assignment (unless `local` is replaced by an alias), so they can be removed.

Huh. That does match the behavior I'm seeing:

```
bash-5.2$ bar='bar quux'; f() { local foo=$bar; echo "$foo"; }; f
bar quux
```

But I don't see anything in https://www.gnu.org/software/bash/manual/bash#index-local about `local` expanding its arguments any differently than a normal command/builtin. From that documentation, I would have expected `local foo=$bar` above to pass two arguments to `local`, `foo=bar` and `quux`. Am I reading that documentation wrong? Or is it documented somewhere else?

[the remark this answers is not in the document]
akinomyoga: > https://www.gnu.org/software/bash/manual/bash#Shell-Parameters
>
> A variable may be assigned to by a statement of the form
> ```
> name=[value]
> ```
>
> If value is not given, the variable is assigned the null string. All values undergo tilde expansion, parameter and variable expansion, command substitution, arithmetic expansion, and quote removal (see [Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/bash#Shell-Parameter-Expansion)). If the variable has its integer attribute set, then value is evaluated as an arithmetic expression even if the $((…)) expansion is not used (see [Arithmetic Expansion](https://www.gnu.org/software/bash/manual/bash#Arithmetic-Expansion)). **Word splitting and filename expansion are not performed. Assignment statements may also appear as arguments to the alias, declare, typeset, export, readonly, and local builtin commands (declaration commands).** When in POSIX mode (see [Bash POSIX Mode](https://www.gnu.org/software/bash/manual/bash#Bash-POSIX-Mode)), these builtins may appear in a command after one or more instances of the command builtin and retain these assignment statement properties.

completions/tmux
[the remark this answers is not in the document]
dseomn: Ok, I disabled it.

completions/tmux
dseomn: I added `_comp_cmd_tmux__run` to do this without `_comp_compgen_help__get_help_lines` for now. Happy to switch to that later if maintainers want to rename it and make it public.

> `_comp_dequote "${comp_args[0]-}" || REPLY=${comp_args[0]-}`

`_comp_dequote`'s documentation says that `REPLY` is an array, so I didn't use that exactly. Should `_comp_compgen_help__get_help_lines` be changed too?

I also updated the semicolon detection to treat `REPLY` as an array too.

completions/tmux
dseomn: Done.

completions/tmux
dseomn: Done, with `_comp_cmd_tmux__run` instead of `$1`.

completions/tmux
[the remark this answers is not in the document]
dseomn: Thanks for the pointer! I removed the double quotes around `local` assignments too.

dseomn:
I think I've addressed all the comments now, let me know if there's anything else I should change. (Or if I missed anything.)

In addition to changes from the comments, I also added one more test case that https://github.com/tmux/tmux/issues/4480 reminded me about:

```
r"foo';'bar",  # semicolon doesn't need to be escaped in the middle
```

completions/tmux
```diff
@@ -0,0 +1,311 @@
...
+}
+
+# Run the tmux command being completed.
+#
+# @param $@ args to tmux
+_comp_cmd_tmux__run()
+{
+    local -a REPLY
+    _comp_dequote "${comp_args[0]-}" || REPLY=("${comp_args[0]-}")
+    LC_ALL=C "${REPLY[@]}" "$@"
```
akinomyoga, line 24: Even if the command name produces multiple words (which happens when pathname expansions produce multiple filenames, etc.), what we want here is the command name, so I think we should use the first word. For example, even if `tmux*` would produce `tmux tmux1 tmux2`, what we want to execute is `tmux --help` instead of `tmux tmux1 tmux2 --help`.

```suggestion
    LC_ALL=C "${REPLY[0]}" "$@"
```

akinomyoga (approved):

completions/tmux
akinomyoga: OK, we can update `_comp_compgen_help__get_help_lines` (`_comp_run_cmd`) in another PR.

akinomyoga: > Should `_comp_compgen_help__get_help_lines` be changed too?

Yeah, and it is actually addressed in https://github.com/scop/bash-completion/commit/3e088b7a8e13a7c352eaad8b17c766baf35a680c in #1357 (where `REPLY` is assigned after `REPLY=()`). That was actually [a fix](https://github.com/scop/bash-completion/compare/372d4992eebecb338ab1b5fdd78bdbd850b91217..ebe65544ee9814985d4dd92bce3c24942787f81f) after [the CI \#2838 failure](https://github.com/scop/bash-completion/actions/runs/14282501108/job/40033674783).

completions/tmux
dseomn: Done. I was thinking that with something like the below it would make sense to include all words, but bash didn't seem to use the custom completion for that.

```
$ tmux='tmux -L foo'
$ $tmux <Tab>
```

completions/tmux
```diff
@@ -0,0 +1,311 @@
...
+{
+    local cur prev words cword comp_args
+    _comp_initialize -- "$@" || return
+
+    local usage
+    usage=$(_comp_cmd_tmux__run -h 2>&1)
+    # Before https://github.com/tmux/tmux/pull/4455 (merged 2025-04-09), `-h`
+    # produced usage information because it was an error, so we have to trim
+    # the error message too.
+    usage=${usage#$'tmux: unknown option -- h\n'}
```
akinomyoga, line 270: I now realized that `tmux --help` worked in a desired way even before `tmux -h` is adjusted as mentioned above. I can confirm `tmux --help` works at least with version `2.9a` (I haven't tried an older version).

```bash
$ ~/.opt/tmux/2.9a/bin/tmux --help
usage: tmux [-2CluvV] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [command [flags]]
$ ~/.opt/tmux/2.9a/bin/tmux -h
tmux: unknown option -- h
usage: tmux [-2CluvV] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [command [flags]]
```

Shouldn't we just use `tmux --help 2>&1` instead of trying to adjust the resutl of `tmux -h 2>&1`?

completions/tmux
akinomyoga: The user might register `complete -F _comp_cmd_tmux '$tmux'`. I also have an unsubmitted branch [`akinomyoga:__load_completion-dequote`](https://github.com/akinomyoga/bash-completion/tree/__load_completion-dequote) that includes commit 7ed6f6c479f80de7c821386c71a2f2b4f6da35e7, which attempts to enable completions for the command line `$tmux`, etc. In that case, I think we want to run `tmux ...` to generate completions (instead of `tmux -L foo ...`).

completions/tmux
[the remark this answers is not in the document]
akinomyoga: This looping behavior ("`# Repeat from the beginning of args.`") seems non-trivial. Now, the support for `display-menu` is explicitly disabled, this looping is actually never used. If this behavior would be left, I think the background about `display-menu` should be described in code comments.

completions/tmux
dseomn: I originally used `--help` in an earlier version of this code before I made the PR, but I don't think its behavior is intentional:

https://github.com/tmux/tmux/blob/36fba4cc54d543bec74a08597b89a019cb6edde3/tmux.c#L382

```
$ tmux --this-is-definitely-not-a-valid-option
usage: tmux [-2CDlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
```

I haven't tested this, but I'm guessing that `getopt` is returning something that hits the `default` branch of the `switch`. I also wouldn't be surprised if some implementations of getopt print an error message in that case.

I figured it was better to add an intentional and documented way to get the usage output, and rely on an easily understood error message in earlier versions, rather than use an undocumented and (I think) unintentional bug/feature.

completions/tmux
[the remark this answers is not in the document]
dseomn: `source-file` still uses it.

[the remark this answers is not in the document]
akinomyoga: According to https://github.com/scop/bash-completion/pull/1364#discussion_r2051850594, it doesn't loop from the beginning, but it just repeats the argument previous to `...`:

> ```
> source-file (source) [-Fnqv] [-t target-pane] path ...
> ```

This is a more natural interpretation than "repeating from the beginning" (although it can *technically* be considered to be repeating from the beginning because there is only one argument). Honestly, I have to say that the current behavior of *repeating from the beginning* would be surprising behavior of the command-line parsing if one doesn't know the special treatment of `display-menu`. Hence, I think some clarification is needed in the code comments. Or if you don't want to leave code comments, could you change the behavior to just repeat the argument previous to `...`?

completions/tmux
akinomyoga: I confirmed that this inconsistency between the unrecognized long and short options is still present with the current `master` branch of `tmux` (commit 36fba4c):

```bash
$ ./tmux --this-is-definitely-not-a-valid-option
usage: tmux [-2CDhlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
$ ./tmux -@
tmux: unknown option -- @
usage: tmux [-2CDhlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
```

dseomn: > Shouldn't this be reported to the upstream tmux? If tmux (before the mentioned fix for `-h`) reported the unrecognized option `-h`, I think `tmux` should also print the error about the unrecognized `--this-is-definitely-not-avalid-option`.

I looked into it a bit more, and I think tmux always uses OpenBSD's getopt implementation, either directly or with https://github.com/tmux/tmux/blob/master/compat/getopt.c. I'm not too familiar with the history of option parsing, but I vaguely remember something about long options being from GNU, and BSDs generally not wanting to copy that style, I think. So I do think it would be nice to get an error message for an unrecognized long option, but it doesn't seem that important to me in this case and (if I'm remembering right) I don't really want to come across as pushing for GNU behavior in a BSD. If you think it's important for bash completion though, I can file a bug.

completions/tmux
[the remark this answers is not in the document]
dseomn: Gotcha. I added a comment, but I didn't go into much detail because https://github.com/tmux/tmux/commit/68ffe654990764ed5bdb7efb88e4d01b921fd3d5 changes the meaning of `...` somewhat. I'll update the comment in another PR along with the changes for that commit.

completions/tmux
akinomyoga: I believe that not showing error messages for unrecognized long options is not forced by the BSD style.

It is understandable that tmux doesn't support long options if tmux is in the BSD style, but it is still strange that tmux doesn't show the error message, given that tmux recognizes the error as it shows the usage. Rather, if it is truely in the BSD style, it should interpret `--this-is-definitely-not-a-valid-option` as `--`, `-t`, `-h`, `-i`, `-s`, `--`, `-i`, `-s`, `--`, `d`, `-e`, `-f`, `-i`, `-n`, `-i`, `-t`, `-e`, `-l`, `-y`, `--`, `-n`, `-o`, `-t`, `--`, `-a`, `--`, `-v`, `-a`, `-l`, `-i`, `-d`, `--`, `-o`, `-p`, `-t`, `-i`, `-o`, `-n` (where "`--`" appearing in a combined option should be treated differently from the independent `--`).

For example, a set of the option flags (excluding `-`) still shows the following error message:

```console
$ tmux -thisisdefinitelynotavalidoption
tmux: unknown option -- t
usage: tmux [-2CDlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
```

Or, it doesn't show the error mesesage for `--@` while it does for `-@`.

```console
$ tmux --@
usage: tmux [-2CDlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
$ tmux -@
tmux: unknown option -- @
usage: tmux [-2CDlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
```

OpenSSH is a part of OpenBSD, but it shows the following error message:

```console
$ ssh --this-is-definitely-not-a-valid-option
unknown option -- -
usage: ssh [-46AaCfGgKkMNnqsTtVvXxYy] [-B bind_interface] [-b bind_address]
           [-c cipher_spec] [-D [bind_address:]port] [-E log_file]
           [-e escape_char] [-F configfile] [-I pkcs11] [-i identity_file]
           [-J destination] [-L address] [-l login_name] [-m mac_spec]
           [-O ctl_cmd] [-o option] [-P tag] [-p port] [-R address]
           [-S ctl_path] [-W host:port] [-w local_tun[:remote_tun]]
           destination [command [argument ...]]
       ssh [-Q query_option]
```

Isn't `getopt.c` used by tmux simply too old?

completions/tmux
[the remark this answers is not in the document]
akinomyoga: Thanks.

completions/tmux
dseomn: Makes sense, I filed https://github.com/tmux/tmux/issues/4491

> Isn't `getopt.c` used by tmux simply too old?

If I'm reading https://cvsweb.openbsd.org/src/lib/libc/stdlib/ correctly (which I might not be, I've never used CVS), it's 21 years old and obsolete.

completions/tmux
```diff
@@ -0,0 +1,314 @@
...
+                _comp_cmd_tmux__log "'tmux $subcommand' first arg is '...'"
+                return
+            elif ((usage_args_index != ${#args[@]} - 1)); then
+                _comp_cmd_tmux__log \
+                    "'tmux $subcommand' usage has '...' before last arg"
+                return
+            fi
+            # In tmux versions before
+            # https://github.com/tmux/tmux/commit/68ffe654990764ed5bdb7efb88e4d01b921fd3d5
+            # (2025-04-09), ... meant to repeat from the beginning of args.
```
akinomyoga, line 234: Does this mean `display-menu` will break the backward compatibility in the next version? As far as I test this with the `master` branch of tmux, `display-menu` still has the current behavior of repeating from the beginning (except for the empty *name* `''` meant for a separator).

If you mean *the empty name `''` as a separator* by the 2025-04-09 change, the behavior actually existed before the 2025-04-09 change despite that the usage did not reflect it.

What I expected in code comments is the fact that the behavior of looping from the beginning of the argument list is used by (or effective with) the argument parsing of `display-menu`.

completions/tmux
akinomyoga: Thank you!

completions/tmux
dseomn: > Does this mean `display-menu` will break the backward compatibility in the next version? As far as I test this with the `master` branch of tmux, `display-menu` still has the current behavior of repeating from the beginning (except for the empty _name_ `''` meant for a separator).
> 
> If you mean _the empty name `''` as a separator_ by the 2025-04-09 change, the behavior actually existed before the 2025-04-09 change despite that the usage did not reflect it.

I don't think the behavior is changing at all, just the way it shows up in the usage.

> What I expected in code comments is the fact that the behavior of looping from the beginning of the argument list is used by (or effective with) the argument parsing of `display-menu`.

Currently, I think they all loop from the beginning, not just `display-menu`. With the new changes, some loop from the beginning (`foo ...`) and some only repeat the previous argument (`[foo ...]`). But `[command [argument ...]]` and `[shell-command [argument ...]]` will need to be handled specially anyway because `argument` is only meaningful in the context of the `command` or `shell-command`.

dseomn: Oh I see. I thought `...` without any grouping like parentheses or brackets meant to repeat from the beginning, but https://man.openbsd.org/style says "Usage statements should take the same form as the synopsis in manual pages" and https://mandoc.bsd.lv/mdoc/intro/synopsis_util.html says "Some commands can optionally take more than one argument of the same kind. This is indicated by an ellipsis trailing the respective Ar macro." So it looks like OpenBSD doesn't require grouping punctuation to make `...` apply to only one argument. I'll take another look at this and update the code or comment.

dseomn: I changed the code to only repeat the last arg instead of all of them. Thanks for catching that, I wasn't too familiar with OpenBSD's usage style.

akinomyoga (approved):
Thank you for all your investigation and updates! LGTM. The following points can be addressed later if necessary:

- Move `_comp_cmd_tmux__log` to `bash_completion` with the name `_comp_log`
- Update `_comp_compgen_help__get_help_lines` for more general use cases

dseomn:
Since the CI tests are failing (hopefully to be fixed by #1373):

```
$ shellcheck completions/tmux && shfmt --diff completions/tmux && podman run --rm -ti -v ~/Code/bash-completion:/home/dseomn/bash-completion:ro localhost/dev-environment:latest bash -c 'cd bash-completion && pytest test/t/test_tmux.py'
===================================================================================================================================================================================== test session starts =====================================================================================================================================================================================
platform linux -- Python 3.13.2, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/dseomn/bash-completion/test
configfile: setup.cfg
plugins: cov-5.0.0, typeguard-4.4.2
collected 34 items                                                                                                                                                                                                                                                                                                                                                                            

test/t/test_tmux.py ..................................                                                                                                                                                                                                                                                                                                                                  [100%]

====================================================================================================================================================================================== warnings summary =======================================================================================================================================================================================
../../../usr/lib/python3/dist-packages/_pytest/cacheprovider.py:475
  /usr/lib/python3/dist-packages/_pytest/cacheprovider.py:475: PytestCacheWarning: could not create cache path /home/dseomn/bash-completion/test/.pytest_cache/v/cache/nodeids: [Errno 30] Read-only file system: '/home/dseomn/bash-completion/test/pytest-cache-files-yf3g6vl0'
    config.cache.set("cache/nodeids", sorted(self.cached_nodeids))

../../../usr/lib/python3/dist-packages/_pytest/stepwise.py:51
  /usr/lib/python3/dist-packages/_pytest/stepwise.py:51: PytestCacheWarning: could not create cache path /home/dseomn/bash-completion/test/.pytest_cache/v/cache/stepwise: [Errno 30] Read-only file system: '/home/dseomn/bash-completion/test/pytest-cache-files-jy64przc'
    session.config.cache.set(STEPWISE_CACHE_DIR, [])

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=============================================================================================================================================================================== 34 passed, 2 warnings in 13.75s ===============================================================================================================================================================================
```

(Ignore the `Read-only file system` warnings, that's just because I mounted the directory read-only.)

dseomn:
Is there a style guide for the function documentation syntax that bash completion uses? I'm starting to work on the changes for https://github.com/tmux/tmux/commit/68ffe654990764ed5bdb7efb88e4d01b921fd3d5 and wrote the below for a new function. Is there a better way to say that `$1` is the arg type and the rest of `$@` is the command and its arguments?

```sh
# @param $@ the arg type of the command (e.g., command or shell-command), then
#     the command itself, then the arguments to the command, ending before the
#     current word to complete
```

akinomyoga:
> `$1` is the arg type and the rest of `$@` is the command and its arguments?

There isn't a precise documentation syntax because currently no tools parse them, but an existing use suggests `$2...` for the above use case:

https://github.com/scop/bash-completion/blob/1cdff0a7aeec83b2b310666c342720ec82f071d4/bash_completion#L3378-L3386

dseomn:
What are the next steps for this PR? Is it just blocked on the CI being fixed?

akinomyoga:
How about the situation with https://github.com/scop/bash-completion/pull/1364#issuecomment-2836143550? I thought you planned to change it.

dseomn:
Oh, sorry for the confusion. That question was about the follow-up that's blocked on this PR, not about this PR. Once this is merged I'll test the code that that question was about, and make another PR.

akinomyoga (approved):

akinomyoga:
Thanks!

dseomn:
Thank you! I'll try to make the PR to update this for the latest unreleased version of tmux within the next few days.

vladfi1:
Am I right that session name completion (`tmux a -t ...`) isn't supported yet? Just want to make sure I haven't installed something incorrectly.

yedayak:
> Am I right that session name completion (`tmux a -t ...`) isn't supported yet? Just want to make sure I haven't installed something incorrectly.

This was added in #1534, it will be available in the next version

vladfi1:
Hm that's odd, I installed from source and still wasn't getting session name completion.

dseomn:
Could you set `BASH_COMPLETION_DEBUG=1`, try completing a session name, and share the output? (Feel free to redact the actual names of the sessions if you want.)

akinomyoga:
> Hm that's odd, I installed from source and still wasn't getting session name completion.

Where did you install it?

What is the result of the following command *after attempting the tmux completion in the session*?

```console
$ (shopt -s extdebug; declare -F _comp_cmd_tmux _tmux)
```

PR #1365, opened 2025-04-13
feat(shtab): add 3rd party fallback completion loader

scop:
https://github.com/iterative/shtab/blob/b83be0581c6cbd46d17b70b7faf235b2e0ba359e/shtab/main.py#L29


akinomyoga:
For the record for myself, the `--print-own-completion` option of `shtab` seems to have been suggested in https://github.com/iterative/shtab/issues/89, added in https://github.com/iterative/shtab/pull/114, and released in [shtab-v1.6.0-beta](https://github.com/iterative/shtab/releases/tag/v1.6.0) (2023-04-02).

The `shtab` framework seems to provide completions to other Python programs as described [here](https://docs.iterative.ai/shtab/use/#library-usage), but it seems we already support it for `keyring` (#1014) (but only for `keyring`). I searched GitHub for `--print-completion bash` [[search 1](https://github.com/search?q=%2F%20--print-completion%20bash%2F&amp;type=code) and [search 2](https://github.com/search?q=%2F+--print-completion+bash%2F+AND+NOT+%2Ftldr%7Ckeyring%7Ctrash-empty%2F&amp;type=code)]. It seems I can find many other tools that use `--print-completion bash` such as `tldr` and `trash{,-empty,-list,-restore,-put}`, `glances`, `btcli`. There are many other minor tools with less stars.

There is an open PR at https://github.com/pygments/pygments/pull/2229.

akinomyoga:
I added completion loaders for glances (421859654c94c9864c8d93ca539295024f480504), tldr (298700a0f34e7cd72b1c7535e0993b8d350e5c2e), and trash-cli (533244c7603b7a139a56549c0b3238c16995a7e7).

akinomyoga (approved):

scop:
Thanks! There's also a class of tools that might work with shtab's [CLI usage](https://docs.iterative.ai/shtab/use/#cli-usage), but those are hard to search and such code structures could be coincidental and thus brittle.

akinomyoga:
> Thanks!

Thank you!

> There's also a class of tools that might work with shtab's [CLI usage](https://docs.iterative.ai/shtab/use/#cli-usage), but those are hard to search and such code structures could be coincidental and thus brittle.

Yeah, we can still search for [`/shtab --shell=bash /`](https://github.com/search?q=%2Fshtab+--shell%3Dbash+%2F&type=code), but the problem with this approach is that one needs to know the internal structure of each tool to identify the `X.Y.Z` part of `shtab --shell=bash -u X.Y.Z`. We cannot use a common loader but need to prepare a loader for each tool. I guess the command `shtab --shell=bash ... | tee /usr/share/bash-completion/completions/<tool>` is supposed to be run by the installer of the tool.

PR #1367, opened 2025-04-14
Run sshd in CI

yedayak:
This should allow us to make better tests for remote files in scp and rsync completions.

CI in this PR fails because the images are only pushed after merging

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,6 +71,7 @@ jobs:
           --env BSD=${{matrix.bsd}}
           --env PYTESTFLAGS="--verbose -p no:cacheprovider"
           --env NETWORK=$NETWORK
+          --env BASH_COMPLETION_TEST_LIVE_SSH_HOST=localhost
           ${NETWORK:+--network $NETWORK}
           --volume $PWD:/usr/src/bash-completion
           --workdir /usr/src/bash-completion
--- a/test/docker/entrypoint.sh
+++ b/test/docker/entrypoint.sh
@@ -13,6 +13,17 @@ oldpwd=$(pwd)
 cp -a . /work
 cd /work
 
+# Dependencies for running sshd and connecting to it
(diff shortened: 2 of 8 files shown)


scop:
Some conflicts to remove, and `ssh-keygen` made available at least on alpine.

test/docker/entrypoint.sh
```diff
@@ -13,6 +13,17 @@ oldpwd=$(pwd)
 cp -a . /work
 cd /work
 
+# Dependencies for running sshd and connecting to it
+ssh-keygen -A
+mkdir -p ~/.ssh/ /var/run/sshd
+echo "localhost $(cat /etc/ssh/ssh_host_ed25519_key.pub)" >>~/.ssh/known_hosts
+ssh-keygen -t ed25519 -N '' -f "$HOME/.ssh/id_ed25519"
```
scop, line 20: At least on alpine:
> `test/docker/entrypoint.sh: line 17: ssh-keygen: not found`

https://github.com/scop/bash-completion/actions/runs/14454159615/job/40533467548?pr=1367#step:4:46

scop: Nevermind, falls in the noted "CI in this PR fails because the images are only pushed after merging" category I suppose, openssh-server does pull in openssh-keygen which has ssh-keygen.

.github/workflows/ci.yaml
```diff
@@ -71,6 +71,7 @@ jobs:
           --env BSD=${{matrix.bsd}}
           --env PYTESTFLAGS="--verbose -p no:cacheprovider"
           --env NETWORK=$NETWORK
+          --env LIVE_SSH_HOST=localhost
```
scop, line 74: Let's rename this envvar to `BASH_COMPLETION_TEST_LIVE_SSH_HOST`, pre-approved with that change.

scop (approved):

PR #1371, opened 2025-04-20
fix(scp,rsync): fix misc bugs and refactor

akinomyoga:
I separated small changes out of #1357. There are still eight commits, but I think they are mostly trivial fixes. If this is still too large as a PR, I'll further split the PR.

For these small changes, there are no ongoing discussions in #1357.

I'm thinking of separating the other *non-trivial* changes into PRs with smaller numbers of commits. I'll submit them after merging this PR.


scop (approved):
Thanks! Looks like shellcheck 0.11.0 does not flag those either.

akinomyoga:
Thank you!

PR #1373, opened 2025-04-24
fix(xfreerdp): try /list:kbd also with dash syntax

yedayak:
This caused failures on recent fedoradev rawhide tests, since it seems to not be built with deprecated options like `--list-kbd` anymore.

This introduces an internal utility function to list kbd layouts, which is used in two places.

I'm not sure about how the usage of the function, I couldn't find other utility functions that don't just complete something right away, but I didn't look very thoroughly. Maybe returning the data through a variable like REPLY makes more sense?

### DIFF
--- a/completions/xfreerdp
+++ b/completions/xfreerdp
@@ -1,5 +1,16 @@
 # xfreerdp completion                                      -*- shell-script -*-
 
+_comp_cmd_xfreerdp__kbd_list()
+{
+    local kbd_list
+    # Trying non deprecated /list:kbd first
+    kbd_list=$("$1" /list:kbd 2>/dev/null) ||
+        # Old syntax, deprecated in 2022-10-19
+        # See https://github.com/FreeRDP/FreeRDP/commit/119b8d4474ab8578101f86226e0d20a53460dd51
+        kbd_list=$("$1" /kbd-list 2>/dev/null)
+    _comp_awk '/^0x/ { print $1 }' <<<"$kbd_list"
+}
+
 _comp_cmd_xfreerdp()
 {
     local cur prev words cword comp_args
@@ -7,8 +18,7 @@ _comp_cmd_xfreerdp()
 
     case $prev in
         -k)
-            _comp_compgen_split -- "$("$1" --kbd-list |
-                _comp_awk '/^0x/ { print $1 }')"
+            _comp_compgen_split -- "$(_comp_cmd_xfreerdp__kbd_list "$1")"
             return
             ;;
         -a)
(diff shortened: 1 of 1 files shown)


completions/xfreerdp
```diff
@@ -1,14 +1,23 @@
 # xfreerdp completion                                      -*- shell-script -*-
 
+_comp_cmd_xfreerdp__kbd_list()
+{
+    local kbd_list
+    # Trying non deprecated /list:kbd first
+    kbd_list=$("$1" /list:kbd 2>/dev/null) ||
+        kbd_list=$("$1" --kbd-list 2>/dev/null) ||
+        kbd_list=$("$1" /kbd-list 2>/dev/null)
```
akinomyoga, line 9: I'm not familiar with xfreerdp, but according to [this page](https://github.com/awakecoding/FreeRDP-Manuals/blob/88c0158c896565214e6b145f67e95efe8dee883f/User/FreeRDP-User-Manual.markdown#syntax), the windows style `/kbd-list` seems to be preferred to the posix-style `--kbd-list`.

Also, `--kbd-list` seems to be a synonym of `/kbd-list`, so maybe do you think we can only test `/kbd-list` and drop `--kbd-list`? Conversely, if we are going to test both the styles, I think we should also check `--list:kbd`.

akinomyoga (approved):

completions/xfreerdp
yedayak: I think we should probably only test stuff that we know works in some places?
`--list:kbd` doesn't seem to work in fedora at least. Regarding only testing the slashes variant, I'm not familiar either, so I don't have a strong opinion, I just tried to keep it consistent with what it did before.

akinomyoga: I tried it in Fedora 41, Arch, and Ubuntu 20 LTS, but none seem to support the POSIX-style options (i.e., neither `--kbd-list` nor `--list:kbd` / `--list kbd`). Maybe the POSIX-style options are removed at all, and the above page is too old.

Anyway, the POSIX-style options seem to have been discouraged even when they were supported, and now they don't seem to work at all. Even if we would keep the test for `--kbd-list`, I think it should be tested after trying `/kbd-list`.

akinomyoga: Thanks!

akinomyoga (approved):

akinomyoga:
> > The commit message needs to be adjusted when merging.
> 
> Does the commit message look better?

Yes!

akinomyoga (approved):

PR #1363, opened 2025-04-09
Fedoradev image: don't install vpnc, update for dnf5

yedayak:
vpnc causes this error:
```
Running transaction
Transaction failed: Rpm transaction failed.
  - file /usr/bin/vpnc conflicts between attempted installs of vpnc-consoleuser-0.5.3-48.svn550.fc41.x86_64 and vpnc-0.5.3-48.svn550.fc41.x86_64
  - file /usr/bin/vpnc-disconnect conflicts between attempted installs of vpnc-consoleuser-0.5.3-48.svn550.fc41.x86_64 and vpnc-0.5.3-48.svn550.fc41.x86_64
```

PR #1375, opened 2025-05-06
ci(fedoradev): Qualify FROM to use fedora registry

yedayak:
By default it seems to pull from docker hub, but the images on docker hub seem to be updated much less frequently then the registry.fedoraproject.org one (hosted on quay.io).

The specific difference which made me notice it is that on the rawhide image in docker hub the PATH includes */sbin parts: `PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`, while the one on registry.fedoraproject.org only has 2 bin paths: `PATH=/usr/local/bin:/usr/bin`.

This caused a test failure for monodevelop, since it depends on the full path its run as, and running it as sbin fails with this error:
```
Cannot open assembly
'/usr/sbin/monodevelop/lib/monodevelop/bin/MonoDevelop.exe': Not a directory.
```

### DIFF
--- a/test/docker/fedoradev/Dockerfile
+++ b/test/docker/fedoradev/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:rawhide
+FROM registry.fedoraproject.org/fedora:rawhide
 
 RUN echo install_weak_deps=False >> /etc/dnf/dnf.conf \
     && sed -i -e /tsflags=nodocs/d /etc/dnf/dnf.conf \


akinomyoga:
I'm not opposed to changing the reference to the container, but does this robustly solve the original problem of `monodevelop`? What if the upstream container again changes the value of `PATH`? Or is the upstream going to maintain consistency between `monodevelop` and `PATH` in the future?

Another concern is that, if `/bin/sbin` and other paths are removed from `PATH`, I'm afraid that tests with `require_cmd` associated with the commands in those paths might be turned off (but I haven't checked if it is the case). What would be the impact?

yedayak:
> I'm not opposed to changing the reference to the container, but does this robustly solve the original problem of `monodevelop`? What if the upstream container again changes the value of `PATH`? Or is the upstream going to maintain consistency between `monodevelop` and `PATH` in the future?

Fedora just unified */sbin and */bin in the [latest release](https://fedoraproject.org/wiki/Releases/42/ChangeSet#Unify_/usr/bin_and_/usr/sbin) (f42), so not having `sbin` in the path makes sense. 
There is also some [relevant discussion](https://bugzilla.redhat.com/show_bug.cgi?id=2267022) about issues this caused with installing some packages (ruby specifically). See another issue this caused us in https://github.com/scop/bash-completion/pull/1363
> Another concern is that, if `/bin/sbin` and other paths are removed from `PATH`, I'm afraid that tests with `require_cmd` associated with the commands in those paths might be turned off (but I haven't checked if it is the case). What would be the impact?

New installation of fedora will have the directories symlinked, so there won't be any binaries that are only in one of them.

scop (approved):
Seems fine based on https://docs.fedoraproject.org/en-US/containers/#_where_to_find_the_containers

> Container images built and released by the Fedora Project are available on our [registry](https://registry.fedoraproject.org/).

PR #1376, opened 2025-05-07
test(fprintd): xfail --finger in container

yedayak:
Containers usually don't have fingerprint readers

### DIFF
--- a/test/t/test_fprintd_enroll.py
+++ b/test/t/test_fprintd_enroll.py
@@ -1,5 +1,7 @@
 import pytest
 
+from conftest import in_container
+
 
 @pytest.mark.bashcomp(
     cmd="fprintd-enroll",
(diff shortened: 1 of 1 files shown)


test/t/test_fprintd_enroll.py
```diff
@@ -13,6 +15,9 @@ def test_basic(self, completion):
     def test_options(self, completion):
         assert completion
 
+    @pytest.mark.xfail(
```
scop, line 18:
```suggestion
    @pytest.mark.skipif(
```

I think we should be skipping and not xfailing tests in scenarios like this. We seem to have the same problem in at least some of our xfail vs skip uses and should review and fix them all up in another PR. (For example, in strict xfail mode, an unexpected pass would cause the tests to fail.)

- https://docs.pytest.org/en/stable/how-to/skipping.html
- https://docs.pytest.org/en/stable/how-to/skipping.html#strict-parameter

Pre-approved with that addressed.

scop (approved):

yedayak:
Created https://github.com/scop/bash-completion/pull/1380 for the ubuntu 14 fail, merging.

PR #1377, opened 2025-05-07
fix(convert): Try calling magick, for IM7

yedayak:
In image magick 7 running `convert` prints a warning. Try running magick first.

```
WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert"
```

Another alternative is to just swallow the warning from `convert`, but this seems a bit more future proof.

### DIFF
--- a/completions/convert
+++ b/completions/convert
@@ -49,7 +49,13 @@ _comp_cmd_convert__common_options()
             return
             ;;
         -format)
-            _comp_compgen_split -- "$(convert -list format | _comp_awk \
+            # FIXME: We should probably accept the actual command being used
+            # here and run that, instead of calling `magick` or `convert`
+            # explicitly.
+            # `magick` is the new way to call the command in ImageMagick 7
+            _comp_compgen_split -- "$({
+                magick -list format 2>/dev/null || convert -list format
+            } | _comp_awk \
                 '/ [r-][w-][+-] / { sub("[*]$","",$1); print tolower($1) }')"
             return
             ;;


completions/convert
```diff
@@ -49,7 +49,11 @@ _comp_cmd_convert__common_options()
             return
             ;;
         -format)
-            _comp_compgen_split -- "$(convert -list format | _comp_awk \
+            # FIXME: We should probably accept the actual command being used
+            # here and run that, instead of calling `magick` or `convert`
+            # explicitly.
+            # `magick` is the new way to call the command in image magick 7
+            _comp_compgen_split -- "$(magick -list format 2>/dev/null || convert -list format 2>/dev/null | _comp_awk \
```
akinomyoga, line 56:
```suggestion
            _comp_compgen_split -- "$({ magick -list format 2>/dev/null || convert -list format 2>/dev/null; } | _comp_awk \
```

`a || b | c` would be treated as `a || { b | c; }`. In addition, formatting by `shfmt` might be needed.

completions/convert
```diff
@@ -49,7 +49,11 @@ _comp_cmd_convert__common_options()
             return
             ;;
         -format)
-            _comp_compgen_split -- "$(convert -list format | _comp_awk \
+            # FIXME: We should probably accept the actual command being used
+            # here and run that, instead of calling `magick` or `convert`
+            # explicitly.
+            # `magick` is the new way to call the command in image magick 7
```
akinomyoga, line 55:
```suggestion
            # `magick` is the new way to call the command in ImageMagick 7
```

yedayak: Done

completions/convert
yedayak: Did the first suggestion. Regarding the second, I accidentally left over the `2>/dev/null` on the `convert` from a different thing I checked, removed it.

akinomyoga (approved):

PR #1378, opened 2025-05-07
ci: Make distcheck not cancel jobs once one fails

yedayak:
It's pretty common that a specific distro fails the check, and since it currently cancels the rest of the matrix, you have to fix it first before seeing failures in other distros.

This enables the fail-fast option:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,6 +50,7 @@ jobs:
             network: none
           - dist: fedoradev
           - dist: ubuntu14
+      fail-fast: false
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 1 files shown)


akinomyoga (commented):
This would be useful for me.

scop (approved):
Yes, this is good, as long as it doesn't cause the matrix to succeed in case any of its entries fail. I think it doesn't, but the linked docs don't seem quite clear on it on a quick read.

We may want to say this _disables_ the fail-fast option in the commit message rather than enables though.

yedayak:
> Yes, this is good, as long as it doesn't cause the matrix to succeed in case any of its entries fail. I think it doesn't, but the linked docs don't seem quite clear on it on a quick read.

Just looking at the checks of this PR, some checks succeeded but still the CI is marked as failed.
 
> We may want to say this _disables_ the fail-fast option in the commit message rather than enables though.

Nice catch, fixed

PR #1379, opened 2025-05-08
docs(README): Suggest using `completionsdir` as a fallback

jtojnar:
If the `bash-completion.pc` file is not available during build,
the build system would fall back to using the legacy `compatdir`.

Let’s update the fallback to make it consistent with the optimistic branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ A. [ Disclaimer: Here, how to make the completion code visible to
 
    ```m4
    PKG_CHECK_VAR(bashcompdir, [bash-completion], [completionsdir], ,
-     bashcompdir="${sysconfdir}/bash_completion.d")
+     bashcompdir="${datadir}/bash-completion/completions")
    AC_SUBST(bashcompdir)
    ```
 
@@ -207,12 +207,13 @@ A. [ Disclaimer: Here, how to make the completion code visible to
    `bash-completion-config-version.cmake` files. Example usage:
 
    ```cmake
+   include(GNUInstallDirs)
    find_package(bash-completion)
    if(BASH_COMPLETION_FOUND)
      message(STATUS
        "Using bash completion dir ${BASH_COMPLETION_COMPLETIONSDIR}")
    else()
-     set (BASH_COMPLETION_COMPLETIONSDIR "/etc/bash_completion.d")
+     set (BASH_COMPLETION_COMPLETIONSDIR "${CMAKE_INSTALL_DATADIR}/bash-completion/completions")
      message (STATUS
        "Using fallback bash completion dir ${BASH_COMPLETION_COMPLETIONSDIR}")
    endif()


README.md
scop, line 1: Makes sense to me, but in the same answer we later refer to data*root*dir instead of datadir. I've read the descriptions of the two, became confused, but was left with the feeling that datadir would actually be what we should be using (mostly because the datadir description mentions it actually being the dir for _installing_ things into, datarootdir does not). We also use datadir for the replacements we do for the .pc file, not datarootdir.

This is not the fault of this change, but as it makes the difference so obvious, I'd like to discuss here what to do about it. I guess I'm suggesting this (to be done in another commit/PR):
```diff
--- a/README.md
+++ b/README.md
@@ -236 +236 @@ A. [ Disclaimer: Here, how to make the completion code visible to
-   bashcompdir = $(datarootdir)/bash-completion/completions
+   bashcompdir = $(datadir)/bash-completion/completions
@@ -243 +243 @@ A. [ Disclaimer: Here, how to make the completion code visible to
-   install(FILES your-completion-file DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/bash-completion/completions")
+   install(FILES your-completion-file DESTINATION "${CMAKE_INSTALL_DATADIR}/bash-completion/completions")
```
@akinomyoga your thoughts on this? Based on git log you were involved with the commit that added these two mentions, 530017d126f2368fa07e2a1aa9d7b8ec65aeb362

akinomyoga: I'm unfamiliar too, but according to the documentations, `datadir` seems better.

- [install — CMake 4.1.0 Documentation](https://cmake.org/cmake/help/latest/command/install.html)
- [Standard Directory Variables (automake)](https://www.gnu.org/software/automake/manual/html_node/Standard-Directory-Variables.html)

`datarootdir` seems to be intended as the root directory for various directories with special functions, such as `share/info`, `share/locale`, `share/man`, `share/doc`, and `share` itself. I agree with applying these changes. In addition, I think we need to add `include(GNUInstallDirs)` for the CMake example.

scop (approved):
Thanks!

README.md
scop: https://github.com/scop/bash-completion/pull/1415

jtojnar: Reading [Directory Variables (GNU Coding Standards)](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html), I think `datarootdir` makes sense for third party projects providing completions:

> This is usually the same place as '`datarootdir`', but we use the two separate variables so that you can move these program-specific files without altering the location for Info files, man pages, etc.

So it allows distributor to override `datadir` for a third party project without installing the completion file to a path bash-completion will not be able to find it.

scop: I read the same document a couple of days ago and came to the opposite conclusion. They could have overridden datadir for bash-completion and in that case installing 3rd party completions to datarootdir would install to a location that's not looked up from.

Anyway, anything that's not using the .pc file to figure out locations to install is doing it the fragile way, and gets to keep both pieces when it breaks :). If this would be a robust way to do it, I suppose there would be no need for the .pc file in the first place.

jtojnar: > I read the same document a couple of days ago and came to the opposite conclusion. They could have overridden datadir for bash-completion and in that case installing 3rd party completions to datarootdir would install to a location that's not looked up from.

Right, it cuts both ways.

Though overriding `datadir` for `bash-completion` makes even less sense than doing that for programs, in my opinion:

- It would require overriding `completionsdir` for every program trying to install completions without the use of pkg-config file. (I guess autotools introducing `datarootdir` predates `pkg-config`.)
- I can imagine someone wanting to build a program with different `--datadir` in order allow simultaneous installations of different versions of software. I do not see a reason to do that for [nexus](https://www.devever.net/~hl/nexuses)-establishing software like `bash-completion` other than ~~vanity~~ organisation.

> Anyway, anything that's not using the .pc file to figure out locations to install is doing it the fragile way, and gets to keep both pieces when it breaks :). If this would be a robust way to do it, I suppose there would be no need for the .pc file in the first place.

Right. `datarootdir` is precisely what prevents this mismatch when overriding `datadir` for programs, nexūs should be still found in `datarootdir`. Though we do not actually support `datarootdir` in our pkg-config file so we would encounter the same issue here:

https://github.com/scop/bash-completion/blob/c55ee7f6fb75300786cb522261f68eb80366c41f/bash-completion.pc.in#L6

Programs are still [expected to override variables in `pkg-config` paths](https://www.bassi.io/articles/2018/03/15/pkg-config-and-paths/) so they could use `pkg-config` to support `datarootdir` with <code>pkg-config bash-completion --variable=completionsdir --define-variable=datadir=<strong>$datarootdir</strong></code>. But without guidance from us, this is pretty unlikely.

If we want to fix that, we could modify the pkg-config file, introducing `datarootdir=@datarootdir@` and replacing `datadir` value with `${datarootdir}`. This is a somewhat common pattern I have seen in pkg-config files ([example](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/a9144f5313ce8a9af2735209db88bbd6bb096ee1/subprojects/gst-plugins-base/meson.build#L466)) before projects started switching en masse to Meson, which [does not support `datarootdir`](https://github.com/mesonbuild/meson/issues/7616).

Alternately, we could decide supporting separate `datarootdir` is not worth the trouble until someone comes up with an actual convincing use case.

Issue #1389, opened 2025-06-10
License clarification


Adeline-Toader:
Hello,

Can you please clarify if the licence of the project is GPL-2.0-or-later or GPL-2.0-only? It could be useful to mention the exact version in the COPYING file (instead of the text of the GPL which is already included in the License file).  GPL 2.0 or later would be really helpful to avoid many incompatibilities (such as Apache 2.0 or GPL 3.0) particularly that version 2 is a deprecated one, with several shortcomings. Many thanks for your support!

akinomyoga:
*From https://github.com/scop/bash-completion/issues/1246#issuecomment-2336710620*
> > At a glance, a few questions come up:
> > 
> > * is the license `SPDX-License-Identifier: GPL-2.0-or-later` or `SPDX-License-Identifier: GPL-2.0-only`?
> 
> Please check the copyright notice. It's `GPL-2.0-or-later`.
> 
> https://github.com/scop/bash-completion/blob/e8dc253ddda79f83532551e45deeeab0c3afc5f4/bash_completion#L3-L11

scop:
https://github.com/scop/bash-completion/pull/1391 adds the discussed SPDX id. tmux related ones have one of their own.

PR #1391, opened 2025-06-15
docs: add SPDX license identifier

scop:
Related docs are not specific whether any amount of whitespace is acceptable before the id stanza, but I suppose it is. Going with more than one as it aligns better with the other commentary we have.

* https://spdx.dev/learn/handling-license-info/#how
* https://github.com/spdx/using/blob/17d56be4f4b99df87a52ae4a352810214707f861/docs/using-SPDX-short-identifiers-in-source-files.md

Closes https://github.com/scop/bash-completion/issues/1389

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -19,6 +19,8 @@
 #   along with this program; if not, write to the Free Software Foundation,
 #   Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
+#   SPDX-License-Identifier: GPL-2.0-or-later
+#
 #   The latest version of this software can be obtained here:
 #
 #   https://github.com/scop/bash-completion


bash_completion
```diff
@@ -19,6 +19,8 @@
 #   along with this program; if not, write to the Free Software Foundation,
 #   Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
+#   SPDX-License-Identifier: GPL-2.0-or-later
```
scop, line 22: I did not bother adding this to every source file, I suppose it's good enough this way. But no objections if someone wants to add it to more files, it's just an oneliner anyway.

akinomyoga (approved):

bash_completion
akinomyoga: If it is added to all files, I think the copyright notices should also be added to all files at least.

https://github.com/scop/bash-completion/blob/3ddb4452dcff00d48456c6edde1d649440e0e495/bash_completion#L5-L6

PR #1393, opened 2025-06-15
interface: handle name@link from iproute2

bjacquin:
Since #1090, iproute2 is preferred over ifconfig, however iproute2 provide the list of network interface in name@link in case an interface is bound to another one (such as macvtap for example), while the real interface name is not.

  $ ip link show
  2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DEFAULT group default qlen 1000
      link/ether 48:21:0b:52:19:22 brd ff:ff:ff:ff:ff:ff
  4: eth0.128@eth0: <BROADCAST,MULTICAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
      link/ether 48:21:0b:52:19:22 brd ff:ff:ff:ff:ff:ff

bash-completion using name@link as provided by iproute2 output then provide such name in auto completion list, leading to invalid interface name when auto completing tcpdump arguments.

  $ tcpdump -i <TAB>
  eth0      eth0.128@eth0

This commit handle this use case to drop link name after name@.

  $ tcpdump -i <TAB>
  eth0      eth0.128

See: https://github.com/scop/bash-completion/pull/1090
See: https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/utils.c?h=v6.15.0#n1306


bash_completion
```diff
@@ -1821,7 +1821,7 @@ _comp_compgen_available_interfaces()
             ip -c=never link show || ip link show || ifconfig -a
         fi
     } 2>/dev/null | _comp_awk \
-        '/^[^ \t]/ { if ($1 ~ /^[0-9]+:/) { print $2 } else { print $1 } }')" &&
+        '/^[^ \t]/ { if ($1 ~ /^[0-9]+:/) { split($2, link, "@"); print link[1] } else { print $1 } }')" &&
```
akinomyoga: I'd rather use `sub(/@.*/, "", $2); print $2`, or another option is to include `@` as a part of the separators, `-F '[ \t@]'`.

yedayak: I think I prefer the `sub` option, but no strong preference

yedayak (approved):

bash_completion
bjacquin: I believe this is fair statement, providing new revision now.

yedayak:
Can you fix the commit title and body? The title needs to be in the conventional commit syntax, `fix(<something): `. The body just needs to be wrapped at 80 characters.

yedayak:
I created https://github.com/scop/bash-completion/pull/1395 (depends on https://github.com/scop/bash-completion/pull/1394) which should test this works as expected.

PR #1396, opened 2025-06-16
fix(scp): fix a bug that `-F<prefix>[TAB]` did not complete at all

akinomyoga:
This was a part of #1371 but was separated.

Details are described in the commit message.


akinomyoga:
Thanks!

PR #1398, opened 2025-06-16
[Waiting for #1397] fix(rsync,scp): remove file-type marks from "ls -F" properly

akinomyoga:
This was a part of #1371 but was separated. Waiting for #1397.

Details are described in the commit message.

### DIFF
--- a/test/t/test_scp.py
+++ b/test/t/test_scp.py
@@ -3,7 +3,12 @@ from itertools import chain
 
 import pytest
 
-from conftest import assert_bash_exec, assert_complete, bash_env_saved
+from conftest import (
+    assert_bash_exec,
+    assert_complete,
+    bash_env_saved,
+    prepare_fixture_dir,
+)
 
 LIVE_HOST = os.environ.get(
     "BASH_COMPLETION_TEST_LIVE_SSH_HOST", default="bash_completion"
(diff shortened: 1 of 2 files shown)


test/t/test_scp.py
```diff
@@ -152,3 +157,22 @@ def test_xfunc_remote_files(self, bash):
...
+
+    @pytest.fixture
+    def tmpdir_mkfifo(self, request, bash):
+        tmpdir, _, _ = prepare_fixture_dir(request, files=[], dirs=[])
+
+        try:
+            assert_bash_exec(bash, "mkfifo '%s/local_path_1-pipe'" % tmpdir)
+        except Exception:
+            pytest.skip(
+                "The present system does not allow creating a named pipe."
```
scop: I wonder if there's a valid setup that would allow creating these, but just doesn't have `mkfifo` installed. The end result would be the same, but the skip message would be misleading. Maybe include the stringified exception in it and generalize a bit, something like `f"Could not create a named pipe: {ex}"`?

Pre-approved with this addressed some way.

akinomyoga: > I wonder if there's a valid setup that would allow creating these, but just doesn't have `mkfifo` installed.

I anticipated the opposite situation, where `mkfifo` is installed, yet creating a named pipe is not allowed. For example, MSYS 1.0 doesn't support creating a named pipe:

```console
$ mkfifo a.pipe
mkfifo: cannot create fifo `a.pipe': Function not implemented
```

Although I haven't tried, I also wonder if we can create a named pipe inside a USB memory stick (whose filesystem is usually FAT32).

scop: With #1397 merged, is this in a draft state for some other reason, perhaps the above discussion?

akinomyoga: Sorry, I've simply been busy. I'll take a look again soon.

akinomyoga: I added a clarification in [`b3e01d6..bc07307`](https://github.com/scop/bash-completion/compare/b3e01d61130059a8b43553335075f115ee54ca63..bc07307f2a200bff33a162f87b5f48f60d5a016d).

akinomyoga:
I've merged this.

Issue #1328, opened 2025-02-12
Documentation of old and new APIs


akinomyoga:
We have released the new API for the completion settings in 2.12 (after discussions in #539 and related Issues and PRs).

Now, new external completions in the wild have started to use the new API. However, the `bash-completion` package that they are going to depend on might not be as up-to-date as 2.12. As we have discussed in https://github.com/scop/bash-completion/discussions/537#discussioncomment-7636168, we actually think that the external completions can still depend on the previous API for now. At least, the old API would not go away until all the `bash-complete` packages in major distributions are updated to 2.12 or above, and we predict that it will take about a decade. After that, we can recommend the external completions to use the new API so that they work everywhere with a `back-completion` package. Even then, we do not have a plan to remove the older API.

However, our stance is not so obvious to external people, so we sometimes need to inform them about it, e.g. as in https://github.com/scop/bash-completion/issues/1317#issuecomment-2619702773.

In addition, there seems to be confusion about the relationship with an even older API of v1.0 (which we still provide for backward compatibility) as in #1220 where the OP seems to consider the newer API as the one for `bash-completion` v2. Another confusion is that the relationship between the functions of different API sets, such as `have` (from the oldest one), `_have` (until v2.11), and `_comp_have` (added in v2.12), as in https://github.com/libsixel/libsixel/pull/89.

I think it would be useful to document these matters in one place.

## Suggestion: API version numbering

I suggest giving a number to the older and newer API sets. Naively, we could call an API set with the version number of `bash-completion` itself, such as "v2.11" vs "v2.12". However, the current situation, where we implement "v2.12" but still recommend external completions based on `bash-completion` v2.12 to use "v2.11", would be confusing.

Since the new API would be backward-incompatible with the older API, I think it would make sense to call the newer API as "v3". When the `bash-completion` packages in major distributions are entirely replaced with 2.12 or above, we can also update the version of `bash-completion` itself to "v3.0". This is my idea.

Then, the older API used until v2.11 can be called "API v2". An even older one can be called "API v1".

## Suggestion: Examples of completion settings based on API v2

Another issue is that we do not have completion settings based on API *v2* in the latest version of `bash-completion`. This means that external completion writers don't find official references for writing completions based on API v2, although we still recommend that external completion writers use API v2 if they want it to work everywhere.

[...]

PR #1399, opened 2025-06-22
Support sourcing bash_completion.d from /usr/share/ and move ours

bluca:
In distributions we are in the process of moving data shipped
by upstream projects to /usr/ and out of /etc/, as the latter
is really meant to be for local changes, not for upstream
projects to ship programs/data.

Add support for sourcing compat scripts from /usr/share/bash_completion.d/
together with /etc/bash_completion.d, and move 000_bash_completion_compat.bash
to the former.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2,8 +2,7 @@ SUBDIRS = completions doc helpers test
 
 pkgdata_DATA = bash_completion
 
-# Empty, but here just to get the compat dir created with install
-compatdir = $(sysconfdir)/bash_completion.d
+compatdir = $(datadir)/bash-completion/compat
 compat_DATA = bash_completion.d/000_bash_completion_compat.bash
 
 profiledir = $(sysconfdir)/profile.d
@@ -34,9 +33,12 @@ EXTRA_DIST = CHANGELOG.md $(pkgdata_DATA) bash_completion.sh.in .dir-locals.el \
 	bash-completion-config-version.cmake.in setup-symlinks.sh \
 	$(compat_DATA)
 
+# Create and install empty directory for local completion scripts in sysconfdir
 install-data-hook:
 	tmpfile=`mktemp $${TMPDIR:-/tmp}/bash_completion.XXXXXX` && \
-	$(SED) -e 's|(/etc/bash_completion\.d|($(compatdir)|' \
+	$(SED) -e 's|(/etc/bash_completion\.d|($(sysconfdir)/bash_completion.d|' \
+	    -e 's|(/usr/share/bash-completion/compat|($(compatdir)|' \
 	    $(DESTDIR)$(datadir)/$(PACKAGE)/bash_completion >$$tmpfile && \
 	cat $$tmpfile >$(DESTDIR)$(datadir)/$(PACKAGE)/bash_completion && \
 	rm $$tmpfile
+	$(MKDIR_P) $(DESTDIR)$(sysconfdir)/bash_completion.d
--- a/bash_completion
+++ b/bash_completion
@@ -3492,6 +3492,7 @@ _comp__init_collect_startup_configs()
         compat_dirs+=("$BASH_COMPLETION_COMPAT_DIR")
     else
         # Keep in sync with Makefile.am install-data-hook
+        compat_dirs+=(/usr/share/bash-completion/compat)
         compat_dirs+=(/etc/bash_completion.d)
         # Similarly as for the "completions" dir, look up from relative to
         # bash_completion, primarily for installed-with-prefix and
@@ -3503,7 +3504,14 @@ _comp__init_collect_startup_configs()
         else
             compat_dir=$_comp__base_directory/bash_completion.d
         fi
-        [[ ${compat_dirs[0]} == "$compat_dir" ]] ||
+        local dir
+        for dir in "${compat_dirs[@]}"; do
+            if [[ $dir == "$compat_dir" ]]; then
+                compat_dir=""
+                break
+            fi
+        done
+        [[ $compat_dir ]] &&
             compat_dirs+=("$compat_dir")
     fi
     for compat_dir in "${compat_dirs[@]}"; do


akinomyoga:
This is related to the discussion at https://github.com/scop/bash-completion/discussions/1329#discussioncomment-12163014.

I don't think-`/usr/share/bash_completion.d/` is the right directory. It should be somewhere under `/usr/share/bash-completion/` just like the completions for each command is arranged in `/usr/share/bash-completion/completions/`. Also, the naming of `xxx.d` should have come from `/etc/profile.d`, so it is not a naming convention used under `/usr/share`.

> In distributions we are in the process of moving data shipped
> by upstream projects to /usr/ and out of /etc/, as the latter

I'm not familiar with the recent movements in distributions, but which distributions are you referring to? For example, `/etc/profile.d` seems to be still widely used. Are these removed from most Linux/BSD/etc distributions in the future? Or are you talking about a specific distribution?

bluca:
> This is related to the discussion at [#1329 (comment)](https://github.com/scop/bash-completion/discussions/1329#discussioncomment-12163014).
> 
> I don't think-`/usr/share/bash_completion.d/` is the right directory. It should be somewhere under `/usr/share/bash-completion/` just like the completions for each command is arranged in `/usr/share/bash-completion/completions/`. Also, the naming of `xxx.d` should have come from `/etc/profile.d`, so it is not a naming convention used under `/usr/share`.

Sure, I've changed it to `/usr/share/bash-completion/compat/`, let me know if you prefer a different name.

> > In distributions we are in the process of moving data shipped
> > by upstream projects to /usr/ and out of /etc/, as the latter
> 
> I'm not familiar with the recent movements in distributions, but which distributions are you referring to? For example, `/etc/profile.d` seems to be still widely used. Are these removed from most Linux/BSD/etc distributions in the future? Or are you talking about a specific distribution?

For example in Debian we recently moved most program's own shell completion scripts from /etc/bash_completion.d/ to /usr/share/bash-completion/completions/

You can find a list of similar efforts at https://github.com/uapi-group/specifications/issues/76

akinomyoga:
> For example in Debian we recently moved most program's own shell completion scripts from /etc/bash_completion.d/ to /usr/share/bash-completion/completions/
> 
> You can find a list of similar efforts at [uapi-group/specifications#76](https://github.com/uapi-group/specifications/issues/76)

This doesn't seem to be the official decision led by the maintainers of Debian's base system (that covers the basic parts of `/etc`, including `/etc/profile.d`). The provided link to a discussion in `uapi-group` doesn't seem to mention Debian, and the relationship between "The Linux Userspace API Group" (UAPI) and Debian is unclear. I searched for [*The Linux Userspace API Group "Debian"*](https://www.google.co.jp/search?q=The+Linux+Userspace+API+Group+%22Debian%22&amp;peek_pws=0) in Google, and it points to [this discussion](https://lists.debian.org/debian-devel/2024/12/msg00444.html), which exactly talks about `/etc` in Debian but with some tension. I also found and read [Debian#1051371](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1051371), where some tensions happened again. The original post of this PR started with "*In distributions we are ...*", and another comment contained "*For example in Debian we recently moved ...*", which sounded as if this is the movement led by the main maintainers of major distributions (or Debian) under the general consensus. Also, I wonder why only a link to the UAPI discussion was provided, while the links to the Debian discussions were not. I feel this PR is unconsciously trying to mislead and put pressure on us. People do this when they anticipate that the PR might not be convincing after honest discussions, which itself isn't bad but certainly makes us wary. Also, I do not find mentions about completion scripts in the provided link to a UAPI discussion or the search in Debian's mailing lists. Another thing is that the migration from `/etc/bash_completion.d` to `/usr/share/bash-completion/completions` is not necessarily for the proper `/etc` arrangement, but rather, we've recommended this for the newer bash-completion API. I cannot confirm that the main reason for the mentioned migration to `bash-completion/completions` was indeed the `/etc` policy. Actually, this migration should happen in the upstream projects that provide the completion scripts (rather than at the packaging stage in the Debian distribution) because it requires rewriting of the completion scripts using the new API. I shall assess this PR by looking at the code changes genuinely.

bash_completion
```diff
@@ -3435,7 +3437,7 @@ _comp__init_collect_startup_configs()
         else
             compat_dir=$_comp__base_directory/bash_completion.d
         fi
-        [[ ${compat_dirs[0]} == "$compat_dir" ]] ||
+        [[ ${compat_dirs[0]} == "$compat_dir" ]] || [[ ${compat_dirs[1]} == "$compat_dir" ]] ||
```
akinomyoga:
```suggestion
        local dir
        for dir in "${compat_dirs[@]}"; do
            if [[ $dir == "$compat_dir" ]]; then
                compat_dir=""
                break
            fi
        done
        || $compat_dir ]] &&
```

akinomyoga (approved):
I think one thing I should point out is that the `/etc/bash_completion.d` is anyway for backward compatibility (although there is a subtlety as discussed in https://github.com/scop/bash-completion/discussions/1329#discussioncomment-12163014). Then, a naive question would be whether we should introduce incompatible changes in the code kept for backward compatibility (which doesn't seem to make sense at first glance). If it breaks the compatibility, we could also argue that `compatdir` should rather be removed.
- To completely get rid of `/etc/bash_completion.d`, all the programs that provides completion scripts with the old API need to be updated or adjusted in the packaging stage.
- If the change is only applied at the packaging stage of distributions (without updating the scripts to use the new API), the package maintainers can always adjust the location of the scripts (including `000_bash_completion_compat.bash`) by setting a system default `BASH_COMPLETION_COMPAT_DIR` (or applying a patch to `bash_completion`).
- If an upstream project of a specific command has a chance to change the location of the default install location from `/etc/bash_completion.d` to another place, it should rather update its completion script to use the newer API and put it in `/usr/share/bash-completion/completions`.
  - In this context, if the project author sets up the installation of the completion script using autoconf or cmake following our `README.md`, if we change the compatdir in pkg-config, the installation location should automatically be adjusted even if the project is not actively maintained. In this sense, changing the default location of `compatdir` now still has some effects. A little concern is how widely our pkg-config settings are installed in the system in distributions.

Even if this PR doesn't have much benefits in the practical situations, I think I can accept the PR.

bluca:
> This doesn't seem to be the official decision led by the maintainers of Debian's base system (that covers the basic parts of `/etc`, including `/etc/profile.d`).

Yes, I mentioned them as separate and independent examples as I thought you were looking for more than one occurrence, sorry for the confusion

bluca:
> If it breaks the compatibility, we could also argue that `compatdir` should rather be removed.

My intention here was to keep backward compatibility intact - the existing directory is still checked for scripts. Do you see any use case that would be broken by these changes?

scop:
My gut feeling about this is that compatdir being a deprecated thing, we should consider functionality related to it as frozen, including not making it any easier/cleaner to use.

We don't mention compatdir being deprecated too prominently though, which is something we could improve. I'm not sure if there's any other mention of it at the moment than https://github.com/scop/bash-completion/blob/98207f4710c45128c96757023f16929828071451/README.md?plain=1#L185-L186

bluca:
> My gut feeling about this is that compatdir being a deprecated thing, we should consider functionality related to it as frozen, including not making it any easier/cleaner to use.

We still need it to ensure our completion scripts work with older versions, and likely will for many years, so I'd be a lot happier with this change included. That way we can build usr-only images where everything works out of the box. Thanks.

scop:
> We still need it to ensure our completion scripts work with older versions, and likely will for many years [...]

Just to confirm I understand correctly, older versions here refers to bash-completion < 2.0 (which was released in Jun 2012)?

bluca:
> > We still need it to ensure our completion scripts work with older versions, and likely will for many years [...]
> 
> Just to confirm I understand correctly, older versions here refers to bash-completion < 2.0 (which was released in Jun 2012)?

Sorry, let me add more details: in this case I am explicitly referring to `_command_offset` which is deprecated in 2.12 from 2024 and requires `bash_completion.d/000_bash_completion_compat.bash` to work since that version. Its alternative was introduced in the same version so can't be used on versions earlier than last year.

We use it here:

https://github.com/systemd/systemd/blob/main/shell-completion/bash/systemd-run#L51
https://github.com/systemd/systemd/blob/main/shell-completion/bash/run0#L46

scop (changes requested):
Hmm, I see, but I still don't like it. Let me think about it some more, but anyway let's get this settled before the next release one way or another. (Adding a placeholder "request changes" to mark that.)

akinomyoga:
I'm still unsure about whether the provided reasons make sense. If a distribution want to put the compat directory in another location, the distribution can always set `BASH_COMPLETION_COMPAT_DIR` when loading `bash_completion`. This is always possible because when the distribution arranges `bash_completion`, the distribution needs to prepare an entry point (such as `/etc/profile.d/bash-completion.sh` or whatever location that fits the distribution's policy). Then, `BASH_COMPLETION_COMPAT_DIR` can be set there. Or the distribution can directly modify `bash_completion` by patching the `install-data-hook` target in Makefile.


> We still need it to ensure our completion scripts work with older versions, and likely will for many years, so I'd be a lot happier with this change included. That way we can build usr-only images where everything works out of the box. Thanks.

I think the distribution can set `BASH_COMPLETION_COMPAT_DIR`, or directly modify `bash_completion` through `install-data-hook`.

> Sorry, let me add more details: in this case I am explicitly referring to `_command_offset` which is deprecated in 2.12 from 2024

Actually, they are not officially deprecated. `_command_offset` is a part of API v2, and we introduced API v3 in 2.12. API v2 are deprecated as *internal APIs*, and we updated to use API v3 in bash-completion's internal implementations. However, we still recommend external completion writers to use API v2. API v2 is supposed to be deprecated in the future, but we probably do not remove API v2 even after they are deprecated. The stub `_comp_deprecate_*` in `000_bash_completion_compat.bash` are currently merely a placeholder for the future deprecation messages.

> and requires `bash_completion.d/000_bash_completion_compat.bash` to work since that version.

This is one of the problems of the current setup. This is discussed in https://github.com/scop/bash-completion/discussions/1329#discussioncomment-12163014. Originally, `/etc/bash_completion.d` was for API v1 compatibility, but `000_bash_completion_compat.bash` is a fallback for the API v2. In this sense, we probably want to move `000_bash_completion_compat.bash` to a new place (while keeping API v1 directory at `/etc/bash_completion.d`).

> Its alternative was introduced in the same version so can't be used on versions earlier than last year.

For this reason, we still recommend external completion writers to use API v2 (which was introduced in bash-completion-2.0). When you write completion settings and don't want to switch between two APIs, v2 and v3, depending on bash-completion's version, you are recommended to use API v2. We intend to recommend API v2 for about ten years. See #1328 for details.

----

I wrote that, forgetting about the provided reasons, I could accept the PR, but I'm actually neutral about accepting or rejecting the PR. As I already discussed in #1329, I believe we want to sort out the directories. In the process of re-arranging the directories, I guess we would finally want to introduce another "compat" directory, so I thought we might introduce the `compat` directory in this PR as a first step. However, another option is to introduce the directory changes (including `startup`, `default`, etc.) all at once after sufficient discussions.

> but anyway let's get this settled before the next release one way or another. (Adding a placeholder "request changes" to mark that.)

It may be better to discuss #1329 at the same time, which would take time. Then, we might want to make a decision on this PR after the next release.

Another thing is that, as I discussed in https://github.com/scop/bash-completion/pull/1399#pullrequestreview-2950696842, the change in this PR seems harmless to me (though I'm not sure if there are actual benefits for the context provided by the OP). We still look at `/etc/bash_completion.d` in addition to the new `<prefix>/share/bash-completion/compat`, so existing installations of the v1 completions continue to be recognized.

bluca:
> I'm still unsure about whether the provided reasons make sense. If a distribution want to put the compat directory in another location, the distribution can always set `BASH_COMPLETION_COMPAT_DIR` when loading `bash_completion`. This is always possible because when the distribution arranges `bash_completion`, the distribution needs to prepare an entry point (such as `/etc/profile.d/bash-completion.sh` or whatever location that fits the distribution's policy). Then, `BASH_COMPLETION_COMPAT_DIR` can be set there. Or the distribution can directly modify `bash_completion` by patching the `install-data-hook` target in Makefile.

Having to do local patches or modifications is strictly worse than simply fixing the behaviour for everyone out of the box. Everyone needs to notice, know and ship the changes. There are no downsides to just changing it upstream, and only advantages. Even from the design point of view, shipping code in /etc/ is wrong and has always been wrong:

> The /etc hierarchy contains configuration files. A "configuration file" is a local file used to control the operation of a program; it must be static and cannot be an executable binary.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html

This is runnable code, so it does not belong in /etc/ according to all standards and specifications.

> However, we still recommend external completion writers to use API v2.

But without 000_bash_completion_compat.bash, it's not possible to use `_command_offset` - if that's not deprecated, then this file is not solely for backward compat?

> though I'm not sure if there are actual benefits for the context provided by the OP

There are. As already mentioned, we are already building usr-only images in many use cases. All the upstream code and data should be in usr.

> the change in this PR seems harmless to me (though I'm not sure if there are actual benefits for the context provided by the OP). We still look at /etc/bash_completion.d in addition to the new <prefix>/share/bash-completion/compat, so existing installations of the v1 completions continue to be recognized.

Yes indeed, as mentioned I made sure not to harm backward compatibility. Files shipped in /etc/ have precedence and it looks there first.

akinomyoga:
> Having to do local patches or modifications is strictly worse than simply fixing the behaviour for everyone out of the box. Everyone needs to notice, know and ship the changes. There are no downsides to just changing it upstream, and only advantages.

I generally appreciate the idea of organizing `/etc` better. However, I'm not sure if the suggested setup would finally be the one that everyone wants. Then, from your point of view, with this patch, you become free from patching and modifications to the package source code, but all the other people who want a different setup still need to apply patches. In addition, this also affects all the projects that currently install a script in `/etc/bash_completion.d`. They need to be changed to comply your suggestion, but it takes time because the projects still using `/etc/bash_completion.d` (API v1) are likely to be defunct except for the projects that intentionally load their scripts eagerly.

I think we can agree that we should organize it in some way, but the concrete setup of how it should be reorganized needs to be determined by the community. So far, I do not see any major movements in distributions to carry it out in a collective way. The suggested changes seem to be promoted only in an organization called UAPI. They raised discussions in Debian, but they didn't seem to be fully welcomed. Otherwise, there don't seem to be public discussions. In addition, I'm not even sure that this decision to move this specific `/etc/bash_completion.d` to another place is based on the whole consensus of the UAPI members. I wonder if there is a proposal-review-vote process involving all the members in the UAPI. Even if there were, it is strange that the bash-completion maintainers didn't have a chance to attend that process.

We should wait until the consensus is formed to avoid unnecessary switching forward and backward. So far, most distributions are satisfied with the current setup. If one wants to move away from the current way of using `/etc`, I think we should first discuss whether and how `/etc/profile.d` would be changed in the distributions. Then, bash-completion can follow it. Honestly, I think most distributions will think it is fine to keep `/etc/profile` and `/etc/profile.d` (or a similar directory) in `/etc`. One realistic possibility would be to make `/etc/profile.d` and `/etc/bash_completion.d` the places to put symbolic links to `/usr/share/xxx/shells/xxx.bash`.

> Even from the design point of view, shipping code in /etc/ is wrong and has always been wrong:

So, in your point of view, `init` had always been wrong? If so, the Unix systems couldn't even start correctly until systemd appeared for Linux. With the traditional System V init, `/etc/sysconfig` contained many shell scripts to configure the settings. The service/daemon scripts were located in `/etc/init.d/*`, and `/etc/rc*.d/init.d` contained symbolic links to the script files, etc. BSD systems still use the configuration based on System V init, so one cannot remove these files unless you force the {Free,Open,Net,...}BSD systems to abandon the `init`, which is one of the most essential components of the operating system.

> > The /etc hierarchy contains configuration files. A "configuration file" is a local file used to control the operation of a program; it must be static and cannot be an executable binary.
> 
> https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html
> 
> This is runnable code, so it does not belong in /etc/ according to all standards and specifications.

So you haven't read the FHS specification carefully. You've dropped the footnote link *[2]* in the quoted explanation of `/etc`. You should read the footnote *[2]*. Even if one missed the footnote, everyone knows that shell scripts are not counted as executable binaries. If one knows the internal structure of `/etc`, I expect one must know that there have been many important shell scripts in `/etc` traditionally. In addition, FHS 3.0 explicitly enumerates `/etc/profile` and `/etc/csh.login`. You may want to change FHS first before trying to change `bash-completion`.

> But without 000_bash_completion_compat.bash, it's not possible to use `_command_offset` - if that's not deprecated, then this file is not solely for backward compat?

The directory is meant for backward compatibility, but the file is not. This is what I explained in my previous reply:

> *From https://github.com/scop/bash-completion/pull/1399#issuecomment-3017344705*
>
> > and requires `bash_completion.d/000_bash_completion_compat.bash` to work since that version.
>
> This is one of the problems of the current setup. This is discussed in https://github.com/scop/bash-completion/discussions/1329#discussioncomment-12163014. Originally, `/etc/bash_completion.d` was for API v1 compatibility, but `000_bash_completion_compat.bash` is a fallback for the API v2. In this sense, we probably want to move `000_bash_completion_compat.bash` to a new place (while keeping API v1 directory at `/etc/bash_completion.d`).

bluca:
> I generally appreciate the idea of organizing `/etc` better. However, I'm not sure if the suggested setup would finally be the one that everyone wants. Then, from your point of view, with this patch, you become free from patching and modifications to the package source code, but all the other people who want a different setup still need to apply patches. 

I really don't think that's true. I see no evidence of anybody intentionally wanting to ship upstream code in /etc/, all occurrences are just where it happens to be for no reason at all, other than legacy setup just having been like that. Do you have any evidence of that?

> In addition, this also affects all the projects that currently install a script in `/etc/bash_completion.d`. They need to be changed to comply your suggestion, but it takes time because the projects still using `/etc/bash_completion.d` (API v1) are likely to be defunct except for the projects that intentionally load their scripts eagerly.

It doesn't? Backward compatibility is kept, intentionally, any existing script will still be loaded from the previous location too and work as expected.

> So far, I do not see any major movements in distributions to carry it out in a collective way.

There is. We do, and anybody else doesn't really care one way or the other. Every distribution that matters adopted usr-merge for this (and many others) reason.

> They raised discussions in Debian

No, we didn't. Please do not mistake some random thread by some random person in some random mailing list as "discussions" - debian-devel is not for discussions, it's just random spam by random accounts, and it's best ignored. Look at changelogs if you want an actual idea of what's going on there.

> We should wait until the consensus is formed to avoid unnecessary switching forward and backward.

As above, this already happened for those who care - and those who don't, simply don't mind one way or another anyway

> So, in your point of view, init had always been wrong?

Yes! One of the many, many wrong things with it

> So you haven't read the FHS specification carefully.

I have - it's a footnote for a reason. It's an outdated concept that vendor-provided scripts should be editable locally - this hasn't been the case for a really long time. Or are you saying that you expect random users to do personalized, local modifications to `/etc/bash_completion.d/000_bash_completion_compat.bash`?

> The directory is meant for backward compatibility, but the file is not.

Then why is the file itself called "compat"?

bluca:
Look, I just noticed two bugs while building images and wanted to be a good citizen, and provided a fix for both, so that things can work out of the box in our use cases, without negatively affecting anybody. If you really feel so strongly about it because of how things happened to organically get mashed together in the 1980s, then feel free to just close this, and keep the package broken for such use cases. It just means it will get either worked around, or most likely, dropped as non-functional. I'll just hit unsubscribe and stop replying now, thanks for your attention so far.

akinomyoga:
This PR has been full of deception from beginning to end, and the OP has finally rage-quitted. They pretended to be authoritative and tried to force us to follow them, but they haven't provided any reference except for a UAPI issue. I appreciate the goal of the OP, but their investigation and survey were insufficient. Their arguments lacked knowledge about modern setups (as well as the historical ones) and were based on misunderstandings about basics. I won't reply to them because they don't hear us anymore (but not because I don't have replies). I'll just pick the first commit.

septatrix:
> This PR has been full of deception from beginning to end, and the OP has finally rage-quitted.

I don't agree with the accusation that his retreat from this discussion is fueled by "rage". IMO abstaining from a discussion where one notices that the two parties are in an "immovable object vs unstoppable force" situation to spent the time where it is better invested is very reasonable.

> They pretended to be authoritative and tried to force us to follow them, but they haven't provided any reference except for a UAPI issue.

They are a Debian maintainer (and UAPI member) which carries a bit of weight but I am not sure if he claimed to be an "authoritative" entity himself. For that he referred to the Spec of the UAPI Group which I would personally call some kind of authoritative source, similar to freedesktop.org. They represent a multitude of interest groups which lay down current, de-facto standards or develop future standards, and have members from many Distros, companies and projects.

> I appreciate the goal of the OP, but their investigation and survey were insufficient.

Please allow me to add a bit more to the surves/investigation: You might not have found too much information about this config thing because not everything is necessarily written down, and even if it is, may not be indexed by Google (e.g. Fedoras Pagure is badly represented). Things like this UAPI spec are often though out in-person, in this case primarily at the "Image-Based Linux Summit" as discussions are easiest that way.

One of the initiators of moving vendor config to `/usr/...` and keeping `/etc` only for local stuff was SUSE and they have made the largest progress therein so far. Regarding references: There are the [meeting minutes on the website of the UAPI group](https://uapi-group.org/docs/minutes/2022-10-05__image-based-linux-summit/) as well as a [writeup on LWN](https://lwn.net/Articles/912774/). I linked 2022 but the `/etc` issue is also discussed 2023 and 2024 (both have meeting minutes and an LWN summary).

Parties represented during those summits (2022-2024 combined) where among others: Ubuntu, Debian, Fedora, Arch Linux, SUSE, NixOS, some hyperscalers, and some smaller distros. Of the listed distros it was often the working group for immutable variants that were present but not exclusively. If these distros all come forwards and create a specification for how configuration should be handled I think it should be given weight.

> Their arguments lacked knowledge about modern setups (as well as the historical ones) and were based on misunderstandings about basics.

You cannot expect that everyone has a deep understanding of your project. Luca saw a package where he wanted to improve the status quo to advance the adoption of the UAPI spec and created a PR. He may have already working with this project before in the context of systemd but that is only slight exposure at the surface level. One should not expect him to know which exact API is deprecated, planned to be deprecated, or is the way forward.

When it comes to modern setups I also agree with him. I too have deployed a few immutable/image based appliances which boot up with only `/usr` present. This is fantastic and has many advantages. It might not be the everyday modern server deployment but saying that such deployments do not exist is simply wrong. Currently, we are forced to set up a symlink for bash-completions but it would be great if upstream, i.e. you, were to simply put them somewhere under `/usr`.

You also stated that the proposed changes seem harmless to you - and they are very small. In lie of having every distro carry this as a downstream patch it seems reasonable to me to accept this.

akinomyoga:
To clarify first, I didn't reject the suggestion, though the OP seems to claim in the UAPI thread that we've rejected it. From the provided information, I couldn't judge whether the provided reasons make sense. The OP didn't provide any references for the relevant details, despite implying that something about `/etc` was discussed somewhere in some distributions (including Debian) in some way. As far as they are hidden, I'm unsure whether the reasons make sense. However, the OP has quit, so we cannot continue the discussion. @septatrix I see you are from UAPI and seem to know the relevant details. Can we continue the discussion?

> One of the initiators of moving vendor config to `/usr/...` and keeping `/etc` only for local stuff was SUSE and they have made the largest progress therein so far. Regarding references: There are the [meeting minutes on the website of the UAPI group](https://uapi-group.org/docs/minutes/2022-10-05__image-based-linux-summit/) as well as a [writeup on LWN](https://lwn.net/Articles/912774/). I linked 2022 but the `/etc` issue is also discussed 2023 and 2024 (both have meeting minutes and an LWN summary).

Thank you for the information. Those kinds of links are what I expected.

If any, can you point me to where to look at what's going on with `/etc/profile` and `/etc/profile.d` under the UAPI spec? I mentioned `/etc/profile.d` in https://github.com/scop/bash-completion/pull/1399#issuecomment-2994748330, https://github.com/scop/bash-completion/pull/1399#issuecomment-2997078268, https://github.com/scop/bash-completion/pull/1399#issuecomment-2997103165, and https://github.com/scop/bash-completion/pull/1399#issuecomment-3017789386, but the OP didn't respond to any of them. As I've written above, an alternative setup could be to make `/etc/profile.d` and `/etc/bash_completion.d` the places to symbolic links to the physical files in `/usr/share`, which seems to be what `systemd` does for services and other targets. However, the OP ignored this without any counterarguments. Is it specified by the UAPI that a setup like the symbolic links in `/etc/bash_completion.d` should not be used? What is the difference from the use by `systemd`?

> Parties represented during those summits (2022-2024 combined) where among others: Ubuntu, Debian, Fedora, Arch Linux, SUSE, NixOS, some hyperscalers, and some smaller distros. Of the listed distros it was often the working group for immutable variants that were present but not exclusively. If these distros all come forwards and create a specification for how configuration should be handled I think it should be given weight.

Are BSD distributions included in this list? I can understand it even if the BSD distributions are excluded from the UAPI because the UAPI seems to target Linux. However, bash-completion targets also BSD, and we are interested in what the UAPI thinks of the different setups of `/etc` in different types of operating systems.

> When it comes to modern setups I also agree with him. I too have deployed a few immutable/image based appliances which boot up with only `/usr` present. This is fantastic and has many advantages. It might not be the everyday modern server deployment but saying that such deployments do not exist is simply wrong. Currently, we are forced to set up a symlink for bash-completions but it would be great if upstream, i.e. you, were to simply put them somewhere under `/usr`.

Based on the arguments provided until now, it seems to me that the current movements on `/etc` are very much related to what is called the image-based Linux, but are these related to the Linux distributions for desktops, workstations, and servers? Are these related to BSD? I'm not going to say that I reject it if it is for the image-based Linux, but I just want to make it clear. If this reorganization of `/etc` is intended to smoothly create the image-based Linux, that's fine, but you can't say the traditional configurations were wrong from the beginning. System V `init` just didn't target the image-based Linux.

> You also stated that the proposed changes seem harmless to you - and they are very small.

The size of the code change is irrelevant now. We are supposed to carefully consider the impact on users and maintainability as maintainers. We are not supposed to blindly accept the PR, no matter how small the code change is.

> In lie of having every distro carry this as a downstream patch it seems reasonable to me to accept this.

I'm unsure whether the statement that *every* distro applies *the same* patch is correct. If the symbolic links in `/etc/bash_completion.d` could be used, we don't have to change anything. One remaining issue is that `000_bash_completion_compat.bash` is installed in the "deprecated" directory `/etc/bash_completion.d`, but this is a known issue discussed in #1329 and should be resolved within #1329. I mentioned #1329 multiple times, but I doubt that the OP has checked it.

<details><summary>Off topic</summary>

> > They pretended to be authoritative and tried to force us to follow them, but they haven't provided any reference except for a UAPI issue.
> 
> They are a Debian maintainer (and UAPI member) which carries a bit of weight but I am not sure if he claimed to be an "authoritative" entity himself.

The OP hasn't explicitly stated that, but the way of writing has been misleading. The OP initially appeared to be a representative of distributions in general, but then a representative of Linux distributions, then a representative of the Debian distribution, then a representative of `/usr`-only images of Debian, and finally *a citizen*:

Initially, this PR started with "*In distributions we ...*", which sounded as if the OP is a representative of some distributions of operating systems. The *distributions* were not precisely specified, but `bash-completion` targets different operating systems such as Linux, BSD, and macOS. Then, the OP didn't appear to be a core developer of the base systems in the threads I found by Google, and the OP appeared to be a representative of UAPI. However, this seems to imply that the *distributions* are limited to Linux and don't appear to care about BSD and other systems. You clarified that the OP is actually [a Debian *maintainer*](https://wiki.debian.org/DebianMaintainer), but it seems to be different from [the Debian developers](https://wiki.debian.org/DebianDeveloper) who have some rights on the final decision about the base system. Furthermore, the implied existing examples seem to be limited to Debian. References or other examples weren't provided. Then, as the discussion goes on, it appeared to be related to *`/usr`-only* images of Linux. Finally, the OP appeared to claim that the OP just tried to build images and came here to fix problems as merely *a citizen*.

This is strange. Also, the PR title appears to imply that the main purpose is to fix a trivial regression, but the actual purpose is to propose a change in the usage of `/etc`. In the first place, ideally, fixes of different purposes shouldn't be mixed, in particular when one of them introduces an interface change. The OP appeared to be reluctant to provide references to the detailed discussions. The OP has removed the footnote link [2] in the quote from the FHS spec, which contradicts the OP's claim. There are other small things. I believe the OP didn't do this on purpose, but I felt the OP's habit of hiding information and choosing the words to cause convenient misconceptions.

> For that he referred to the Spec of the UAPI Group which I would personally call some kind of authoritative source, similar to freedesktop.org. They represent a multitude of interest groups which lay down current, de-facto standards or develop future standards, and have members from many Distros, companies and projects.

I agree that one can say that UAPI and FHS specs have some kind of authority, but what is unclear is if the OP was really a representative of those groups or merely a member. Without references, I can't tell e.g. whether the core Debian developers are members of the UAPI and whether the OP's claim matches a consensus under the UAPI. (I actually doubt the general consensus in Debian from the limited information.) I see that there are indeed movements on reorganizing `/etc` in some way, but it was unclear which part of the OP's claim is the official decision of the UAPI and FHS and which is not. The OP seems to claim that scripts do not belong to `/etc` in *all standards and specifications* including the FHS spec, but I doubt this. At least, the FHS spec lists `/etc/profile`. I'm not sure what the OP meant by *ALL STANDARDS AND SPECIFICATIONS*, but from this point, the OP's survey/investigation appeared to be insufficient. Also, I'm not sure whether the OP has investigated how they work with e.g. init-based BSD systems.

> > I appreciate the goal of the OP, but their investigation and survey were insufficient.
> 
> Please allow me to add a bit more to the surves/investigation: You might not have found too much information about this config thing because not everything is necessarily written down, and even if it is, may not be indexed by Google (e.g. Fedoras Pagure is badly represented).

Yes, that is the problem, but in this case, I expect it to be the PR submitter's responsibility to provide enough information and references when requested. The OP didn't provide any links but the UAPI thread.

> > Their arguments lacked knowledge about modern setups (as well as the historical ones) and were based on misunderstandings about basics.
> 
> You cannot expect that everyone has a deep understanding of your project.

I didn't talk about the knowledge in bash-completion. I talked about the basic knowledge related to Unix and standards. The OP claimed that the shell scripts are executable binaries. The OP claimed that the footnotes in specifications are what are discouraged. The OP didn't know about the traditional configuration of `init`, or the OP thinks System V `init` was wrong from the beginning. I can understand that the traditional configurations have problems from the modern point of view, but it wasn't wrong from the beginning. The OP claimed that the mailing list debian-devel is not for discussions but for spams from random people, while [the wiki](https://wiki.debian.org/DebianDeveloper) says debian-devel is for "*discussion of technical development topics*".

</details>

<details><summary>Comments on the last reply from the OP</summary>

https://github.com/scop/bash-completion/pull/1399#issuecomment-3021288485

> > I generally appreciate the idea of organizing `/etc` better. However, I'm not sure if the suggested setup would finally be the one that everyone wants. Then, from your point of view, with this patch, you become free from patching and modifications to the package source code, but all the other people who want a different setup still need to apply patches.
> 
> I really don't think that's true. I see no evidence of anybody intentionally wanting to ship upstream code in /etc/,

I talked about a different setup, but the different setup doesn't necessarily require installing upstream codes in `/etc`. I had e.g. the symbolic links in `/etc/bash_completion.d` in my mind (as mentioned above), but other setups (that I'm unaware of) may also be possible.

> > In addition, this also affects all the projects that currently install a script in `/etc/bash_completion.d`. They need to be changed to comply your suggestion, but it takes time because the projects still using `/etc/bash_completion.d` (API v1) are likely to be defunct except for the projects that intentionally load their scripts eagerly.
> 
> It doesn't? Backward compatibility is kept, intentionally, any existing script will still be loaded from the previous location too and work as expected.

I wrote "*to comply your suggestion*". You are right that the change of this PR doesn't break the existing completions. However, my point is that even if `bash-completion` is patched, all projects shipping their completion in `/etc/bash_completion.d` still need to be patched at the package level *to remove `/etc/bash_completion.d` from the distributions*. To become completely free from patches, one needs to request changes in the (possibly defunct) upstream projects shipping outdated completions.

> > So far, I do not see any major movements in distributions to carry it out in a collective way.
> 
> There is. We do, and anybody else doesn't really care one way or the other. Every distribution that matters adopted usr-merge for this (and many others) reason.

I expected references to discussions about `/etc/profile`, `/etc/profile.d`, `/etc/bash_completion.d`, etc., which haven't yet been provided. Then, I'm not sure if I can trust the information, in particular after seeing those threads in Debian.

> > They raised discussions in Debian
> 
> No, we didn't. Please do not mistake some random thread by some random person in some random mailing list as "discussions"  - debian-devel is not for discussions, it's just random spam by random accounts, and it's best ignored.

The first thread that I linked in https://github.com/scop/bash-completion/pull/1399#issuecomment-2997078268 explicitly mentioned the UAPI as the group behind it. Another thread is started by the OP of this PR. The above statement implies that the OP is one of random accounts posting random spam to the Debian mailing list. This is strange. If that's true, I need to be careful about the possibility that this PR is also spam from random accounts in the UAPI (instead of an official representative of the UAPI).

> > > But without 000_bash_completion_compat.bash, it's not possible to use `_command_offset` - if that's not deprecated, then this file is not solely for backward compat?
> >
> > The directory is meant for backward compatibility, but the file is not.
> 
> Then why is the file itself called "compat"?

To write it down explicitly, the directory is meant for backward compatibility with the deprecated interface v1, but the file is for the compatibility with v2, which hasn't been deprecated. Anyway, the file `000_bash_completion_compat.bash` is not for the deprecated things as I've repeatedly linked as https://github.com/scop/bash-completion/discussions/1329#discussioncomment-12163014.

septatrix:
> @septatrix I see you are from UAPI and seem to know the relevant details. Can we continue the discussion?

I am not connected to UAPI in any way apart from following and commenting on some of their issues and deploying images which benefit from their specs.

> If any, can you point me to where to look at what's going on with `/etc/profile` and `/etc/profile.d` under the UAPI spec? [...] As I've written above, an alternative setup could be to make `/etc/profile.d` and `/etc/bash_completion.d` the places to symbolic links to the physical files in `/usr/share`, which seems to be what `systemd` does for services and other targets. However, the OP ignored this without any counterarguments. Is it specified by the UAPI that a setup like the symbolic links in `/etc/bash_completion.d` should not be used? What is the difference from the use by `systemd`?

Ideally `/etc/profile` et al. would also gain support for lookup under `/usr` but no one has taken that on yet. That conversion will also take a lot longer because it involves many more dependents. Symlinks are also viable but native support is preferred because something (like systemd-tmpfiles) needs to ensure that the symlinks exist at boot.

For more complex systems there is the argument that config files in `/etc` overwrite those with the same name under `/usr` (think stuff like Apache site config, systemd services, sysctl.d etc.) though I do not think this will be necessary here.

> > Parties represented during those summits (2022-2024 combined) where among others: [...]
> Are BSD distributions included in this list? I can understand it even if the BSD distributions are excluded from the UAPI because the UAPI seems to target Linux. However, bash-completion targets also BSD, and we are interested in what the UAPI thinks of the different setups of `/etc` in different types of operating systems.

Not to my knowledge but adopting the config spec does not break compatibility with them.

> > When it comes to modern setups I also agree with him. [...]
> 
> Based on the arguments provided until now, it seems to me that the current movements on `/etc` are very much related to what is called the image-based Linux, but are these related to the Linux distributions for desktops, workstations, and servers? Are these related to BSD? I'm not going to say that I reject it if it is for the image-based Linux, but I just want to make it clear. If this reorganization of `/etc` is intended to smoothly create the image-based Linux, that's fine, but you can't say the traditional configurations were wrong from the beginning. System V `init` just didn't target the image-based Linux.

Apart from the layering, priority, merging aspects of the config spec which applies less to bash-completion the most important aspect is that deploying/updating systems becomes easier. This is very important in the context of image-based deployments (those who update e.g. whole partitions instead of using a package manager) but also in other contexts. Let me expand on that:

1. The whole of `/etc` can be made a tmpfs. More so,  it is possible to boot from a partition only containing `/usr` (so called usr-only, or hermetic-usr). This is ideal for deployments where persistent changes are undesired, or to eliminate some ways for malware to persist.
2. If everything is self-container in `/usr` this makes it possible to update the OS by updating a single partition which can be otherwise readonly and protected by a checksum. Files outside of this partition would not get updated any longer
3. Users avoid touching stuff under `/usr`. They have been tought that modifying configs under `/etc` is fine but this may result in broken systems. E.g. all hell could break lose if someone tinkers with `/etc/bash_completion.d/000_bash_completion_compat.bash`. And even if they make a harmless change, normal package managers would refuse to update that file afterwards resulting in potentially broken configuration.
4. A "factory-reset" functionality which can prune everything under /etc without bricking the system
5. There are probably a few more small niceties...

Point 3 is relevant for all distros, 1 only for a selected few because support is still somewhat bad (openSUSE and Fedora already can do this for the most part, Debian and Ubuntu fall apart last I tested), 2 is relevant for distros like Ubuntu Core, Fedora Silverblue/Kinoite, NixOS, PostmarketOS, etc.

> > You also stated that the proposed changes seem harmless to you - and they are very small.
> 
> The size of the code change is irrelevant now. We are supposed to carefully consider the impact on users and maintainability as maintainers. We are not supposed to blindly accept the PR, no matter how small the code change is.
> 
> > In lie of having every distro carry this as a downstream patch it seems reasonable to me to accept this.
> 
> I'm unsure whether the statement that _every_ distro applies _the same_ patch is correct. If the symbolic links in `/etc/bash_completion.d` could be used, we don't have to change anything. One remaining issue is that `000_bash_completion_compat.bash` is installed in the "deprecated" directory `/etc/bash_completion.d`, but this is a known issue discussed in #1329 and should be resolved within #1329. I mentioned #1329 multiple times, but I doubt that the OP has checked it.

They should work but then something would have to create them, i.e. you would need to ship a tmpfiles.d config. The use-case you should have in mind is that booting with an empty /etc should eventually work out-of-the-box.

> Off topic
> 
> Comments on the last reply from the OP

GitHub seems to have swallowed both of these when quoting/replying, I'll leave them be if that is fine with you unless you would like some comments on that

Makefile.am
```diff
@@ -3,7 +3,7 @@ SUBDIRS = completions doc helpers test
 pkgdata_DATA = bash_completion
 
 # Empty, but here just to get the compat dir created with install
-compatdir = $(sysconfdir)/bash_completion.d
+compatdir = $(datadir)/bash-completion/compat
```
scop: This calls for a couple of updates
- we should create both the old compat dir and the new one on install
- the comment should be updated along the new implementation (it's bitrotten as it stands as the dir is not empty, but with the new impl one of them is)

scop (changes requested):
I won't go back and read the discussion, but I'm amenable to accept this if other members feel the same. Just needs a rebase and addressing a comment I'm leaving here. Not sure of the mechanics if the original submitter is no longer following, maybe a new PR needs to be done if some non-member takes care of the updating?

I aim to push a new release soon, so this has a chance to make it if we can get it sorted rather quickly.

Makefile.am
bluca: Updated as requested, thanks

akinomyoga:
I've corrected the PR title, reflecting the actual change of the current version. The first commit has been merged in #1447.

akinomyoga:
After that, I found the page "[Configuration Files Specification | UAPI Group Specifications](https://uapi-group.org/specifications/specs/configuration_files_specification/)" in the UAPI repository by myself, and it told me the details that hadn't been revealed or whose reference hadn't been given in this thread. After reading it, my "*impression*" was that the change suggested in this PR doesn't directly try to implement their rule, but this PR tried to resolve a problem while trying to apply their spec. However, *the problem* seemed to be exactly one of the problems discussed in #1329. Then, my naive conclusion is that this should be discussed along with #1329, and the change should be applied under a consistent rule.

However, I forgot the details after the three-month blank period, and the above is just *an impression* remaining in my mind. I need to take time to think about it again.

> I aim to push a new release soon, so this has a chance to make it if we can get it sorted rather quickly.

I feel that this PR should be discussed along with #1329 after the release (unless we try to settle #1329 quickly before the next release). @scop Do you have any thoughts on #1329?

scop:
Yeah, let's get #1329 resolved the way we want and punt this along with it to the next release. Left some comments over there earlier this week.

akinomyoga:
#1558, which partly addresses #1329 and contains changes relevant to this thread, has been merged. In #1558, a new location for startup scripts (which are eagerly loaded in the initialization of bash-completion), `share/bash-completion/startup{,-core}` has been prepared, and `/etc/bash_completion.d/000_bash_completion_compat.bash` has been moved to `startup-core`. Although it has never been mentioned in this PR, #1558 also borrowed an idea from UAPI's [configuration file spec](https://uapi-group.org/specifications/specs/configuration_files_specification/) (see ff1bf722c6ecbf0d6fe491d345c08fa0c981035e and e.g., https://github.com/scop/bash-completion/pull/1558#discussion_r2728425654 for details). [ Note: we only borrowed an idea of the mechanism of shadowing a configuration by putting an empty file, but we don't use the specific directory structures that the UAPI prepares. ]

This PR tries to move the location of `/etc/bash_completion.d/000_bash_completion_compat.bash`, while still loading scripts installed in `/etc/bash_completion.d` by external projects. This has already been achieved by #1558 as a part of #1329.

Also, this PR prepares a new location `share/bash-completion/compat` for completion settings (which have been installed in `/etc/bash_completion.d`) and updates the pkgconfig variable `compatdir`. There are two types of completion settings installed in `/etc/bash_completion.d` by external projects. One is the completion settings written for an old API of `bash-completion-1.*` (which I think was deprecated more than 14 years ago). The other type of completion settings is the startup scripts that want to adjust the global setting (e.g., `complete -D`) such as `fzf` and `_python-argcomplete`. While the latter should now go to `share/bash-completion/startup`, I'm not sure if we need to prepare a new directory for the former, the old completion settings for API v1.

- First, external projects that install completions with API v1 are likely stale, and even if we provide a new location, there is no chance for them to be updated to use the new location. If there is any chance, the completion scripts should be updated to use API v2. Actually, we already removed some interfaces of API v1, so they are not guaranteed to work under the present version of `bash-completion`.
- Or some external projects intentionally keep the completion settings for API v1 to support Bash 3.2 and bash-completion 1.* for macOS. In that case, they still need to make the completion settings to work with bash-completion, so they need to continue to use `/etc/bash_completion.d` no matter how we change the setup for bash-completion 2.*.

Some of those projects shipping API-v1 completions may reference the pkgconfig parameter `compatdir` to choose the install target, but there are still many projects that directly install the completion settings to `/etc/bash_completion.d`. The issue wouldn't be solved simply by changing `compatdir`.

My feeling is that the changes in #1558 are already enough to resolve the original issue that this PR tried to fix.

PR #1403, opened 2025-07-18
fix: completion generation inside $()

### LINKED ISSUE #630: bash-completion fails to generate path completions with `$(`
## Describe the bug

If `bash-completion` package is installed, bash does not generate completions with `$(`.

## To reproduce

1. Execute `touch test.out` in current directory.
2. Execute `for i in $(cat <TAB>`.
3. No completions generated for `cat` command.

## Expected behavior

There should be completions listed for `cat` command.

- [x] Operating system name/distribution and version: Fedora release 33 (Thirty Three)
- [x] bash version, `echo "$BASH_VERSION"`: 5.0.17(1)-release
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.8

## Additional context

It is caused by a behavior change introduced in bash-4.3 (by [this commit](http://git.savannah.gnu.org/cgit/bash.git/commit/?id=ba4ab0554b02cf9d5bb98a633971692deed94c02) to `bashline.c`).

I discussed this change with Chet over e-mail and this is what he mentioned:

> It looks like the completer for `for' doesn't know how to handle getting
the command substitution as a single word (since that's what it is). I
guess it must be trying to parse that out on its own, using COMP_WORDS,
since you get this when you run an instrumented function as the completer:
>
> touch /tmp/test.txt
> for f in $(cat /tmp/te[TAB]
>
> args: for /tmp/te in
> declare -a COMP_WORDS=([0]="for" [1]="f" [2]="in" [3]="\$(cat /tmp/te")
> COMP_CWORD=3

> Note that the word readline is trying to complete is always passed to the
completion function as $2, and that is /tmp/te. But the words on the line
are passed as COMP_WORDS, and ${COMP_WORDS[$COMP_CWORD]} is as above.

> Bash-4.2 looked `into' the command substitution and passed

> args: cat /tmp/te cat
> declare -a COMP_WORDS='([0]="cat" [1]="/tmp/te")'
> COMP_CWORD=1

> and so ran the completion for `cat'.

> But that was fragile, and didn't take into account other command
substitutions on the line. For instance, when finding the command name to
be completed, we want to skip over command substitutions in assignment
statements, something bash-4.2 had real problems with. Here's a bug report
on that:

> https://lists.gnu.org/archive/html/bug-bash/2011-12/msg00053.html

> So bash-4.3 changed this. The actual change was 12/16/2011 to bashline.c:
find_cmd_start(), which now considers command substitutions as single
words, or parts of single words, as the rest of the shell does. So the
words that a completer gets are closer to the words that the parser will
eventually produce. That is why bash-4.2 thinks the command name is `cat'
and bash-4.3 thinks it's `for'.

> If bash-completion has a function that understands how to complete words
beginning with `$(', possibly recursively, that could work. Or the function
could look at $2 to see the word that readline thinks should be completed
(readline's pretty generic, and just goes back to the whitespace).

richardkmichael:
Closes #630 

- `python -m pytest test/t/unit` has two failing tests and two skipped tests in `test_unit_compgen_filedir.py`.  This is the same as on `main`, so I do not believe they are related to this patch.

- `python -m mypy .` and `python -m ruff chec` pass

- I am unsure what to make of the `runLint` output, but it has exit status `0`, so I presume it is OK


bash_completion
```diff
@@ -1483,6 +1483,27 @@ _comp_initialize()
...
 
+    # If the current word is a command substitution $(, reset the completion context to be the
+    # command line *inside* the substitution.
+    if [[ $cur == '$('* ]]; then
+        local inner_line=${cur#\$\(}
+
+        # Replace the completion variables with ones that reflect the inner context.
+        COMP_LINE=$inner_line
+        read -ra COMP_WORDS <<<"$inner_line"
+        COMP_CWORD=$((${#COMP_WORDS[@]} - 1))
```
yedayak: This can become negative if there isn't anything inside the `$()`:
```
echo $(<TAB>bash: COMP_WORDS: bad array subscript
```

yedayak (commented):
In main `echo $(ip <TAB>` completes all files (the default completion, it should complete sub-commands), but with this PR it doesn't complete anything. Maybe you can use `_comp_command_offset` for this?

richardkmichael:
@yedayak Thanks for the quick review.  I'll investigate.

richardkmichael:
@yedayak 

> In main `echo $(ip <TAB>` completes all files (the default completion, it should complete sub-commands), but with this PR it doesn't complete anything. Maybe you can use `_comp_command_offset` for this?

Thanks for mentioning `_comp_command_offset`.

All completions should now be correct, e.g.,

```
bash-5.3$ echo $(git describe --<TAB>
--abbrev         --broken         --debug          --exclude=       --match=         --tags
--all            --candidates=    --dirty          --first-parent   --no-...
--always         --contains       --exact-match    --long           --no-contains


# Empty command substitution completion is empty, consistent with empty prompt completion.
bash-5.3$ echo $(<TAB>
bash-5.3$ <TAB>
```

Handling nesting and completion position is a little complicated, so I commented the code somewhat more than other areas.  Let me know if I should reduce them -- I'd find it helpful being new to the code base, but it might be verbose for experts.

- I added more test cases
- I checked lint and reviewed the code style and API guidelines
- I ran the [GitHub Actions in my fork](https://github.com/richardkmichael/bash-completion/actions/runs/21849729836?pr=2)

test/t/unit/test_unit_command_substitution.py
```diff
@@ -0,0 +1,62 @@
...
+    ignore_env=r"^[+-](COMPREPLY|REPLY)=",
+)
+class TestUnitCommandSubstitution:
+    """Test completion inside $() command substitutions.
+
+    Each test sends real keystrokes via assert_complete, comparing
+    completions at the bare prompt with completions inside $().
+    """
+
+    wordlist = sorted(["alpha", "bravo"])
```
yedayak: Is there a reason to sort here explicitly?

test/t/unit/test_unit_command_substitution.py
```diff
@@ -0,0 +1,62 @@
...
+            '  COMPREPLY=($(compgen -W "%s"'
+            ' -- "${COMP_WORDS[COMP_CWORD]}"));'
+            "}; "
+            "complete -F _csub_compfunc csub_cmd" % " ".join(self.wordlist),
+        )
+
+    def test_normal_not_affected(self, bash, functions):
+        """Completion without $( still works."""
+        assert assert_complete(bash, "csub_cmd ") == self.wordlist
+
```
yedayak: I don't think this is necessary, basically all other tests (in other files) will fail if this does

test/t/unit/test_unit_command_substitution.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_single_level(self, bash, functions):
+        """echo $(csub_cmd <TAB> produces same completions."""
```
yedayak: You don't need to repeat the cmdline in assert_complete for each test, just the essence of the test if you think its not obvious.

bash_completion
```diff
@@ -1626,6 +1626,83 @@ _comp_initialize()
...
+                    fi
+                fi
+            done
+
+            # No unclosed $( found -- we're at the innermost level.
+            [[ $nested_idx ]] || break
+
+            # Rebuild inner_line from the nested word onward, then
+            # strip its leading $( for the next iteration.
+            inner_line="${COMP_WORDS[*]:nested_idx}"
```
yedayak:
```suggestion
            inner_line="${COMP_WORDS[@]:nested_idx}"
```

test/t/unit/test_unit_command_substitution.py
```diff
@@ -0,0 +1,62 @@
...
+
+@pytest.mark.bashcomp(
+    cmd=None,
+    ignore_env=r"^[+-](COMPREPLY|REPLY)=",
+)
+class TestUnitCommandSubstitution:
+    """Test completion inside $() command substitutions.
+
+    Each test sends real keystrokes via assert_complete, comparing
+    completions at the bare prompt with completions inside $().
```
yedayak:
```suggestion
```

test/t/unit/test_unit_command_substitution.py
richardkmichael: No, it's noise; removed.

richardkmichael:
> Maybe add test for something like `echo $(csub_cmd) ` where the substitution is closed?

Added.  I noticed the fixture directory, so I used it for stable `echo` completion.  I considered `csub_cmd $(echo hi) `, but I wondered if the `echo <TAB>` suggestion was to have one real completion.

> Edit: Also maybe one where one subsitution is closed but the outer one is still opened, `echo $(csub_cmd $(csub_cmd) `

Added, and thank you, this caught a bug.  I hadn't considered the `_comp_split` returning `COMP_WORDS` with `$(` and its `)` complement in two different words, e.g., `(... "$(echo" "hi)" ...)`, so the counting and traversal was wrong.

bash_completion
richardkmichael: I believe either will work for the scalar assignment.  However, pre-commit shellcheck fails with `@`.  Should it be disabled?

```bash
# shellcheck disable=SC2124
inner_line="${COMP_WORDS[@]:nested_idx}"
```

akinomyoga: `[*]` is affected by `IFS`, while `[@]` always joins the words by whitespace. Actually, one can set `local IFS=$' \t\n'` and use `[*]`.

richardkmichael: Thanks for the explanation.  Given the preference for `@`, what should be done about shellcheck?

akinomyoga: Actually, we have never used `@` in this context. I think we just used `IFS` and `*` for this type of processing.

yedayak: Yeah sorry, I didn't know about the shellcheck rule. `[*]` with setting IFS locally is fine

akinomyoga: I searched the codebase. It seems we haven't had similar cases (with `[*]` or `[@]` in the right-hand side of a scalar assignment) so far. The following cases explicitly set `IFS` to a special value:

```console
./completions-core/gnokii.bash-229-    local IFS='|'
./completions-core/gnokii.bash:230:    local regex_main_cmd="(${main_cmd[*]})($|[^_[:alnum:]])"
--
./completions-core/invoke-rc.d.bash-20-        local IFS='|'
./completions-core/invoke-rc.d.bash:21:        local exclude="@(${words[*]})"
--
./completions-core/lintian.bash-33-        local IFS='|'
./completions-core/lintian.bash:34:        todisable="@(${todisable[*]})"
--
./completions-core/mr.bash-23-    local IFS='|'
./completions-core/mr.bash:24:    local glob_commands="@(${commands[*]})"
```

The following case is not accompanied by any adjustment of `IFS`, so it's broken with a custom `IFS`:

```console
./completions-core/tree.bash:37:    local line="${words[*]}"
```

If we search for `${a[*]}` not restricted in the right-hand sides of assignments, there seem to be many cases affected by custom `IFS`, which I think we can work on in another PR.

There was no occurrence of `"${a[@]}"` in the right-hand sides of assignments.

richardkmichael: Thank you 👍 

I did a similar exploration after this feedback.  

I was unsure about using `IFS`, due to scope effects elsewhere  -- especially since there is no RHS scalar assignment like this in `bash_completion`, and `_comp_initialize` is a function called quite early.  For example, it will affect `_comp_command_offset`.

I also wondered if, when `IFS` is used, it's set at the top of a function (i.e., for reader visibility), or instead near the site needed / affected.  The approaches seems evenly split.  For example, `_comp_split` is at the top, so is `ssh-keygen.bash`.  Usage in `_comp_compgen` is at the usage site with a comment, so I've set it just above usage as well.  (I also considered a new function, to confine IFS scope, but it seemed unnecessary.) 

Aside from this PR --

Since lint can't be used for this, an addition to the quoting section of the style guide could be helpful.  Perhaps something like:

```diff
diff --git a/doc/styleguide.md b/doc/styleguide.md
index cff6d95b..436a631b 100644
--- a/doc/styleguide.md
+++ b/doc/styleguide.md
@@ -171,6 +171,10 @@ exceptions where the quoting is still needed for other reasons.
   suggests that e.g. `var= cmd` is confusing with `var=cmd`.
 - `$*` and `${array[*]}` need to be always quoted because they can be affected
   by the word splitting in bash <= 4.2 even in the above contexts.
+- When joining array elements into a scalar variable, prefer `"${array[*]}"`
+  over `"${array[@]}"`, and ensure `IFS` is set to the intended separator
+  (typically `local IFS=$' \t\n'`), because `[*]` joins by the first character
+  of `IFS`.
 - In the following contexts, double-quoting of shell expansions is needed
   unless the result of expansions is intentionally treated as glob patterns or
   regular expressions.
```

I used an LLM for code exploration and review (help ensuring adherence to the style guide and API naming documentation), and such an explanation would have caught this usage.

akinomyoga: > For example, `_comp_split` is at the top,

That is because `IFS` is determined by the options specified to `_comp_split`, and we need to give the default value at the top of `_comp_split` before starting the argument parsing.

> so is `ssh-keygen.bash`.

I think that is because `${words[*]}` and `${protocols[*]}` are used many times (8 times) in the function, and the smallest common scope seems to be the function scope.

However, there doesn't seem to be a particular reason for `IFS` set at the top of `_comp_load`. Git Blame reveals that it was introduced in commit d9082d2c8d, but I don't remember the reason that I put it at the top. The `IFS` is necessary for the line `$(complete -p "$cmd" 2>/dev/null || echo false) "\\$cmd" && return 0`, so it can be set just above this line technically.

> (I also considered a new function, to confine IFS scope, but it seemed unnecessary.)

Yeah, that is another possibility, something like `_comp_join`.

bash_completion
```diff
@@ -1626,6 +1626,85 @@ _comp_initialize()
...
+                    local without_opens=${word//\$\(/}
+                    ((open_count += (${#word} - ${#without_opens}) / 2))
+                    if ((open_count > close_count)); then
+                        nested_idx=$i
+                        # re-parse from this word onward;
+                        # inner_line is assembled below
+                        break
+                    fi
+                fi
+            done
```
akinomyoga, line 1687: The current implementation falls into an infinite loop with the following completion attempt:

```console
$ cmd1 $(cmd2 a$(cmd3)b$(cmd4 [tab]
```

Also,  the command name is wrongly extracted to be `pwd)/$(cmd3` with the following completion attempt:

```console
$ cmd1 $(cmd2 $(pwd)/$(cmd3 [tab]
```

Actually, I have difficulty understanding the choice of the present algorithm. Even if the above issues are fixed somehow, is this algorithm stable and robust in general? Since I currently don't see a clear path or picture to understand the construction of the current algorithm, the algorithm appears to be a random trial currently.

Can you explain what is the assumptions of the algorithm, and how it is ensured to work under the given assumptions?

richardkmichael: Thank you!

Given the failures, yes, one assumption that words do not contain unbalanced pairs is a mistake, and another being that the first `$(` in the `COMP_WORD` is the correct one.  Evidently not robust. :) I was trying to find the word right-adjacent to the last open `$(` but the pair-wise counting by line length difference won't work, and the strip is to the wrong `$(`.  I wouldn't say "random trial" :-) but definitely too naive.  I was experimenting with test cases on my command line, and hoping to avoid writing a parens-matcher with character position.

I'll go through it more carefully, for another approach.

akinomyoga: Thank you for the explanation!

Thank you for thinking about another approach! If it turns out to be difficult, we don't necessarily have to support all possible cases. I think the important thing is that we want to gracefully fail in the unsupported cases (instead of causing an infinite loop or producing a strange result); for example, if the required assumptions are clear, we may examine the contents of `cur` and give up the special treatment when the assumptions are unsatisfied.

PR #1404, opened 2025-08-02
bsdtar and unzip should support .apk, .ipa and .hap

eagleoflqj:
.hap is the extension of HarmonyOS app package, which is a zip.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -3157,7 +3157,7 @@ _comp__init_install_xspec()
 }
 # bzcmp, bzdiff, bz*grep, bzless, bzmore intentionally not here, see Debian: #455510
 _comp__init_install_xspec '!*.?(t)bz?(2)' bunzip2 bzcat pbunzip2 pbzcat lbunzip2 lbzcat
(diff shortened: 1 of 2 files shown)


akinomyoga:
I'm not sure if we should add all *actually-zip-archive* files, but I see that many such extensions have already been added. Should Google Chrome Extension (`.crx`) be added too?

scop:
I don't see a problem adding more of them. crx would fit in well if it's a zip, we have xpi already which is corresponding thing for Firefox.

scop:
But re these, would be great to provide URLs to some sample files with which reviewers could verify that they work with these tools. I couldn't locate a .hap to download in a 10 second web search, and I've no idea what an .ipa is (but I do see it's included for unzip and zipinfo).

eagleoflqj:
> if we should add all actually-zip-archive files

At least case-by-case I guess? As an amateur app developer I did have times that stuff inside package is missing and I need `bsdtar tf xxx.hap/ipa`, where auto-completion doesn't work and that's why I submit this.

eagleoflqj:
>  provide URLs to some sample files

Sure, here is a .hap https://github.com/mgz0227/legado-Harmony/releases/tag/V1.0.1.6
and here is a .ipa https://github.com/Ingan121/FSUntether/releases/tag/fsug
Android apk ≈ iOS ipa ≈ HarmonyOS hap

Ideally bsdtar's zip rules could share with unzip, but I'm not quite familiar with the bash art so I just did some addition here.

scop (approved):
Thanks!

PR #1406, opened 2025-08-03
feat(unzip,zipinfo): generate *.crx for Google Chrome Extension's ZIP archives

akinomyoga:
See the commit message.

> https://github.com/scop/bash-completion/pull/1404#issuecomment-3147310228
>
> I don't see a problem adding more of them. crx would fit in well if it's a zip, we have xpi already which is corresponding thing for Firefox.
>
> https://github.com/scop/bash-completion/pull/1404#issuecomment-3147329632
>
> But re these, would be great to provide URLs to some sample files with which reviewers could verify that they work with these tools. 

I've included the reference URL in the commit message as usual, but it might be harder to track the changes later because many extensions are included in a single line where `git blame` isn't useful to identify a commit that added a specific filename extension. Maybe we should include the reference information directly in the code comment inside the source code `bash_completion`. What do you think?

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -3157,7 +3157,7 @@ _comp__init_install_xspec()
 }
 # bzcmp, bzdiff, bz*grep, bzless, bzmore intentionally not here, see Debian: #455510
 _comp__init_install_xspec '!*.?(t)bz?(2)' bunzip2 bzcat pbunzip2 pbzcat lbunzip2 lbzcat
-_comp__init_install_xspec '!*.@(zip|[aegjkswx]ar|exe|pk3|wsz|zargo|xpi|s[tx][cdiw]|sx[gm]|o[dt][tspgfc]|od[bm]|oxt|?(o)xps|epub|cbz|apk|apk[ms]|aab|xapk|ipa|hap|do[ct][xm]|p[op]t[mx]|xl[st][xm]|pyz|vsix|whl|[Ff][Cc][Ss]td)' unzip zipinfo
+_comp__init_install_xspec '!*.@(zip|[aegjkswx]ar|exe|pk3|wsz|zargo|xpi|crx|s[tx][cdiw]|sx[gm]|o[dt][tspgfc]|od[bm]|oxt|?(o)xps|epub|cbz|apk|apk[ms]|aab|xapk|ipa|hap|do[ct][xm]|p[op]t[mx]|xl[st][xm]|pyz|vsix|whl|[Ff][Cc][Ss]td)' unzip zipinfo
 _comp__init_install_xspec '*.Z' compress znew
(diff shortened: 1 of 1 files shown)


akinomyoga:
The following commits have added filename extensions for `unzip` and `zipinfo`.

- 92f558d582c452e96872f480ebf86cc3c08f0a87
- 3c3ab1781bf6061607d327a72a77ea93cf8aaedf
- fcaf3d4118af251973a57028302bf057a9681f1f
- 29c8644f6e86094f4f8acf5152cabe657f40a92f
- 4c1cf9dd6725b38cf7f9c35bf9be0e0a9feac708
- 8273ce3138708ad1db851b74eafba38d6683ca0a
- 31b5cbc8016b181675e10dd068c92008782d4196
- c263a3fc8ca098cb42eab88026e1ffa2fa809ec0
- df8bc32be0989cef6c4baaa30230848f9d1ea4ad
- e5ee408a5aee2f1c9d898150e256540d020fc5ee
- 7b9dc46c6a26a1a8e99409bb7752dfe153f2de0f
- a5761607bd37a201a2512f10bbc7be646e51b2fb
- f0a3147179ea06bdcd463e2d160df97255656e95
- ca533452195134de6ceb5c3f894e824fac53849a (Note: this removed the code comment for the explanation on some filename extensions)
- 7c7b5608c8505633aa55e106f98d2a943937d41a
- 6a71ee504c11b798125b9c85359d5b7367a27dd4
- 8890abfdfb716692bf1242ac6d2701160ffc5ee6
- 0ceca807c50f9cec823999e497f6842c3f16e6e5
- 9566f5994afc9ee40a7561b0f0ad1422bcedca97
- ba9cacc77c12d72854a04f3fee01b1b9b5c59084
- 9cbd3ceab20f9b6ac244e28ae018f9e91929491b
- 95662ec19a7f9fd3aa4045899a2f09e9e3fb24ba
- 0b2be3fc7b8a3bfdc176b83cb6160375861d8d3e
- 82224e9b28ff14d755c8465634af3b6c6f88699c
- b64e48ad037b2f8ccf0f4c18784b54a75ae654f6
- 5b55df4156593a4d8e6b2f214faad3e892c5992f
- 7c16b7956ddc827d4bdf46a2e57b7aded86db140
- 63efd13c161ddd74a3d9fec9b4a4077b8931f034
- 5fa59bcce25294e6451a742b520c8bcb6edb8c69
- d1dfcfb8aaff76bd98fb1a6f173bc5818700fc98
- c442af0caa89a9ff2fd70b5971aaa77387acd4e6

scop:
Agree that git blame isn't too useful for these lines, but I think it's good enough if we clearly note each extension in the commit message. With the preceding dot it should be searchable enough from git log.

akinomyoga:
Thanks!

PR #1408, opened 2025-08-03
test(sshfs,scp): create files on the fly to work around Windows filesystem

akinomyoga:
This addresses the remaining issue with PR 1397, https://github.com/scop/bash-completion/pull/1397#discussion_r2164546105.

e8aeed702 Also, the second commit adds tests for the regression of PR 1397 with the `scp` command, which was not added in PR 1397 because I thought it could not be tested. As commented in https://github.com/scop/bash-completion/pull/1397#discussion_r2165093062, it was actually possible. I'm adding tests using `completion.output` in this patch.

### DIFF
--- a/test/t/test_scp.py
+++ b/test/t/test_scp.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from itertools import chain
 
 import pytest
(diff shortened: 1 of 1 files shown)


test/t/test_sshfs.py
scop, line 1: Should we not remove the `test/fixtures/sshfs` dir and its contents as part of this change? It would seem to be unused after this, and would still cause problems if around I guess.

akinomyoga: Ah, right. I completely forgot that. I'll address it.

akinomyoga: [`3c07ffd..8841957a`](https://github.com/scop/bash-completion/compare/3c07ffdb80b0bafdc8cb4a375f6faa24a1263ef7..8841957a3d7317d12d508e9ce622db6505c680bf)

akinomyoga:
Thank you!

PR #1412, opened 2025-08-09
refactor(scp): factorize common codes

akinomyoga:
This is another PR separated from #1357. This commit is just trivial refactoring, but the next commit is nontrivial. I decided to submit this single commit before submitting the rest.

### DIFF
--- a/completions/ssh
+++ b/completions/ssh
@@ -457,6 +457,54 @@ _comp_cmd_sftp()
 # things we want to backslash escape in scp paths
 _comp_cmd_scp__path_esc='[][(){}<>"'"'"',:;^&!$=?`\\|[:space:]]'
 
+# Escape shell special characters in filenames by backslash.  This also
+# suffixes a space or a slash based on the file type.
+#
+# Note: With a non-empty prefix ($1 of _comp_xfunc_scp_compgen_local_files),
+# Bash will not recognize any filenames, so we need to perform the proper
+# quoting manually.  We also need to manually suffix a space or a slash based
+# on the file type because "-o nospace" is specified.  One might think of using
+# "compopt +o nospace" instead, but it would suffix a space to directory names
+# unexpectedly.
+#
+# Options:
+#     -d  Only directory names are selected.
+# @param $1 escape_replacement - If a non-empty value is specified, special
+#     characters are replaced with the specified value (instead of the default
+#     '\\&').
+# @stdin List of filenames in the "ls -1F" format, where filenames are
+#     separated by newlines, and characters /=@|* are suffixed based on the
+#     types of the files.
+_comp_cmd_scp__escape_path()
+{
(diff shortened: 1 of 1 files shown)


akinomyoga:
I realized that fe7a3f97c is a related commit, which was added after @yedayak's comment at https://github.com/scop/bash-completion/pull/1357#discussion_r2029957266, so I pushed the commit. If a fixed positional argument is preferred, the commit can be dropped. Otherwise, this extra commit can finally be squashed at the merging time.

completions/ssh
```diff
@@ -461,6 +461,54 @@ _comp_cmd_sftp()
...
+    local escape_replacement=${1:-'\\&'}
+
+    if [[ $dirs_only ]]; then
+        # escape problematic characters; remove non-dirs
+        command sed \
+            -e '/[^/]$/d' \
+            -e 's/'"$_comp_cmd_scp__path_esc"'/'"$escape_replacement"'/g'
+    else
+        # escape problematic characters; remove executables, aliases, pipes
+        # and sockets; add space at end of file names
```
scop: Not introduced by this PR, but I suppose we could improve it here in another commit:
```suggestion
        # escape problematic characters; remove executable, symlink, pipe,
        # socket and door indicators; add space at end of file names
```

Ref https://www.gnu.org/software/coreutils/manual/html_node/General-output-formatting.html#index-_002dF-3

Re "door", see another comment below.

completions/ssh
```diff
@@ -461,6 +461,54 @@ _comp_cmd_sftp()
...
+        command sed \
+            -e 's/[*@|=]$//g' \
```
scop: I don't know what a "door" is, but it does have an indicator of its own, see above ref.
```suggestion
            -e 's/[*@|=>]$//g' \
```

akinomyoga: I don't know either, but it seems to be a special type of files in Solaris.

https://en.wikipedia.org/wiki/Doors_(computing)

completions/ssh
```diff
@@ -461,6 +461,54 @@ _comp_cmd_sftp()
...
+# "compopt +o nospace" instead, but it would suffix a space to directory names
+# unexpectedly.
+#
+# Options:
+#     -d  Only directory names are selected.
+# @param $1 escape_replacement - If a non-empty value is specified, special
+#     characters are replaced with the specified value (instead of the default
+#     '\\&').
+# @stdin List of filenames in the "ls -1F" format, where filenames are
+#     separated by newlines, and characters /=@|* are suffixed based on the
```
akinomyoga:
```suggestion
#     separated by newlines, and characters /*@|=> are suffixed based on the
```

akinomyoga:
I applied the door changes. I also squashed the second commit (the change for the `-d` command-line flag) into the first commit, while rebasing the PR on top of the latest `main` branch.

akinomyoga:
Thanks!

PR #1415, opened 2025-08-10
docs(README): external install tip improvements

scop:
Per discussion in https://github.com/scop/bash-completion/pull/1379#discussion_r2264550285 and onwards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,12 +207,12 @@ A. [ Disclaimer: Here, how to make the completion code visible to
    `bash-completion-config-version.cmake` files. Example usage:
 
    ```cmake
-   include(GNUInstallDirs)
    find_package(bash-completion)
    if(BASH_COMPLETION_FOUND)
      message(STATUS
        "Using bash completion dir ${BASH_COMPLETION_COMPLETIONSDIR}")
    else()
+     include(GNUInstallDirs)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -207,12 +207,12 @@ A. [ Disclaimer: Here, how to make the completion code visible to
    `bash-completion-config-version.cmake` files. Example usage:
 
    ```cmake
-   include(GNUInstallDirs)
    find_package(bash-completion)
    if(BASH_COMPLETION_FOUND)
      message(STATUS
        "Using bash completion dir ${BASH_COMPLETION_COMPLETIONSDIR}")
    else()
+     include(GNUInstallDirs)
```
scop, line 215: This is fairly cosmetic I suppose and I don't know if it actually works, but would make sense to me. It's a separate commit we can remove if it doesn't.

akinomyoga: I don't use cmake, so I'm not familiar with it. AI says it works, and AI is probably right, but I'm not sure whether there is anything that we should be careful with. I think the original author @jtojnar should know this better.

jtojnar: I believe conditional imports are fine in CMake should probably work. Though, in practice, it [should be used elsewhere](https://github.com/jtojnar/cmake-snips#hardcoding-the-installation-paths) in the CMake script too.

PR #1416, opened 2025-08-10
chore(renovate): use best-practices preset

scop:
Our previous "config:base" was actually these days called "config:recommended":
https://docs.renovatebot.com/presets-config/#configrecommended

"config:best-practices":
https://docs.renovatebot.com/presets-config/#configbest-practices

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:base",
+    "config:best-practices",
     ":enablePreCommit",
     ":prConcurrentLimit20",
     ":preserveSemverRanges",
@@ -10,7 +10,8 @@
   "ignorePresets": [
     ":dependencyDashboard",
     ":ignoreModulesAndTests",
-    ":semanticPrefixFixDepsChoreOthers"
+    ":semanticPrefixFixDepsChoreOthers",
+    "docker:pinDigests"
   ],
   "semanticCommits": "enabled",
   "commitMessageTopic": "{{depName}}",


.github/renovate.json
```diff
@@ -10,7 +10,8 @@
   "ignorePresets": [
     ":dependencyDashboard",
     ":ignoreModulesAndTests",
-    ":semanticPrefixFixDepsChoreOthers"
+    ":semanticPrefixFixDepsChoreOthers",
+    "docker:pinDigests"
```
scop, line 14: Docker images are used only for tests and my thinking is that we want the latest of each as it follows the tag without separate PR's controlling them, so I don't think we want to pin their digests.

akinomyoga (approved):

akinomyoga:
I'm not sure if this change affected, but renovates are recently very noisy. For every new push to the `main` branch (through PRs), it queues 32 GitHub workflows for CI and consumes large resources. I guess Renovate doesn't assume the usage of keeping the renovate PRs open, but it assumes merging as soon as possible as far as no problems are detected while CI testing.

Can we consider more [aggressive grouping](https://docs.renovatebot.com/noise-reduction/#package-grouping)? We might even consider [`group:all`](https://docs.renovatebot.com/presets-group/#groupall).

yedayak:
I'm actually not sure why it does this, all of the jobs are because it rebases on `main`, but I don't see why it does this: https://docs.renovatebot.com/configuration-options/#rebasewhen
I don't see we set a custom value for this, and since we don't have auto-merge enabled it shouldn't rebase if there isn't a conflict.

akinomyoga:
Thank you for the information. `rebaseWhen:conflicted` seems the desired behavior for us, though I'm not sure either why it currently rebases every time.

yedayak:
Made https://github.com/scop/bash-completion/pull/1449

scop:
Thanks!

Would be also good to look into refactoring the CI stuff some so that various checks are run only when files related to those checks change.

PR #1423, opened 2025-08-31
complete doas commands

ghost:
so that a command  completes when prefixed by doas, like this `doas command`

This is not for completing doas it self.

it seams there's was confusion:
#837 was about to do the same thing, but they thought that it was a duplicate of #766  but that's for doas it self.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -3073,7 +3073,7 @@ _comp_root_command()
     local _comp_root_command=$1
     _comp_command
 }
-complete -F _comp_root_command fakeroot gksu gksudo kdesudo really
+complete -F _comp_root_command doas fakeroot gksu gksudo kdesudo really
 
 # Return true if the completion should be treated as running as root
 #


akinomyoga:
> #766 but that's for doas it self.

#766 is a superset of #837 and this PR. The feature #837 and this PR try to add is supposed to be provided by #766 on [this line](https://github.com/scop/bash-completion/pull/766/files#diff-70a867c1b686c721c196fa7a96722e4a7b1f35549fd89b2e27a697c45f4355afR14).

@exponentialmatrix If you don't think so, could you please explain it?

ghost:
I don't understand the complexities of bash completion, i saw all the sudo like stuff there, i added doas and it started working.
766 is there for 3 years and it seams it will never be merged.

It seams that the line of this PR adds support for sudo like apps without a dedicated profile. In that sense it's correct.

I'm proposing to merge this one and add a correcting commit to 766 for when it will be merged.

akinomyoga:
The main reason that #766 is not merged is not because it is more complicated. It's pending because there is no action or response about the confirmation with the upstream as requested in `CONTRIBUTING.md`. This PR #1423 has the same issue.

PR #1425, opened 2025-09-09
feat(smartctl): update for smartmontools 7.5

chrfranke:
Backward compatible with even 5.* (<= 2012) versions of smartctl, except for the newly added options.

### DIFF
--- a/completions/smartctl
+++ b/completions/smartctl
@@ -1,32 +1,32 @@
 # bash completion for smartctl(8)                          -*- shell-script -*-
+# Updated for smartmontools 7.5 (released 2025-04-30)
 
 _comp_cmd_smartctl__device()
 {
     case $cur in
-        areca* | 3ware* | megaraid* | cciss*)
+        3ware,* | cciss,* | areca,* | megaraid,*)
             _comp_compgen -- -W '${cur%%,*},{0..31}'
             ;;
-        hpt*)
+        aacraid,*)
+            _comp_compgen -- -W 'aacraid,{0..1},{0..1},{0..7}'
+            ;;
+        hpt,*)
             _comp_compgen -- -W 'hpt,{1..4}/{1..8} hpt,{1..4}/{1..8}/{1..5}'
             ;;
+        sssraid,*)
+            _comp_compgen -- -W 'sssraid,{0..1},{0..31}'
+            ;;
+        jmb39x,* | jmb39x-q,* | jmb39x-q2,* | jms56x,*)
+            _comp_compgen -- -W '${cur%%,*},{0..4}'
+            ;;
         *)
-            _comp_compgen -- -W 'ata scsi sat usbcypress usbjmicron usbsunplus
-                marvell areca 3ware hpt megaraid cciss auto test'
-            case "${COMPREPLY[@]}" in
-                areca | 3ware | hpt | megaraid | cciss)
-                    compopt -o nospace
-                    ;;
-            esac
+            _comp_compgen -- -W 'auto test ata scsi nvme{,\,}
+                sat{,\,auto}{,\,12} usb{jmicron,prolific,sunplus}
+                snt{asmedia,jmicron,realtek}{,/sat} 3ware, aacraid, areca,
+                cciss, hpt, megaraid, sssraid, jmb39x{,-q,-q2}, jms56x,'
+            [[ ${COMPREPLY-} == *, ]] && compopt -o nospace
             ;;
     esac
-}
-_comp_cmd_smartctl__test()
-{
-    [[ $cur == @(pending|scttempint|vendor), ]] && return
-    _comp_compgen -- -W 'offline short long conveyance select, select,redo
-        select,next afterselect,on afterselect,off pending, scttempint,
-        vendor,'
-    [[ ${COMPREPLY-} == *, ]] && compopt -o nospace
 }
 _comp_cmd_smartctl__drivedb()
 {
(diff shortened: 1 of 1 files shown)


chrfranke:
Force-pushed two fixes: added missing `=` after `--test`, removed unneeded `compopt -o space` from `--format=` handling.

completions/smartctl
```diff
@@ -101,15 +132,41 @@ _comp_cmd_smartctl()
...
+            _comp_cmd_smartctl__set 'on off'
             return
             ;;
     esac
 
     [[ $was_split ]] && return
 
     if [[ $cur == -* ]]; then
-        _comp_compgen_help
+        # _comp_compgen_help # not used because result is incomplete
```
akinomyoga, line 162: We try to avoid hardcoding as much as possible because a hardcoded list cannot reflect version differences. Couldn't the output of `smartctl --help` be fixed so that a complete list of options can be extracted?

completions/smartctl
```diff
@@ -101,15 +132,41 @@ _comp_cmd_smartctl()
...
+        _comp_compgen -- -W '--abort --all --attributes --badsum=
+            --capabilities --captive --device= --drivedb= --firmwarebug=
+            --format= --get= --health --help --identify{,=} --info --json{,=}
+            --log= --nocheck= --offlineauto= --presets= --quietmode= --report=
+            --saveauto= --scan{,-open} --set= --smart= --test= --tolerance=
+            --vendorattribute= --version --xall -A -B -C -F -H -P -S -T -V -X
+            -a -b -c -d -f -g -h -i -j -l -n -o -q -r -s -t -v -x'
```
akinomyoga, line 169: Please check [`CONTRIBUTING.md`](https://github.com/scop/bash-completion/blob/c55ee7f6fb75300786cb522261f68eb80366c41f/CONTRIBUTING.md?plain=1#L144-L161); remove short options that have corresponding long options.

completions/smartctl
```diff
@@ -38,13 +38,42 @@ _comp_cmd_smartctl__drivedb()
...
+            _comp_compgen -- -W '${cur}, {100..255},'
+            ;;
+        [1-9])
+            _comp_compgen -- -W '${cur}, ${cur}{0..9}'
+            ;;
+        *)
+            _comp_compgen -- -W '{1..9} N, help'
+            ;;
+    esac
+    [[ ${COMPREPLY-} == *, ]] && compopt -o nospace
```
akinomyoga, line 60: - Doesn't this cause insertion of a space for `{1..9}` (`$cur in *`) and `${cur}{0..9}` (`$cur in [1-9]`) with `menu-complete`, etc?

completions/smartctl
```diff
@@ -38,13 +38,42 @@ _comp_cmd_smartctl__drivedb()
...
+        [1-9N],*|[1-9][0-9],*|1[0-9][0-9],*|2[0-4][0-9],*|25[0-5],*)
+            fmt='raw{8,16,48,56,64},hex{48,56,64},raw24/raw{24,32}'
+            fmt+=',msec24hour32,{sec,min,halfmin}2hour,temp10x,tempminmax'
+            _comp_compgen -- -W '${cur%%,*},{'"$fmt"'}{,\,,}'
+            ;;
+        [1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])
+            _comp_compgen -- -W '${cur}, {100..255},'
+            ;;
+        [1-9])
+            _comp_compgen -- -W '${cur}, ${cur}{0..9}'
```
akinomyoga, line 54:
```suggestion
            _comp_compgen -- -W '"$cur", "$cur"{0..9}'
```

completions/smartctl
```diff
@@ -38,13 +38,42 @@ _comp_cmd_smartctl__drivedb()
     _comp_compgen_filedir h && [[ $prefix ]] &&
         _comp_compgen -Rv COMPREPLY -- -P "$prefix" -W '"${COMPREPLY[@]}"'
 }
+_comp_cmd_smartctl__vendorattribute()
```
akinomyoga, line 41: This function seems to try to assist in inputting a number `{0..255}` by suggesting one digit at a time, but it can be confusing. For example, when `$cur` is empty, it only lists `{1..9}` (which is actually merely the first digit of possible numbers `{0..255}`), but it looks like the accepted number range is `{1..9}`.

completions/smartctl
```diff
@@ -1,33 +1,33 @@
...
 _comp_cmd_smartctl__device()
 {
     case $cur in
-        areca* | 3ware* | megaraid* | cciss*)
+        3ware,* | cciss,* | areca,* | megaraid,*)
             _comp_compgen -- -W '${cur%%,*},{0..31}'
             ;;
-        hpt*)
+        aacraid,*)
+            _comp_compgen -- -W 'aacraid,{0..1},{0..1},{0..7}'
```
akinomyoga, line 11: This isn't specific to this PR but also applies to existing completions for `hpt,*`, but I'm wondering if it is possible to let the user input each section (separated by commas) incrementally.

completions/smartctl
chrfranke: Yes - for the long term as this is already planned for next release. But next release 8.0 will take plenty of time due to planned refactoring and enhancements after we finally moved the project from SF svn to GH three month ago.

So please accept the hardcoding for now to support existing releases.

Please note that the current coverage of options is really outdated. For example the `-v, --vendorattribute` completions only cover options which are deprecated since I added the more flexible `-v ID,RAW_FORMAT,ATTR_NAME` syntax in smartmontools 5.39. This version was released in December 2009.

completions/smartctl
chrfranke: Of course. Sorry, I missed that note.

completions/smartctl
chrfranke: Indeed. Let me re-think this part...

completions/smartctl
chrfranke: For all uses of `$cur` within the `_comp_compgen -- -W ...` parameter or only this one?

completions/smartctl
chrfranke: Could you point to an example in another completion file?

completions/smartctl
chrfranke: Alternative: Drop the completion of digits and continue completion after `-v {1..255},` or `-v N,` is present:

```sh
_comp_cmd_smartctl__vendorattribute()
{
    local fmt
    case $cur in
        [1-9N],*|[1-9][0-9],*|1[0-9][0-9],*|2[0-4][0-9],*|25[0-5],*)
            fmt='raw{8,16,48,56,64},hex{48,56,64},raw24/raw{24,32}'
            fmt+=',msec24hour32,{sec,min,halfmin}2hour,temp10x,tempminmax'
            _comp_compgen -- -W '${cur%%,*},{'"$fmt"'}{,\,,}'
            ;;
        *)
            _comp_compgen -- -W 'N, help'
            ;;
    esac
    [[ ${COMPREPLY-} == *, ]] && compopt -o nospace
}
```
This would also fix the _insertion of a space for ... with `menu-complete`_ you mentioned above.

completions/smartctl
chrfranke: ... see new `_comp_cmd_smartctl__vendorattribute()` below.

completions/smartctl
akinomyoga: All.

completions/smartctl
akinomyoga: https://github.com/scop/bash-completion/pull/1425#discussion_r2345039085

By "*incrementally*", I mean that, as you did in `_comp_cmd_smartctl__vendorattribute`, it first generate the completions `"${cur%,*}",{0..1},` when `cur` starts with `aacraid,` but without any additional comma. When there is already `aacraid,[01],` in the command line, it generates the next completions `"${cur%,*}",{0..1},`. When there is `aacraid,[01],[01],` in the command line, it generates the next completions `"${cur%,*}",{0..7},`.

completions/smartctl
akinomyoga: I'm talking about the above line of removing `_comp_compgen_help`. I'm not talking about the list of possible values for option arguments of each option (though it would be better if the `smartctl` binary could provide a list of possible values in 8.0). I've checked the output of `_comp_compgen_help`, and the only missing options are `--drivedb`, `--nocheck`, and `--set`, where the following lines in the output `--help` seem to fail to be parsed.

```
  -n MODE[,STATUS[,STATUS2]], --nocheck=MODE[,STATUS[,STATUS2]] (ATA, SCSI)
  -s NAME[,VALUE], --set=NAME[,VALUE]
  -B [+]FILE, --drivedb=[+]FILE                                       (ATA)
```

This may be fixed at the side of `_comp_compgen_help` by allowing a wider format of the option argument name [at line 1697 of `bash_completion`](https://github.com/scop/bash-completion/blob/c55ee7f6fb75300786cb522261f68eb80366c41f/bash_completion#L1697).

completions/smartctl
akinomyoga: Or 

```diff
             ;;
+        [1-9])
+            _comp_compgen -- -W '{1..255},'
+            ;;
         *)
             _comp_compgen -- -W 'N, help'
             ;;
```

though `1`<kbd>TAB</kbd> will surpass the default `bind 'set completion-query-items 100'`.

completions/smartctl
chrfranke: See #1430.

chrfranke:
Thanks for reviewing and various hints. #1430 should be addressed first. I will provide a reworked version of `completions/smartctl` afterwards. Therefore I close this PR for now.

PR #1427, opened 2025-09-10
fix(_comp_compgen): go to the original directory on compgen failure

### LINKED ISSUE #1426: Completion can change working directory
## Describe the bug

Current working directory is not restored when using `_comp_compgen -C` in some cases.

## To reproduce

Using `removepkg` as an example because it uses `_comp_compgen -C`:

1. Run `mkdir -p tmp/var/log/packages && touch tmp/var/log/packages/{foo,bar,baz} && cd tmp`.
2. Run `ROOT=$PWD` (that's setup for `removepkg`).
3. Then try completing with a prefix that is not found, for example: `removepkg a`<kbd>\<TAB\></kbd>.
4. After that, you can type <kbd>Ctrl-U</kbd> or <kbd>Ctrl-C</kbd> to clear the line.
5. Run `pwd` and see that your current working directory has changed to `./var/log/packages` (`$PWD` hasn't changed).
6. To clean-up: `cd "$ROOT"/.. && unset ROOT && rm -rf tmp`

## Expected behavior

After step 3, the shell's working directory should be the same as before.

- [x] Operating system name/distribution and version: macOS 15.6.1
- [x] bash version, `echo "$BASH_VERSION"`: 5.3.3(1)-release
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2 16 0

## Debug trace

[debug-trace.log](https://github.com/user-attachments/files/22252694/debug-trace.log)

akinomyoga:
Commit b6de4915213377fb406be6fe0cfdd3b944f667f9 adds a check for the current working directory. Checking `PWD` isn't sufficient because `PWD` can be different from the actual current working directory.

Then, the second commit fixes the issue reported at #1426. `command cd -- "$_original_pwd"` was missing for the early return with the completion error by `builtin compgen`.

Fixes #1426.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -809,6 +809,8 @@ if ((BASH_VERSINFO[0] > 5 || BASH_VERSINFO[0] == 5 && BASH_VERSINFO[1] >= 3)); t
         # pat, the effect of -X '' is overwritten by the specified one.
         IFS=$_ifs compgen -V _result -X '' "$@" ${_cur:+-- "$_cur"} || {
             _comp_compgen__error_fallback
+            # shellcheck disable=SC2164
+            [[ $_dir ]] && command cd -- "$_original_pwd"
             return
         }
 
--- a/test/config/bashrc
+++ b/test/config/bashrc
@@ -64,6 +64,12 @@ add_comp_wordbreak_char()
 
 _comp__test_get_env()
 {
+    # Note: PWD is supposed to reflect the current working directory, but the
+    # real current working directory of the process can be different, e.g. when
+    # PWD is manually altered.  Therefore, we also check the result of $(pwd)
+    # explicitly via a new variable `_comp_test_pwd`.
+    local _comp_test_pwd=$(pwd)
+
     (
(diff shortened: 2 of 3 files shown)


scop (approved):
Is there a valid scenario where `$PWD` would be changed? I can see that it _could_ be, but then again so could one modify the `pwd` command to do something that we don't expect.

I'm generally not quite comfortable in trying to cater for _everything_ possible one might want do to break things, so I'd rather see us only addressing scenarios we know are "valid" and exist in the wild, but I don't feel strongly enough about this particular case to request changes.

Anyway the substance of the PR LGTM, thanks!

akinomyoga:
> Is there a valid scenario where `$PWD` would be changed? I can see that it _could_ be, but then again so could one modify the `pwd` command to do something that we don't expect.

#1426 has been a case where `local PWD=${PWD-}` was used in `_comp_compgen__call_builtin`, but I forgot to return to the original directory on a error. This could have been detected earlier if the actual current working directory `$(pwd)` had been tested. Is it an *invalid* case?

PR #1430, opened 2025-09-14
feat(_comp_compgen_help): enhance for smartctl

chrfranke:
A followup to #1425 with two commits addressing independent topics.

Should now detect these properly:
```
  -n MODE[,STATUS[,STATUS2]], --nocheck=MODE[,STATUS[,STATUS2]] (ATA, SCSI)
  -s NAME[,VALUE], --set=NAME[,VALUE]
  -B [+]FILE, --drivedb=[+]FILE                                       (ATA)
```

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -1694,7 +1694,7 @@ _comp_compgen_help()
     for _line in "${_lines[@]}"; do
         [[ $_line == *([[:blank:]])-* ]] || continue
         # transform "-f FOO, --foo=FOO" to "-f , --foo=FOO" etc
-        while [[ $_line =~ ((^|[^-])-[A-Za-z0-9?][[:space:]]+)\[?[A-Z0-9]+([,_-]+[A-Z0-9]+)?(\.\.+)?\]? ]]; do
+        while [[ $_line =~ ((^|[^-])-[A-Za-z0-9?][[:space:]]+)(\[,?)?[A-Z0-9+]+([,_-]+[A-Z0-9]+)?(\.\.+)?\]* ]]; do
             _line=${_line/"${BASH_REMATCH[0]}"/"${BASH_REMATCH[1]}"}
         done
         _comp_compgen_help__parse "${_line// or /, }"
--- a/test/t/unit/test_unit_parse_help.py
+++ b/test/t/unit/test_unit_parse_help.py
@@ -203,6 +203,20 @@ class TestUnitParseHelp:
         output = assert_bash_exec(bash, "_parse_help fn", want_output=True)
         assert output.split() == "--exclude=".split()
 
+    def test_33(self, bash):
+        """Pattern found in the output of 'smartctl --help'."""
+        assert_bash_exec(
+            bash, "fn() { echo '-f A1[,A2[,A3]] --foo=A1[,A2[,A3]]'; }"
+        )
+        output = assert_bash_exec(bash, "_parse_help fn", want_output=True)
+        assert output.split() == "--foo=".split()
+
+    def test_34(self, bash):
+        """Pattern found in the output of 'smartctl --help'."""
+        assert_bash_exec(bash, "fn() { echo '-f [+]ARG --foo=[+]ARG'; }")
+        output = assert_bash_exec(bash, "_parse_help fn", want_output=True)
+        assert output.split() == "--foo=".split()
+
     def test_custom_helpopt1(self, bash):
         assert_bash_exec(bash, "fn() { [[ $1 == -h ]] && echo '-option'; }")
         output = assert_bash_exec(bash, "_parse_help fn -h", want_output=True)


bash_completion
```diff
@@ -1694,7 +1694,7 @@ _comp_compgen_help()
     for _line in "${_lines[@]}"; do
         [[ $_line == *([[:blank:]])-* ]] || continue
         # transform "-f FOO, --foo=FOO" to "-f , --foo=FOO" etc
-        while [[ $_line =~ ((^|[^-])-[A-Za-z0-9?][[:space:]]+)\[?[A-Z0-9]+([,_-]+[A-Z0-9]+)?(\.\.+)?\]? ]]; do
+        while [[ $_line =~ ((^|[^-])-[A-Za-z0-9?][[:space:]]+)(\[,?)?(\+\]?)?[A-Z0-9]+([,_-]+[A-Z0-9]+)?(\.\.+)?\]* ]]; do
```
akinomyoga: ~~Does this correctly handle `MODE[,STATUS[,STATUS2]]`? In the updated regular expression, `\[` only appears at the beginning and is not contained in any repetition (like `*`, `+`, and `{n,m}`), so `\[` is allowed only at the beginning.~~ **edit**: Never mind, it is inside a `while` loop.

Also, the special rule for `[+]`  seems specific to the case of `smartctl`. Or is this pattern `[+]` common for the output of `--help`? What does it actually mean? Is it a literal `+`, which can be specified optionally?

chrfranke: I guess `[+]` is specific to `smartctl`. It is a literal `+`. If specified, the new drive database entries are added (prepended) to the default ones instead of replacing these.

An alternative would be to drop [c422102](https://github.com/scop/bash-completion/pull/1430/commits/c422102eec01e02b3c69c83c83cba4a44df09a85) and fix the result of `_comp_compgen_help` after the call in `completions/smartctl`: remove `-B` and add `--drivedb=` (if there is an official way to do this).

chrfranke: A flexible approach to handle such special cases without the need to enhance the regex would be an ability to pass some `FILTER` function to `_comp_compgen_help`:
```bash
comp_cmd_smartctl__help_filter()
{
    eval $1=\${$1//\\[+\\]/}
}

_comp_cmd_smartctl()
{
    # ...
        _comp_compgen_help -f _comp_cmd_smartctl__help_filter
    # ...
}
```
```bash
_comp_compgen_help()
{
    # ...
    for _line in "${_lines[@]}"; do
        [[ -n "$filter" ]] && "$filter" _line
        # ...
    done
    # ...
}
```

akinomyoga: > An alternative would be to drop [c422102](https://github.com/scop/bash-completion/pull/1430/commits/c422102eec01e02b3c69c83c83cba4a44df09a85) and fix the result of `_comp_compgen_help` after the call in `completions/smartctl`: remove `-B` and add `--drivedb=` (if there is an official way to do this).

That's still a kind of hardcoding.

> A flexible approach to handle such special cases without the need to enhance the regex would be an ability to pass some `FILTER` function to `_comp_compgen_help`:

I don't think it's worth such a complication. What I had in mind is to allow `+` as one of the possible characters in the option-argument description (along with other `A-Z0-9`), i.e., `[A-Z0-9+]+`. The exact prefix `[+]` seems specific to `smartctl`, but in a more general point of view, it implies that `+` can be included in the option-argument description.

chrfranke:
> I don't think it's worth such a complication. What I had in mind is to allow + as one of the possible characters in the option-argument description...

Force-pushed new version of 2nd commit.

test/t/unit/test_unit_parse_help.py
```diff
@@ -203,6 +203,16 @@ def test_32(self, bash):
         output = assert_bash_exec(bash, "_parse_help fn", want_output=True)
         assert output.split() == "--exclude=".split()
 
+    def test_33(self, bash):
```
akinomyoga, line 206: Could you include the description that the pattern `A1[,A2[,A3]]` is found in the output of `smartctl --help`?

```suggestion
    def test_33(self, bash):
        """ <description> """
```

`test_34` for `[+]ARG` is similar.

bash_completion
akinomyoga: > Force-pushed new version of 2nd commit.

Thanks.

test/t/unit/test_unit_parse_help.py
chrfranke: Done

akinomyoga: Thank you.

chrfranke:
Sorry for CI failures. Will address these in the next days.

chrfranke:
Force-pushed new version, now combined in one commit. This reformats (as suggested by pre-commit) and fixes the testcases.
```diff
--- a/test/t/unit/test_unit_parse_help.py
+++ b/test/t/unit/test_unit_parse_help.py
@@ -205,15 +205,19 @@ class TestUnitParseHelp:

     def test_33(self, bash):
         """Pattern found in the output of 'smartctl --help'."""
-        assert_bash_exec(bash, "fn() { echo '-f A1[,A2[,A3]] --foo=A1[,A2[,A3]]'; }")
+        assert_bash_exec(
+            bash, "fn() { echo '-f A1[,A2[,A3]] --foo=A1[,A2[,A3]]'; }"
+        )
         output = assert_bash_exec(bash, "_parse_help fn", want_output=True)
-        assert output.split() == "--foo".split()
+        assert output.split() == "--foo=".split()

     def test_34(self, bash):
         """Pattern found in the output of 'smartctl --help'."""
-        assert_bash_exec(bash, "fn() { echo '-f [+]ARG --foo=[+]ARG'; }")
+        assert_bash_exec(
+            bash, "fn() { echo '-f [+]ARG --foo=[+]ARG'; }"
+        )
         output = assert_bash_exec(bash, "_parse_help fn", want_output=True)
-        assert output.split() == "--foo".split()
+        assert output.split() == "--foo=".split()

     def test_custom_helpopt1(self, bash):
         assert_bash_exec(bash, "fn() { [[ $1 == -h ]] && echo '-option'; }")

```

chrfranke:
Obviously my assumption about max allowed title length (4 indentation + 74 title = 78) was wrong.  Force-pushed a new version using the original title of this PR. The code is unchanged.

chrfranke:
Undoing reformatting of `test_34`. Last try for this week to make pre-commit happy :-)

akinomyoga:
The CI error in fedoradev seems to have started to happen after the latest [`update-docker-images`](https://github.com/scop/bash-completion/actions/runs/17840804329) and is unrelated to this PR.

PR #1432, opened 2025-09-18
Analyze github workflows with zizmor

yedayak:


scop (approved):
Interesting, TIL & LGTM.

I've had the intention to work on this area and other security related things, looks like this will blend in nicely. One related thing I just submitted is https://github.com/scop/bash-completion/pull/1433

yedayak:
One thing I will say about this is I would love to somehow test that releases will work correctly with this PR, is there a way to simulate the release please process somewhere?

scop:
The first thing springing to mind is I guess you could tweak your own fork to do a release there. But as far as this PR is concerned, while that would be nice, I don't think that's required -- I hope I'll get to adding some other goodies which will require some eyeballing at release time anyway so I'll be aware.

PR #1441, opened 2025-09-28
test(glob): Normalize some strings to compare

yedayak:
Some of the files have unicode characters that can be represented in multiple ways, so we normalize them. Specifically: 'e\u0301' ('e' + 'Combining Acute Accent') doesn't equal '\xe9' ('LATIN SMALL LETTER E WITH ACUTE') unless you normalize them.

Cherry-picked from #1339

### DIFF
--- a/test/t/unit/test_unit_expand_glob.py
+++ b/test/t/unit/test_unit_expand_glob.py
@@ -1,6 +1,16 @@
+import unicodedata
+
 import pytest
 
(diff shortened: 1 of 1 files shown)


akinomyoga (approved):
Ah, yeah. I recently read an article about the problems caused by the UTF-8-MAC/NFD conversions performed by macOS filesystems.

PR #1442, opened 2025-09-28
fix(patch, truncate): Get help correctly on macos

yedayak:
The usage output on macos is in bsd style, not gnu style
Cherry picked from #1339


completions/patch
```diff
@@ -50,7 +50,11 @@ _comp_cmd_patch()
     [[ $was_split ]] && return
 
     if [[ $cur == -* ]]; then
-        _comp_compgen_help
+        if [[ $OSTYPE == *@(darwin)* ]]; then
```
akinomyoga: How about *BSD systems? If this is only in macOS, the enclosing `@()` should be removed.

completions/patch
```diff
@@ -50,7 +50,11 @@ _comp_cmd_patch()
     [[ $was_split ]] && return
 
     if [[ $cur == -* ]]; then
-        _comp_compgen_help
+        if [[ $OSTYPE == *@(darwin|bsd)* ]]; then
+            _comp_compgen_usage
+        else
+            _comp_compgen_help
+        fi
```
scop: To me, the preferred pattern for cases like these is
```suggestion
        _comp_compgen_help || _comp_compgen_usage
```

This frees us from hardwiring "known" OS dependencies, as well as helps with cases for which detection isn't that feasible, such as Alpine systems where a command might be a GNU one supporting long options (_comp_compgen_help applies), or a busybox one that supports only short ones (_comp_compgen_usage applies).

completions/truncate
```diff
@@ -20,7 +20,11 @@ _comp_cmd_truncate()
     [[ $was_split ]] && return
 
     if [[ $cur == -* ]]; then
-        _comp_compgen_help
+        if [[ $OSTYPE == *@(darwin|bsd)* ]]; then
```
scop: As above.

scop (changes requested):
(Pre-approved with requested changes.)

completions/patch
yedayak: Yeah that makes sense, also there are some linux distros that come with BSD tools. Will update

PR #1444, opened 2025-09-30
doc improvements

### LINKED ISSUE #1443: how to avoid double-sourcing?
Please document this in the `README`.

I've been wondering why my Bash startup time is so slow on Termux compared to Debian, while using identical dotfiles.

> [!note]
> For reference: Debian doesn't source any completions at all, while Termux sources **all** (most?) of the completion files installed by the package.

Turns out I was double-sourcing 2 files. Here's an excerpt from my `~/.bashrc`:
```bash
bc="${TERMUX__PREFIX:-/usr}/share/bash-completion"

[[ -f $bc/bash_completion ]] && \
	\. "$bc/bash_completion"

if [[ -f $bc/completions/git ]]; then
	\. "$bc/completions/git"
	__git_complete g __git_main
fi

unset bc
```

I commented-out  that, and the startup-time went from 0.53s to 0.29s (both `real`), according to `time bash -i -c exit`.

So instead of checking if the rc is running on Termux or Debian, I decided to use a cross-platform approach:
```bash
bc="${TERMUX__PREFIX:-/usr}/share/bash-completion"

[[ -z ${BASH_COMPLETION_VERSINFO:-} && -f $bc/bash_completion ]] && \
	\. "$bc/bash_completion"

[[ $(type -t __git_complete) != function ]] && \
[[ -f $bc/completions/git ]] && \
	\. "$bc/completions/git"
__git_complete g __git_main

unset bc
```

I copied the `BASH_COMPLETION_VERSINFO` check from `/etc/profile.d/bash_completion.sh`, and since the var is not prefixed by `_` I assume this is API-stable and not an implementation detail. However, I'm afraid `type` is not the proper way to check if Git completions are loaded

Rudxain:
- change markdown code-block lang-tags from `shell` to `bash` (only where appropriate)
- check if completions are already sourced in example `.bashrc` (in `README`)

fixes #1443

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,7 +99,7 @@ Also, please bear the following coding guidelines in mind:
 
   can be replaced by:
 
-  ```shell
+  ```bash
   bar=${foo//bar/baz}
   ```
 
@@ -132,7 +132,7 @@ Also, please bear the following coding guidelines in mind:
   expansions will be unexpectedly performed, which becomes a vulnerability.  In
   the latter case, checks by shellcheck and shfmt will not be performed inside
   `'...'`.  Also, `_comp_compgen_split` is `IFS`-safe.
-  
+
   Avoid using `_comp_compgen -- -G "pattern"` to generate completions.  The
   result is not filtered by the current word `cur` due to the Bash design of
   `compgen`.  Also, this cannot be used to generate filenames with a specified
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ need to source it from either `/etc/bashrc` or `~/.bashrc` (or any
 other file sourcing those). If you have _only_ bash >= 4.2 installed, you can
 do this by simply using:
 
-```shell
+```bash
 # Use bash-completion, if available
 [[ $PS1 && -f /usr/share/bash-completion/bash_completion ]] && \
     . /usr/share/bash-completion/bash_completion
@@ -70,7 +70,7 @@ standard way is to configure `~/.bash_profile` to source `~/.bashrc` and write
 interactive settings in `~/.bashrc`.  You can source `~/.bashrc` in
 `~/.bash_profile` in the following way:
 
-```shell
+```bash
 # ~/.bash_profile
 
 if [[ -f ~/.bashrc ]]; then
@@ -90,7 +90,7 @@ the entry point of `bash-completion` to
 `$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh`.  We can source it by
 adding the following to our startup file `~/.bashrc`:
 
-```shell
+```bash
 if [[ -s $HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh ]]; then
   . "$HOMEBREW_PREFIX/etc/profile.d/bash_completion.sh"
 fi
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -21,11 +21,19 @@ need to source it from either `/etc/bashrc` or `~/.bashrc` (or any
 other file sourcing those). If you have _only_ bash >= 4.2 installed, you can
 do this by simply using:
 
-```shell
+```bash
 # Use bash-completion, if available
 [[ $PS1 && -f /usr/share/bash-completion/bash_completion ]] && \
     . /usr/share/bash-completion/bash_completion
 ```
+or if your `bashrc` will be shared across distros (and major upgrades):
```
akinomyoga: I think we can directly update the existing code (instead of adding an alternative example).

README.md
```diff
@@ -21,11 +21,19 @@ need to source it from either `/etc/bashrc` or `~/.bashrc` (or any
...
+```bash
+# Use bash-completion, if available, and avoid double-sourcing
+[[ $PS1 && \
+  -z ${BASH_COMPLETION_VERSINFO:-} && \
+  -f /usr/share/bash-completion/bash_completion ]] && \
```
akinomyoga: Actually, '\' are unnecessary here.
```suggestion
[[ $PS1 &&
  -z ${BASH_COMPLETION_VERSINFO:-} &&
  -f /usr/share/bash-completion/bash_completion ]] &&
```

akinomyoga (commented):
LGTM, basically.

README.md
Rudxain: I was unsure at first, but I agree. Users can simply remove the extra check if they know in advance that it's never sourced

CONTRIBUTING.md
```diff
@@ -99,13 +99,16 @@ Also, please bear the following coding guidelines in mind:
...
-  ```shell
+  ```bash
   bar=${foo//bar/baz}
   ```
 
   These forms of parameter substitutions can also be used on arrays,
   which makes them very powerful (if a little slow).
 
+  See ["Pure Bash"](https://github.com/dylanaraps/pure-bash-bible)
+  for more examples.
```
akinomyoga: I'm not sure if we should mention Pure Bash Bible. It contains many useful tricks, but it also contains inefficient/non-portable/broken tricks, and it is hard for typical users to distinguish them.

Rudxain: ok, I'll move the commit to another branch for a separate PR. I'll `drop` (force-push implied) the commit here instead of `revert`, to keep the log clean

akinomyoga: > to keep the log clean

If you want this PR to be merged without squashing, could you squash the commits from the second to the fourth? The fourth commit is a fixup commit for the change in the second commit.

Rudxain: I was considering to squash them, but wanted to ask first. Thanks for the confirmation!

yedayak (approved):

README.md
```diff
@@ -19,11 +19,13 @@ list of operating system distributions, package names, and available versions.
...
-do this by simply using:
+do this by using:
 
-```shell
-# Use bash-completion, if available
-[[ $PS1 && -f /usr/share/bash-completion/bash_completion ]] && \
+```bash
+# Use bash-completion, if available, and avoid double-sourcing
+[[ $PS1 &&
+  -z ${BASH_COMPLETION_VERSINFO:-} &&
```
akinomyoga: This doesn't follow our coding style. The coding style is meant for the actual codebase for `bash-completion`, but I think there's no reason to use other styles in documents.

```suggestion
  ! ${BASH_COMPLETION_VERSINFO:-} &&
```

akinomyoga (commented):
Other than the below, LGTM. Please squash the style change into the last commit.

Issue #1435, opened 2025-09-21
fedoradev CI fail in test_unit_filedir


yedayak:
Since the update-docker-images job updated the fedoradev test container image, `test_26` in `test/t/unit/test_unit_compgen_filedir.py` started failing there, with these errors:
```
=========================== short test summary info ============================
FAILED ../../../../test/t/unit/test_unit_compgen_filedir.py::TestUnitCompgenFiledir::test_26[f] - assert <CompletionResult ['\x08\x08\x08\x08$b/h"']> == 'h"'
FAILED ../../../../test/t/unit/test_unit_compgen_filedir.py::TestUnitCompgenFiledir::test_26[f2] - assert <CompletionResult ['\x08\x08\x08\x08$b/h"']> == 'h"'
= 2 failed, 2028 passed, 113 skipped, 23 xfailed, 6 xpassed in 1096.06s (0:18:16) =
```
Looking at it, this seems to have been an issue in the past since the test, they were disabled for Centos 6 in a1a4eb82050084a8447c4d2e9e248da9131d55b0, and seem to have been fixed in 726e4c08a96b1f0be1c931179f1b3edb31913917:
```
-        # endswith: in CentOS 6 container we get something like this, accept it:
-        #     a\$b/\x08\x08\x08\x08\x08/work/test/fixtures/_filedir/a\$b/h
-        assert completion == r"a$b/h" or completion.output.strip().endswith(
-            "\b\b\b\b%s/_filedir/a\\$b/h" % bash.cwd
-        )
+        assert completion == "h"
```
Basically the issue as I understand it is that bash "de-quotes" the arguments be removing slashes and double quotes.
Running in the fedoradev test container:
```
$ f "a\$<TAB>
$ f "a$b"/<TAB>
$ f a\$b/h 
```
Presumably bash sends backspaces in order to de-quote the arguments.

This test also failed in #1339 which installs bash from homebrew, `bash--5.3.3.arm64_sequoia.bottle.1.tar.gz`
I can reproduce the behavior locally (but not the failing test) with bash 5.3.3, but not 5.2.37, so presumably something in a bash 5,3 changed.

akinomyoga:
It seems this has been reported to Bash last month [1], but this hasn't yet been fixed in the `devel` branch of Bash.

- [1] https://lists.gnu.org/archive/html/bug-bash/2025-08/msg00107.html

akinomyoga:
[...]

PR #1446, opened 2025-10-03
test(filedir): Workaround bug in bash 5.3

yedayak:
To fix CI failures, see https://github.com/scop/bash-completion/issues/1435 and https://lists.gnu.org/archive/html/bug-bash/2025-09/msg00332.html

I made it catch the assertion so we won't have to run the version check for all tests.

### DIFF
--- a/test/t/unit/test_unit_compgen_filedir.py
+++ b/test/t/unit/test_unit_compgen_filedir.py
@@ -228,7 +228,21 @@ class TestUnitCompgenFiledir:
         completion = assert_complete(
             bash, r'%s "a\$b/' % funcname, cwd="_filedir"
         )
-        assert completion == 'h"'
+        try:
+            assert completion == 'h"'
+        except AssertionError:
+            bash_version = assert_bash_exec(
+                bash,
+                r'printf "%s.%s\n" "${BASH_VERSINFO[0]}" "${BASH_VERSINFO[1]}"',
+                want_output=True,
+            ).strip()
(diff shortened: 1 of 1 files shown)


test/t/unit/test_unit_compgen_filedir.py
```diff
@@ -228,7 +228,21 @@ def test_26(self, bash, functions, funcname):
...
+        try:
+            assert completion == 'h"'
+        except AssertionError:
+            bash_version = assert_bash_exec(
+                bash,
+                r'printf "%s.%s\n" "${BASH_VERSINFO[0]}" "${BASH_VERSINFO[1]}"',
+                want_output=True,
+            ).strip()
+            # This is workaround for https://github.com/scop/bash-completion/issues/1435, see
+            # https://lists.gnu.org/archive/html/bug-bash/2025-08/msg00107.html
```
akinomyoga: After that, I posted a follow-up email to bug-bash, but Chet's reply [1] was that the issue reported in the above link (`2025-08/msg00107.html`) is similar, but a different issue (though I'm not fully convinced). I think it would be better to link [1] rather than the above one.

[1] https://lists.gnu.org/archive/html/bug-bash/2025-09/msg00332.html

akinomyoga (commented):
Thanks!

test/t/unit/test_unit_compgen_filedir.py
yedayak: Thanks, Updated

PR #1447, opened 2025-10-03
build: fix install-data-hook for bash_completion.d

yedayak:
It no longer matches bash_completion so it was not substituting for the actual compatdir due to a refactor.

Follow-up for cdd8a157abdbb4ce2e74df3fd8131a2b199152e6

Originally submitted as part of #1399

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -3490,6 +3490,7 @@ _comp__init_collect_startup_configs()
     if [[ ${BASH_COMPLETION_COMPAT_DIR-} ]]; then
         compat_dirs+=("$BASH_COMPLETION_COMPAT_DIR")
     else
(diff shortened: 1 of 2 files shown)


akinomyoga (approved):

bash_completion
```diff
@@ -3490,6 +3490,7 @@ _comp__init_collect_startup_configs()
     if [[ ${BASH_COMPLETION_COMPAT_DIR-} ]]; then
         compat_dirs+=("$BASH_COMPLETION_COMPAT_DIR")
     else
+        # Keep in sync with install-data-hook at Makefile.am
```
scop:
```suggestion
        # Keep in sync with Makefile.am install-data-hook 
```

PR #1448, opened 2025-10-03
test: Don't sleep before sending data to bash

yedayak:
pexpect has a default delay before sending data, to let applications turn off TTY echo for example, see
https://pexpect.readthedocs.io/en/stable/commonissues.html?highlight=delaybeforesend#timing-issue-with-send-and-sendline

We don't need this for basically all of our tests, and setting delaybeforesend to 0 makes running the unit tests around 90% faster, from 3m18s to 16s on my machine.


akinomyoga:
Doesn't this increase the frequency of the timeout issue discussed in #706?

(Note: I'm purely asking. I'm totally unsure about whether this could affect #706.)

yedayak:
> Doesn't this increase the frequency of the timeout issue discussed in #706?

Maybe? I only ever encountered issues with timeouts when running in parallel.
I'm not sure if this was mentioned somewhere, but when running in parallel I get hundreds of these error:
```
  /usr/lib/python3.12/pty.py:95: DeprecationWarning: This process (pid=28895) is multi-threaded, use of forkpty() may lead to deadlocks in the child.
    pid, fd = os.forkpty()
```
From reading [this thread](https://discuss.python.org/t/concerns-regarding-deprecation-of-fork-with-alive-threads/33555) it seems that there is an inherent issue with calling `fork` (forkpty here) inside something that was forked (presumably by pytest-xdist), before it calls `exec` (which I guess pytest-xdist doesn't do?). I think this is at least partly the problem with running the tests in parallel.

This PR does make errors happen more when running in parallel must more frequent, but I guess its just because everything happens much faster. Running it not in parallel I didn't manage to get any errors.

akinomyoga:
That multi-thread issue seems independent of #706. Issue #706 is only observed in CI. It usually does not happen, but it happened frequently depending on the day or the week. I'm not sure what the factor would be, but I haven't recently seen this issue, so the situation might have improved.

Since it doesn't usually happen, it's hard to test before actually applying the change. I think we can merge this and see if it would cause the problem. If it causes the problem, we can revert this.

akinomyoga (approved):

scop:
Oh, this would be awsome, thanks! :four_leaf_clover:

PR #1451, opened 2025-10-04
Add completion for keychain

mikkoi:
Copied from https://github.com/mikkoi/keychain-bash-completion

Keychain is part of https://www.funtoo.org/Funtoo:Keychain


akinomyoga:
The same applies to this PR as https://github.com/scop/bash-completion/pull/1452#issuecomment-3368606970.

akinomyoga:
The upstream project seems to be https://github.com/danielrobbins/keychain, where I don't see any codes or discussions about completions.

mikkoi:
https://github.com/danielrobbins/keychain/issues/186

mikkoi:
The upstream project has not commented in the issue https://github.com/danielrobbins/keychain/issues/186

Can we still go ahead with this PR or should we abandon it?

scop:
I suggest we wait some more before going either way. Could try pinging upstream just in case the issue went unnoticed.

mikkoi:
The upstream project has implemented Bash Completion: https://github.com/danielrobbins/keychain/issues/186

I have deprecated my own previous implementation and archived the repository [mikkoi/keychain-bash-completion](https://github.com/mikkoi/keychain-bash-completion).

PR #1453, opened 2025-10-07
Inconflict with uutils/coreutils

### LINKED ISSUE #1385: Do no conflict with uutils project
Provide an easy method to split files for `coreutils`.

Default `make` options provides `/usr/share/bash-completion/completions/chmod` which conflicts with [compat package](https://github.com/uutils/coreutils/issues/8039#issuecomment-2924639130).

(Ofcause best solution may moving them to GNU coreutils project...)

- [x] This is a request about packaging layout

oech3:
Add _ prefix to completions for coreutils.
Closes #1385

I think this is minimal effort for distributions which is needed to install every files under `/usr`.


akinomyoga:
Could you check #1386 and #1329? This is a general problem and should be resolved in a consistent way for all completions.

oech3:
I saw them for few month ago. But this approach is reasonable and there is nothing to block fixing partially.

akinomyoga:
As pointed out by https://github.com/scop/bash-completion/discussions/1329#discussioncomment-13488992, `_*` are used for both deprecated and default completions, so we need to check all existing `_*` to judge whether each is deprecated or default. I don't want to increase the cost of judging by increasing the number of `_*`.

Also, it takes time for bash-completion packages to be replaced with the versions without conflicts with uutils/coreutils, so I'm not sure if this PR actually solves the current problem soon. If it doesn't solve the problem quickly, it doesn't seem to be an urgent change.

akinomyoga:
Actually, I want inputs from others for #1329 before I create PRs. Somehow, there are no responses about the general idea, and I'm totally unsure if it is welcomed or not.

oech3:
Closing since this could be replaced by better solution

oech3:
@akinomyoga @scop What do you think about moving completion files to GNU coreutils project?

akinomyoga:
If the upstream GNU coreutils are willing to accept and maintain completion settings, we can provide our completion files to GNU coreutils. However, we will still have to keep our existing completion files (for commands not specific to GNU coreutils) as fallbacks because the commands may be provided by other implementations.

oech3:
So the issue:
>I don't want to increase the cost of judging by increasing the number of _*.

still active even those flles are moved to GNU coreutils?

akinomyoga:
If they were *moved* (meaning that we remove our files), the issue would disappear, but I said we cannot remove our files in the above reply.

oech3:
Yes. They cannot be removed. But we can append `_` at least for.

akinomyoga:
> But we can append `_` at least for.

That's this PR, and I already replied to it: https://github.com/scop/bash-completion/pull/1453#issuecomment-3376376493. That could be one solution, and it's been the solution so far. However, we want to reconsider the approach as I suggested in #1329 and @scop mentioned at https://github.com/scop/bash-completion/issues/1385#issuecomment-2945529883.

PR #1456, opened 2025-10-08
fix(rsync): Try getting rsync arguments BSD style

yedayak:
There is a BSD licensed version of rsync at https://github.com/kristapsdz/openrsync, and its what comes with macOS

Cherry picked from https://github.com/scop/bash-completion/pull/1339


completions/rsync
```diff
@@ -85,11 +85,15 @@ _comp_cmd_rsync()
...
             # "(-h) --help                  show this help (-h is --help only if used alone)"
-            _comp_compgen -Rv tmp help - <<<"$("$1" --help 2>&1 | command sed -e 's/^([^)]*)//')"
+            if _comp_compgen -Rv tmp help - <<<"$("$1" --help 2>&1 | command sed -e 's/^([^)]*)//')"; then
 
-            _comp_compgen -- -W '"${tmp[@]}"
-                --daemon --old-d{,irs}
-                --no-{blocking-io,detach,whole-file,inc-recursive,i-r}' -X '--no-OPTION'
+                _comp_compgen -- -W '"${tmp[@]}"
+                        --daemon --old-d{,irs}
+                        --no-{blocking-io,detach,whole-file,inc-recursive,i-r}' -X '--no-OPTION'
```
akinomyoga:
```suggestion
                    --daemon --old-d{,irs}
                    --no-{blocking-io,detach,whole-file,inc-recursive,i-r}' -X '--no-OPTION'
```

test/t/test_rsync.py
```diff
@@ -19,7 +19,8 @@ def test_3(self, completion):
 
     @pytest.mark.complete("rsync --", require_cmd=True)
```
scop: I wonder if `require_longopt` would be in order here, instead of `require_cmd`?

yedayak: Yeah it makes sense, I don't think I knew we had that option

PR #1458, opened 2025-10-09
docs(CONTRIB): link to Pure-Bash-Bible

Rudxain:
https://github.com/scop/bash-completion/pull/1444#discussion_r2404641951

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,6 +106,9 @@ Also, please bear the following coding guidelines in mind:
   These forms of parameter substitutions can also be used on arrays,
   which makes them very powerful (if a little slow).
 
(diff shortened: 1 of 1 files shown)


akinomyoga:
As I've written in https://github.com/scop/bash-completion/pull/1444#discussion_r2404096619, I don't think Pure Bash Bible is a good resource, so I doubt it's worth including in our documents. However, others may have a different opinion.

scop:
Thanks for the suggestion. Given that the project repo was archived in Apr 2024, it seems it's unlikely it'll receive updates for example for new versions of bash. Therefore it's also subject to becoming outdated with regards to bash versions we may use/require at some point. While this is likely to take quite a while, I don't see adding the link it a net positive. Given that and the other raised concerns, let's not add it now.

PR #1463, opened 2025-10-12
test(docker): upgrade ubuntu14 Python to standalone 3.13.3

scop:
20250517/3.13.3+20250517 was latest found working as of today.

This is mainly to get a currently supported one for tests and tooling. In particular, current mypy no longer supports < 3.9 as python_version.

But depending on what we do with bash version support soon, chances are that we don't need to go here at all, https://github.com/scop/bash-completion/discussions/1479

### DIFF
--- a/test/docker/ubuntu14/Dockerfile
+++ b/test/docker/ubuntu14/Dockerfile
@@ -10,8 +10,9 @@ RUN export DEBIAN_FRONTEND=noninteractive \
         software-properties-common \
         openssh-server \
         xvfb \
-    && python3.4 -c "import urllib.request; urllib.request.urlretrieve('https://github.com/pyston/pyston/releases/download/pyston_2.3.1/pyston_2.3.1_portable_v2.tar.gz', '/tmp/pyston.tar.gz')" \
-    && tar xCf /usr/local /tmp/pyston.tar.gz --strip-components=1
+    && : 20250517/3.13.3+20250517 was latest found working as of 2025-10-12 \
+    && python3.4 -c "import urllib.request; urllib.request.urlretrieve('https://github.com/astral-sh/python-build-standalone/releases/download/20250517/cpython-3.13.3+20250517-x86_64_v2-unknown-linux-gnu-install_only_stripped.tar.gz', '/tmp/python.tar.gz')" \
+    && tar xCf /usr/local /tmp/python.tar.gz --strip-components=1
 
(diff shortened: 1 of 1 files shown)


test/docker/ubuntu14/Dockerfile
```diff
@@ -10,16 +10,17 @@ RUN export DEBIAN_FRONTEND=noninteractive \
...
 ADD test-cmd-list.txt \
     requirements.txt \
     docker/ubuntu14/install-packages.sh \
     /tmp/
 
 RUN set -x \
-    && pyston3 -m pip install -Ir /tmp/requirements.txt
+    && python3 -m pip install -Ir /tmp/requirements.txt
 
 RUN /tmp/install-packages.sh </tmp/test-cmd-list.txt \
```
scop, line 25: Build fails currently with
```
[...]
991.4 Setting up mailman (1:2.1.16-2ubuntu0.5) ...
991.5 Looking for enabled languages (this may take some time) ... done.
991.6   File "<string>", line 1
991.6     import sys; print sys.version[:3]
991.6                 ^^^^^^^^^^^^^^^^^^^^^
991.6 SyntaxError: Missing parentheses in call to 'print'. Did you mean print(...)?
991.6 dpkg: error processing package mailman (--configure):
991.6  subprocess installed post-installation script returned error exit status 1
[...]
1064.6 Errors were encountered while processing:
1064.6  mailman
1064.8 E: Sub-process /usr/bin/dpkg returned an error code (1)
------
Dockerfile:24
--------------------
  23 |     
  24 | >>> RUN /tmp/install-packages.sh </tmp/test-cmd-list.txt \
  25 | >>>     && rm -r /tmp/* /root/.cache/pip /var/lib/apt/lists/*
  26 |     
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c /tmp/install-packages.sh </tmp/test-cmd-list.txt     && rm -r /tmp/* /root/.cache/pip /var/lib/apt/lists/*" did not complete successfully: exit code: 123
```

So we may want to look into doing our python install only after installing the OS packages, as the one in /usr/local/bin becomes the system one, and will start to interfere with package installs etc as shown here.

In fact, it is quite possibly that it can interfere with packages at runtime as well, and consequently our invocations of them for --help parsing and such. We may want to switch installing our python so that it doesn't become the system default one.

PR #1464, opened 2025-10-13
style: Enforce no trailing white spaces

yedayak:
Also fix all current ones


test/fixtures/compgen/t1.txt
akinomyoga, line 1: These trailing white spaces in `compgen/t{1..3}.txt` are produced by Bash's `declare`. However, the files `compgen/t{1..3}.txt` don't seem to be used by tests. They were added by commit c9c98da36edeb9305e24e7ea2d8ccf8b0934ce70, but I'm not sure the reason that these files are added. At least, even if I remove those files, the following tests still pass:

```console
$ pytest test/t/test_screen.py
$ pytest test/t/test_compgen.py
$ pytest test/t/unit
```

Can we remove these files?

test/fixtures/shared/bin/ip
akinomyoga, line 1: This is what `ip addr` produces. Shouldn't we use the output of `ip addr` as is for testing purposes? For example, to ensure that the completion generation still works with trailing white spaces. `shared/bin/ifconfig` for `ifconfig -a` is similar.

yedayak:
> I think I'd rather not do this. At least some of the affected files are intentional as-is copies of outputs etc that are better left as-is. For cases where it "matters" more we already have shfmt taking care of it for shell code, and ruff for python.

Yeah makes sense, thanks for the review

PR #1471, opened 2025-10-23
feat(smartctl): update for smartmontools 7.5

chrfranke:
This is a retry of #1425.

Uses a different approach to let `menu-complete` only append a space if all alternatives allow this.

### DIFF
--- a/completions/smartctl
+++ b/completions/smartctl
@@ -1,32 +1,66 @@
 # bash completion for smartctl(8)                          -*- shell-script -*-
+# Updated for smartmontools 7.5 (released 2025-04-30)
 
+_comp_cmd_smartctl__set_nospace_if_incomplete()
+{
+    local i
+    for i in "${!COMPREPLY[@]}"; do
+        if [[ ${COMPREPLY[i]} == *[,=] ]]; then
+            compopt -o nospace
+            break
+        fi
+    done
+}
 _comp_cmd_smartctl__device()
 {
     case $cur in
-        areca* | 3ware* | megaraid* | cciss*)
-            _comp_compgen -- -W '${cur%%,*},{0..31}'
+        3ware,* | areca,* | cciss,* | megaraid,*)
+            _comp_compgen -P "${cur%%,*}," -- -W '{0..31}'
             ;;
-        hpt*)
-            _comp_compgen -- -W 'hpt,{1..4}/{1..8} hpt,{1..4}/{1..8}/{1..5}'
+        aacraid,*,*,*,*) ;;
+        aacraid,[0-7],[0-7],*)
+            _comp_compgen -P "${cur%,*}," -- -W '{0..31}'
+            ;;
+        aacraid,*,*,*) ;;
+        aacraid,[0-7],*)
+            _comp_compgen -P "${cur%,*}," -- -W '{0..7},'
+            compopt -o nospace
+            ;;
+        aacraid,*)
+            _comp_compgen -P "aacraid," -- -W '{0..7},'
+            compopt -o nospace
+            ;;
+        hpt,*/*/*/*) ;;
+        hpt,[1-4]/[1-8]/*)
+            _comp_compgen -P "${cur%/*}/" -- -W '{1..8}'
+            ;;
+        hpt,[1-4]/*)
+            _comp_compgen -P "${cur%%/*}/" -- -W '{1..8}{,/}'
+            compopt -o nospace
+            ;;
+        hpt,*)
+            _comp_compgen -P "hpt," -- -W '{1..4}/'
+            compopt -o nospace
+            ;;
+        jmb39x,* | jmb39x-q,* | jmb39x-q2,* | jms56x,*)
+            _comp_compgen -P "${cur%%,*}," -- -W '{0..4}'
+            ;;
+        sssraid,*,*,*) ;;
+        sssraid,[0-7],*)
+            _comp_compgen -P "${cur%,*}," -- -W '{0..31}'
+            ;;
+        sssraid,*)
+            _comp_compgen -P "sssraid," -- -W '{0..7},'
+            compopt -o nospace
             ;;
         *)
-            _comp_compgen -- -W 'ata scsi sat usbcypress usbjmicron usbsunplus
-                marvell areca 3ware hpt megaraid cciss auto test'
-            case "${COMPREPLY[@]}" in
-                areca | 3ware | hpt | megaraid | cciss)
-                    compopt -o nospace
-                    ;;
-            esac
+            _comp_compgen -- -W 'auto test ata scsi nvme{,\,}
+                sat{,\,auto}{,\,12} usb{jmicron,prolific,sunplus}
+                snt{asmedia,jmicron,realtek}{,/sat} 3ware, aacraid, areca,
+                cciss, hpt, megaraid, sssraid, jmb39x{,-q,-q2}, jms56x,'
+            _comp_cmd_smartctl__set_nospace_if_incomplete
             ;;
     esac
-}
-_comp_cmd_smartctl__test()
-{
-    [[ $cur == @(pending|scttempint|vendor), ]] && return
-    _comp_compgen -- -W 'offline short long conveyance select, select,redo
-        select,next afterselect,on afterselect,off pending, scttempint,
-        vendor,'
-    [[ ${COMPREPLY-} == *, ]] && compopt -o nospace
 }
 _comp_cmd_smartctl__drivedb()
 {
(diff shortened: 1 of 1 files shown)


completions/smartctl
```diff
@@ -1,33 +1,60 @@
 # bash completion for smartctl(8)                          -*- shell-script -*-
+# Updated for smartmontools 7.5 (released 2025-04-30)
 
+_comp_cmd_smartctl__set_nospace_if_any()
+{
+    local i
+    for i in "${!COMPREPLY[@]}"; do
+        # shellcheck disable=SC2053
+        if [[ ${COMPREPLY[i]} == $1 ]]; then
```
akinomyoga: I think we can fix the pattern (and remove the argument). The intent of the function is clearer this way.

```suggestion
        if [[ ${COMPREPLY[i]} == *[,=] ]]; then
```

The name of the function may also be adjusted.

completions/smartctl
```diff
@@ -38,13 +65,39 @@ _comp_cmd_smartctl__drivedb()
...
         _comp_compgen -Rv COMPREPLY -- -P "$prefix" -W '"${COMPREPLY[@]}"'
 }
+_comp_cmd_smartctl__vendorattribute()
+{
+    local fmt
+    case $cur in
+        [1-9N],* | [1-9][0-9],* | 1[0-9][0-9],* | 2[0-4][0-9],* | 25[0-5],*)
+            fmt='raw{8,16,48,56,64},hex{48,56,64},raw24/raw{24,32}'
+            fmt+=',msec24hour32,{sec,min,halfmin}2hour,temp10x,tempminmax'
+            _comp_compgen -- -W '"${cur%%,*}",{'"$fmt"'}{,\,,}'
```
akinomyoga: `{,\,,}` generates two empty suffixes. I guess

```suggestion
            _comp_compgen -- -W '"${cur%%,*}",{'"$fmt"'}{,\,,:}'
```

completions/smartctl
```diff
@@ -38,13 +65,39 @@ _comp_cmd_smartctl__drivedb()
     _comp_compgen_filedir h && [[ $prefix ]] &&
         _comp_compgen -Rv COMPREPLY -- -P "$prefix" -W '"${COMPREPLY[@]}"'
 }
+_comp_cmd_smartctl__vendorattribute()
+{
+    local fmt
+    case $cur in
+        [1-9N],* | [1-9][0-9],* | 1[0-9][0-9],* | 2[0-4][0-9],* | 25[0-5],*)
+            fmt='raw{8,16,48,56,64},hex{48,56,64},raw24/raw{24,32}'
```
akinomyoga: `man smartctl` in my environment (Arch Linux with smartctl 7.5) also contains `raw16(raw16)`, `raw16(avg16)`, and `raw24(raw8)`. What are they, and why are they excluded?

completions/smartctl
```diff
@@ -64,32 +117,34 @@ _comp_cmd_smartctl()
...
+                tapealert tapedevstat xerror{,\,,\,error}
+                xselftest{,\,,\,selftest} zdevstat'
+            _comp_cmd_smartctl__set_nospace_if_any '*,'
             return
             ;;
         --vendorattribute | -${noargopts}v)
-            _comp_compgen -- -W 'help 9,minutes 9,seconds 9,halfminutes 9,temp
-                192,emergencyretractcyclect 193,loadunload 194,10xCelsius
-                194,unknown 198,offlinescanuncsectorct 200,writeerrorcount
-                201,detectedtacount 220,temp'
```
akinomyoga, line 88: The man page says "`The following old arguments to '-v' are also still valid:`". Shouldn't we suggest these for users who are familiar with the old ones? Or maybe we can generate these as a fallback in the case that no completions are generated for the new arguments.

completions/smartctl
```diff
@@ -1,33 +1,60 @@
...
+            _comp_compgen -- -W 'aacraid,{0..3},{0..7},'
+            compopt -o nospace
+            ;;
+        sssraid,*,*)
+            _comp_compgen -- -W 'sssraid,{0..3},{0..31}'
+            ;;
+        aacraid,* | sssraid,*)
+            _comp_compgen -- -W '"${cur%%,*}",{0..3},'
+            compopt -o nospace
+            ;;
```
akinomyoga: How about using `_comp_compgen -P`? Also, interleaving the completions for `aacraid` and `sssraid` seems a bit tricky to understand at a glance.

```suggestion
        aacraid,*,*,*,*) ;;
        aacraid,*,*,*)
            _comp_compgen -P "${cur%,*}," -- -W '{0..31}'
            ;;
        aacraid,*,*)
            _comp_compgen -P "${cur%,*}," -- -W '{0..7},'
            compopt -o nospace
            ;;
        aacraid,*)
            _comp_compgen -P "${cur%,*}," -- -W '{0..3},'
            compopt -o nospace
            ;;
        sssraid,*,*,*) ::
        sssraid,*,*)
            _comp_compgen -P "${cur%,*}," -- -W '{0..31}'
            ;;
        sssraid,*)
            _comp_compgen -P "${cur%,*}," -- -W '{0..3},'
            compopt -o nospace
            ;;
```

completions/smartctl
chrfranke: OK.

completions/smartctl
chrfranke: Good catch.

completions/smartctl
chrfranke: Again, good catch.

These options prints the raw value like `123` if the extra words are zero else like `123 (456 789)` , see [atacmds.cpp](https://github.com/smartmontools/smartmontools/blob/3c8755fa8e3c/lib/atacmds.cpp#L2006).

These were excluded because the `(...)` require shell quoting which didn't work. If added to `_comp_compgen -- -W` as `raw16\\(raw16\\)`, it correctly returns `raw16\(raw16\)` but further completion does not work then.  I also tried to pass `raw16'('raw16')'` with some `\`. Sorry if I missed something.

Please note that completeness is not very important here because these options were primarily added for the [drivedb.h](https://github.com/smartmontools/smartmontools/blob/3c8755fa8e3c/lib/drivedb.h#L83) file and are rarely used from command line. Otherwise I wouldn't have chosen such syntax.

completions/smartctl
chrfranke: Sorry, no. It is planned for 8.0 to print a deprecation warning and later remove these in 8.2.

completions/smartctl
chrfranke: Sorry, no. This happily completes nonsense like: 
```
$ /usr/sbin/smartctl -d aacraid,foo,bar,[TAB][TAB]
aacraid,foo,bar,0   aacraid,foo,bar,16  aacraid,foo,bar,23  aacraid,foo,bar,30
...
aacraid,foo,bar,15  aacraid,foo,bar,22  aacraid,foo,bar,3   aacraid,foo,bar,9
```

completions/smartctl
akinomyoga: OK.

completions/smartctl
akinomyoga: Yes, I know that behavior, but I'm not sure whether we really need to care about the parts other than the currently completed part. Anyway, if you want to stop the completion when the previous part is wrong, it is also a valid decision. I was just concerned about the fact that `aacraid,{0..3},{0..7},{0..31}` internally generates 1000 candidates, which will finally be filtered out to less than or equal to 32 candidates always.

akinomyoga: Anyway, the point you raised can be easily fixed by slightly modifying the patterns in the suggested code, e.g. from `aacraid,*,*,*` to `aacraid,[0-3],[0-7],*`.

chrfranke: OK.

BTW, even much larger expansions are still reasonably fast:
```
someone@debian-12:~$ time bash -c 'x=($(echo aacraid,{0..7},{0..15},{0..127})) ; echo "${x[@]}" | wc'
      1   16384  254208
real    0m0.032s
user    0m0.031s
sys     0m0.004s
```

completions/smartctl
akinomyoga: > it correctly returns `raw16\(raw16\)` but further completion does not work then.

I fetched your fork and checked the behavior. It seems the problem can be traced down to the fact that the `compgen` builtin performs the filtering under the assumption that the words specified to `-W` are the ones before the shell quoting.

```console
$ compgen -W 'raw16\\(raw16\\)' -- 'raw16\('
$ compgen -W 'raw16\\(raw16\\)' -- 'raw16\\('
raw16\(raw16\)
```

One way is to generate it without quoting (i.e., `-W 'raw16(raw16)'`) with `compopt -o fullquote`, but `compopt -o fullquote` is only available in Bash >= 5.3.

chrfranke: Bash >= 5.3 has been released end of July 2025. Even Ubuntu unstable ('questing') does not provide it yet.  I suggest to exclude these three options for now.

akinomyoga: Could you leave code comments about the three values?

chrfranke: Of course, something like this?
```bash
        [1-9N],* | [1-9][0-9],* | 1[0-9][0-9],* | 2[0-4][0-9],* | 25[0-5],*)
            # 'raw16(raw16)', 'raw16(avg16)' and 'raw24(raw8)' are excluded
            # because these would require 'compopt -o fullquote' (Bash >= 5.3).
            # TODO: add these if Bash >= 5.3 is detected.
            fmt='raw{8,16,48,56,64},hex{48,56,64},raw24/raw{24,32}'
```

akinomyoga: Yeah, but for lower versions of Bash, I'm actually thinking about incorporating [`__bat_escape_completions`](https://github.com/sharkdp/bat/blob/4e13c3eb5bd5e78de8cafb4b21f646b880216c2d/assets/completions/bat.bash.in#L13) (e.g. into our `_comp_compgen` with a flag `-q`), so you don't have to describe how it would be solved in 5.3. But if you want to, you can.

scop:
Just a note, I plan to release the next bash-completion version really soon, as soon as tomorrow (Oct 27). Would be nice to get this included in that release if we can get it merged in time (can also wait for a couple of more days if that's what it takes, but would rather not).

chrfranke:
Thanks for the info, will push a new version today.

chrfranke:
Done. Now always use `_comp_compgen -P` for prefixes, IMO easier to read.

completions/smartctl
akinomyoga: Thank you for updating.

completions/smartctl
akinomyoga: Thanks for fixing this.

completions/smartctl
akinomyoga: Thank you.

completions/smartctl
akinomyoga: Thank you also for addressing this point.

completions/smartctl
```diff
@@ -38,13 +72,42 @@ _comp_cmd_smartctl__drivedb()
...
+{
+    local fmt
+    case $cur in
+        [1-9N],* | [1-9][0-9],* | 1[0-9][0-9],* | 2[0-4][0-9],* | 25[0-5],*)
+            # TODO: add 'raw16(raw16)', 'raw16(avg16)' and 'raw24(raw8)'.
+            # Excluded for now because 'compopt -o fullquote' (Bash >= 5.3)
+            # or similar bash_completion functionality would be required.
+            fmt='raw{8,16,48,56,64},hex{48,56,64},raw24/raw{24,32}'
+            fmt+=',msec24hour32,{sec,min,halfmin}2hour,temp10x,tempminmax'
+            _comp_compgen -P "${cur%%,*}," -- -W '{'"$fmt"'}{,\,,:}'
```
akinomyoga: Showing three variations for all of the possible values (16x3 or 19x3) may be too much. It would be tedious to select one in `menu-complete`. What do you think about the idea of showing the three variations only when the candidate is unique?
```suggestion
            _comp_compgen -P "${cur%%,*}," -- -W '{'"$fmt"'}'
            if ((${#COMPREPLY[@]} == 1)); then
                _comp_compgen -R -- -W '"$COMPREPLY"{,\,,:}'
            fi
```

If we adopt this, `'{'"$fmt"'}'` should be expanded in the actual code.

chrfranke: Makes plenty of sense && done

akinomyoga: Thank you! I confirmed the behavior of the new version in my environment.

akinomyoga (approved):

PR #1478, opened 2025-10-25
chore(ci): Stop using apt-add-repository

yedayak:
We switched to using archive.debian.org in https://github.com/scop/bash-completion/pull/1469, but apt-add-repository adds the old ones itself even when not specified. Use sed to add `contrib` and `non-free` instead.


scop (approved):
Approved, but let's note debian10 in the commit message before merging as ubuntu14 still uses apt-add-repository.

PR #1483, opened 2025-10-27
fix(_comp_dequote): set the literal value to REPLY as a fallback

akinomyoga:
See the commit message. This is separated from PR #1357.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -190,11 +190,11 @@ _comp_dequote__initialize()
 _comp_dequote__initialize
 
 # This function expands a word using `eval` in a safe way.  This function can
-# be typically used to get the expanded value of `${word[i]}` as
-# `_comp_dequote "${word[i]}"`.  When the word contains unquoted shell special
-# characters, command substitutions, and other unsafe strings, the function
-# call fails before applying `eval`.  Otherwise, `eval` is applied to the
-# string to generate the result.
+# be typically used to get the expanded value of `${word[i]}` as `_comp_dequote
+# "${word[i]}"`.  When the word contains unquoted shell special characters,
+# command substitutions, and other unsafe strings, the function call fails
+# before applying `eval` and REPLY is set to be the literal string.  Otherwise,
+# `eval` is applied to the string to generate the result.
 #
 # @param    $1  String to be expanded.  A safe word consists of the following
 #               sequence of substrings:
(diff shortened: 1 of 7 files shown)


akinomyoga:
There is an unresolved conversation in the original PR:

https://github.com/scop/bash-completion/pull/1357#discussion_r2030264755 by yedayak

> This change seems reasonable to me, do you think this might cause breakage in external uses of the function?

https://github.com/scop/bash-completion/pull/1357#discussion_r2030283770 by akinomyoga

> Yeah, that is a good point.
> 
> After this change, the cases where it failed will continue to fail. Some of the cases where it succeeded and produced an empty result [e.g., `REPLY=()` for `_comp_dequote '$empty_variable'`] will start to fail after this change. Since the caller should have considered failing cases even with the current `main` branch, I expect that increasing another case of failing wouldn't cause a problem.
> 
> Rather, an empty `REPLY=()` would have caused a problem when the caller uses `$REPLY` after `_comp_dequote` succeeds.
> 
> ```shell
> some=""
> if _comp_dequote '$some'; then # -> results in REPLY=()
>   use "$REPLY" # <- This failed for "set -u" before the present change
> fi
> ```
> 
> Another change is in the value of `REPLY` when `_comp_dequote` failed. Since `REPLY` was not supposed to be used when `_comp_dequote` failed (`REPLY` was actually empty, `REPLY=()`), I expect the change in `REPLY` in the failing case wouldn't cause a problem, though I admit that there might be a case that the user expects the empty `REPLY` in principle. I think this change would rather improve the compatibility in the case where the caller isn't careful enough to check the exit status:
> 
> ```shell
> _comp_dequote "$some2"
> use "$REPLY" # <- This failed for "set -u", or if "set -u" is not set, an empty value was used.
> ```

bash_completion
```diff
@@ -237,8 +242,15 @@ _comp_dequote__initialize
 _comp_dequote()
 {
     REPLY=() # fallback value for unsafe word and failglob
```
yedayak: This isn't true anymore, maybe just don't initialize it yet? Or set it to the default value if we don't eval.

akinomyoga: Ah, yeah. This isn't a fallback value for an unsafe word anymore, though it is still a fallback value for `failglob`. On `failglob`, the state of the array before `eval "REPLY=($1)"` remains. For the PR, I tried to minimize the number of changed lines, but I'll move this line under the branch performing `eval` as it seems we need to adjust the code comment anyway.

akinomyoga: 64853cbe1262445c15582d7566a91b1394ed1c47

akinomyoga: No, we actually needed `REPLY=()` even for the case of an unsafe word. `REPLY=()` cleared `REPLY` as an array.

In an additional commit d3bfb19, I changed the assignment for the unsafe word to use the full assignment of an array.

PR #1485, opened 2025-10-27
fix(java,rsync,ssh): complete syntactically incomplete cur

### LINKED ISSUE #1255: Escaping of remote filenames containing spaces is broken for rsync

akinomyoga:
This is also separated from #1357. See the commit messages. This fixes #1255.

### DIFF
--- a/test/t/unit/test_unit_dequote.py
+++ b/test/t/unit/test_unit_dequote.py
@@ -54,7 +54,7 @@ class TestDequote:
 
     def test_7_quote_3(self, bash, functions):
         output = assert_bash_exec(
-            bash, "__tester \\$\\'a\\\\tb\\'", want_output=True
+            bash, r"__tester \$\'a\\tb\'", want_output=True
         )
         assert output.strip() == "<a\tb>"
 
@@ -66,7 +66,7 @@ class TestDequote:
 
     def test_7_quote_5(self, bash, functions):
         output = assert_bash_exec(
-            bash, "__tester \\'abc\\'\\\\\\'\\'def\\'", want_output=True
+            bash, r"__tester \'abc\'\\\'\'def\'", want_output=True
         )
         assert output.strip() == "<abc'def>"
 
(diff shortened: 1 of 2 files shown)


completions/java
```diff
@@ -114,7 +114,7 @@ _comp_cmd_java__packages()
     local -a sourcepaths=("${REPLY[@]}")
 
     local REPLY
-    _comp_dequote "$cur"
+    _comp_dequote_incomplete "$cur"
```
scop, line 117: It's not immediately clear to me why we'd do this when completing java packages, the commit messages nor the referenced issues don't mention it either, and there's seemingly no test suite coverage for it. Is it necessary, and if yes, can we somehow clarify why and add test cases?

akinomyoga: With `javadoc 'a[tab]`, `cur` becomes an incomplete word `cur="'a"`. If we use `_comp_dequote`, nothing is completed because modules starting with `'a` are not found. In dealing with a value coming from `cur`, we need to use `_comp_dequote_incomplete` in general.

- I rebased on top of the latest `main`, and I added a test case in `test/t/test_javadoc.py`.
- I described this in the commit message of the first commit.
- I mentioned it in the code documentation of `_comp_dequote` and `_comp_dequote_incomplete`.

test/fixtures/scp/local_path/backslash-a\ b.txt
scop, line 1: Missed in the review and already merged, but I think this is a filename that will cause problems e.g. on Windows, and it would be better created dynamically.

scop: https://github.com/scop/bash-completion/pull/1489, https://github.com/scop/bash-completion/issues/1490

PR #1491, opened 2025-10-31
test(scp): create files on the fly to work around Windoes filesystem

### LINKED ISSUE #1490: Backslash filenames in repo
> Missed in the review and already merged, but I think this is a filename that will cause problems e.g. on Windows, and it would be better created dynamically. 

 _Originally posted by @scop in [#1485](https://github.com/scop/bash-completion/pull/1485/files#r2478861790)_

I'd like this addressed before releasing 2.17.0.

See also https://github.com/scop/bash-completion/pull/1489

akinomyoga:
This fixes the issue pointed out in https://github.com/scop/bash-completion/pull/1485#discussion_r2478882097. I'm sorry that I caused this again. I caused the same issue before in https://github.com/scop/bash-completion/pull/1397#discussion_r2164546105, which was fixed in #1408. I realized that the previous issue was pointed out this June, but #1357 (i.e., the original PR of #1485) was created this April. I created the original change before the previous issue but failed to properly update the changes.

In this PR, I removed the affected files and used the same approach to the test as #1408.

Fixes #1490.

### DIFF
--- a/test/t/test_scp.py
+++ b/test/t/test_scp.py
@@ -33,7 +33,7 @@ class TestScp:
                     )
                 ),
                 # Local filenames
-                ["config", "known_hosts", "local_path/", r"spaced\ \ conf"],
+                ["config", "known_hosts", r"spaced\ \ conf"],
             )
         )
         assert completion == expected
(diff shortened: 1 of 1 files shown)


scop (approved):
No problem, this is an easy thing to miss and we should have had checks for it way earlier already. Thanks for addressing it quickly!

akinomyoga:
Thank you.

PR #1501, opened 2025-11-07
fix(cargo): removed local from a variable declaration outside a function

safiyat:
Ever since upgrading bash from 5.2.37 (released 2024-09-23) to  5.3.3 (released 2025-07-03), getting the warning:
```
-bash: local: can only be used in a function
```

This PR fixes that.

### DIFF
--- a/completions/_cargo
+++ b/completions/_cargo
@@ -2,8 +2,7 @@
 #
 # This serves as a fallback in case the completion is not installed otherwise.
 
-# shellcheck disable=SC2168 # "local" is ok, assume sourced by _comp_load
-local rustup="${1%cargo}rustup" # use rustup from same dir
+rustup="${1%cargo}rustup" # use rustup from same dir
 eval -- "$("$rustup" completions bash cargo 2>/dev/null)"
 
 # ex: filetype=sh


akinomyoga:
Have you noticed the code comment that you removed? The file `completions/_cargo` is supposed to be sourced by the shell function `_comp_load`, so the error shouldn't happen. If you see the message, it implies that you source the file `completions/_cargo` in a wrong way.

safiyat:
Oh, that never crossed my mind. Thanks!

I am still curious though how just bumping up my bash version led to that warning message. Any ideas?
I am using the [brew package manager](https://github.com/Homebrew/homebrew-core/blob/main/Formula/b/bash.rb), if that helps.

akinomyoga:
> I am still curious though how just bumping up my bash version led to that warning message. Any ideas?

I don't have an idea. In the general case, I'd suspect that some other packages were also updated or installed at the same time. If you still see the warning messages, when is the error message printed? Is the message displayed when a Bash session starts? Or is it shown when you attempt completion of arguments to `cargo`?

safiyat:
I still see that warning message. Is is displayed whenever I start a new bash session (new window or new tab of the terminal).

Doesn't come when attempting the autocompletion of `cargo`.

akinomyoga:
Thanks for the information. If the message is displayed on the startup, it implies that the file `_cargo` is placed in a wrong place, or `BASH_COMPLETION_COMPAT_DIR` is set to a wrong value. Can you identify the absolute path of the file `_cargo` in your system? Also, what is the result of the following command in a Bash session (where the error message was printed on startup)?

```console
$ declare -p BASH_COMPLETION_COMPAT_DIR
```

safiyat:
`_cargo` is placed at `/opt/homebrew/share/bash-completion/completions/_cargo`.

```
Last login: Tue Nov 11 17:35:20 on ttys000
-bash: local: can only be used in a function
11 17:47:16 safiyat@Safiyats-MacBook-Pro:~$ declare -p BASH_COMPLETION_COMPAT_DIR
-bash: declare: BASH_COMPLETION_COMPAT_DIR: not found
11 17:58:08 safiyat@Safiyats-MacBook-Pro:~$ find /opt/homebrew/share/bash-completion/completions/_cargo
/opt/homebrew/share/bash-completion/completions/_cargo
11 17:58:32 safiyat@Safiyats-MacBook-Pro:~$
```

akinomyoga:
Thank you. Hmm, I don't have an idea. Could you temporarily rewrite `/opt/homebrew/share/bash-completion/completions/_cargo` to inspect the information on the loading time? (If the following requires an administrator privilege, please do it with the administrator privilege.)

```console
$ file=/opt/homebrew/share/bash-completion/completions/_cargo
$ cp "$file"{,.backup}
$ echo 'declare -p BASH_SOURCE BASH_LINENO _comp_init_startup_config BASH_COMPLETION_VERSINFO >/dev/tty' > "$file"
```

With this setup, what is shown on the startup?

After testing, you can revert the setting:

```console
$ file=/opt/homebrew/share/bash-completion/completions/_cargo
$ mv "$file"{.backup,}
```

PR #1504, opened 2025-11-11
chore(renovate): extend scop/common preset

scop:
I'm consolidating my Renovate configs (and soon some other common things) to a common repo, cf. ref https://github.com/scop/common/blob/main/default.json

No plans to make any drastic changes and I'll update extending projects if I figure out I may be doing something in the common that conflicts with them.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,20 +1,12 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:best-practices",
-    ":enablePreCommit",
-    ":prConcurrentLimit20",
-    ":preserveSemverRanges",
-    ":semanticPrefixChore"
+    "github>scop/common"
   ],
   "ignorePresets": [
-    ":dependencyDashboard",
     ":ignoreModulesAndTests",
-    ":semanticPrefixFixDepsChoreOthers",
     "docker:pinDigests"
   ],
-  "semanticCommits": "enabled",
-  "commitMessageTopic": "{{depName}}",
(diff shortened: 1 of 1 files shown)


scop:
The usual JSON does not, however Renovate's parser does, and they explicitly recommend using the .json suffix when there are comments nevertheless (and they don't support .jsonc at least according to docs) :shrug: 

https://docs.renovatebot.com/configuration-options/
> Renovate supports JSONC for .json files and any config files without file extension (e.g. .renovaterc). We also recommend you prefer using JSONC within a .json file to using a .json5 file if you want to add comments.

.github/renovate.json
```diff
@@ -1,20 +1,12 @@
...
-    ":prConcurrentLimit20",
-    ":preserveSemverRanges",
-    ":semanticPrefixChore"
+    "github>scop/common"
   ],
   "ignorePresets": [
-    ":dependencyDashboard",
     ":ignoreModulesAndTests",
-    ":semanticPrefixFixDepsChoreOthers",
     "docker:pinDigests"
```
scop, line 8: Hm this did not work the way I thought it would, it appears `ignorePresets` was overridden rather than being merged, as the dep dashboard popped up again, https://github.com/scop/bash-completion/issues/589. Need to have a look.

akinomyoga:
> The usual JSON does not, however Renovate's parser does, and they explicitly recommend using the .json suffix when there are comments nevertheless (and they don't support .jsonc at least according to docs) 🤷
> 
> https://docs.renovatebot.com/configuration-options/

Thank you!

PR #1517, opened 2025-11-30
feat(ssh-keygen): more option arg completions

scop:
See individual commits for details.

### DIFF
--- a/completions/ssh-keygen
+++ b/completions/ssh-keygen
@@ -145,11 +145,14 @@ _comp_cmd_ssh_keygen()
             local pathcmd protocols
             pathcmd=$(type -P -- "$1") && local PATH=${pathcmd%/*}:$PATH
             _comp_compgen -v protocols -x ssh query protocol-version
-            local types='dsa ecdsa ecdsa-sk ed25519 ed25519-sk rsa'
+            local -a types=(
+                dsa ecdsa ecdsa-sk ed25519 ed25519-sk rsa
+                rsa-sha2-256 rsa-sha2-512 ssh-rsa
+            )
             if [[ ${protocols[*]} == *1* ]]; then
-                types+=' rsa1'
+                types+=(rsa1)
             fi
-            _comp_compgen -- -W "$types"
+            _comp_compgen -- -W '"${types[@]}"'
             return
             ;;
         -${noargopts}Y)


completions/ssh-keygen
```diff
@@ -141,15 +145,25 @@ _comp_cmd_ssh_keygen()
             local pathcmd protocols
             pathcmd=$(type -P -- "$1") && local PATH=${pathcmd%/*}:$PATH
             _comp_compgen -v protocols -x ssh query protocol-version
-            local types='dsa ecdsa ecdsa-sk ed25519 ed25519-sk rsa'
+            local -a types=(
+                dsa ecdsa ecdsa-sk ed25519 ed25519-sk rsa
```
yedayak, line 149: Maybe we should remove dsa? OpenSSH [removed support](https://www.openssh.org/releasenotes.html#:~:text=This%20release%20removes%20support%20for%20the%20weak%20DSA%20signature%20algorithm) completely in April 2025

completions/ssh-keygen
```diff
@@ -141,15 +145,25 @@ _comp_cmd_ssh_keygen()
...
         -${noargopts}Y)
-            _comp_compgen -- -W 'find-principals check-novalidate sign verify'
+            _comp_compgen -- -W \
+                'find-principals match-principals check-novalidate sign verify'
+            return
+            ;;
+        -${noargopts}Z)
+            local ciphers
+            _comp_compgen -v ciphers -x ssh query ciphers
+            _comp_compgen -- -W '"${ciphers[@]}"'
```
akinomyoga:
```suggestion
            _comp_compgen -x ssh query ciphers
```

Is there a reason to store the results in an array?

scop: Not really, thanks for the catch.

completions/ssh-keygen
scop: Maybe. I suppose we could also drop support for protocol version 1 which is long gone as well, but the difference is that supported protocol versions are queryable, key types are not. Anyway I won't include either in this PR.

completions/ssh-keygen
akinomyoga: Thanks!

akinomyoga (approved):

PR #1529, opened 2025-12-28
fix(cd): Complete from . on empty CDPATH entry (#1527)

### LINKED ISSUE #1527: Empty string entry in CDPATH completes from /
## Describe the bug

CDPATH entries may be empty, which according to the bash documentation is equivalent to `.`; however it is slightly different in practice because `.` will echo the directory name while the empty string will not.

When CDPATH includes an empty string (e.g., `CDPATH=:~/cdpath`) the completion includes directories under the root directory `/`.

This appears to be a working one-line fix:
```
diff --git a/completions/cd b/completions/cd
index 8de903a6..51692079 100644
--- a/completions/cd
+++ b/completions/cd
@@ -38,6 +38,7 @@ _comp_cmd_cd__compgen_cdpath()
     local paths dirs _d
     _comp_split -F : paths "$CDPATH"
     for _p in "${paths[@]}"; do
+        _p="${_p:-.}"
         # create an array of matched subdirs
         _comp_compgen -v dirs -c "$_p/$cur" -- -d
         for _d in "${dirs[@]}"; do
```

## To reproduce

```
$ env -i bwrap --ro-bind / / --dev /dev --ro-bind /etc/skel $HOME bash
[joe@doughboy joe]$ ls
[joe@doughboy joe]$ CDPATH=:$HOME
[joe@doughboy joe]$ cd s[TAB][TAB]
sbin/  srv/   sys/   
```
This is basically a default empty home directory with CDPATH set to `:/home/joe`.

## Expected behavior

The current home directory is empty so there should be no completions.

- [ ] Operating system name/distribution and version: Fedora 43
- [ ] bash version, `echo "$BASH_VERSION"`: 5.3.0(1)-release (x86_64-redhat-linux-gnu)
- [ ] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2 16 0 (also reproduced at commit e8a3d339)

jmou:
Fixes #1527  (edit by akinomyoga)

### DIFF
--- a/test/t/test_cd.py
+++ b/test/t/test_cd.py
@@ -51,3 +51,15 @@ class TestCd:
             bash_env.write_variable("foo2", "shared/default")
             completion = assert_complete(bash, "cd f")
             assert completion == ["foo1", "foo2"]
+
+    @pytest.mark.complete("cd ", cwd="shared/default", env=dict(CDPATH=":.."))
+    def test_cdpath_leading_colon(self, completion):
+        assert completion == [
+            ".ssh/",
+            "bar bar.d/",
+            "bin/",
+            "default/",
+            "empty_dir/",
+            "foo.d/",
+            "ld.so.conf.d/",
+        ]
--- a/completions-core/cd.bash
+++ b/completions-core/cd.bash
@@ -38,6 +38,7 @@ _comp_cmd_cd__compgen_cdpath()
     local paths dirs _d
     _comp_split -F : paths "$CDPATH"
     for _p in "${paths[@]}"; do
+        _p="${_p:-.}"
         # create an array of matched subdirs
         _comp_compgen -v dirs -c "$_p/$cur" -- -d &&
             for _d in "${dirs[@]}"; do


jmou:
For some reason `test_cdable_vars` fails on my setup with or without my change, but it seems unrelated.

akinomyoga:
> For some reason `test_cdable_vars` fails on my setup with or without my change, but it seems unrelated.

Fixes are included as https://github.com/scop/bash-completion/pull/1530/commits/73958857530074ef0f3caa7723deff73c3df2ce7 and https://github.com/scop/bash-completion/pull/1530/commits/8899a06ecb7525f1512f667dc046081c88c5b008 in #1530

test/t/test_cd.py
```diff
@@ -42,3 +42,9 @@ def test_cdable_vars(self, bash):
             bash_env.write_variable("foo6", "shared/default/bar")
             completion = assert_complete(bash, "cd f")
             assert completion == ["foo1", "foo2"]
+
+    @pytest.mark.complete(
+        "cd ", cwd="shared/empty_dir", env=dict(CDPATH=":../default")
```
scop: With this, we're testing we no longer complete anything from `/` which is good. I think it would be even better if we'd cwd somewhere that has subdirs, in order to test that we actually _do_ offer completions from that dir due to the empty entry?

jmou:
Good point, I switched them around. I don't see any other directories in shared with subdirectories, so it's now only testing completion from the empty CDPATH entry and not the second entry. I'm happy to change the second CDPATH entry to another directory with subdirectories, if someone can suggest an appropriate directory to depend on for this test.

Rebased with the fix for test_cdable_vars

scop:
I think we can use for example `shared/` as the other dir, no need to make it fancier than that. (It does contain `bin/` which is a thing in `/` too, but I don't think that's a problem.)

jmou:
Updated

PR #1531, opened 2025-12-29
refactor: separate external/bundled completion directories

akinomyoga:
This implements https://github.com/scop/bash-completion/discussions/1329#discussioncomment-12163012.

The first commit b4dda6d12 contains purely renaming of files (plus necessary changes to `.pre-commit-config.yaml` to make `git commit` to pass). This is to make `git` correctly recognize the move of the files. See the commit message for details.
- In this PR, the name `-core` (instead of `-dist` initially suggested in [#1329](https://github.com/scop/bash-completion/discussions/1329#discussioncomment-12163012))  is adopted. Also, fallback completions are currently separated into `-fallback`. I think the naming of the directories needs to be discussed before merging.
  - The background for the choice of `core` is explained in [this reply](https://github.com/scop/bash-completion/discussions/1329#discussioncomment-14771602): `dist` seems to be used by both the core and third-party packages in the case of `texmf` and used to exclude the files installed by the users and administrators of a specific machine.
  - Currently fallback completions are separated into `completions-fallback`, but it was suggested in [a reply](https://github.com/scop/bash-completion/discussions/1329#discussioncomment-14766343) to not distinguish the primarily maintained completions and fallback/compat completions and mix the files in a directory `completions-core`. However, I would like to keep the distinction as argued in [a reply](https://github.com/scop/bash-completion/discussions/1329#discussioncomment-14771885).

The second commit 29355f8cf adjusts the file contents for the file renaming. The search order of `_comp_load` is also adjusted. New search ordering is described in the updated `README.md`.

- The search ordering may also be a topic to be discussed.
- To simplify the PR, the following substantial changes were separated into the next (planned) PR.
  - I plan to change the precedence of `completions/<cmd>` vs `completions/<cmd>.bash` in the next PR (so that `<cmd>.bash` will have the higher priority than `<cmd>`)
  - I plan to remove the support for `completion/_<cmd>` because `_<cmd>` has been reserved for internal uses of `bash-completion` and is not supposed to be used by external completions, as has been described in `README.md`.
  - The pkg-config variable `helpersdir` will be described in `README`.

The third commit 961b8be83 removes `test/fallback/completions`. See the commit message for the details.


scop (approved):
This makes sense to me, including the proposed naming. I have not read everything through, but I'm comfortable going with this if @yedayak agrees. Whether we go with this or not, let's not keep this cooking for too long as this stuff is very prone to conflicts.

test/Makefile.am
```diff
@@ -1,8 +1,12 @@
...
+EXTRA_DIST = config/bashrc \
+	     config/inputrc \
 	     fixtures \
 	     setup.cfg
 
 CLEANFILES = \
 	fixtures/make/extra_makefile
+
+clean-local:
+	-rm -rf config/bash-completion config/bash-completion-fallback
```
yedayak: nit: Should be in .gitignore maybe

completions/Makefile.am
```diff
@@ -1,1575 +1,22 @@
...
 .PHONY: symlinks
 
 SETUP_SYMLINKS = $(srcdir)/../setup-symlinks.sh
 
 all-local: ss = $(SETUP_SYMLINKS) .
 all-local: symlinks
 
-install-data-hook: ss = $(SETUP_SYMLINKS) $(DESTDIR)$(bashcompdir)
+install-data-hook: ss = $(SETUP_SYMLINKS) $(DESTDIR)$(bashcompdir_fallback)
 install-data-hook: symlinks
```
yedayak: We should probably still create the directory when installing, so external completions can be placed here easily
Maybe using something [like this](https://stackoverflow.com/a/8531548)?

Same for `helpers/` if users are expected to put their own stuff there.

completions/Makefile.am
yedayak, line 1: We should probably still create the directory when installing, so external completions can be placed here easily
Maybe using something [like this](https://stackoverflow.com/a/8531548)?

Same for `helpers/` if users are expected to put their own stuff there.

test/config/bashrc
```diff
@@ -1,5 +1,15 @@
 # bashrc file for bash-completion test suite
 
+_comp__test_link()
+{
+    local src=$1 dst=$2
+    if [[ -L $dst || -e $dst ]]; then
+        rm -rf "$dst"
+    fi
+    mkdir -p "${dst%/*}"
```
yedayak: This makes running `make distcheck` fail for me, it doesn't manage to create the directories in since the directories in `bash-completion-2.17.0/` are created without write permissions. Not sure how it works in CI though

<details>
<summary>Error</summary>

```
E               AssertionError: Unexpected output from "source '/home/yedaya/dev/bash-completion/bash-completion-2.17.0/test/config/bashrc'": exit status=0, output="
E                 mkdir: cannot create directory ‘/home/yedaya/dev/bash-completion/bash-completion-2.17.0/test/config/bash-completion’: Permission denied
E                 ln: failed to create symbolic link '/home/yedaya/dev/bash-completion/bash-completion-2.17.0/test/config/bash-completion/completions': No such file or directory
E                 mkdir: cannot create directory ‘/home/yedaya/dev/bash-completion/bash-completion-2.17.0/test/config/bash-completion-fallback’: Permission denied
E                 ln: failed to create symbolic link '/home/yedaya/dev/bash-completion/bash-completion-2.17.0/test/config/bash-completion-fallback/completions': No such file or directory"
```

</details>

akinomyoga:
CI started to fail with the committed change (#1525). It seems the maximum length of the commit subject became 50 (instead of the previous 70), which seems too short to me. Can we explicitly set the maximum length to 70, or is this change intentional?

```
f8b9f78: error Commit subject is too long, 66 exceeds the max length of 50
```

test/Makefile.am
akinomyoga: Thank you. I added the directories to `.gitignore` ([`f8b9f78..946457e`](https://github.com/scop/bash-completion/compare/f8b9f781a0b3a2acadda7962e392bb30e4fa8f6f..946457e30a499262df48bb74f2f55b5e9298aff0)).

completions/Makefile.am
akinomyoga: I also added the instructions to create/remove the directories in the installation target ([`f8b9f78..946457e`](https://github.com/scop/bash-completion/compare/f8b9f781a0b3a2acadda7962e392bb30e4fa8f6f..946457e30a499262df48bb74f2f55b5e9298aff0)).

test/config/bashrc
akinomyoga: OK, so `make distcheck` seems to require that the check does not modify anything in the source tree. One way is to create the directories in a temporary directory, but I'd like to think about a better way.

akinomyoga: I decided to create the directories as a PyTest fixture ([`946457e..92d9a8d`](https://github.com/scop/bash-completion/compare/946457e30a499262df48bb74f2f55b5e9298aff0..92d9a8d6b1262fdcb02d7859c555a5862b97f1e5)).

scop:
Re commit subject length, https://github.com/scop/bash-completion/pull/1532

test/config/bashrc
yedayak: This looks good, a follow up might be to convert usages of `prepare_fixture_dir` to also use `tmp_path_factory`, maybe it will help with https://github.com/scop/bash-completion/pull/1339#discussion_r2442558142, But this isn't needed for this PR.

yedayak (approved):
For a followup: some exclusions in pre-commit for completions-core are probably not needed anymore, since all of the files end with `.bash` now. Also some file filters could be simplified to just catch `*.bash` files.

akinomyoga:
> For a followup: some exclusions in pre-commit for completions-core are probably not needed anymore, since all of the files end with `.bash` now. Also some file filters could be simplified to just catch `*.bash` files.

Thank you. I updated: [`92d9a8d..6fb0bf9`](https://github.com/scop/bash-completion/compare/92d9a8d6b1262fdcb02d7859c555a5862b97f1e5..6fb0bf929e44fb258741fd117f3288e4ef864936). Actually, some of `completions-core` in `.pre-commit-config.yaml` should have been `completions-(core|fallback)`. I added `\.bash` and removed some `exclude`s.

akinomyoga:
Rebased ([`6fb0bf9..a973571`](https://github.com/scop/bash-completion/compare/6fb0bf929e44fb258741fd117f3288e4ef864936..a9735711f4b687af68efa4511146e479bc01d4d2)).

test/config/bashrc
akinomyoga: 7b8bd9c30f8ca30fb00b7f3ab245fca8243b14f6 I updated to use `tmp_path_factory` in `prepare_fixture_dir` (and other places that have used `tmpfile`).

akinomyoga: > a follow up might be to convert usages of `prepare_fixture_dir` to also use `tmp_path_factory`,

I initially thought we could quickly change `prepare_fixture_dir` to use `tmp_path_factory`, but, as you initially suggested, I decided to separate the PR for an easier review.

I dropped the commit from this PR. I'm going to merge this PR after the CI passes.

PR #1532, opened 2025-12-30
chore(committed): increase subject length to 72

scop:
50 is a tad too short, 72 reaps benefits well enough. 72 is what we had with gitlint, too.

Note that we'll still have 72 for commit body lines (I think we used to have 80). 72 is a better default for that, as it keeps the message display tidy on 80-char terminal, taking git log indentation into account.


akinomyoga:
Thanks!

PR #1533, opened 2025-12-30
test(ifup, ifdown): Require cmd for completions

yedayak:
fedora 43 doesn't seem to have ifup. Presumably it didn't fail in CI because we have an xfail for containers.

### DIFF
--- a/test/t/test_ifdown.py
+++ b/test/t/test_ifdown.py
@@ -11,7 +11,7 @@ class TestIfdown:
         reason="There won't be any configured interfaces without network",
     )
     @pytest.mark.xfail(in_container(), reason="Probably fails in a container")
-    @pytest.mark.complete("ifdown ")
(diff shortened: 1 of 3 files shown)


test/t/test_ifdown.py
```diff
@@ -11,7 +11,7 @@ class TestIfdown:
         reason="There won't be any configured interfaces without network",
     )
     @pytest.mark.xfail(in_container(), reason="Probably fails in a container")
-    @pytest.mark.complete("ifdown ")
+    @pytest.mark.complete("ifdown ", require_cmd=True)
```
scop, line 14: Needs an entry in test/test-cmd-list.txt, see CI fail.

yedayak: Forgot to setup pre-commit... updated

PR #1534, opened 2025-12-30
feat(tmux): complete target-session arguments

jessedhillon:
This adds completion for session names. Commands which take a running session's name as an argument, for example `attach-session`, can now have those options completable. For example:

```bash
$ tmux list-sessions -F '#{session_name}'
bash-cmp
nix-config
$ tmux attach-session -t <tab>
bash-cmp    nix-config
```

I checked the tmux repo, and they do not maintain bash completions, so I'm submitting this here.

### DIFF
--- a/completions-core/tmux.bash
+++ b/completions-core/tmux.bash
@@ -111,6 +111,10 @@ _comp_cmd_tmux__value()
         file | *-file | path | *-path)
             _comp_compgen_filedir
             ;;
+        target-session)
+            _comp_compgen_split -l -- "$(_comp_cmd_tmux__run \
+                list-sessions -F '#{session_name}')"
+            ;;
     esac
 }
 


completions/tmux
```diff
@@ -111,6 +111,10 @@ _comp_cmd_tmux__value()
         file | *-file | path | *-path)
             _comp_compgen_filedir
             ;;
+        target-session | src-session)
```
yedayak: Did you see `src-session` in output from tmux? I only found it in the [tmux source](https://github.com/tmux/tmux/blob/00030aabf739ea40aae25570b3f12ffaa15c0018/tmux.h#L2220), but it doesn't seem to be used currently. If we do want to have it here anyway, we should probably also have `dst-session`

jessedhillon: no, I asked an AI to look at tmux output and propose those options. I see that `target-session` is the only option that is used in the usage output, so we can trim this down to one option name.

akinomyoga:
The `main` branch has conflicting changes, so I rebased the PR head (and squashed the second commit into the first) last week. As the second commit is a fix for the problem in the first one, it should anyway be squashed into the first commit before merging. Now, you seem to have simply reverted to the state before the conflict resolution, so we cannot merge this.

jessedhillon:
I've squashed the tree and rebased onto main HEAD

akinomyoga (approved):
Thank you.

PR #1538, opened 2025-12-31
test(service): Skip if we don't have init tools or dirs

yedayak:
Fails in a container without any services or init related tools.

### DIFF
--- a/test/t/test_service.py
+++ b/test/t/test_service.py
@@ -2,12 +2,19 @@ import sys
 
 import pytest
 
+from conftest import assert_complete, bash_env_saved
+
 
 class TestService:
     @pytest.mark.xfail(
         sys.platform == "darwin",
         reason="Service completion not available on macOS",
     )
-    @pytest.mark.complete("service ")
-    def test_1(self, completion):
-        assert completion
+    def test_1(self, bash):
+        with bash_env_saved(bash) as bash_env:
+            bash_env.write_variable(
+                "_comp__test_inetd_dir",
+                "%s/_comp_compgen_xinetd_services/xinetd.d" % bash.cwd,
+            )
+            completion = assert_complete(bash, "service ")
+            assert all(x in completion for x in ("arp", "ifconfig"))
--- a/bash_completion
+++ b/bash_completion
@@ -144,6 +144,7 @@ _comp_userland()
 _comp_sysvdirs()
 {
     sysvdirs=()
+    [[ -d ${_comp__test_inetd_dir-} ]] && sysvdirs+=("$_comp__test_inetd_dir")
     [[ -d /etc/rc.d/init.d ]] && sysvdirs+=(/etc/rc.d/init.d)
     [[ -d /etc/init.d ]] && sysvdirs+=(/etc/init.d)
     # Slackware uses /etc/rc.d


test/t/test_service.py
```diff
@@ -8,6 +8,14 @@ class TestService:
         sys.platform == "darwin",
         reason="Service completion not available on macOS",
     )
-    @pytest.mark.complete("service ")
+    @pytest.mark.complete(
+        "service ",
+        # Skip if we don't have a way to find services
+        skipif="! (type systemctl || type service || type initctl"
```
akinomyoga: Do you intend this?

```suggestion
        skipif="! (type systemctl || type service || type initctl || "
```

akinomyoga (commented):
Maybe we should provide a fixture file to replace one of `/etc/init.d`, `/etc/rc.d/init.d`, etc. as done in `test/t/unit/test_unit_compgen_xinetd_services.py`.

https://github.com/scop/bash-completion/blob/5b5b9ccc8166c7872c1ee4c01927fd1e9ef9ae24/bash_completion#L2185

https://github.com/scop/bash-completion/blob/5b5b9ccc8166c7872c1ee4c01927fd1e9ef9ae24/test/t/unit/test_unit_compgen_xinetd_services.py#L20

**edit**: cf c2cafae0769a82513266b25d4c9f50974525b93e

test/t/test_service.py
yedayak: Fixed, thanks

yedayak:
> Maybe we should provide a fixture file to replace one of `/etc/init.d`, `/etc/rc.d/init.d`, etc. as done in `test/t/unit/test_unit_compgen_xinetd_services.py`.
> 
> https://github.com/scop/bash-completion/blob/5b5b9ccc8166c7872c1ee4c01927fd1e9ef9ae24/bash_completion#L2185
> 
> https://github.com/scop/bash-completion/blob/5b5b9ccc8166c7872c1ee4c01927fd1e9ef9ae24/test/t/unit/test_unit_compgen_xinetd_services.py#L20
> 
> **edit**: cf [c2cafae](https://github.com/scop/bash-completion/commit/c2cafae0769a82513266b25d4c9f50974525b93e)

Maybe, I'm not really sure what the format of those files are supposed to be.

akinomyoga:
Init scripts are just shell scripts, but I think the contents are irrelevant in generating the service names. A service name is just a filename. Actually, I think the existing fixture `_comp_compgen_xinetd_services` can be used. I pushed an example implementation 0adde50b3.

Note: I'm not sure if it is a good idea to embed a code for test in the first place, but this is the way in which `_comp_compgen_xinetd_services` has been tested. It seems to have been introduced in ea1cfa1da4.

test/t/test_service.py
```diff
@@ -2,12 +2,19 @@
...
-    def test_1(self, completion):
-        assert completion
+    def test_1(self, bash):
+        with bash_env_saved(bash) as bash_env:
+            bash_env.write_variable(
+                "_comp__test_inetd_dir",
+                "%s/_comp_compgen_xinetd_services/xinetd.d" % bash.cwd,
+            )
+            completion = assert_complete(bash, "service ")
+            assert completion
```
scop: Maybe with this, assert what we "know" are in the fixture dir? Pre-approved & good to merge with this if you agree.
```suggestion
            assert all(x in completion for x in ("arp", "ifconfig"))
```

yedayak:
Thanks both of you, rebased with your changes

scop:
rumdl fail is unrelated.

PR #1543, opened 2026-01-13
chore: switch from pymarkdown to rumdl

scop:


README.md
```diff
@@ -323,13 +323,13 @@ A. If your `command_not_found_handle` function is not intended to address
...
+   > We recommended checking whether `COMP_LINE` is _set_, which still works if
    > properly done with e.g. `[[ ${COMP_LINE+set} ]]`.  However, if you have
-   > been checking if `COMP_LINE` is *non-empty* with `[[ ${COMP_LINE:-} ]]`,
-   > it may fail to detect programmable completion with the setting `complete
-   > -E` (in Bash >= 4.1) because `COMP_LINE` can be empty in this context.  It
-   > is safer to test `COMP_POINT` as one does not need to care about the
-   > differences between the set and non-empty states of variables.
+   > been checking if `COMP_LINE` is _non-empty_ with `[[ ${COMP_LINE:-} ]]`,
+   > it may fail to detect programmable completion with the setting
+   > `complete -E` (in Bash >= 4.1) because `COMP_LINE` can be empty in this
```
akinomyoga, line 330: Does this imply that `rumdl` doesn't allow breaking line breaking `` `...` ``? The formatting inside GNU Emacs `markdown-mode` behaves as if line breaking can happen inside the inline code span `` `...` ``. We know that GFM allows breaking lines inside inline code spans, and [CommonMarks seems to also allow it](https://spec.commonmark.org/0.31.2/#example-337).

scop: I no longer remember, could be, but I also think it could have also been confused with the -E starting a line. Anyway, I think the modified version reads better, so I'm not concerned about the workaround.

akinomyoga: > I also think it could have also been confused with the -E starting a line.

OK, I also sometimes experienced an option name at the beginning of a line causing an issue in Emacs, though I'm not sure about the exact condition under which the problem happens (I couldn't reproduce the behavior now in Emacs).

> Anyway, I think the modified version reads better, so I'm not concerned about the workaround.

OK. I just asked, so that's fine.

PR #1553, opened 2026-01-25
refactor: use `_comp_compgen -C` for service name generation

akinomyoga:
5e6dc86817f9d011223419e3643e937a5fefd349 and 655371b09ff9ba92e90ffa38847085261d8f34d2 use `_comp_compgen -C "$sysvdir" -- -f` instead of first generating filenames with `"$svsvdir"/*` and stripping the parent directory name by `"{tmp[@]#$sysvdir/}"`.

e21dbb87f and 354f9c522 are additional refactoring.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2216,7 +2216,7 @@ _comp_compgen_services()
     _comp_sysvdirs || return 1
 
     local services
-    _comp_expand_glob services '${sysvdirs[0]}/!($_comp_backup_glob|functions|README)'
+    _comp_compgen -v services -C "${sysvdirs[0]}" -- -f -X "@($_comp_backup_glob|functions|README)"
 
     local _generated=$({
         systemctl list-units --full --all ||
@@ -2230,7 +2230,7 @@ _comp_compgen_services()
     fi
 
     ((${#services[@]})) || return 1
-    _comp_compgen -U services -U sysvdirs -- -W '"${services[@]#${sysvdirs[0]}/}"'
+    _comp_compgen -U services -U sysvdirs -- -W '"${services[@]}"'
 }
 
 # This completes on a list of all available service scripts for the
--- a/bash_completion
+++ b/bash_completion
@@ -2209,28 +2209,30 @@ _comp_compgen_xinetd_services()
 
 # This function completes on services
 #
+# @return 0 if at least one completion is generated, or 1 otherwise.
 # @since 2.12
 _comp_compgen_services()
 {
     local sysvdirs
     _comp_sysvdirs || return 1
 
-    local services
-    _comp_compgen -v services -C "${sysvdirs[0]}" -- -f -X "@($_comp_backup_glob|functions|README)"
+    local status=1
+    _comp_compgen -U sysvdirs -C "${sysvdirs[0]}" -- -f -X "@($_comp_backup_glob|functions|README)" &&
+        status=0
 
     local _generated=$({
         systemctl list-units --full --all ||
             systemctl list-unit-files
     } 2>/dev/null |
         _comp_awk '$1 ~ /\.service$/ { sub("\\.service$", "", $1); print $1 }')
-    _comp_split -la services "$_generated"
+    _comp_compgen -a split -l -- "$_generated" && status=0
 
     if [[ -x /sbin/upstart-udev-bridge ]]; then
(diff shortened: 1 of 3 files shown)


bash_completion
```diff
@@ -2199,22 +2199,18 @@ _comp_compgen_services()
...
-    _comp_split -la services "$_generated"
+        _comp_awk 'sub(/\.service$/, "", $1) { print $1 }')
+    _comp_compgen -a split -l -- "$_generated"
 
     if [[ -x /sbin/upstart-udev-bridge ]]; then
-        _comp_split -la services "$(initctl list 2>/dev/null | cut -d' ' -f1)"
+        _comp_compgen -a split -l -- "$(initctl list 2>/dev/null | cut -d' ' -f1)"
     fi
-
-    ((${#services[@]})) || return 1
```
scop, line 2232: On a quick skim it seems removing this changes the behavior of `_comp_compgen_services` so that it no longer returns 1 when there are no services found. We don't seem to rely on that behavior in any in-tree uses, but should we not preserve it as the function is part of the API?

bash_completion
```diff
@@ -2199,22 +2199,18 @@ _comp_compgen_services()
...
-    _comp_expand_glob services '${sysvdirs[0]}/!($_comp_backup_glob|functions|README)'
+    _comp_compgen -U sysvdirs -C "${sysvdirs[0]}" -- -f -X "@($_comp_backup_glob|functions|README)"
 
     local _generated=$({
         systemctl list-units --full --all ||
             systemctl list-unit-files
     } 2>/dev/null |
-        _comp_awk '$1 ~ /\.service$/ { sub("\\.service$", "", $1); print $1 }')
-    _comp_split -la services "$_generated"
+        _comp_awk 'sub(/\.service$/, "", $1) { print $1 }')
```
scop: I think simple edits like this are usually done using sed instead of awk, maybe do that here?

bash_completion
akinomyoga: OK, I checked the uses of this function inside bash-completion, but I didn't care about the possibility of external dependencies.

I updated it: [`c4c79ad..a02d487`](https://github.com/scop/bash-completion/compare/c4c79ad91e470fb21276f59768e966ff999b31fa..a02d4877ff817957e5e90ee0eefc862e207d3057).

bash_completion
akinomyoga: The non-trivial part is the extraction of `$1`. We may still implement it in sed as follows, but I feel it's worse.

```suggestion
        command sed -n 's/^[[:blank:]]*\([^[:blank:]]\{1,\}\)\.service\([[:blank:]].*\)\{0,1\}$/\1/p'
```

Another option is to use both, but I'd rather do it in a single invocation of `awk`.

```suggestion
        _comp_awk '{print $1}' | command sed -n 's/\.service$//p'
```

----

However, I noticed in the output of `systemctl list-units --full --all` in my system that the service name isn't necessarily included as `$1`. Running services are prefixed by `●`. For example,

```console
$ systemctl list-units --full --all | grep -F iptables.service
● iptables.service                                       not-found inactive   dead      iptables.service
```

The current implementation fails to extract those lines.

<details><summary>Alternative implementations are ...</summary>

```suggestion
        _comp_awk '{ for (i = 1; i <= 2; i++) if (sub(/\.service$/, "", $i)) { print $i; next } }')
```

```suggestion
        _comp_awk '{ if (sub(/\.service$/, "", $1)) print $1; else if (sub(/\.service$/, "", $2)) print $2 }')
```

```suggestion
        _comp_awk 'sub(/\.service$/, "", $1) { print $1; next } sub(/\.service$/, "", $2) { print $2 }')
```

```suggestion
        command sed -n 's/^\([^[:blank:]]\{1,\}[[:blank:]]\)\{0,1\}[[:blank:]]*\([^[:blank:]]\{1,\}\)\.service\([[:blank:]].*\)\{0,1\}$/\2/p')
```

```suggestion
        _comp_awk 'sub(/\.service([[:blank:]].*)?$/, "") && NF <= 2 { print $NF }')
```

</details>

akinomyoga: I decided to use this: [`a02d487..8082cf8`](https://github.com/scop/bash-completion/compare/a02d4877ff817957e5e90ee0eefc862e207d3057..8082cf8a78a40d08ea89d5c844112898cb30a9bb)

scop (approved):

PR #1554, opened 2026-01-25
perf(_comp_compgen_filedir_xspec): avoid double negation `!(!(...))`

akinomyoga:


akinomyoga:
> Some example numbers would be interesting to see (also in the related commit message), if you have some readily available.

Ah, sorry. I have to clarify that I haven't faced the actual performance issue in `bash-completion`.

I just know that Bash's implementation of `fnmatch` doesn't use a good algorithm, and also that even if one uses a good one, the negation pattern matching is heavy in general, either in time complexity (with an NFA implementation) or in spatial complexity (with a DFA implementation). There are many hypothetical cases where the negation becomes very slow in Bash, but I'm not sure if any patterns actually used by `bash-completion` trigger the problem.

This isn't an actual pattern used in `bash-completion`, but e.g., `+(0)+(?).txt` becomes slow with the following case. I here compare the times of `@(...)` and `!(!(...))`.

```console
$ mkdir tmp
$ cd tmp
$ touch $(printf "foo%0100dbar.txt")
$ time bash -O extglob -c 'echo @(+(0)+(?).txt)'
0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000.txt

real    0m0.003s
user    0m0.000s
sys     0m0.003s
$ time bash -O extglob -c 'echo !(!(+(0)+(?).txt))'
0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000.txt

real    0m0.991s
user    0m0.988s
sys     0m0.002s
```

However, the length of filenames that cause perceptible decays largely depends on the pattern, so one cannot say even a typical order of time or the filename lengths. For example, if one uses an inefficient pattern `+(+(0)).txt`, the problem arises with relatively short filenames:

```console
$ mkdir tmp2
$ cd tmp2
$ touch $(printf "foo%020dbar.txt")
$ time bash -O extglob -c 'echo @(+(+(0)).txt)'
00000000000000000000.txt

real    0m0.003s
user    0m0.001s
sys     0m0.002s
$ time bash -O extglob -c 'echo !(!(+(+(0)).txt))'
00000000000000000000.txt

real    0m4.537s
user    0m4.529s
sys     0m0.003s
```

akinomyoga:
I added an explanation in the commit message.

PR #1556, opened 2026-01-25
test(env): exclude _omp_*

scop:
Ref https://ohmyposh.dev,
https://github.com/JanDeDobbeleer/oh-my-posh/blob/572c99b2ec6a57a18d4a911cf8d577805002811e/src/shell/scripts/omp.bash#L11-L18

### DIFF
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -746,7 +746,8 @@ def diff_env(before: List[str], after: List[str], ignore: str):
         if not re.search(r"^(---|\+\+\+|@@ )", x)
         # Ignore variables expected to change:
         and not re.search(
-            r"^[-+](_|PPID|BASH_REMATCH|(BASH_)?LINENO)=",
+            # _omp_*: https://ohmyposh.dev
+            r"^[-+](_|PPID|BASH_REMATCH|(BASH_)?LINENO|_omp_.+)=",
             x,
             re.ASCII,
         )


akinomyoga:
I'm curious about the situation where this matters during the test process. In my understanding, we initialize the Bash sessions in our test with `test/config/bashrc`, where we do not source external settings. Is there a case where OMP affects the test Bash sessions (e.g., through exported functions, `BASH_ENV`, `ENV` &amp; `POSIXLY_CORRECT`, etc., though I'm not sure if they can actually be used to inject settings in the test sessions)? Or is there a future plan to load external settings during the test process?

test/t/conftest.py
```diff
@@ -746,7 +746,8 @@ def diff_env(before: List[str], after: List[str], ignore: str):
         if not re.search(r"^(---|\+\+\+|@@ )", x)
         # Ignore variables expected to change:
         and not re.search(
-            r"^[-+](_|PPID|BASH_REMATCH|(BASH_)?LINENO)=",
+            # _omp_*: https://ohmyposh.dev
+            r"^[-+](_|PPID|BASH_REMATCH|(BASH_)?LINENO|_omp_.+)=",
```
akinomyoga, line 750: Also, to reduce the size of the output, I think environment variables are better to be filtered in `_comp__test_get_env` (`test/config/bashrc`):

https://github.com/scop/bash-completion/blob/c4a8fd00cf7ccfb87a211b6ed2db3dd250c407b3/test/config/bashrc#L79

`_`, `PPID`, `BASH_REMATCH`, `BASH_LINENO`, and `LINENO` are excluded in `conftest.py` because they are special variables that cannot be `unset`. Technically, `BASH_REMATCH` can be unset in recent versions of Bash, but it can't in Bash < 5.1.

scop: Re the original question, TBH I did not research the "why" nor can I answer it offhand. It's just something I saw when doing a `pytest test/t/unit/test_unit_compgen_filedir_xspec.py` while having the branch related to that change checked out and oh-my-posh active, every time. I can't reproduce right now on main though, so drafting for now and will reopen if I see it again.

akinomyoga: OK, I know that *apparently* strange things sometimes happen. There could be a specific setup that OMP affects the test sessions, though I don't have an idea for now.

scop:
Haven't come across this again, closing.

PR #1536, opened 2025-12-31
fix: miscellaneous fixes after separation of directories

akinomyoga:
I described the details in the respective commit messages.

PR #1558, opened 2026-01-25
feat: support the `startup{,-core}` directory

akinomyoga:
For startup scripts discussed in the separation of directories #1329.


README.md
```diff
@@ -142,9 +142,8 @@ A. Install a local completion of your own appropriately for the desired
    command, and it will take precedence over the one shipped by us. See the
    next answer for details where to install it, if you are doing it on per user
    basis. If you want to do it system wide, you can install eagerly loaded
-   files in `compatdir` (see a couple of questions further down for more
-   info. To get the path of `compatdir` for the current system, the output of
-   `pkg-config bash-completion --variable compatdir` can be used) and install a
+   files in `<startupdir>`, whose value in the current system can be retrieved
+   by `pkg-config bash-completion --variable=startupdir`, and install a
    completion for the commands to override our completion for in them.
```
yedayak: Maybe this should be split up? Installing system wide doesn't necessarily imply eagerly loaded.
Perhaps `How can I override...` and `How to load completions eagerly on startup?`

bash_completion
```diff
@@ -3582,6 +3582,30 @@ _comp__init_collect_startup_configs()
...
+        [[ -d $startup_dir && -r $startup_dir && -x $startup_dir ]] || continue
+        _comp_expand_glob startup_files "$startup_dir/*.bash" || continue
+        for startup_file in "${startup_files[@]}"; do
+            [[ ! -d $startup_file && -r $startup_file ]] || continue
+
+            # Disable the startup files with the same name as one of already
+            # selected startup files.  This allows users to override a startup
+            # file with another file with the same name placed in an earlier
+            # directory in "${startup_files[@]}".
+            local name=${startup_file##*/?([0-9][0-9][0-9]_)}
```
yedayak, line 3691: Why is this processing needed?

startup/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+This directory `startup` is the place to put **third-party startup scripts**
+for initializing [bash-completion](https://github.com/scop/bash-completion).
+
+The `bash-completion` framework has been providing `/etc/bash_completion.d` as
+a directory to put scripts that are eagerly loaded on the initialization stage
+of `bash-completion`.  However, the historical purpose of
+`/etc/bash_completion.d` has been the place to put completion files for
+`bash-completion < 2.0`, which did not support the dynamic loading through
+`complete -D` (`bash >= 4.1`).  This original usage was already deprecated.
```
yedayak:
```suggestion
`bash-completion < 2.0`, which did not support dynamic loading through
`complete -D`.  This original usage was already deprecated.
```
I don't think it's relevant why it didn't support it

yedayak (commented):
Can you explain the reason for keeping track of what we loaded to allow for overriding? Is there a specific use-case you're thinking about?

README.md
akinomyoga: Thank you for your review.

Yeah, that makes much more sense. It's been mixed, so I just updated locally, but I agree to separate the question.

bash_completion
akinomyoga: I think this comment discusses the same thing as the following cover comment, so let me answer together in this sub-thread.

> Can you explain the reason for keeping track of what we loaded to allow for overriding? Is there a specific use-case you're thinking about?

The original motivation that I thought of this mechanism is #1399. Although the OP of #1399 didn't explain or show a reference to any technical details that they tried to use to reason the PR, I found [the UAPI spec](https://uapi-group.org/specifications/specs/configuration_files_specification/) and other materials myself. They explain the usage of `/etc/<app>` and `/usr/etc/<app>` (or `/usr/lib/<app>`, `/usr/share/<app>`, or whatever). I'm neutral on whether to adopt their specific use of the specific directories, `/etc` and `/usr/etc`, but I found the underlying idea useful.

Meanwhile, I've been thinking it is very annoying that some distribution packages put in `/etc/bash_completion.d` completion settings that affect the global behavior of completion, such as `fzf` and `_python-argcomplete`. For example, even if I don't intend to use `fzf` for Bash completion, when some application packages require the `fzf` package as a dependency in order to provide a menu for their UI, the fzf completion is unintentionally turned on through `/etc/bash_completion.d/fzf` installed by dependency. I believe the `fzf` package and the package for the fzf-bash integration should be separated ideally, but I guess the package maintainers of `fzf` consider `fzf` a tool for shell completion (rather than a general selector).

This also becomes an issue in a shared system, for which I do not have the administrator privilege. If there are any files in `/etc/bash_completion.d/*`, they are all forcibly loaded. Even if I want to disable any of them *partially*, there is *no way* to realize that. Actually, I ended up doing [a very ugly hack of overriding the `source` builtin and the dot (`.`) builtin](https://github.com/akinomyoga/akinomyoga.dotfiles/blob/4e50ab09b337fec5c3dae07105db3dd7489c8b24/bashrc#L171-L184), but I would rather avoid such a hack if possible. Then, if those system-wide startup completion settings are moved to `/usr/share/bash-completion/startup/<xxx>.bash`, and if bash-completion provides a way to disable each setting by putting an empty file at e.g. `~/.local/share/bash-completion/startup/<xxx>.bash` (or a way to override the behavior by putting a non-empty file), it will be useful.

Even if we forget about those arguments, when different versions of the same application install the same name of startup completion files at different paths, such as `/usr/share/bash-completion/startup/xyz.bash`, `/usr/local/share/bash-completion/startup/xyz.bash`, and `~/.local/share/bash-completion/startup/xyz.bash`, I doubt we should source all of them. I think it would make more sense to pick up one of them, particularly the one installed by the current user.

> Why is this processing needed?

If you mean by "*this*" the entire processing described by the code comment, I think I answered it above.

If you mean by "*this*" particularly the stripping of `?([0-9][0-9][0-9]_)`, it is to remove the prefix used to control the loading order. We already have `000_` prefixed to our `000_bash_completion_compat.bash`, so other startup settings may also be prefixed by numbers. However, such a prefix number isn't controlled by the upstream project of the application, but the package distributions, or the system administrator may also change the number. Thus, when multiple startup completions of the same application are installed to different paths, they are not guaranteed to have the matching prefix numbers (if any). Then, the prefix numbers used to control the loading order should be ignored in identifying the initialization of the same application.

startup/README.md
akinomyoga: > I don't think it's relevant why it didn't support it

I wanted to describe the migration from `/etc/bash_comletion.d`, particularly about what type of files should be moved from `/etc/bash_completion.d` to `/usr/share/bash-completion/startup` and what type of files should not. To enlighten that two different types of files are actually mixed in `/etc/bash_completion.d`, I wanted to explain the historical background.

Or if you specifically talk about the reason to include the Bash version that introduced `complete -D`, there is no strong reason. I don't remember by heart the version in which each Bash feature was introduced, but I always want to care about Bash version compatibility. I can't immediately answer with the 100% confidence whether each Bash feature is ensured to exist in the Bash versions that `bash-completion` currently covers. In short, I can say I included the version information as a reminder to *myself*. Is there a reason that we shouldn't include the version information?

bash_completion
akinomyoga: Maybe the name "eagerly loaded settings" is not a good description for startup files that can be overridden by users?

startup/README.md
```diff
@@ -0,0 +1,26 @@
+# Third-party startup directory - bash-completion
```
akinomyoga:
```suggestion
# External startup directory - bash-completion
```

I forgot to update the README contents following [the suggestion in the previous PR](https://github.com/scop/bash-completion/pull/1536#discussion_r2657919344). The body text also contains some "third-party".

startup/README.md
yedayak: I was saying that for the users all that matters is that those old bash-completion versions didn't support it, but I guess its fine, gives more context

bash_completion
akinomyoga: I realized I haven't considered users' directories in looking up the users' startup files. I added the support for it in commit a45e2ab0b. I'll squash this commit into the first commit later.

startup/README.md
akinomyoga: [`8161db8..19a1302`](https://github.com/scop/bash-completion/compare/8161db842855462de20224c865f8a38c6a542eb7..19a130216b4ca7e61aa47cbbff143f3b7c33362a)

startup/README.md
akinomyoga: [`8161db8..19a1302`](https://github.com/scop/bash-completion/compare/8161db842855462de20224c865f8a38c6a542eb7..19a130216b4ca7e61aa47cbbff143f3b7c33362a)

README.md
akinomyoga: I reorganized related questions and answers. fc6787f01

bash_completion
akinomyoga: I also included the description of this behavior in fc6787f01741689c24abacdd71d607ac3ec5975a.

akinomyoga (commented):
I probably need to add tests.

akinomyoga:
> I probably need to add tests.

I added tests in 4990f89e6a1abcfa502a195ae5ff460df118d90d. I also adjusted the loading order of startup files from different directories in commit 011affa1727d7ca56b8f083380cc9d724a1d8944.

test/config/bashrc
```diff
@@ -57,6 +57,12 @@ unset -v \
     COMP_KNOWN_HOSTS_WITH_HOSTFILE \
     COMP_TAR_INTERNAL_PATHS
 
+if [[ $PWD == */_comp__init_collect_startup_configs ]]; then
+    # To test the loading order and masking of startup files, we specify
+    # directories containing the subdirectory "startup".
+    BASH_COMPLETION_USER_DIR=$PWD/usr:$PWD/sys:$BASH_COMPLETION_USER_DIR
+fi
```
yedayak, line 64: Can you set the variable using bash_env.write_variable, like in https://github.com/scop/bash-completion/blob/main/test/t/unit/test_unit_load.py#L52C13-L56 ?

test/fixtures/_comp__init_collect_startup_configs/user/startup/000_foo.bash
yedayak, line 1: Maybe change the name of the directory from `/usr`? kind of an overloaded name, maybe `local` or `user`?

bash_completion
yedayak: I think this isn't needed anymore since we don't ship `bash_completion.d` in the repo itself anymore

Makefile.am
```diff
@@ -1,11 +1,11 @@
 SUBDIRS = doc completions-core completions-fallback helpers-core \
-	test completions helpers
+	startup-core test completions helpers startup
 
 pkgdata_DATA = bash_completion
 
 # Empty, but here just to get the compat dir created with install
 compatdir = $(sysconfdir)/bash_completion.d
-compat_DATA = bash_completion.d/000_bash_completion_compat.bash
+compat_DATA =
```
yedayak, line 8: We still need to create `${sysconfdir}/bash_completion.d`

test/config/bashrc
akinomyoga: Do you have an idea how to use `bash_env.write_variable` in the present case? To initialize `bash_env`, we need the fixture `bash`. To initialize the fixture `bash`, we source `bash_completion` in `test/config/bashrc`. For the startup-file lookup, the variable is used when `bash_completion` is sourced, so it's too late to set the variable when `bash_env` is available.

test/fixtures/_comp__init_collect_startup_configs/user/startup/000_foo.bash
akinomyoga: Ah, that's a good point. I'll change the name to `user` (as `local` also implies system-wide location `/usr/local`). Probably `sys` should also be updated to `system`.

bash_completion
akinomyoga: Good catch.

Makefile.am
akinomyoga: Do we? #1399 seemed to want us not to create anything in `/etc`. I thought the projects that want to put any file in `/etc/bash-completion.d` can create a directory themselves, but does it actually fail in a typical setup of autoconf? I thought it would automatically create a directory. Is there another reason that we need to create it?

bash_completion
akinomyoga: Thank you for pointing out the existence of the configuration variable, `BASH_COMPLETION_COMPAT_IGNORE`! I haven't thought about using it to select valid startup files rather than buggy ones, (though it's not sufficient to process the UAPI-type overriding and selection from multiple versions of the same startup file).

Makefile.am
akinomyoga: > I thought the projects that want to put any file in `/etc/bash-completion.d` can create a directory themselves, but does it actually fail in a typical setup of autoconf? I thought it would automatically create a directory.

When creating a package, one would do `make install` into a clean directory. So I guess `make install` is rather required to work without any existing directory structures.

bash_completion
akinomyoga: Thank you. I've fixed this in push [`d89b679..5a1f187`](https://github.com/scop/bash-completion/compare/d89b6791c388d9a46203013d13c95ac07039bf48..5a1f1872d6e3794360940eca63971c94dfa5a884).

test/fixtures/_comp__init_collect_startup_configs/user/startup/000_foo.bash
akinomyoga: I fixed in the push [`d89b679..5a1f187`](https://github.com/scop/bash-completion/compare/d89b6791c388d9a46203013d13c95ac07039bf48..5a1f1872d6e3794360940eca63971c94dfa5a884). I decided to change `usr => user` and `sys => host`.

test/config/bashrc
yedayak: Yeah that's a very good point :)

Makefile.am
yedayak: Yeah that's a good point, any other packages that install there will create the directory there themselves, especially since they can be installed even if bash-completion isn't.
My only worry is for users following guides that tell them to put the completions in `/etc/bash-completion.d/`. Maybe this calls for another README :upside_down_face: 
I think creating the directory will be fine from the perspective of the UAPI concerns, since we're not dependent anymore on any files in that directory for a fully working system. Maybe someone from the original PR wants to chime in? @septatrix @bluca

bash_completion
```diff
@@ -3584,7 +3584,53 @@ _comp__init_collect_startup_configs()
...
+        ((${#selected_startup_files[@]})) || continue
+
+        # Note: bash < 4.4 has a bug that all elements are connected with
+        # ${a[@]+"${a[@]}"} when IFS does not contain whitespace.
+        local IFS=$' \t\n'
+        # Prepend the selected startup files because we want to source the
+        # system startup files earlier.
+        _comp__init_startup_configs=(
+            "${selected_startup_files[@]}"
+            ${_comp__init_startup_configs[@]+"${_comp__init_startup_configs[@]}"}
```
yedayak, line 3713: Side note: I see we have this pattern in other places, but I don't understand why it's needed, doesn't just doing 
```
_comp__init_startup_configs=( "${selected_startup_files[@]}" ${_comp__init_startup_configs[@]} )
```
work?

akinomyoga: Bash 4.2 has a bug that `"${arr[@]}"` fails with `set -u` for `arr=()`.

Makefile.am
akinomyoga: > My only worry is for users following guides that tell them to put the completions in `/etc/bash-completion.d/`. Maybe this calls for another README 🙃

I think I already removed that part with [`6a78aa3`](https://github.com/scop/bash-completion/pull/1558/changes/6a78aa3bbd605f6a3415a959906cce1671578614#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5). Is there still a place to modify?

bash_completion
scop: Just a random off-context comment, as I haven't been following this MR yet, ref https://github.com/scop/bash-completion/discussions/1479#discussioncomment-15273385: I think we should drop support for bash < 5.0 (or _at least < 4.3_) for the next release.

akinomyoga: So they can finally be removed, but I believe it should be done in a separate PR for consistency (and for later research in case something happens). In the first place, the code doesn't pass the CI tests without the workaround in the current setup of the test environment.

Makefile.am
yedayak: > > My only worry is for users following guides that tell them to put the completions in `/etc/bash-completion.d/`. Maybe this calls for another README 🙃
> 
> I think I already removed that part with [`6a78aa3`](https://github.com/scop/bash-completion/pull/1558/changes/6a78aa3bbd605f6a3415a959906cce1671578614#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5). Is there still a place to modify?

I mean guides outside of the repo, for example this stack overflow question: https://askubuntu.com/questions/68175/how-to-create-script-with-auto-complete, or the [cobra13 docs](https://cobra.dev/docs/how-to-guides/shell-completion/). Maybe adding a bit more friction for doing the "wrong thing" does make sense though.

akinomyoga: I see. It can sometimes take a long time to update the community's knowledge. Until recently, there have been many articles recommending the use of the old-style command substitutions `` `...` `` that were ~~deprecated~~discouraged by the standard about 30 years ago.

bash_completion
akinomyoga: I added a clarification in [`71ae643..6c1af0e`](https://github.com/scop/bash-completion/compare/71ae6438b15f57be261336d8cba44d521081f7b8..6c1af0ef29bc59445d31218c598068ad4fb14a2a).

I also considered adding the same comments to other cases, but I gave it up because there seem to be many, and I didn't feel that we needed to bother with what would soon be removed for the version requirement change. There have been 13 cases, and none of them had comments:

```bash
./bash_completion:975:    _comp_compgen -F "$_ifs" -U input -- ${_compgen_options[@]+"${_compgen_options[@]}"} -W '$input'
./bash_completion:1101:    local "$2" "$3" "$4" && _comp_upvars -a"${#words[@]}" "$2" ${words[@]+"${words[@]}"} \
./bash_completion:1379:    _comp_compgen -U toks set ${toks[@]+"${toks[@]}"}
./bash_completion:2657:    _comp_compgen_known_hosts ${options[@]+"${options[@]}"} -- "$cur"
./bash_completion:3413:    elif . "$compfile" "$cmd" ${source_args[@]+"${source_args[@]}"}; then
./completions-core/gdb.bash:32:                find ${path_array[@]+"${path_array[@]}"} . -name . -o \
./completions-core/ri.bash:23:            "$(ri ${classes[@]+"${classes[@]}"} 2>/dev/null | ruby -ane \
./completions-core/ri.bash:31:            "$(ruby -W0 "$ri_path" ${classes[@]+"${classes[@]}"} 2>/dev/null | ruby -ane \
./completions-core/strace.bash:60:                                '${syscalls[@]+"${!syscalls[@]}"} file process
./completions-core/update-rc.d.bash:20:        _comp_compgen -- -W '"${options[@]}" ${services[@]+"${services[@]}"}' \
./startup-core/000_bash_completion_compat.bash:338:        args=(-c "$REPLY" ${opt[@]+"${opt[@]}"})
./startup-core/000_bash_completion_compat.bash:362:        args=(-c "$REPLY" ${opt[@]+"${opt[@]}"})
./test/t/unit/test_unit_compgen.py:22:            '_comp__test_words() { local -a input=("${@:1:$#-1}"); _comp__test_compgen -c "${@:$#}" -- -W \'${input[@]+"${input[@]}"}\'; }',
```

README.md
akinomyoga: @yedayak Could you check my replies? Or if you are already satisfied, could you confirm that or close the conversation? The same for all the following discussions.

README.md
```diff
@@ -136,48 +136,85 @@ A. No. If needed just once in a while,
...
-
-A. [ Disclaimer: Here, how to make the completion code visible to
-   bash-completion is explained.  We do not require always making the
-   completion code visible to bash-completion.  In what condition the
-   completion code is installed should be determined at the author/maintainers'
-   own discretion. ]
+**Q2. How can I override a completion shipped by bash-completion or install a
+   new completion for a user account?**
+
+A. To override a completion for a specific command, you can simply install your
```
yedayak:
```suggestion
A. To override a completion for a specific command, you can install your
```

startup/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+In `bash-completion >= 2.18`, we started to provide the `startup` directory as
+a dedicated directory for such scripts that need to be loaded on the
+initialization stage.
+
+* Eagerly loaded scripts should be placed in this directory instead of
+  `/etc/bash_completion.d`.
+* External completion files of API v1 should be updated to use API v2 or v3 and
+  should be moved to `completions/<cmd>.bash`.  Although we recommend to update
+  the completion file to use newer API, the completion file relying on API v1
```
scop:
```suggestion
  the completion file to use newer API, completion files relying on API v1
```

README.md
```diff
@@ -136,48 +136,85 @@ A. No. If needed just once in a while,
...
+   Completion settings that should be loaded on initialization of
+   bash-completion should be defined in a startup completion file.  A startup
+   completion file can be added in the directory `<userdir>/startup` (see the
+   previous paragraph for `<userdir>`).  A prefix of the form
+   `[0-9][0-9][0-9]_` may be prepended to the filename of the startup files to
+   control the loading order.  To override a existing startup file installed
+   systemwide or shipped by us, you can put your own version with the same
+   filename in `<userdir>/startup`.  In identifying the overridden startup
+   file, the prefixes `[0-9][0-9][0-9]_` are ignored.  If you want to disable a
+   startup completion file, you can put an empty file.
```
scop:
```suggestion
   startup completion file, you can use an empty file.
```

bash_completion
```diff
@@ -3660,7 +3660,60 @@ _comp__init_collect_startup_configs()
...
+    # collect startup files
+    local -A startup_visited=()
+    for startup_dir in "${startup_dirs[@]}"; do
+        [[ -d $startup_dir && -r $startup_dir && -x $startup_dir ]] || continue
+
+        local -a startup_files=()
+        _comp_expand_glob startup_files "$startup_dir/*.bash" || continue
+
+        local -a selected_startup_files=()
+        local startup_dir
```
scop: I think this `local` should be further above, before the loop, to avoid a leak.

scop (commented):
Did not go through with a fine comb, but makes sense and LGTM with a few small comments.

Good to merge as far as I'm concerned after @yedayak has found time to check feedback to his earlier findings.

bash_completion
akinomyoga: Thank you. This seems a mistake in duplicating the `for`-loop for `startup_dir` and `startup_file`. `local startup_dir` must be placed earlier as suggested, and here we must have `local startup_file` instead.

startup/README.md
akinomyoga: [`58bfc42..754401b`](https://github.com/scop/bash-completion/compare/58bfc4223462492bbb4902d2c9933b818cdcf176..754401b104f304e5431f1f7e996a8a8823ff84a9)

README.md
akinomyoga: [`58bfc42..754401b`](https://github.com/scop/bash-completion/compare/58bfc4223462492bbb4902d2c9933b818cdcf176..754401b104f304e5431f1f7e996a8a8823ff84a9)

bash_completion
akinomyoga: [`58bfc42..754401b`](https://github.com/scop/bash-completion/compare/58bfc4223462492bbb4902d2c9933b818cdcf176..754401b104f304e5431f1f7e996a8a8823ff84a9)

README.md
yedayak:
```suggestion
A. To override a completion for a specific command, you can install your
```

yedayak (approved):

README.md
akinomyoga: Thank you for checking. I've applied the change. [`754401b..b09700d`](https://github.com/scop/bash-completion/compare/754401b104f304e5431f1f7e996a8a8823ff84a9..b09700dda29b56fb89812e1215956a631cc7c848)

akinomyoga:
Thank you for the reviews. I'll later merge this after waiting for a few days.

akinomyoga:
I merged the PR, but I realized that the README is not rendered as expected.

> image

akinomyoga:
> but I realized that the README is not rendered as expected.

Since this is a trivial mistake, I've directly pushed the fix b8e7fe62a to the `main` branch.

PR #1564, opened 2026-01-31
fix: fix non-POSIX sed expressions

akinomyoga:


completions-core/invoke-rc.d.bash
```diff
@@ -17,9 +17,12 @@ _comp_cmd_invoke_rc_d()
 
     if [[ $cword -eq 1 || $prev == --* ]]; then
         # generate valid_options
+        local IFS='|'
+        local regex_options="${options[*]}"
+        IFS=$' \t\n'
         _comp_compgen_split -- "$(
             tr " " "\n" <<<"${words[*]} ${options[*]}" |
-                command sed -ne "/$(command sed 's/ /\\|/g' <<<"${options[*]}")/p" |
+                command grep -E "$regex_options" |
```
scop:
```suggestion
                command grep -E -- "$regex_options" |
```

`--`  works at least with the GNU, busybox, and FreeBSD grep versions I tried.

test/runLint
```diff
@@ -33,6 +33,9 @@ gitgrep "$cmdstart"'sed\b.*\\[?+]' \
 gitgrep "$cmdstart"'sed\b.*\\\|' \
     "sed with \|, use POSIX BRE (possibly multiple sed invocations) or another tool instead"
 
+gitgrep '('"$cmdstart"'sed\b| =~ ).*\\[<>]' \
+    "\< or \> in regular expressions. Anchors \< and \> are not defined by POSIX.  No POSIX alternatives."
```
scop, line 37: I wonder if we should check this for grep and awk, too.

completions-core/invoke-rc.d.bash
akinomyoga: Ah, right! Or, actually, the option `-e regexp` exists for exactly this kind of situation.

test/runLint
akinomyoga: I added grep and awk in the push [`75e5fd7..2d8b329`](https://github.com/scop/bash-completion/compare/75e5fd771fb5e0e4b1624950a9db5036c9a181ea..2d8b3299eaebb9fc02dfab9b6e1f222250821b39). I also added a check for `\b`.

completions-core/invoke-rc.d.bash
akinomyoga: [`75e5fd7..2d8b329`](https://github.com/scop/bash-completion/compare/75e5fd771fb5e0e4b1624950a9db5036c9a181ea..2d8b3299eaebb9fc02dfab9b6e1f222250821b39).

akinomyoga: I modified this part by looking only at how the regular expression is used, but I now realized that the entire processing here tries to generate the elements of `"${options[@]}"` that haven't yet appeared in `"${words[@]}"`. In the current implementation, when an element in `words` contains any of `"${options[@]}"` as a substring, the element in `words` is also generated as a candidate, which seems strange. If we change this behavior, we can significantly simplify the entire processing as

```bash
local IFS='|'
_comp_compgen -- -W '"${options[@]}"' -X "@(${words[*]})"
```

completions-core/iptables.bash
```diff
@@ -27,17 +27,17 @@ _comp_cmd_iptables()
                 _comp_compgen -- -W '$targets'
                 _comp_compgen -a split -- "$("$1" ${table:+-t "$table"} -nL \
                     2>/dev/null | command sed -ne "$chain" \
-                    -e 's/INPUT|OUTPUT|FORWARD|PREROUTING|POSTROUTING//')"
+                    -e 's/INPUT//;s/OUTPUT//;s/FORWARD//;s/PREROUTING//;s/POSTROUTING//')"
```
scop, line 30: Just a general note, the replacement is not equivalent -- the former would have replaced the first occurrence of the alternatives, while the latter replaces first occurrence of _each_ alternative. Maybe not a problem here.

But having had a look at this, I wonder if this has ever worked the way it was intended, as the latter expression (which was modified here) is a no-op both before and after the changes. To have any effect it should be for example in another sed invocation altogether, like `sed -ne "$chain" | sed -e 's/INPUT...`.

And even with that, I'm not sure it still does what was intended, for example with chains named `DOCKER-FORWARD` present.

Maybe `command sed -ne "$chain" | command grep -vEx 'INPUT|OUTPUT|FORWARD|PREROUTING|POSTROUTING'`?

And similarly for others below in this file and in ebtables.

akinomyoga: OK, that is a good point. In this PR, I just tried to fix the wrong/non-portable usage of `sed`, but I haven't tried to examine the behavior. Actually, `_comp_compgen_split` accepts an option `-X '@(INPUT|OUTPUT|FORWARD|PREROUTING|POSTROUTING)'`.

completions-core/mutt.bash
```diff
@@ -101,7 +101,7 @@ _comp_cmd_mutt__query()
     [[ $cur ]] || return 0
     local muttcmd=${words[0]}
 
-    local querycmd="$("$muttcmd" -Q query_command 2>/dev/null | command sed -e 's|^query_command=\"\(.*\)\"$|\1|' -e 's|%s|'"$cur"'|')"
+    local querycmd="$("$muttcmd" -Q query_command 2>/dev/null | command sed -e 's|^query_command="\(.*\)"$|\1|' -e 's|%s|'"$cur"'|')"
```
scop, line 104: I believe these backslashed double quotes were there to escape them from the surrounding quoted `"$( ... )"`. But IDK if that's necessary in the first place, or further if actually quoting the surrounding thing is ever necessary (i.e. if it could be written just as `$( ... )` instead of `"$( ... )"`). But if so, let's look into addressing/cleaning up that in another PR.

completions-core/iptables.bash
scop: I'm fine with leaving fixing that to another PR (maybe even better that way, but no strong opinions as long as we remember to fix it).

akinomyoga: ~~389029413~~ This will be discussed in the next PR.

completions-core/invoke-rc.d.bash
akinomyoga: ~~3a22b2130~~ This will be discussed in the next PR.

completions-core/mutt.bash
akinomyoga: The inside of `$(...)` is an independent syntactic context and it's unaffected by the quoting outside of it. ~~I guess the command substitution was originally in the old style `` `...` ``, where the quoting was necessary, but the quoting was left after switching to `$(...)`.~~ I've identified the commit: 755041cb5f, where it seems to have been POSIX'ly-incorrect from the beginning, or it might have been copied from an even older code.

Yeah, the outside double-quoting `"$()"` is also unnecessary on the right-hand side of assignments.

completions-core/iptables.bash
akinomyoga: > I'm fine with leaving fixing that to another PR (maybe even better that way, but no strong opinions as long as we remember to fix it).

OK, I haven't noticed the message. I noticed the message after reloading the page. I'll drop the change and also the change for https://github.com/scop/bash-completion/pull/1564#discussion_r2749219149 (**edit**: I dropped these).

scop (approved):

scop:
Thanks!

akinomyoga:
Thank you!

completions-core/invoke-rc.d.bash
akinomyoga: > ~[3a22b21](https://github.com/scop/bash-completion/commit/3a22b21308b40c9afff0bd8b44834884058d4a66)~ This will be discussed in the next PR.

#1566

completions-core/iptables.bash
akinomyoga: > ~[3890294](https://github.com/scop/bash-completion/commit/38902941311c3ac49454ff7f71fc599e1ecea0a6)~ This will be discussed in the next PR.

#1566

completions-core/mutt.bash
akinomyoga: #1573

PR #1566, opened 2026-02-02
fix: fix uses of sed/awk scripts

akinomyoga:
These are broken regardless of POSIX.


scop (approved):
LGTM on a quick read, thanks!

GH actions is not feeling great at the moment, good to merge after we get a clean/green CI run, https://www.githubstatus.com/incidents/xwn6hjps36ty

akinomyoga:
> GH actions is not feeling great at the moment, good to merge after we get a clean/green CI run, https://www.githubstatus.com/incidents/xwn6hjps36ty

Oh, OK. Thanks for the information.

PR #1567, opened 2026-02-03
refactor: simplify uses of sed

akinomyoga:


akinomyoga:
Thank you.

> I have a nagging feeling that some POSIX features of basic tools cannot be taken for granted e.g. on Solaris,

`tail -n +2` was already used in many places in the codebase, so I assumed that it runs on a wide environment, but I have a Solaris virtual machine, so I can test it later. As for the `sed` expressions, I have the feeling that Solaris `sed` is compliant (unlike Solaris `awk`), but I haven't checked that the sed expressions actually work in Solaris. Although I'm not sure if I can set up the necessary environment for the full test, I can check the behavior of the sed expressions separately in Solaris at least.

akinomyoga:
I checked the behavior in Solaris 11.3. It turned out that Solaris `tail` doesn't support the option `-n`. It seems to require us to write as `tail +2` instead of `tail -n +2`. However, other compliant implementations of `tail` don't accept Solaris's `tail +2`. As usual, Solaris provides the compliant version at `/usr/xpg4/bin/tail`, which I confirmed accepts `-n +2`.

I haven't checked the behavior of Solaris 11.4 because I don't have a virtual machine currently, but I found an article (written in Japanese [^1]) saying that Solaris 11.4 `/usr/bin/awk` seems to have been updated to be a compliant version. Solaris 11.3 ships with Bash 4.1, and Solaris 11.4 ships with Bash 4.4. This means that we already dropped the support for Solaris 11.3 technically. If we can confirm that Solaris 11.4 `awk` and `tail` are indeed POSIX-compliant, maybe we can drop the `/usr/xpg4` workaround. Or maybe we should still provide the workaround because the Bash version is actually independent of the version of the standard utilities; users can build and install newer Bash versions in old operating systems, while we reference the operating-system version only for surveying the popular Bash versions in the market.

By the way, do we care about the support for Solaris 11.4? Solaris isn't mentioned in #1429 nor in [Minimum bash versions - Wiki](https://github.com/scop/bash-completion/wiki/Minimum-bash-versions). Solaris 11.4 support seems to [continue until 2037](https://www.theregister.com/2024/01/29/oracle_extends_solaris_support/). If we care about it, we may need to continue to support Bash 4.4 until 2037. We probably want to intentionally ignore Solaris 11.4.

[^1]: [なぜawkの乱数関数でUnix時間を取得できるのか？ ～ 乱数の話とUnix時間の深い関係 #UNIX - Qiita](https://qiita.com/ko1nksm/items/9c20739c68d88857ce4c#posix-issue-8-%E3%81%A7%E8%A6%8F%E5%AE%9A%E3%81%95%E3%82%8C%E3%82%8B%E5%8B%95%E4%BD%9C:~:text=Solaris%2011.3%20%E3%81%A7%E3%82%82%20/usr/xpg4/bin/awk%20%E3%81%AB%20POSIX%20awk%20%E3%81%8C%E3%81%82%E3%82%8B%E3%81%AE%E3%81%A7%E3%81%93%E3%81%A1%E3%82%89%E3%82%92%E4%BD%BF%E3%81%A3%E3%81%A6%E3%81%8F%E3%81%A0%E3%81%95%E3%81%84%E3%80%82%E3%81%BE%E3%81%9F%E6%9C%80%E6%96%B0%E3%81%AE%20Solaris%2011.4%20%E3%81%A7%E3%81%AF%20/usr/bin/awk%20%E3%81%AF%20POSIX%20awk%20%E3%81%AB%E6%9B%B4%E6%96%B0%E3%81%95%E3%82%8C%E3%81%A6%E3%81%84%E3%82%8B%E3%82%88%E3%81%86%E3%81%A7%E3%81%99%E3%80%82)

completions-core/7z.bash
```diff
@@ -99,7 +99,7 @@ _comp_cmd_7z()
         if [[ ${words[1]} == d ]]; then
             _comp_compgen_split -l -- "$(
                 "$1" l "${words[2]}" -slt 2>/dev/null | command sed -n \
-                    '/^Path =/s/^Path = \(.*\)$/\1/p' 2>/dev/null | tail -n+2
+                    's/^Path = \(.*\)$/\1/p' 2>/dev/null | tail -n+2
```
scop, line 102: I don't know if it matters at all, just happened to notice now, but in other places we have a space between `-n` and `+2`.

bash_completion
```diff
@@ -2072,12 +2072,12 @@ if [[ $OSTYPE == *@(solaris|aix)* ]]; then
     # This function completes on process IDs.
     _comp_compgen_pids()
     {
-        _comp_compgen_split -- "$(command ps -efo pid | command sed 1d)"
+        _comp_compgen_split -- "$(command ps -efo pid | tail -n +2)"
```
scop, line 2075: Given the info on `tail -n +2` not working on Solaris, this change and the one below are unfortunate in the sense that they are in a `if $OSTYPE` block that specifically targets Solaris (and AIX).

bash_completion
```diff
@@ -2072,12 +2072,12 @@ if [[ $OSTYPE == *@(solaris|aix)* ]]; then
...
     }
 
     _comp_compgen_pgids()
     {
-        _comp_compgen_split -- "$(command ps -efo pgid | command sed 1d)"
```
scop, line 2080: This being the other change mentioned above.

scop:
I think we should care about what we are willing to maintain and can test. Or in other words, what we care about :)

As far as I'm personally concerned, that no longer includes any Solaris version as I don't have access to one nor can I test with one these days. I have no problem if others want to support it as long as it doesn't get too badly on the way of keeping things working on other "more supported" platforms.

completions-core/7z.bash
akinomyoga: Thank you. A fix is in #1572.

bash_completion
akinomyoga: This is also addressed in #1572.

bash_completion
akinomyoga: #1572

akinomyoga:
> I think we should care about what we are willing to maintain and can test. Or in other words, what we care about :)

As mentioned in #1572, for now, I think we can support Solaris 11.3 tail with a simple compatibility wrapper.

PR #1569, opened 2026-02-05
fix: load from more new style completion dirs

scop:
This might not be spot on off the bat, but it restores some of the way we used to load completions from.

In particular, before this, completions-core and completions-fallback from the bash_completion location were not attempted, which is a regression at least for my workflow: I rely on being able to modify files in my bash-completion working dir and have them take precendence over any system installed ones.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -3450,8 +3450,7 @@ _comp_load()
     done
     shift "$((OPTIND - 1))"
 
-    local cmd=$1 cmdname=${1##*/} dir compfile
-    local -a paths
+    local cmd=$1 cmdname=${1##*/}
     [[ $cmdname ]] || return 1
 
     local backslash=
@@ -3469,55 +3468,108 @@ _comp_load()
         cmd=$REPLY
     fi
 
-    local -a dirs=()
+    # Begin lookup
 
-    # Lookup order:
-    # 1) From BASH_COMPLETION_USER_DIR (e.g. ~/.local/share/bash-completion):
-    # User installed completions.
-    if [[ ${BASH_COMPLETION_USER_DIR-} ]]; then
-        _comp_split -F : dirs "$BASH_COMPLETION_USER_DIR" &&
-            dirs+=("${paths[@]/%*(\/)//completions}")
-    else
-        dirs=("${XDG_DATA_HOME:-$HOME/.local/share}/bash-completion/completions")
-    fi
-
-    # 2) From the location of bash_completion: Completions relative to the main
-    # script. This is primarily for run-in-place-from-git-clone setups, where
-    # we want to prefer in-tree completions over ones possibly coming with a
-    # system installed bash-completion. (Due to usual install layouts, this
-    # often hits the correct completions in system installations, too.)
-    dirs+=("$_comp__base_directory"/completions{,-core,-fallback})
-
-    # 3) From bin directories extracted from the specified path to the command,
-    # the real path to the command, and $PATH
-    paths=()
-    [[ $cmd == /* ]] && paths+=("${cmd%/*}")
-    _comp_realcommand "$cmd" && paths+=("${REPLY%/*}")
-    _comp_split -aF : paths "$PATH"
-    for dir in "${paths[@]%/}"; do
-        [[ $dir == ?*/@(bin|sbin) ]] &&
-            dirs+=("${dir%/*}/share/bash-completion"/completions)
-    done
-
-    # 4) From XDG_DATA_DIRS or system dirs (e.g. /usr/share, /usr/local/share):
-    # Completions in the system data dirs.
-    _comp_split -F : paths "${XDG_DATA_DIRS:-/usr/local/share:/usr/share}" &&
-        dirs+=("${paths[@]/%//bash-completion}"/completions)
-
-    # Look up and source
-    shift
+    shift # cmd
     local -a source_args=("$@")
+    local -a dirs
+    local dir
 
+    # There are two kinds of completion subdirs, main and fallback.
+    # Main subdirs are first looked up from all supported base locations,
+    # followed by fallback subdir lookups in certain locations.
+
+    # Typically, there is only one main completion subdir, "completions",
+    # and one fallback completion subdir, "completions-fallback".
+
+    # The location of the main bash_completion script, i.e. the "home" location
+    # of the bash-completion setup in use, is special in a few respects.
+    # It is primarily for run-in-place-from-git-clone setups or ones unpacked
+    # from tarballs, typically user setups where we want to prefer in-tree
+    # completions over ones possibly coming with a system installed
+    # bash-completion.
+    #
+    # It supports an additional "completions-core" main subdir to look up from.
+    # This subdir contains completion snippets that come with that
+    # bash-completion version, and snippets in it are looked up after the usual
+    # "completions" dir. The "completions" dir in this location is for other
+    # completion snippets, such as ones installed by packages shipping
+    # completions of their own.
+    #
+    # Completions in the "completions-core" subdir may make use of
+    # bash-completion's internal API that is highly version specific.
+    # Therefore we do not look up from any other "completions-core" subdir
+    # besides this one.
+    #
+    # This location is also the only one from which we look up fallback
+    # completions from. These may also make use of bash-completion's internal
+    # API, and as it is assumed there's little if any need for a fallback
+    # hierarchy, we don't do it "just in case" to avoid some lookups.
+
+    # Filenames we look up from in usual main subdirs have the name of the
+    # command we are looking up a completion for, with .bash suffix appended
+    # (recommended) or no suffix. From others (i.e. ones containing files
+    # that are part of the bash-completion version in use), only the one with
+    # .bash appended is considered.
+
+    # 1) From user locations
+
+    if [[ ${BASH_COMPLETION_USER_DIR-} ]]; then
+        _comp_split -F : dirs "$BASH_COMPLETION_USER_DIR"
+    else
+        dirs=("${XDG_DATA_HOME:-$HOME/.local/share}/bash-completion")
+    fi
     for dir in "${dirs[@]}"; do
-        [[ -d $dir ]] || continue
-        for compfile in "$cmdname.bash" "$cmdname"; do
-            _comp_load__visit_file "$dir/$compfile" && return 0
-        done
+        _comp_load__visit_file "$dir/completions/$cmdname.bash" && return 0
+        _comp_load__visit_file "$dir/completions/$cmdname" && return 0
     done
 
-    # Look up simple "xspec" completions
+    # 2) From the bash-completion "home" location
+
+    _comp_load__visit_file \
+        "$_comp__base_directory/completions/$cmdname.bash" && return 0
+    _comp_load__visit_file \
+        "$_comp__base_directory/completions/$cmdname" && return 0
+    _comp_load__visit_file \
+        "$_comp__base_directory/completions-core/$cmdname.bash" && return 0
+
+    # 3) From bin directories extracted from a specified path to the command,
+    # the real path to the command, and $PATH
+
+    dirs=()
+    [[ $cmd == /* ]] && dirs+=("${cmd%/*}")
+    _comp_realcommand "$cmd" && dirs+=("${REPLY%/*}")
+    _comp_split -aF : dirs "$PATH"
+    for dir in "${dirs[@]%/}"; do
+        if [[ $dir == ?*/@(bin|sbin) ]]; then
+            _comp_load__visit_file "${dir%/*}/share/bash-completion/completions/$cmdname.bash" && return 0
+            _comp_load__visit_file "${dir%/*}/share/bash-completion/completions/$cmdname" && return 0
+        fi
+    done
+
+    # 4) From bash-completion subdir in XDG_DATA_DIRS or system dirs
+    # (e.g. /usr/share, /usr/local/share) -- typically meaning a system
+    # installed one if the one in use is not that, see 2) above
+
+    _comp_split -F : dirs "${XDG_DATA_DIRS:-/usr/local/share:/usr/share}"
+    for dir in "${dirs[@]}"; do
+        _comp_load__visit_file \
+            "$dir/bash-completion/completions/$cmdname.bash" && return 0
+        _comp_load__visit_file \
+            "$dir/bash-completion/completions/$cmdname" && return 0
+    done
+
+    # 5) From the fallback completion dir
+
+    _comp_load__visit_file \
+        "$_comp__base_directory/completions-fallback/$cmdname.bash" && return 0
+
+    # 6) Simple "xspec" completions
+
     [[ -v _comp_xspecs[$cmdname] || -v _xspecs[$cmdname] ]] &&
         complete -F _comp_complete_filedir_xspec "$cmdname" "$backslash$cmdname" && return 0
+
+    # Last) fallback to minimal if applicable
 
     if [[ $flag_fallback_default ]]; then
         complete -F _comp_complete_minimal -- "$origcmd" && return 0
--- a/bash_completion
+++ b/bash_completion
@@ -3475,9 +3475,10 @@ _comp_load()
     # 1) From BASH_COMPLETION_USER_DIR (e.g. ~/.local/share/bash-completion):
     # User installed completions.
     if [[ ${BASH_COMPLETION_USER_DIR-} ]]; then
-        _comp_split -F : dirs "$BASH_COMPLETION_USER_DIR"
+        _comp_split -F : dirs "$BASH_COMPLETION_USER_DIR" &&
+            dirs+=("${paths[@]/%*(\/)//completions}")
     else
-        dirs=("${XDG_DATA_HOME:-$HOME/.local/share}/bash-completion")
+        dirs=("${XDG_DATA_HOME:-$HOME/.local/share}/bash-completion/completions")
     fi
 
     # 2) From the location of bash_completion: Completions relative to the main
@@ -3485,7 +3486,7 @@ _comp_load()
     # we want to prefer in-tree completions over ones possibly coming with a
     # system installed bash-completion. (Due to usual install layouts, this
     # often hits the correct completions in system installations, too.)
-    dirs+=("$_comp__base_directory")
+    dirs+=("$_comp__base_directory"/completions{,-core,-fallback})
 
     # 3) From bin directories extracted from the specified path to the command,
     # the real path to the command, and $PATH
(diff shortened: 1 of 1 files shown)


bash_completion
```diff
@@ -3496,28 +3496,24 @@ _comp_load()
...
+        dirs+=("${paths[@]/%//bash-completion}"/completions{,-core,-fallback})
 
     # Look up and source
     shift
     local -a source_args=("$@")
 
-    local i
-    for i in "${!dirs[@]}"; do
-        dir=${dirs[i]}/completions
+    for dir in "${dirs[@]}"; do
```
scop, line 3555: The result of this is that we end up loading completions without the .bash suffix from more dirs we did before this change. We may want to tighten that up some.

Could check if the dir ends with -core or -fallback and load only .bash for them. That isn't 100% correct in the case if someone has for whatever reason configured their BASH_COMPLETION_USER_DIR to have one of those suffixes (we'd want to load both with and without .bash from there for backwards compatibility).

bash_completion
```diff
@@ -3478,15 +3478,15 @@ _comp_load()
     if [[ ${BASH_COMPLETION_USER_DIR-} ]]; then
         _comp_split -F : dirs "$BASH_COMPLETION_USER_DIR"
```
akinomyoga, line 3518: `BASH_COMPLETION_USER_DIR` is supposed to contain a path like `~/.local/share/bash-completion`, so we need to suffix `/completion` to them too. I think we can use `_comp_compgen`. For example, something like this: `_comp_compgen -Rv dirs -F : -- -W '$BASH_COMPLETION_USER_DIR' -S /completions`, though I haven't tested this carefully.

bash_completion
```diff
@@ -3496,28 +3496,24 @@ _comp_load()
...
         [[ $dir == ?*/@(bin|sbin) ]] &&
-            dirs+=("${dir%/*}/share/bash-completion")
+            dirs+=("${dir%/*}/share/bash-completion"/completions{,-core,-fallback})
     done
 
     # 4) From XDG_DATA_DIRS or system dirs (e.g. /usr/share, /usr/local/share):
     # Completions in the system data dirs.
     _comp_split -F : paths "${XDG_DATA_DIRS:-/usr/local/share:/usr/share}" &&
-        dirs+=("${paths[@]/%//bash-completion}")
+        dirs+=("${paths[@]/%//bash-completion}"/completions{,-core,-fallback})
```
akinomyoga: Should we search the system `completions-{core,fallback}` after none were found in the in-tree `completions-{core,fallback}` (i.e., `$_comp__base_directory/completions-{core,fallback}` in the tree of the current `bash_completion`)? I think all default/fallback completions should already be found in the in-tree `completions-{core,fallback}`. Are there cases where the completion file is not found in the in-tree `completions-{core,fallback}` but found in the system `completions-{core,fallback}`?

I also think loading external `completions-{core,fallback}` would cause the compatibility problem with the internal API. We've been assuming that we can use the latest internal APIs defined in `bash_completion` inside the in-tree completions of the same version. When the in-tree `completions-{core,fallback}` doesn't contain the completion for a command while the system `completions-{core,fallback}` contains one, I think that implies the system `bash-completion` is newer than the current `bash-completion`. In this case, we should avoid loading the system `completions-{core,fallback}` because it is likely to contain uses of new internal APIs.

In short, I think the completions in `completions-{core,fallback}` are tightly bound, in the sense of internal API, to the exact version of `bash_completion` in the same tree. It's not safe to load `completions-{core,fallback}` outside the tree.

bash_completion
```diff
@@ -3496,28 +3496,24 @@ _comp_load()
     _comp_split -aF : paths "$PATH"
     for dir in "${paths[@]%/}"; do
         [[ $dir == ?*/@(bin|sbin) ]] &&
-            dirs+=("${dir%/*}/share/bash-completion")
+            dirs+=("${dir%/*}/share/bash-completion"/completions{,-core,-fallback})
```
akinomyoga: For the same reason, I think we don't want to load `completions-{core,fallback}` in the bin directories.

Also, as I mentioned above, I would like to suggest exchanging the ordering of `2)` and `3)`.

akinomyoga (commented):
Thank you for creating the PR! This is exactly the issue that I described in https://github.com/scop/bash-completion/discussions/1557#discussioncomment-15600873.

> When using the new version, I also noticed a search-order problem of `_comp_load` when the system `bash-completion < 2.18` and the user's `bash-completion >= 2.18` coexist. I'll later describe it. I would like to discuss it and fix issues of `_comp_load` including it.

I was naively thinking of swapping the directory precedence of `2)` and `3)`, and inserting the search for `completions-core` before `4)`, but I haven't yet considered it deeply enough.

bash_completion
akinomyoga: I agree to check the suffix `-{core,fallback}` (probably inside `_comp_load__visit_file`).

akinomyoga:
Another issue is that when, the current `bash_completion`'s tree (`$_comp__base_directory`) is different from the system's one (`/usr/share/bash-completion`), with the suggested search ordering, external completions installed in the system directory `/usr/share/bash-completion/completions/<cmd>.bash` will be masked by our completions in `completions-{core,fallback}`. This implies that the way to override `bash-completion`'s in-tree completion is missing, even if the upstream decides to start hosting the completion file.

- I think `completions-fallback` should at least be checked at the very last step as before this PR.
- We should also be aware of the version time lag of the move of the file `completions-{core => fallback}`; even if we move a completion file to a fallback directory on request from the upstream, our in-tree completion continues to take precedence until the bash-completion version in the market is entirely replaced (which takes several to ten years).

  For this reason, I think the search ordering that we currently discuss in this PR should be a temporary one; when new bash-completion versions (having in-tree completions in `completions-{core,fallback}`) are available in distributions, I think we should move back the search order of the in-tree `completions-core` after `1)..4)` directories.

akinomyoga:
> * For this reason, I think the search ordering that we currently discuss in this PR should be a temporary one; when new bash-completion versions (having in-tree completions in `completions-{core,fallback}`) are available in distributions, I think we should move back the search order of the in-tree `completions-core` after `1)..4)` directories.

Or, maybe we can keep the current `master`, and set something like `XDG_DATA_DIRS=/dev/null` for testing purposes.

Or, another possibility is to search i) `1-4` with `.bash`, ii) in-tree `completions-core`, iii) `1-4` without suffix, iv) `completions-fallback`. The biggest problem currently is that completions in `/usr/share/bash-completion/completions/<cmd>` provided by the system `bash-completion` package have higher precedence than the in-tree completions. In the current search ordering, we need to wait until distributed versions of `bash-completion` are updated to be the ones with the new directory structure. Loading of old completions provided by the system `bash-completion` package can be avoided by checking the `.bash` suffix. That's the idea.

akinomyoga:
> Or, another possibility is to search i) `1-4` with `.bash`, ii) in-tree `completions-core`, iii) `1-4` without suffix, iv) `completions-fallback`.

Currently, I like this idea, though I need to think about it more later. It seems to solve the main issues cleanly with the minimum change.

```diff
diff --git a/bash_completion b/bash_completion
index e67cc3d51..d9c1d1cce 100644
--- a/bash_completion
+++ b/bash_completion
@@ -3508,15 +3508,17 @@ _comp_load()
     shift
     local -a source_args=("$@")

-    local i
-    for i in "${!dirs[@]}"; do
-        dir=${dirs[i]}/completions
-        [[ -d $dir ]] || continue
-        for compfile in "$cmdname.bash" "$cmdname"; do
-            _comp_load__visit_file "$dir/$compfile" && return 0
-        done
+    local dir
+    for dir in "${dirs[@]}"; do
+        _comp_load__visit_file "$dir/completions/$cmdname.bash" && return 0
     done
+
     _comp_load__visit_file "$_comp__base_directory/completions-core/$cmdname.bash" && return 0
+
+    for dir in "${dirs[@]}"; do
+        _comp_load__visit_file "$dir/completions/$cmdname" && return 0
+    done
+
     _comp_load__visit_file "$_comp__base_directory/completions-fallback/$cmdname.bash" && return 0

     # Look up simple "xspec" completions
```

bash_completion
scop: I haven't thought of this too deeply, but I think we do not need to do that. This dir is a single directory to load completions from, solely under the user's control, I don't think we need to add any more structure or segments to it. Adding would also be a breaking change.

scop: Hm README.md seems to disagree with me, I misremembered. `$BASH_COMPLETION_USER_DIR` is by the way not documented in doc/configuration.md, only its `$XDG_CONFIG_HOME/bash_completion` fallback is. We should fix that.

bash_completion
scop: Yep, agreed.

scop:
> Currently, I like this idea, though I need to think about it more later. It seems to solve the main issues cleanly with the minimum change.

I like how it looks too, but it contains a behavioral change: `*.bash` from the base completions-core now has higher precedence than plain non-`*.bash` from other dirs such as the user specified one. Strictly speaking that sounds like a breaking change.

akinomyoga:
> > Currently, I like this idea, though I need to think about it more later. It seems to solve the main issues cleanly with the minimum change.
> 
> I like how it looks too, but it contains a behavioral change: `*.bash` from the base completions-core now has higher precedence than plain non-`*.bash` from other dirs such as the user specified one. Strictly speaking that sounds like a breaking change.

That is a good point.

Then, how about using different rules for the user-specified completion files and the package-provided completion files?

1. We first look for `<cmd>.bash` and `<cmd>` in the user directories (specified through `$BASH_COMPLETION_USER_DIR`).
2. Then, we look for `completions/<cmd>.bash` in system directories,
3. the in-tree `completions-core/<cmd>.bash`,
4. `completions/<cmd>` in system directories, and
5. the in-tree `completions-fallback/<cmd>.bash`, in order.

Example Implementation: bc5dfffff...4366ead5e

<details><summary><b>Rationale for checking <code>completion/&lt;cmd&gt;</code> in system directories after the in-tree <code>completions-core/&lt;cmd&gt;.bash</code></b></summary>

I think we don't have to care about the completion file `<cmd>` provided by the upstream `<cmd>` project. While users may put their custom version at `$BASH_COMPLETION_USER_DIR/completions/<cmd>` to override the one provided by `bash-completion` or the upstream `<cmd>` project, I think the upstream `<cmd>` project has been supposed to notify us that they start to ship the completion file so that we rename the file `{ => _}<cmd>`.

- When we had `completions/<cmd>` before the directory structure change, while the upstream provided its completion at `completions/<cmd>` in a different install prefix, the situation is unchanged. Our in-tree `completions/<cmd>` had higher precedence than the upstream one, so it isn't a new problem that our `completions-core/<cmd>.bash` masks the upstream `completions/<cmd>` in the new directory structure.
- When we renamed the in-tree completion to `completions/_<cmd>` for the upstream `completions/<cmd>`, the problem doesn't arise. That's because we moved `completions/_<cmd>` to `completions-fallback/<cmd>.bash`, and the upstream `completions/<cmd>` still has higher precedence than `completions-fallback`.
- When the upstream `<cmd>` project starts to provide their completion file in the future, they can install it at `completions/<cmd>.bash` with a suffix `.bash`. Then, it has higher precedence than the in-tree `completions-core/<cmd>.bash` in the new directory structure.
- There might still be a project newly installing a completion file at `completions/<cmd>`, but it causes a conflict even with the old directory structure, or it is masked by our in-tree completion with the old directory structure.

</details>

scop:
That sounds mostly sane, but I think (offhand) that reading system locations before in-tree ones makes it harder to work on bash-completion as well as to replace the system installed one with a user one.

scop:
https://github.com/scop/bash-completion/pull/1579 is one example suffering from our current load dir precedence issues.

akinomyoga:
> That sounds mostly sane, but I think (offhand) that reading system locations before in-tree ones makes it harder to work on bash-completion

Could you explain it in a bit more detail?

I think that is the original intent of the present separation of the directory. The original change for the separation of in-tree completions into `completions-core` intended to allow a system-installed completion to override the in-tree completion; it previously required renaming the in-tree completion from `<cmd>` to `_<cmd>` to avoid the conflict on the path `completions/<cmd>`, but that doesn't work until bash-completion in the market is entirely replaced with the new version.

For the test suite of `bash-completion`, I think we already [set `BASH_COMPLETION_USER_DIR` and `XDG_DATA_DIR`](https://github.com/scop/bash-completion/blob/45530b095fd0af2d4e6b2430f2eb2558c6ef658c/test/config/bashrc#L40-L42) to disable users' and system completions. If it still loads the system-installed one intead of the in-tree completion, it is probably found by `3)` (i.e., from the prefix where the command file exists).

> as well as to replace the system installed one with a user one.

In my understanding, replacing the system-installed one with a user one is easy because the user-provided one is searched frist in the suggestion https://github.com/scop/bash-completion/pull/1569#issuecomment-3891108369.

> #1579 is one example suffering from our current load dir precedence issues.

As I mentioned above, we already set `BASH_COMPLETION_USER_DIR` and `XDG_DATA_DIR`, so  I guess they are found through `3)` in `_comp_load`. We might add a configuration variable to turn off the search `3)`.

scop:
> In my understanding, replacing the system-installed one with a user one is easy because the user-provided one is searched frist in the suggestion

That suggestion notes this happens through `$BASH_COMPLETION_USER_DIR`. This variable was not required before for this to happen. I have run my personal setup for ages so that I do have the distro/system bash-completion installed, but I have my bash-completion working tree checked out somewhere in my home dir, and I source it with
```
. ~/path/to/bash-completion/bash_completion
```
...in my bashrc, and it has Just Worked: things in my home dir one get precedence without having to set any other variables.

scop:
I've reworked this PR pretty much altogether.

It documents my thoughts regarding this fairly completely I hope, and implements them in a way that is easier to read and reason. Could be it's also faster in some cases because it returns as soon as it finds what needs to be found instead of computing dirs that might never be even looked at.

This is very lightly tested as it stands.

akinomyoga:
Thank you for updating the PR. I have a question. In the new version, `<cmd>.bash` in system directories are still searched after the in-tree `completions-core`, or in other words, steps 2 and 3 in https://github.com/scop/bash-completion/pull/1569#issuecomment-3891108369 are swapped. Is this intentional?

The current version still has the problem described in https://github.com/scop/bash-completion/pull/1569#issuecomment-3857087231 with a user installation of `bash-completion`; even if the upstream starts to offer the completion setting at `<cmd>.bash` (as suggested in https://github.com/scop/bash-completion/issues/1385#issuecomment-3450778050) to replace an in-tree completion in `completions-core`, it continues to be masked until we move the in-tree completion to `completions-fallback` and the user updates `bash-completion` to the latest version. Then, when the user installs `bash-completion` in a home directory, there is technically no way for the upstream to replace the completion setting without waiting for `bash-completion` to be updated.

scop:
In a sense, preferring _everything_ from in-tree over system and executable based locations by default out of the box is intentional. I do see the problem for user installs though, I just don't know if we should do anything about it, and if, what.

Another aspect of this I'm torn about is if we should load any non-.bash from other `completions/` dirs besides that of the currently running bash_completion's "in-tree" one. Reason for this is the same why we don't load anything from other `completions-core/` dirs: they may make use of internal API of some other bash-completion version that is subject to change between versions. (If we'd never remove any completions from core, this should not happen, but I'm not convinced we should rely on that.) But: if we stopped doing that for other `completions/` dirs, that would mean user installs would not get non-.bash completions installed by other things besides bash-completion in system dirs loaded by default.

However, users are not dead in the water for either of the above cases: they can always install/symlink completions that would not be loaded by default otherwise in their `$BASH_COMPLETION_USER_DIR`s.

Mandating users do that is not great, and it is a breaking change in a sense, but I'm not sure I would prefer any other options that I can think of now, as all of them bear some risk of more unexpected breakage and complexity.

What I intend to do is to start encouraging bash completion snippet shippers to use the .bash suffix anyway, as it should not have any backwards compatibility issues, and does most likely have forward compatibility benefits.

scop:
Related: https://github.com/scop/bash-completion/pull/1610

akinomyoga:
> In a sense, preferring _everything_ from in-tree over system and executable based locations by default out of the box is intentional. I do see the problem for user installs though, I just don't know if we should do anything about it, and if, what.

If that's indeed the intention, knowing that the upstream completion may be masked by `bash-completion` for a long time, that's fine. We may discuss the issue later (and I think we don't have to address it for 2.18).

> Another aspect of this I'm torn about is if we should load any non-.bash from other `completions/` dirs besides that of the currently running bash_completion's "in-tree" one. Reason for this is the same why we don't load anything from other `completions-core/` dirs: they may make use of internal API of some other bash-completion version that is subject to change between versions.
>
> [...]
>
> Mandating users do that is not great, and it is a breaking change in a sense, but I'm not sure I would prefer any other options that I can think of now, as all of them bear some risk of more unexpected breakage and complexity.

Probably we don't have to worry about the "*internal*" API because `completions-core/*.bash` doesn't seem to use the *internal* API. They only use the *public* API (By the way, `_comp_comptpen` is my typo of `_comp_compgen` in commit 18723c8102 **edit**: submitted a fix in #1620):

```console
$ grep -Eho '\b_comp_[[:alnum:]]+' completions-core/*.bash | sort -u
_comp_awk
_comp_backup
_comp_cmd
_comp_command
_comp_compgen
_comp_complete
_comp_comtpen
_comp_count
_comp_delimited
_comp_deprecate
_comp_dequote
_comp_dpkg
_comp_expand
_comp_get
_comp_have
_comp_initialize
_comp_looks
_comp_ltrim
_comp_quote
_comp_readline
_comp_realcommand
_comp_root
_comp_split
_comp_tail
_comp_try
_comp_unlocal
_comp_userland
_comp_variable
_comp_xfunc
```

Actually, the issue I raised in https://github.com/scop/bash-completion/pull/1569#discussion_r2771632605 isn't about the "internal/private" API, but new *public* APIs introduced in specific versions of `bash-completion`.

I think we can safely load non-`.bash` files with regard to the API. We stop putting in-tree completions (which are tightly bound to the corresponding bash-completion version through new *public* APIs) in `completions`. Then, the non-`.bash` files in `completions` are supposed to be using only the bash-completion public APIs before the completion directory separation. As long as the public APIs are maintained, there shouldn't be issues.

akinomyoga (approved):

bash_completion
```diff
@@ -3470,59 +3469,109 @@ _comp_load()
...
+
+    # There are two kinds of completion subdirs, main and fallback.
+    # Main subdirs are first looked up from all supported base locations,
+    # followed by fallback subdir lookups in certain locations.
+
+    # Typically, there is only one main completion subdir, "completions",
+    # and one fallback completion subdir, "completions-fallback".
+
+    # The location of the main bash_completion script, i.e. the "home" location
+    # of the bash-completion setup in use, is special a few respects.
```
yedayak:
```suggestion
    # of the bash-completion setup in use, is special in a few respects.
```

yedayak (approved):

scop:
Thanks! I'm not sure if all this is spot on "correct", but at least it is now documented in detail.

PR #1570, opened 2026-02-05
feat(syncthing): add 3rd party fallback completion loader

scop:
https://syncthing.net (no online documentation for installing completions that I can find)

### DIFF
--- a/completions-fallback/Makefile.am
+++ b/completions-fallback/Makefile.am
@@ -58,6 +58,7 @@ cross_platform = adb.bash \
 		svn.bash \
 		svnadmin.bash \
 		svnlook.bash \
+		syncthing.bash \
 		task.bash \
 		tokio-console.bash \
 		udevadm.bash \
--- a/completions-fallback/syncthing.bash
+++ b/completions-fallback/syncthing.bash
@@ -1,0 +1,9 @@
+# 3rd party completion loader for commands emitting        -*- shell-script -*-
+# their completion using "$cmd install-completions". For example,
+# some Go programs using https://github.com/WillAbides/kongplete do.
+#
+# This serves as a fallback in case the completion is not installed otherwise.
+
+eval -- "$("$1" install-completions 2>/dev/null)"
+
+# ex: filetype=sh


akinomyoga:
The completion support seems to be based on [Kongplete](https://github.com/WillAbides/kongplete/blob/master/completion_install.go). Maybe we can mention this in the commit message and also in the completion file.

However, the feature to print the completion setting by `install-completions` seems to be provided by the unreleased Kongplete-0.3.0. The Kongplete documentation doesn't mention the usage of `install-completions` at all. According to https://github.com/WillAbides/kongplete/issues/14, the feature to print the completion setting (so that one can `eval` or `source` the output) is only in 0.3.0 and missing in the latest official release Kongplete-0.2.0.

- Syncthing seems to rely on [Kongplete 0.4.0](https://github.com/syncthing/syncthing/blob/1bba07a15210fcbdb21b98f12211cd9d8dc07ca5/go.mod#L41).
- Another project [Terramate](https://github.com/terramate-io/terramate) also seems to offer [the Kongplete-based completion support](https://github.com/terramate-io/terramate/blob/a22c73f383b4f0d250b6bef71563881e4c56db96/ui/tui/cli_spec.go#L179). ~~However, the usage of `terramate install-completions` described in [its documentation](https://terramate.io/docs/cli/installation#auto-completion) seems different from the one in this PR. A similar description is also found in [this page](https://terramate.io/docs/cli/reference/cmdline/install-completions) in the documentation. The completion support was added in [terramate-io/terramate#96](https://github.com/terramate-io/terramate/pull/96).~~ I realized that Terramate seems to depend on [Kongplete 0.2.0](https://github.com/terramate-io/terramate/blob/a22c73f383b4f0d250b6bef71563881e4c56db96/go.mod#L33). The usage difference seems to come from the difference in the Kongplete version.
- Even another project [oalders/is](https://github.com/oalders/is)[^1] also uses [the Kongplete-based completion](https://github.com/oalders/is/blob/a64c3c29f17b12109da106fe58d99d932769c7df/main.go#L31). The usage described in [its documentation](https://github.com/oalders/is/blob/a64c3c29f17b12109da106fe58d99d932769c7df/README.md#install-completions) is actually consistent with this PR. This depends on [Kongplete 0.3.0](https://github.com/oalders/is/blob/a64c3c29f17b12109da106fe58d99d932769c7df/go.mod#L13).


[^1]: this project `oalders/is` has only 47 stars but has a conflicting command name with [microsoft/inshellisense](https://github.com/microsoft/inshellisense), so I don't think we should cover `oalders/is` for the third-party completion loader. I've mentioned this project just for the usage of `install-completions` described in its documentation.

scop:
Thanks for the detective work. Fixed the commit message typo and added a mention in the completion file, merging with those.

PR #1572, opened 2026-02-12
fix: add compatibility wrapper for tail

akinomyoga:
This fixes the uses of `tail -n +2` in Solaris mentioned in https://github.com/scop/bash-completion/pull/1567#issuecomment-3851588026.

See the commit message of 994dfde72 for details.

----

> https://github.com/scop/bash-completion/pull/1567#discussion_r2770801269
>
> I don't know if it matters at all, just happened to notice now, but in other places we have a space between `-n` and `+2`.

Yeah, I noticed the difference in the style. Technically, both `-n +2` and `-n+2` should be supported based on POSIX, but it would be better to normalize the style in the codebase. I changed it in commit 994dfde72.

> https://github.com/scop/bash-completion/pull/1567#discussion_r2770809286
>
> Given the info on `tail -n +2` not working on Solaris, this change and the one below are unfortunate in the sense that they are in a `if $OSTYPE` block that specifically targets Solaris (and AIX).
>
> https://github.com/scop/bash-completion/pull/1567#discussion_r2770810471
>
> This being the other change mentioned above.

I also checked the online manuals of AIX, but AIX tail seems to be fine; it supports the option `-n +<num>`.

> https://github.com/scop/bash-completion/pull/1567#issuecomment-3855795895
> I think we should care about what we are willing to maintain and can test. Or in other words, what we care about :)
> 
> As far as I'm personally concerned, that no longer includes any Solaris version as I don't have access to one nor can I test with one these days. I have no problem if others want to support it as long as it doesn't get too badly on the way of keeping things working on other "more supported" platforms.

It seems we can support it just with `_comp_tail`, so the maintenance cost seems small. If there are any problems, we may remove the wrapper `_comp_tail` together with `_comp_awk` to drop the support for Solaris.

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2072,12 +2072,12 @@ if [[ $OSTYPE == *@(solaris|aix)* ]]; then
     # This function completes on process IDs.
     _comp_compgen_pids()
     {
-        _comp_compgen_split -- "$(command ps -efo pid | tail -n +2)"
+        _comp_compgen_split -- "$(command ps -efo pid | _comp_tail -n +2)"
     }
 
     _comp_compgen_pgids()
     {
-        _comp_compgen_split -- "$(command ps -efo pgid | tail -n +2)"
+        _comp_compgen_split -- "$(command ps -efo pgid | _comp_tail -n +2)"
     }
     _comp_compgen_pnames()
     {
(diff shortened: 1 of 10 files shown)


akinomyoga:
> Just wondering if we should add something related to this in test/runLint.

Thank you! I completely forgot about `test/runLint`. I added it in the push [`994dfde..f6fee8a`](https://github.com/scop/bash-completion/compare/994dfde72508061b469b1a702b9d00f0dabb79ff..f6fee8af6d744c1ddbcc4a947a125a897bbaaa6e).

PR #1573, opened 2026-02-12
style: remove unnecessary quoting of comsub in assignment rhs

akinomyoga:
As suggested in https://github.com/scop/bash-completion/pull/1564#discussion_r2749378798.

After merging this, I'm going to submit another PR to refactor some of the commands inside those command substitutions.


akinomyoga:
Thank you!

PR #1574, opened 2026-02-12
refactor: simplify sed/awk/grep uses, etc.

akinomyoga:


completions-fallback/modules.bash
```diff
@@ -53,8 +53,8 @@ _comp_cmd_module()
         # First parameter on line -- we expect it to be a mode selection
 
         local options
-        options=$(module help 2>&1 | command grep -E '^[[:space:]]*\+' |
-            _comp_awk '{print $2}' | command sed -e 's/|/ /g' | sort)
+        options=$(module help 2>&1 | _comp_awk '/^[[:space:]]*\+/ {
+            $0 = $2; gsub(/\|/, " "); print }' | sort)
```
scop, line 57: I have a nagging feeling that there is an awk implementation in existence that does not support gsub, but I can't find it right now and gsub is in POSIX at least down to 2008, so maybe it's all good.

There is one gsub use in the tree in chronyc, but that could be fine if we assume chronyc is Linux specific.

akinomyoga: Isn't that Solaris awk? `Solaris <= 11.3` awk indeed doesn't have `gsub`:

```console
$ awk 'gsub(/a/, "b")' <<< "aaazzz"
awk: syntax error near line 1
awk: bailing out near line 1
```

We use `/usr/xpg4/bin/awk` (which has `gsub`) in Solaris through the wrapper `_comp_awk`, so the problem doesn't happen in Solaris.

I don't have AIX, NetBSD, and OpenBSD, but online manuals seem to suggest that their awk has `gsub`.

- [aix/7.1.0 awk](https://www.ibm.com/docs/ja/aix/7.1.0?topic=awk-command) (Mar. 2021)
- [OpenBSD-2.2/awk.1](https://man.openbsd.org/OpenBSD-2.2/awk.1) (Dec. 1997, OpenBSD seems to have introduced man pages in 2.2)
- [NetBSD-2.0/awk.1](https://man.netbsd.org/NetBSD-2.0/awk.1) (Jun. 2004, NetBSD seems to have switched from gawk to nawk in 2.0)
- [NetBSD-1.3/awk.1](https://man.netbsd.org/NetBSD-1.3/awk.1) (Nov. 1994, gawk)

akinomyoga: - [[PDF] AIX 5L 5.3 Commands Reference a-c](http://public.dhe.ibm.com/systems/power/docs/aix/53/aixcmds1.pdf)/p179 mentions `gsub` too (Aug. 2004)

scop: Could be that it was Solaris awk indeed. Or some old version of mawk. Anyway I think we're good.

PR #1577, opened 2026-02-28
fix(make): properly skip the "Not a target" block in the helper AWK

rnvannatta:
In some obscure cases, `target: VAR = val` can cause `make t` to autocomplete to the incorrect result.

This is happening in one of my projects with a somewhat complex set of makefiles, such that `make w` returns windows, webgl, and a random build file. A silly, though minimal repro is:

```
all: fluff
fluff:
	touch fluff
foobar: OUCH = glitch
```

where `make f` will erroneously list completion possibilities `all fluff`

The root cause is that the foobar 'block' looks like this

```
# makefile (from 'Makefile', line 4)
foobar: OUCH = glitch
# Not a target:
foobar:
#  Implicit rule search has not been done.
#  Modification time never checked.
#  File has not been updated.
# variable set hash-table stats:
# Load=1/32=3%, Rehash=0, Collisions=0/2=0%
```
  
the script matches f against foobar, setting the target flag true, then it encounters `# Not a target` over and over and over again until it hits the all target, and because the target flag is true, it accepts that. The solution is to make `# Not a target` reset the flag as it skips.

You can test the script in isolation with: `make -npq __BASH_MAKE_COMPLETION__=1 -C . .DEFAULT 2>/dev/null | prefix="f" prefix_replace="f" awk -f make-extract-targets.awk`

Tested on Debian Trixie with GNU Make 4.4.1, the latest in Debian.

Here is the output of make -nqp on the minimal repro makefile for your convenience: [make-nqp-output.txt](https://github.com/user-attachments/files/25617961/make-nqp-output.txt)

### DIFF
--- a/helpers-core/make-extract-targets.awk
+++ b/helpers-core/make-extract-targets.awk
@@ -29,7 +29,20 @@ NR == 1, /^# +Make data base/ { next; }
 /^# +Variables/, /^# +Files/ { next; }
 
 # skip not-target blocks
-/^# +Not a target/, /^$/     { next; }
+/^# +Not a target/, /^$/ {
+  # We need to manually clear the state.  Without it, the cleanup associated
+  # with the present block would be skipped because /^$/ on the above line also
+  # consumes the empty line.  This would become an issue with the case where
+  # the "Not a target" line appears in the middle of a block, e.g., for the
+  # case with "foobar: OUCH = glitch" without the actual rule for "foobar".
+  # The leftover information would be associated with the next target block
+  # unexpectedly.
+  #
+  # https://github.com/scop/bash-completion/pull/1577
+  is_target_block = 0;
+  target = "";
+  next;
+}
 
 # The stuff above here describes lines that are not
 #  explicit targets or not targets other than special ones
--- a/.typos.toml
+++ b/.typos.toml
@@ -72,3 +72,8 @@ extend-exclude = ["CHANGELOG.md"]
 [type.test_export]
 extend-glob = ["test_export.py"]
 extend-identifiers = { BASH_VERSIO = "BASH_VERSIO" }
+
+# test/t/test_make.py
+[type.test_make]
+extend-glob = ["test_make.py"]
+extend-words = { ba = "ba" }
--- a/test/fixtures/make/Makefile
+++ b/test/fixtures/make/Makefile
@@ -44,3 +44,8 @@ VARIABLE_LOOKS_A_BIT_LIKE_A_TARGET := fooled-you
 extra_makefile:
 	touch $@
 include extra_makefile
+
+# foobar is not actually a target, the line is setting target-specific variables
+# not defining a target so it shouldn't show up in the completion list
+fluff:
+foobar: OUCH = glitch
--- a/test/fixtures/make/test_github_pr_1577/Makefile
+++ b/test/fixtures/make/test_github_pr_1577/Makefile
@@ -1,0 +1,13 @@
+all: fluff
+fluff:
+	touch fluff
+
+common-proc:
+	echo $(OUCH)
+.PHONY: common-proc
+
+foobar: OUCH = glitch
+foobar: common-proc
+
+barbaz: OUCH = switch
+barbaz: common-proc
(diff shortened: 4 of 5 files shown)


rnvannatta:
Fixed the lint issues with the commit message

yedayak:
How do you feel about adding a test? You could probably just add it to `test/fixtures/make/Makefile` and change `test/t/test_make.py` to test it

helpers-core/make-extract-targets.awk
akinomyoga, line 53: I don't know if this PR is the right fix for the present issue. In the first place, does anyone have an idea of the reason that we start the block with the matching line and include also unmatching lines in the same block? Isn't this the root cause of the present issue? I translated the original sed script in [#1069](https://github.com/scop/bash-completion/pull/1069/changes#diff-1776d90047a6c8d16281efc474f61be0173178c5564bbeddf154f7c4ceb3b07aL35) into AWK so that the behavior wouldn't be changed, but the behavior seems a mystery. This behavior seems to have existed from the beginning since the analysis was introduced in commit [`b28d710`](https://github.com/akinomyoga/bash-completion/commit/b28d7108d3677c61bd01c51ccee8bb1cf9e3bfba#diff-1776d90047a6c8d16281efc474f61be0173178c5564bbeddf154f7c4ceb3b07aR10) by @code5hot.

rnvannatta: It's kind of an obscure issue for a less widely used Makefile feature, that for some reason make outputs a slightly different formatted message for.

Normally 'blocks' that aren't targets have their first line be "Not a target:".

In this case, the "Not a target:" comes after the line that passes the starts_with() awk test.

akinomyoga: I didn't ask why the present PR fixes the issue. I know that this PR does fix the present specific issue, but that doesn't immediately mean that's the *right* fix. If the root cause lies in the original approach at all, the root cause should be fixed instead of patching every downstream symptom separately.

akinomyoga: I just checked what happens with the provided example.

First of all, I misunderstood the output format of `make` as if it would include multiple targets in a section (separated by empty lines). Basically, one section includes a single target name. The provided example seems a special case where `foobar:` appears twice in a single section, but the target names (`foobar`) in a single section are still consistent.

I have now understood the true issue to be that, since all lines of the range `/^# +Not a target/, /^$/` are skipped, the empty line `/^$/` that should have introduced a new section is also removed together when a section including `#  Not a target` is skipped.

akinomyoga (approved):
There might be better handling of the empty lines, but the suggested change appears to make sense now. I'm approving this.

akinomyoga (changes requested):
> How do you feel about adding a test? You could probably just add it to `test/fixtures/make/Makefile` and change `test/t/test_make.py` to test it

Actually, the test should still be added.

rnvannatta:
Sure. Added test which fails without this fix and passes with it.

akinomyoga:
I've rebased on top of the latest `main` branch.

akinomyoga:
I added changes [`7f9a4c2..158ddbe`](https://github.com/scop/bash-completion/compare/7f9a4c291f4829a11caa6e37d962afbd8e531674..158ddbe55f8fed41e9775b4b2ff95c8a2dad3caa).

- ~~The commit message should follow Conventional Commits for the `release-please` bot. In particular, the message should be a verb clause in the imperative form or the bare infinitive.~~ (**edit**: I realized that this is my misunderstanding... I didn't notice this for a long time...) By the way, I noticed that it's "Not a target" (instead of "Not a Target") in the output of `make -p`.
- I added descriptions to the test case and the changed part of the AWK script.
- I added another test case.

I kept the separation of the original two commits, while modifying each. However, I think those two commits are tightly related, so they can be squashed in merging.

akinomyoga:
@rnvannatta Your email address used to sign the commits doesn't seem to be recognized by GitHub. You may optionally add an extra email address through GitHub's Settings for Emails (at `https://github.com/settings/emails`).

akinomyoga (approved):

PR #1609, opened 2026-04-02
chore: switch from pre-commit to prek

scop:
This is an interim change as we may end up revamping pre-commit checks further, but in the meantime, might just as well [reap benefits of prek over pre-commit](https://prek.j178.dev/#why-prek).

To migrate existing pre-commit enabled setups:
  `prek install --overwrite --hook-type pre-commit --hook-type commit-msg`

### DIFF
--- a/.github/requirements.txt
+++ b/.github/requirements.txt
@@ -1,3 +1,3 @@
 -r ../test/requirements-dev.txt
 
-pre-commit>=3.2.0
+prek>=0.3.8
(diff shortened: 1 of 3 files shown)


scop:
Hm prek does not implement perl checks yet, so drafting.

Deferring to docker based checks is an option (maybe better to require docker than perl these days :thinking:). But even that is not quite there yet.

For perltidy a docker image was recently added: https://github.com/perltidy/perltidy/pkgs/container/perltidy

For perlcritic, it's (still) in progress: https://github.com/Perl-Critic/Perl-Critic/issues/999

scop:
Perl support was added in prek 0.4.5, so this is ready to roll. https://github.com/j178/prek/releases/tag/v0.4.5

akinomyoga (approved):

PR #1615, opened 2026-04-13
docs(README): use .bash suffix in autotools and cmake examples

scop:


akinomyoga (approved):
There seem to be CI errors, but I think they should be unrelated to this PR.

PR #1616, opened 2026-04-13
Add iwctl completion

cladmi:
Complete commands/subcommands, devices, networks, SSID with spaces.

No handling of `set-property` and `debug connect|roam` completions.
Not tested with `ap/ad-hoc` devices as my device does not handle it.

Implementation of `hidden-networks completion using quotes to
handle completion on `:`.

#### Split commits

In separate commits, as I am not sure if it is wanted, I added support to not show any completion when nothing should be displayed, and some testing using a mock command to show parsing of real command outputs to find devices/networks.

These could be squashed together in the same commit, just thought it would be easier to review and remove them if needed.

#### Inspiration

I did look at `ip.bash` as an inspiration.

#### Style

If any implementation style is not found appropriate, and there is a better file to look into/change this, please tell and I'll adapt it.

#### Testing

I did interactively test everything supported at some point.
Added plenty of unit tests to try validating what I did. I did not work in TDD with these though as was more learning how to write things, read documentation and testing multiple things at once. So there might be tests that at the end do not test correctly as not finding a completion for other reasons.

```
pytest -q -- test/t/test_iwctl.py 
............................s.......s.................................................. [ 56%]
..................................................................                      [100%]
151 passed, 2 skipped in 0.84s
```

With skipped tests being the one I cannot run on my machine, hoping they would work elsewhere.

```
test/t/test_iwctl.py::TestIwctl::test_ad_hoc_with_interfaces SKIPPED (iwctl ad-hoc list
| command grep -q 'No devices in') 
test/t/test_iwctl.py::TestIwctl::test_ap_with_interfaces SKIPPED (iwctl ap list |
command grep -q 'No devices in')
```

#### Commit-hooks

It seems that the hook for `mypy` just always fails, no information on why.
When run manually it seems happy. Not sure what to do here.

I did try resetting and re-committing everything at once, and only this did fail.
Each of the commit might not pass hooks as I do not know how to re-execute hooks on a `git commit --amend` and all.

### DIFF
--- a/completions-core/iwctl.bash
+++ b/completions-core/iwctl.bash
@@ -68,6 +68,8 @@ _comp_cmd_iwctl()
     local cur prev words cword comp_args was_split
     # No split on ':' for "connect-hidden" address.
     _comp_initialize -n ":" -s -- "$@" || return
+    local prevprev=${words[cword - 2]}
+    local prevprevprev=${words[cword - 3]}
 
     case $prev in
         --username | --password | --passphrase)
@@ -114,6 +116,8 @@ _comp_cmd_iwctl()
                     # No handling of properties completion.
                     ;;
                 *)
+                    [[ ${prevprev} == set-property ]] && return
+                    [[ ${prevprevprev} == set-property ]] && return
                     _comp_compgen_split -- "show set-property"
                     ;;
             esac
@@ -160,6 +164,8 @@ _comp_cmd_iwctl()
                     # No handling of properties completion.
                     ;;
                 *)
+                    [[ ${prevprev} == set-property ]] && return
+                    [[ ${prevprevprev} == set-property ]] && return
                     _comp_compgen_split -- "show set-property"
                     ;;
             esac
@@ -178,6 +184,8 @@ _comp_cmd_iwctl()
                     # No handling of properties completion.
                     ;;
                 *)
+                    [[ ${prevprev} == set-property ]] && return
+                    [[ ${prevprevprev} == set-property ]] && return
                     _comp_compgen_split -- "forget show set-property"
                     ;;
             esac
@@ -192,6 +200,7 @@ _comp_cmd_iwctl()
                     ;;
                 push-button | start-user-pin | start-pin | cancel) ;;
                 *)
+                    [[ ${prevprev} == start-user-pin ]] && return
                     _comp_compgen_split -- "push-button start-user-pin start-pin cancel"
                     ;;
             esac
@@ -226,6 +235,8 @@ _comp_cmd_iwctl()
                     # get-networks | get-hidden-access-points rssi-dbms | rssi-bars
                     ;;
                 *)
+                    [[ ${prevprev} == @(connect|connect-hidden|get-bsses) ]] && return
+                    [[ ${prevprevprev} == @(connect|connect-hidden|get-bsses) ]] && return
                     _comp_compgen_split -- \
                         "connect connect-hidden disconnect get-networks get-hidden-access-points scan show get-bsses"
                     ;;
@@ -262,6 +273,7 @@ _comp_cmd_iwctl()
                     # enroll|configure key
                     ;;
                 *)
+                    [[ ${prevprev} == key ]] && return
                     _comp_compgen_split -- "stop show enroll configure"
                     ;;
             esac
@@ -282,6 +294,7 @@ _comp_cmd_iwctl()
                     # autoconnect on|off
                     ;;
                 *)
+                    [[ ${prevprev} == @(connect|roam) ]] && return
                     _comp_compgen_split -- "connect roam get-networks autoconnect"
                     ;;
             esac
--- a/completions-core/iwctl.bash
+++ b/completions-core/iwctl.bash
@@ -1,0 +1,291 @@
+# iwctl(1) completion                                      -*- shell-script -*-
+
+# Clean and remove all lines that are not a first column of output.
+# @param $1 Column offset. Some values use multiple lines, so only keep lines
+#                          that start at this offset and not more.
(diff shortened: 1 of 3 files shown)


completions-core/iwctl.bash
```diff
@@ -0,0 +1,304 @@
...
+                *)
+                    [[ ${prevprev} == start-user-pin ]] && return
+                    _comp_compgen_split -- "push-button start-user-pin start-pin cancel"
+                    ;;
+            esac
+            ;;
+        station)
+            [[ $subcmd == list ]] && return
+            # All arguments except $prev and trailing empty argument
+            local iwctl_args=("${words[@]::${#words[@]}-2}")
```
cladmi, line 211: Here I need to call `iwctl station wlan0 get-networks` so try to get rely on `words` to get everything except the  `connect` in `iwctl station wlan0 connect`.

There might be options provided in here, as long as the option are given just after `iwctl` all this will work well, but if not, I know this might break but not sure what would be the correct way of handling it. I thought about generating a whole new array with options removed, but was not sure if it it was necessary to add it, or if there was already a preferred way of handling it.

scop:
Thank you for working on making bash completions better. But to verify, did you notice the two few paragraphs of CONTRIBUTING.md as linked in the PR creation view, in particular:

> However, before submitting a completion to us, first consider submitting it to the project that ships the commands your completion is for.

If you considered submitting it upstream, let's include the explanation why submit it here instead. If it was asked upstream and rejected there, let's add a link to the response.

cladmi:
No I did not submit it upstream indeed. I'll try.
Anything that seems obviously wrong before I should submit it upstream?
One thing they would not have I guess is the test infrastructure.

I did notice that it was submitted for fish-shell separately though: https://github.com/fish-shell/fish-shell/pull/9932

completions-core/iwctl.bash
```diff
@@ -0,0 +1,304 @@
+# iwctl(1) completion                                      -*- shell-script -*-
+
+# Clean and remove all lines that are not a first column of output.
+# @param $1 Column offset. Some values use multiple lines, so only keep lines
+#                          that start at this offset and not more.
+_comp_cmd_iwctl__cleanup_output()
+{
+    local offset="$1"
+    shift
+    local remove_colors="s/\x1B\[[0-9;]\+m//g"
```
akinomyoga, line 10:
```suggestion
    local remove_colors="s/\x1B\[[0-9:;]\{1,\}m//g"
```

- `\+` is a GNU extension.
- SGR may also contain `:`.

completions-core/iwctl.bash
```diff
@@ -0,0 +1,304 @@
...
+
+# Extract SSIDs from 'station list' or 'known-networks list'.
+# SSIDs have a maximum of 32 characters, and no leading/trailing spaces.
+# @param $1 Column offset. See _comp_cmd_iwctl__cleanup_output.
+_comp_cmd_iwctl__filter_list_networks()
+{
+    local offset="$1"
+    shift
+    local format_network="s/^\(.\{32\}\)  .*/\1/"
+    local format_rstrip="s/\([^ ]\)  *$/\1/g"
```
akinomyoga, line 41:
```suggestion
    local format_rstrip="s/  *$//"
```

completions-core/iwctl.bash
```diff
@@ -0,0 +1,304 @@
...
+        --username | --password | --passphrase)
+            return
+            ;;
+    esac
+
+    # 'subcmd' can be either 'list' or the command after the 'device' name.
+    local subcword cmd="" has_cmd="" subcmd=""
+    for ((subcword = 1; subcword < cword; subcword++)); do
+        [[ $subcmd == list ]] && break
+        [[ $subcmd ]] && subcmd=${words[subcword]} && break
```
akinomyoga, line 84: In what situation does `subcmd` become non-empty in this context? In this loop, `subcmd` is set only by line 85, but that line performs `break`, so isn't `subcmd` inside the loop always empty?

completions-core/iwctl.bash
```diff
@@ -0,0 +1,304 @@
...
+
+# List devices for given device mode.
+# @param $1 iwctl binary.
+# @param $2 device mode.
+_comp_cmd_iwctl__list_devices()
+{
+    local iwctl="$1"
+    local device_mode="$2"
+    "${iwctl}" "${device_mode}" list | _comp_cmd_iwctl__filter_first_column
+}
```
akinomyoga, line 56: This function is called only in the following form.

```bash
                    _comp_compgen_split -l -- \
                        "$(_comp_cmd_iwctl__list_devices "$1" "$cmd")"
```

Then, shouldn't the entire `_comp_compgen_split ...` be made a function instead of only making the command-substitution part a function?

Also, `_comp_compgen -a -- -W "list"` can be included.

completions-core/iwctl.bash
```diff
@@ -0,0 +1,304 @@
...
+
+_comp_cmd_iwctl__doublequote_compreply_if_has_space()
+{
+    local i
+    for i in "${!COMPREPLY[@]}"; do
+        COMPREPLY[i]=${COMPREPLY[i]//* */\"${COMPREPLY[i]}\"}
```
akinomyoga, line 62:
```suggestion
        COMPREPLY[i]=${COMPREPLY[i]//* */\""${COMPREPLY[i]}"\"}
```

See the description [here](
https://github.com/scop/bash-completion/blob/f4c58c9ee4be5cfbd0a020cad3e267333c578e4e/doc/styleguide.md#patsub_replacement-for-array-elements). Even if you know that `&` will never be included in the values in this context, it is unclear just by looking at the line, so it's not readable.

cladmi:
Thanks for the review, I'll come back to it when I manage. Still struggling for now.

PR #1619, opened 2026-04-19
test(pytest): skip instead of xfail where appropriate

scop:
Skipping is appropriate for cases where there's no point in running the tests for some reason we can detect, whereas expecting failure is more like a TODO thing -- known bug or a missing feature.

Ref https://docs.pytest.org/en/stable/how-to/skipping.html

This only addresses pytest.mark.xfail/skipif for now, not yet our pytest.mark.complete skipif=/xfail= (there's a lot of that in the tree).

### DIFF
--- a/test/t/unit/test_unit_compgen_ip_addresses.py
+++ b/test/t/unit/test_unit_compgen_ip_addresses.py
@@ -41,7 +41,7 @@ class TestUnitCompgenIpAddresses:
         assert completion
         assert all("." in x for x in completion)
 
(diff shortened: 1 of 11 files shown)


test/t/unit/test_unit_compgen_ip_addresses.py
scop, line 1: Just commenting on a random file, not related to this one:

WIP/draft for initial comments because this only addresses pytest.mark.xfail/skipif for now, not yet our pytest.mark.complete skipif=/xfail= (there's a lot of that in the tree).

PR #1620, opened 2026-04-19
fix(nsupdate): fix typo `_comp_com{tp => pg}en`

akinomyoga:
This is a regression introduced by commit 18723c81. I noticed this in https://github.com/scop/bash-completion/pull/1569#issuecomment-4276206193.


scop (approved):

scop:
Thanks!

PR #1625, opened 2026-04-30
ISC bind-utils dig completion with hints helper option file for "@"

mave007:
## Summary
Adding dedicated completion for ISC BIND **dig** (`completions-core/dig.bash`), registers it in the build, adds tests, and removes `dig` from the generic `_comp_complete_known_hosts` registration in `bash_completion` so installed
`dig.bash` can take effect instead of always completing only hostnames.

## Behavior (`completions-core/dig.bash`)
- **Switch-style dig**: query CLASS, RRTYPE, and NAME may appear in any order, plus `+query` options and `@server`
- **One class and one type** for completion purposes: after each is set (bare or via `-c` /`-t` flags) completion stops offering more classes or types. Meanwhile `+` options can be multiple.
- **Bare `ANY`**: classified as RRTYPE first (matches `dig <NAME> ANY`);
- **CHAOS + TXT**: suggests fixed `*.bind`-style names when class is `CH` and type is `TXT`.
- **Query names**: `-q`, dotted tokens, or non–class/type-prefix tokens use `_comp_compgen_known_hosts` where applicable.
- **`compopt +o bashdefault +o default`** to avoid broad default hostname fallback on empty **`COMPREPLY`**.
- **`@server`**: `~/.ssh/known_hosts`, or `BASH_COMPLETION_CMD_DIG_NS_HINTS_FILE` for suffix-based hints (longest pattern wins, `*` defaults; trailing dots on name/pattern normalized). Format documented in comments

### Example of `$BASH_COMPLETION_CMD_DIG_NS_HINTS_FILE`:
```shell
# Declare this in `.bashrc` as:
# export BASH_COMPLETION_CMD_DIG_NS_HINTS_FILE=<full_path_to_this_file>

# Glob useful to declare resolvers. Will match always as auto-completion
* 1.1.1.1 8.8.8.8 9.9.9.9 127.0.0.53

# Root zone and COM name servers
. a.root-servers.net. b.root-servers.net. c.root-servers.net. d.root-servers.net. e.root-servers.net. f.root-servers.net. g.root-servers.net. h.root-servers.net. i.root-servers.net. j.root-servers.net. k.root-servers.net. l.root-servers.net. m.root-servers.net.
com a.gtld-servers.net. b.gtld-servers.net. c.gtld-servers.net. d.gtld-servers.net. e.gtld-servers.net. f.gtld-servers.net. g.gtld-servers.net. h.gtld-servers.net. i.gtld-servers.net. j.gtld-servers.net. k.gtld-servers.net. l.gtld-servers.net. m.gtld-servers.net.

# Internal usage
internal.example.com 192.0.2.1 198.51.100.2 203.0.113.3
```

## Note on removing `dig` from `./bash_completion`

Given that `dig` was registered with `_comp_complete_known_hosts`, that generic hostname completer took precedence over the lazy-loaded `completions-core/dig.bash`, so dedicated completion for class, RR type, @server, and + options never ran.

The other option to avoid this case is to add into `.bashrc` or `.bash_profile` the load of the dig completion after loading bash-completion with `_comp_load` (example in macos with `brew` installation):

```shell
[[ -r "${BREWPREFIX:-/opt/homebrew}/etc/profile.d/bash_completion.sh" ]] && . "${BREWPREFIX:-/opt/homebrew}/etc/profile.d/bash_completion.sh"
_comp_load dig 2>/dev/null
```

## Files
| File | Change |
|------|--------|
| `completions-core/dig.bash` | New dig completion |
| `completions-core/Makefile.am` | Install `dig.bash` |
| `bash_completion` | Stop registering `dig` with `_comp_complete_known_hosts` |
| `test/t/test_dig.py` | Integration tests |
| `test/t/Makefile.am` | Register `test_dig.py` |

## Testing
- `pytest test/t/test_dig.py` (cases with `require_cmd=True` need **dig** on **PATH**).

### DIFF
--- a/bash_completion
+++ b/bash_completion
@@ -2934,8 +2934,9 @@ _comp_compgen_known_hosts__impl()
 
     _comp_compgen -v known_hosts -c "$prefix$cur" ltrim_colon "${known_hosts[@]}"
 }
+# dig: completed by completions-core/dig.bash (lazy-loaded), not generic hosts.
 complete -F _comp_complete_known_hosts traceroute traceroute6 \
-    fping fping6 telnet rsh rlogin ftp dig drill ssh-installkeys showmount
+    fping fping6 telnet rsh rlogin ftp drill ssh-installkeys showmount
 
 # Convert the word index in `words` to the index in `COMP_WORDS`.
 # @param $1           Index in the array WORDS.
--- a/completions-core/Makefile.am
+++ b/completions-core/Makefile.am
@@ -79,6 +79,7 @@ cross_platform = 2to3.bash \
 		desktop-file-validate.bash \
 		dhclient.bash \
 		dict.bash \
+		dig.bash \
 		dmypy.bash \
 		dnssec-keygen.bash \
 		dnsspoof.bash \
--- a/completions-core/dig.bash
+++ b/completions-core/dig.bash
@@ -145,7 +145,7 @@ _comp_cmd_dig__chaos_txt_names()
     return 0
 }
 
-# Append NS-hint targets from $BASH_COMPLETION_DIG_NS_HINTS_FILE for a probe
+# Append NS-hint targets from $BASH_COMPLETION_CMD_DIG_NS_HINTS_FILE for a probe
 # name ($_q_name). Plain-text lines: <pattern> <space-separated server names>
 # (no @ prefix; completion adds @). Lines starting with # are ignored.
 #
@@ -166,8 +166,8 @@ _comp_cmd_dig__chaos_txt_names()
 _comp_cmd_dig__ns_hints()
 {
     local fqdn=${1%.}
-    [[ -z ${BASH_COMPLETION_DIG_NS_HINTS_FILE-} ]] && return
-    [[ -r $BASH_COMPLETION_DIG_NS_HINTS_FILE ]] || return
+    [[ -z ${BASH_COMPLETION_CMD_DIG_NS_HINTS_FILE-} ]] && return
+    [[ -r $BASH_COMPLETION_CMD_DIG_NS_HINTS_FILE ]] || return
 
     local best_targets="" best_len=0 glob_targets=""
     local -A union_seen=()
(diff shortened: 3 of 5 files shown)


scop (commented):
Thank you for working on making bash completions better. But to verify, did you notice the two few paragraphs of CONTRIBUTING.md as linked in the PR creation view, in particular:

> However, before submitting a completion to us, first consider submitting it to the project that ships the commands your completion is for.

If you considered submitting it upstream, let's include the explanation why submit it here instead. If it was asked upstream and rejected there, let's add a link to the response.

mave007:
Still waiting for answer from current maintainer in https://gitlab.isc.org/isc-projects/bind9/-/issues/4879

I think the biggest issue we will face, is if that is uploaded on ISC end, then in this project we will still need to remove the lazy-load on `bash_completion` portion: https://github.com/scop/bash-completion/pull/1625/changes#diff-a4757074ff650000804fd3eaabe9b0a9e02e33040ca5b8afd4c0275fc5f3e136 

If that is not done in this project, it implies the user will need to do the following on their own shell to load it after the load of `bash_completion.sh`:

```bash
_comp_load dig 2>/dev/null
```

which it's not the end of the world... but definitely not intuitive.

akinomyoga:
> I think the biggest issue we will face, is

That's not really an issue. First of all, that "issue" is not specific to `dig`, and a similar situation has been occurring frequently. We have been moving in-tree completion settings to fallback completions when the upstream of the corresponding commands has started to offer their completion settings themselves.

Rencently, I also thought about related changes associated with #1558 and #1329. We may move all these `complete` settings into separate files in `completions-core` or `completions-fallback` so that they can be customized in a way described in https://github.com/scop/bash-completion/pull/1558#discussion_r2728425654 or https://github.com/scop/bash-completion/discussions/1329#discussioncomment-12163014. Also, currently, some completion files define completion settings for multiple commands at the same time, but that may override user-defined completion settings for a command when the completion setting of another command is loaded. To avoid such a situation, the definition of the completion functions and the declaration of `complete ...` should also be separated when one file declares multiple `complete ...`.

PR #1637, opened 2026-05-30
chore: remove most editor mode config comments

scop:
In particular, remove them from files with names containing a typical extension for a language. These should be unnecessary at best, but an unknown mode or alias can also prevent a file from being opened with the correct mode.

This is what happens currently with Zed 1.4.4, ref https://github.com/zed-industries/zed/pull/58124

Giant but trivial/comment-only PR.


yedayak (approved):
I don't really have the context of what uses these comments, but looks good to me

akinomyoga (approved):
I've thought GNU Emacs didn't recognize `.bash`, but I realized that the current version of Bash recognizes it. Perhaps it has recently added support, or I misunderstood.

PR #1639, opened 2026-06-02
Deprecate dpkg completion

### LINKED ISSUE #694: dpkg integration upstream

I'd like to add bash-completion support in dpkg upstream, I've got a branch starting work for that. The problem is that there's already some support for it here, but mixed with completions for dpkg-reconfigure, which is (very confusingly) not a dpkg program, but a debconf one. So to avoid stomping over the support for the debconf command, first it would need to be split into its own file. The dpkg support can remain I guess, for a transitional period, until people have upgraded to a dpkg version that ships with the completions.

I'm not sure how you handle upstream adding support when there was some already present here though?

- [ ] This is a request for new completion
- [ ] Link to upstream project query about shipping the completion:

guillemj:
To make it possible for dpkg to provide its own bash completion, the support here needs to be namespaced. In addition, the support for dpkg-reconfigure (which is actually provided by debconf not dpkg), needs to be split.

Because the dpkg completion provides public interfaces used by other completion scripts, the completion imported into dpkg will preserve those public functions for backwards compatibility for now. Ideally bash-completions would switch those functions into another file. But that's left for later to be able to unblock the integration with the next dpkg upstream release 1.23.8.

Fixes: #694

### DIFF
--- a/completions-core/alternatives.bash
+++ b/completions-core/alternatives.bash
@@ -2,7 +2,7 @@
 
 _comp_cmd_alternatives__installed()
 {
-    local i admindir
+    local i admindir=""
     # find the admin dir
     for i in alternatives rpm/alternatives; do
         [[ -d /var/lib/$i ]] && admindir=/var/lib/$i && break
--- a/completions-core/alternatives.bash
+++ b/completions-core/alternatives.bash
@@ -1,10 +1,10 @@
-# bash completion for update-alternatives
+# bash completion for alternatives
 
-_comp_cmd_update_alternatives__installed()
+_comp_cmd_alternatives__installed()
 {
     local i admindir
     # find the admin dir
-    for i in alternatives dpkg/alternatives rpm/alternatives; do
+    for i in alternatives rpm/alternatives; do
         [[ -d /var/lib/$i ]] && admindir=/var/lib/$i && break
     done
     for ((i = 1; i < cword; i++)); do
@@ -16,7 +16,7 @@ _comp_cmd_update_alternatives__installed()
     [[ -d $admindir ]] && _comp_compgen_split -- "$(command ls "$admindir")"
 }
 
-_comp_cmd_update_alternatives()
+_comp_cmd_alternatives()
 {
     local cur prev words cword comp_args
     _comp_initialize -- "$@" || return
@@ -49,7 +49,7 @@ _comp_cmd_update_alternatives()
                     _comp_compgen_filedir
                     ;;
                 2)
-                    _comp_cmd_update_alternatives__installed
+                    _comp_cmd_alternatives__installed
                     ;;
                 4)
                     # priority - no completions
@@ -63,7 +63,7 @@ _comp_cmd_update_alternatives()
                             _comp_compgen -- -W '--slave'
                             ;;
                         3)
-                            _comp_cmd_update_alternatives__installed
+                            _comp_cmd_alternatives__installed
                             ;;
                     esac
                     ;;
@@ -72,7 +72,7 @@ _comp_cmd_update_alternatives()
         --remove | --set)
             case $args in
                 1)
-                    _comp_cmd_update_alternatives__installed
+                    _comp_cmd_alternatives__installed
                     ;;
                 2)
                     _comp_compgen_filedir
@@ -80,7 +80,7 @@ _comp_cmd_update_alternatives()
             esac
             ;;
         --auto | --remove-all | --display | --config)
-            _comp_cmd_update_alternatives__installed
+            _comp_cmd_alternatives__installed
             ;;
         *)
             _comp_compgen_help - <<<"$(LANG=C "$1" --help 2>&1 | command sed '
@@ -92,4 +92,4 @@ _comp_cmd_update_alternatives()
             ;;
     esac
 } &&
-    complete -F _comp_cmd_update_alternatives update-alternatives alternatives
+    complete -F _comp_cmd_alternatives alternatives
--- a/completions-core/alternatives.bash
+++ b/completions-core/alternatives.bash
@@ -13,7 +13,7 @@ _comp_cmd_alternatives__installed()
             break
         fi
     done
-    [[ -d $admindir ]] && _comp_compgen_split -- "$(command ls "$admindir")"
+    [[ -d $admindir ]] && _comp_compgen -C "$admindir" -- -fX '*/*'
 }
 
 _comp_cmd_alternatives()
--- a/completions-core/dpkg-reconfigure.bash
+++ b/completions-core/dpkg-reconfigure.bash
@@ -1,0 +1,34 @@
+# dpkg-reconfigure(1) completion
+
+_comp_cmd_dpkg_reconfigure()
+{
+    local cur prev words cword comp_args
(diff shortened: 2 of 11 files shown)


guillemj:
Sorry, just noticed now that this was against the now stale master branch, will rebase tomorrow against main instead and force push.

akinomyoga:
We've already moved those files so that no conflicts happen. This PR seems to rename non-existent files, though I'm not sure why GitHub interface doesn't report conflicts. You base on an old commit between 2.13.0--2.14.0. If you intend to submit a patch to a certain version packaged in a distribution, could you please submit the patch to the package maintainer? If you intend to submit it here, could you please rebase the branch on top of the latest `main` branch and re-examine the changes that you try to add?

guillemj:
> We've already moved those files so that no conflicts happen. This PR seems to rename non-existent files, though I'm not sure why GitHub interface doesn't report conflicts. You base on an old commit between 2.13.0--2.14.0. If you intend to submit a patch to a certain version packaged in a distribution, could you please submit the patch to the package maintainer? If you intend to submit it here, could you please rebase the branch on top of the latest `main` branch and re-examine the changes that you try to add?

Right, see my comment which seems crossed at the same time as yours. I've now rebased it and force pushed.

completions-fallback/update-alternatives.bash
```diff
@@ -1,5 +1,8 @@
 # bash completion for update-alternatives
 
+# Use of this file is deprecated.  Upstream completion is available in
+# dpkg >= 1.23.8, use that instead.
```
scop: I think we should rephrase this some, as it's likely not correct advice for example for Fedora systems -- alternatives and update-alternatives on it come from the alternatives package, which originates from https://github.com/fedora-sysv/chkconfig, and it does not come with any completions for either.

guillemj: Ah, was aware of chkconfig, but didn't recall they also provided a compat symlink for update-alternatives. Perhaps the new comment addresses this?

completions-core/dpkg-reconfigure.bash
scop, line 1: This should be named dpkg-reconfigure.bash -- it does not seem to do anything particularly useful when sourced on demand on `debconf <TAB>`, and on the other hand as it stands, it won't be sourced for `dpkg-reconfigure <TAB>`. Let's correct the top level comment while at it.

scop (changes requested):
Another partial review round here, will need to look into this some more still.

As a general note, it seems this PR combines a number of changes in one commit that could have been split to multiple standalone/followup commits. It would be much easier to review it if was submitted as multiple logical commits (could review one commit at a time).

Because the single commit description needs an update anyway (conventional commits, dpkg vs non-dpkg system considerations / some of these are not deprecated everywhere), maybe look into splitting it, too?

completions-core/dpkg-reconfigure.bash
guillemj: I'll try to split it up, this was just carried over from what I had locally, and perhaps that was more of a draft at the time, and when I prepared it for submission it just all seemed to be tied together.

For the debconf name, I think I went with project name as grouping, but I see that does not make sense as you point out it would not trigger any auto-loading. Will rename.

scop: Thanks for the split, will take another look.

The commit messages still don't follow conventional commits, search for it in CONTRIBUTING.md for more info and why getting this right is important. But I can fix it up later if that's the only remaining straggler.

scop (approved):
I split the completions somewhat further and looked through them, looks good to me. Adjusted commit messages while at that. Hopefully the 1.23.8 version timeline will hold :)

Approved as far as I'm concerned, but would not mind another look by @akinomyoga or @yedayak if you happen to have time. If not, I'll merge shortly (assuming CI passes).

completions-core/alternatives.bash
```diff
@@ -0,0 +1,95 @@
+# bash completion for alternatives
+
+_comp_cmd_alternatives__installed()
+{
+    local i admindir
```
akinomyoga: For `localvar_inherit`, local variables need to be explicitly initialized.

```suggestion
    local i admindir=""
```

completions-core/alternatives.bash
```diff
@@ -0,0 +1,95 @@
...
+    for i in alternatives rpm/alternatives; do
+        [[ -d /var/lib/$i ]] && admindir=/var/lib/$i && break
+    done
+    for ((i = 1; i < cword; i++)); do
+        if [[ ${words[i]} == --admindir ]]; then
+            admindir=${words[i + 1]}
+            break
+        fi
+    done
+    [[ -d $admindir ]] && _comp_compgen_split -- "$(command ls "$admindir")"
```
akinomyoga: This just lists the filenames, so there is no need to use the external executable `ls`. I thought `_comp_compgen -C "$admindir" filedir` would work, but I realized that it would also generate a slash `/` when the word matches a directory name.

<details><summary>(Old suggestions)</summary>

One may first call `_comp_compgen_filedir` with an empty `cur` and then filter the results:

```suggestion
    if [[ -d $admindir ]]; then
        local installed
        _comp_compgen -Rv installed -C "$admindir" filedir &&
            _comp_compgen -U installed -- -W '"${installed[@]}"'
    fi
```

Or, actually, it may be simpler to use `_comp_expand_glob`.

```suggestion
    if [[ -d $admindir ]]; then
        local installed
        _comp_expand_glob installed '"$admindir"/*' &&
            _comp_compgen -U installed -- -W '"${installed[@]#"$admindir/"}"'
    fi
```
</details>

**Edit**: Actually, this can be

```suggestion
    [[ -d $admindir ]] && _comp_compgen -C "$admindir" -- -fX '*/*'
```

completions-core/alternatives.bash
```diff
@@ -0,0 +1,95 @@
...
+                    case $((args % 4)) in
+                        0 | 2)
+                            _comp_compgen_filedir
+                            ;;
+                        1)
+                            _comp_compgen -- -W '--slave'
+                            ;;
+                        3)
+                            _comp_cmd_alternatives__installed
+                            ;;
```
akinomyoga, line 67: As far as I see [the command-line parser of the redhat implementation](https://github.com/fedora-sysv/chkconfig/blob/d00ad17ee7a8b8ac4f28d01a7edf3a8ca0d88af4/alternatives.c#L1394-L1503), `alternatives` doesn't fix the orderings of the command-line options. In addition, there are extensions `--family <1 arg>`, `--initscript <1 arg>`, and `--follower <3 args>`. The orderings in the man page must be only for presentation.

We should explicitly see the nearest option name to generate the command line arguments.

akinomyoga (changes requested):

scop:
I'd like to release 2.18.0 soon, basically soon after https://github.com/scop/bash-completion/pull/1558 is merged. Would like to get this in as well if we can get the final bits ironed out and merged soonish.

completions-core/alternatives.bash
scop: I think this was not introduced in this PR, so I've filed it as https://github.com/scop/bash-completion/issues/1678, let's not delay merging because of this.

PR #1643, opened 2026-06-13
feat(ty): add 3rd party fallback completion loader

electricbrass:
I saw uv, uvx, and ruff had the fallbacks here already but not ty, so just wanted to get that one in here too.


akinomyoga (changes requested):
`CLEANFILES` in the `Makefile.am` should also be updated.

akinomyoga (approved):
Thank you.

yedayak (approved):

PR #1651, opened 2026-06-27
refactor(arch,gcc,pkg-get,screen): reduce extra indentation

akinomyoga:
This is a trivial fix of the change of indentation. See the commit message.

For the diff without the whitespace change, see [this](https://github.com/scop/bash-completion/pull/1651/changes?w=1) (with the `?w=1` query in the URL).

### DIFF
--- a/completions-core/arch.bash
+++ b/completions-core/arch.bash
@@ -2,46 +2,47 @@
 
 # Try to detect whether this is the mailman "arch" to avoid installing
 # it for the coreutils/util-linux-ng one.
-_comp_have_command mailmanctl &&
-    _comp_cmd_arch()
-    {
-        local cur prev words cword was_split comp_args
-        _comp_initialize -s -- "$@" || return
+_comp_have_command mailmanctl || return 1
 
-        case $prev in
-            -w | -g | -d | --welcome-msg | --goodbye-msg | --digest)
-                _comp_compgen -- -W 'y n'
-                return
+_comp_cmd_arch()
+{
+    local cur prev words cword was_split comp_args
+    _comp_initialize -s -- "$@" || return
+
+    case $prev in
+        -w | -g | -d | --welcome-msg | --goodbye-msg | --digest)
+            _comp_compgen -- -W 'y n'
+            return
+            ;;
+        --file)
+            _comp_compgen_filedir
+            return
+            ;;
+    esac
+
+    [[ $was_split ]] && return
+
+    if [[ $cur == -* ]]; then
+        _comp_compgen_help
+    else
+        local args=$cword
+        for ((i = 1; i < cword; i++)); do
+            if [[ ${words[i]} == -* ]]; then
+                ((args--))
+            fi
+        done
+        case $args in
+            1)
+                # Prefer `list_lists` in the same dir as command
+                local pathcmd
+                pathcmd=$(type -P -- "$1") && local PATH=${pathcmd%/*}:$PATH
+                _comp_compgen -x list_lists mailman_lists
                 ;;
-            --file)
+            2)
                 _comp_compgen_filedir
-                return
                 ;;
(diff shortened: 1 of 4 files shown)


scop (commented):
If we are doing this (meaning those pkg-get and screen changes primarily), we could as well go through all our files and simplify
```
_comp_cmd_something() {
    # ...
} &&
    complete -F _comp_cmd_something something
```
...to
```
_comp_cmd_something() {
    # ...
}
complete -F _comp_cmd_something something
```

Strictly speaking this is a semantic change though, kicks in at least if the function to be defined was previously set as readonly. But I don't know if we need to care about this scenario or if there are some "more realistic" use cases which we'd care about more.

scop (approved):
Ah I see you noted the same thing already in the commit message. Good to go as far as I'm concerned, unless anyone has second thoughts.

As far as going through all our files, if we want to do it, that can be left to another MR.

akinomyoga:
> If we are doing this (meaning those pkg-get and screen changes primarily), we could as well go through all our files and simplify
> 
> ```
> _comp_cmd_something() {
>     # ...
> } &&
>     complete -F _comp_cmd_something something
> ```
> 
> ...to
> 
> ```
> _comp_cmd_something() {
>     # ...
> }
> complete -F _comp_cmd_something something
> ```

Yeah, this is also what I was wondering about the necessity. With the current setup, if there is a function-definition error, it anyway outputs error messages like:

```console
bash: line 123: _comp_cmd_something: readonly function
```

So, if one really wants to care about the case where the function definition fails, the code needs to be globally updated to use

```bash
{
    _comp_cmd_something() {
        # ...
    } &&
        complete -F _comp_cmd_something something
} 2>/dev/null
```

which is kind of overkill.

Then, it's simpler to just put the function definition and the `complete` setting at the top level flatly. Since we now use the name `_comp_cmd_<cmd>` instead of `_<cmd>`, the conflict probability is significantly lower or effectively zero now. The only case is the one where another completion framework defines the completion function.

> As far as going through all our files, if we want to do it, that can be left to another MR.

Yes.

scop:
It's also not 100% clear if we should even avoid installing a completion if the definition is read only or just go ahead with it, especially given the very low probability of it conflicting something entirely unrelated. But it would be good to be consistent about this.

Syntax errors in function definitions will cause an exit anyway before the `&&` is reached.

Not sure if there are some other cases besides readonly where the `&&` could kick in on function definitions.

akinomyoga:
> It's also not 100% clear if we should even avoid installing a completion if the definition is read only or just go ahead with it, especially given the very low probability of it conflicting something entirely unrelated. But it would be good to be consistent about this.

Currently, we are mostly consistent with `_comp_cmd_something() {...} && complete -F _comp_cmd_something something`, but a small inconsistency seems to lie in `bash_completion`, and a few files stemmed from there:

```console
$ git grep -E '^complete -F'
bash_completion:2959:complete -F _comp_complete_known_hosts traceroute traceroute6 \
bash_completion:3170:complete -F _comp_command aoss command "do" else eval exec ltrace nice nohup padsp \
bash_completion:3180:complete -F _comp_root_command fakeroot gksu gksudo kdesudo really
bash_completion:3254:complete -F _comp_complete_longopt \
bash_completion:3393:complete -F _comp_complete_minimal ''
completions-core/cd.bash:80:complete -F _comp_cmd_cd -o nospace cd pushd
completions-fallback/mtr.bash:6:complete -F _comp_complete_known_hosts mtr
```

> Syntax errors in function definitions will cause an exit anyway before the `&&` is reached.

We are supposed to check the syntax before the push, but the user may still flip `extglob`, which is an issue.

> Not sure if there are some other cases besides readonly where the `&&` could kick in on function definitions.

I'm not sure about other cases either.

PR #1662, opened 2026-06-28
feat(_comp_compgen_filedir): support `filedir -X`

akinomyoga:


bash_completion
```diff
@@ -1289,11 +1289,14 @@ _comp_quote_compgen()
 #           Ignored with `-d`.
 # OPTIONS
 #   -d      Complete only on directories
-
 #   -f      Perform `compopt -o filenames` modifications manually.  This
 #           suffixes a slash to a directory name.  When "-C <dir>" or "-P
 #           <prefix>" is specified to the "_comp_compgen" call, this option
 #           is automatically turned on.
+#   -X PATTERN
+#           Specify the filename pattern to exclude, or the filename pattern to
```
scop: Could add `@since 2.18` here some way.

bash_completion
```diff
@@ -1289,11 +1289,14 @@ _comp_quote_compgen()
...
+#           include prefixed by "!".  When this option is specified, $1 should
+#           not be specified or $1 should be empty.
```
scop:
```suggestion
#           include prefixed by "!".  When this option is specified, $1 should
#           be empty or not specified.
```

bash_completion
```diff
@@ -1543,20 +1551,7 @@ _comp_variable_assignments()
...
-                        unset -v 'COMPREPLY[i]'
-                        continue
-                    elif [[ -d ${COMPREPLY[i]} ]]; then
-                        COMPREPLY[i]+=/
-                        compopt -o nospace
-                    fi
-                    COMPREPLY[i]=${COMPREPLY[i]#/usr/share/zoneinfo/}
-                done
-            fi
+            _comp_compgen -C /usr/share/zoneinfo filedir -fX '*.tab'
```
scop: Not introduced here, but I noticed on Ubuntu 26.04 this should also exclude at least `leapseconds`, `leap-seconds.list` and `tzdata.zi`. I'm fine with it if we want to smuggle that in as another followup commit.

scop (approved):
Nice. Pre-approved with a few minor items addressed.

bash_completion
akinomyoga: I added a commit 97487aaef.

bash_completion
akinomyoga: I'm not sure about the ideal way to put it, but I've just put it at the beginning in [`6937585..97487aa`](https://github.com/scop/bash-completion/compare/693758579edb44e46100ef431f19860b69cbcbe9..97487aaef3eac1c5adda261721ec85d00d91d777). If we would parse those doc comments in the future, we may need to think about how to scope the effect of the directives, and how to properly parse them. However, we can later adjust them when we introduce a parser.

bash_completion
akinomyoga: i've applied.

scop (approved):

bash_completion
scop: LGTM

scop:
Thanks!

PR #1664, opened 2026-06-29
feat(codex): 3rd party fallback completion loader

scop:
https://developers.openai.com/codex/cli/features#shell-completions

### DIFF
--- a/completions-fallback/.gitignore
+++ b/completions-fallback/.gitignore
@@ -31,6 +31,7 @@
 /cloudquery.bash
 /clusterctl.bash
 /cmctl.bash
+/codex.bash
 /colima.bash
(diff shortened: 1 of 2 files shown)


completions-fallback/.gitignore
```diff
@@ -38,6 +38,7 @@
 /constellation.bash
 /consul.bash
 /container-structure-test.bash
+/codex.bash
```
akinomyoga: I think co"d"ex should come earlier in the alphabetical order. The same for the two cases in Makefile.am.

akinomyoga (approved):
Otherwise, approving.

scop:
Oops, thanks!

PR #1666, opened 2026-06-29
feat(encore): 3rd party fallback completion loader

scop:
https://encore.dev

### DIFF
--- a/completions-fallback/.gitignore
+++ b/completions-fallback/.gitignore
@@ -62,6 +62,7 @@
 /dprint.bash
 /driftctl.bash
 /dyff.bash
+/encore.bash
 /esc.bash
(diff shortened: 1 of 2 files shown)


akinomyoga (commented):
> https://encore.dev

I cannot find an explicit mention about the usage `encore completion bash`, but I could find the following installation script.

https://github.com/encoredev/homebrew-tap/blob/8a6de5e6d6bba8f4d8156e97e6ea7af90c6e1ee8/Formula/encore.rb#L34-L36

I think this can be mentioned in the commit message as well.

akinomyoga (approved):
Otherwise, pre-approving.

scop:
Thanks!

PR #1671, opened 2026-07-01
docs(README): clarify that we provide completion contents

### LINKED ISSUE #1623: Docs - demo / better description
The readme doesn't really tell the user what this can do. After reading it back and forth, I still don't know what it can do for me, whether this is what I'm looking for or not. I'd say that it could use improvement - whether it's a better worded description (on the very top) or a visual demo (video/gif).

No, "completion" doesn't cut it. There are countless types of completions, history searches, fish like behaviours, you name it.

akinomyoga:
Closes #1623

In the GitHub interface, it may be harder to identify the changes in the paragraph with the linewise diff, so here are the edits indicated by `<del></del><ins></ins>`:

```md
## Introduction

bash-completion is a collection of <ins>command-specific</ins> command<del> </del><ins>-</ins>line
<del>command</del> completions<ins>, providing completion contents</ins> for the [Bash
shell](https://www.gnu.org/software/bash/)<del>,</del><ins>'s completion system.  This also
provides a</ins> collection of helper functions to assist in creating new completions,
and <ins>a</ins> set of facilities for loading completions automatically on demand, as well
as installing them.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,13 +5,14 @@
 
 ## Introduction
 
-bash-completion is a collection of recipes for [command-line
-completions](https://en.wikipedia.org/wiki/Command-line_completion) of a number
-of commands for the [Bash shell](https://www.gnu.org/software/bash/)'s
-completion system.  This enhances Bash's plain completion of filenames to
-provide better suggestions of options and arguments for many commands.  Also
-provided is a collection of helper functions to assist in creating new
-completions, and a set of facilities for loading completions automatically on
+bash-completion is a collection of
+[command-line completion](https://en.wikipedia.org/wiki/Command-line_completion)
+recipes for a number of commands for the completion system of the
+[Bash shell](https://www.gnu.org/software/bash/).
+These recipes enhance Bash's typical plain completion of filenames to
+provide better suggestions of options and arguments for many commands.
+Also provided is a collection of helper functions to assist in creating new
+recipes, and a set of facilities for loading recipes automatically on
 demand, as well as installing them.
 
 ## Installation
--- a/README.md
+++ b/README.md
@@ -5,10 +5,12 @@
 
 ## Introduction
 
-bash-completion is a collection of command line command completions for the
-[Bash shell](https://www.gnu.org/software/bash/), collection of helper
-functions to assist in creating new completions, and set of facilities for
-loading completions automatically on demand, as well as installing them.
+bash-completion is a collection of recipes for [command-line
+completions](https://en.wikipedia.org/wiki/Command-line_completion) of a number
+of commands for the [Bash shell](https://www.gnu.org/software/bash/)'s
+completion system.  Also provided is a collection of helper functions to assist
+in creating new completions, and a set of facilities for loading completions
+automatically on demand, as well as installing them.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


scop (changes requested):
The concept of "completion contents" reads alien to me, this is the first time I've heard it, and I think I wouldn't understand what it means if I wasn't so deeply involved with these things (and even if I am, I'm still not quite sure).

But yeah, "collection of command line command completions" really is a mouthful. Would "collection of specific completions for a number of commands" be any better?

Maybe it'd also be helpful if we'd link to a general page describing the concept of completions, e.g. https://en.wikipedia.org/wiki/Command-line_completion?

akinomyoga:
> The concept of "completion contents" reads alien to me, this is the first time I've heard it, and I think I wouldn't understand what it means if I wasn't so deeply involved with these things (and even if I am, I'm still not quite sure).

I haven't actually heard "completion contents", but I thought it would make sense. Somehow #1623 seems to misunderstand this project to provide the interface of the completion system, but this project doesn't provide a completion system. Rather, this project provides the words and strings that would be completed by the completion system. It's not like the medium of a dictionary, such as a book, but it's like the dictionary contents (though ChatGPT told me that it should be a singular form, "the dictionary content" in this case). That was my idea. For example, if I rephrase it as the singular form `completion content`, does that make sense to you? Or the connected form "completion content" might sound like an established idiom, so `content of completion` may be better?

> But yeah, "collection of command line command completions" really is a mouthful. Would "collection of specific completions for a number of commands" be any better?

That is better than the original wording. I updated to use that form.

However, that's actually orthogonal to #1623, and doesn't clarify what confuses #1623. For now, I removed the phrase `, providing completion contents` but do you have a better alternative for `contents`?

> Maybe it'd also be helpful if we'd link to a general page describing the concept of completions, e.g. https://en.wikipedia.org/wiki/Command-line_completion?

I've added a link. However, this doesn't solve #1623 either. #1623 doesn't ask about the concept of completions.

README.md
```diff
@@ -5,10 +5,12 @@
...
 ## Introduction
 
-bash-completion is a collection of command line command completions for the
-[Bash shell](https://www.gnu.org/software/bash/), collection of helper
-functions to assist in creating new completions, and set of facilities for
-loading completions automatically on demand, as well as installing them.
+bash-completion is a collection of specific [command-line
+completions](https://en.wikipedia.org/wiki/Command-line_completion) for a
+number of commands for the [Bash shell](https://www.gnu.org/software/bash/)'s
+completion system.  This also provides a collection of helper functions to
```
scop:
```suggestion
completion system.  Also provided is a collection of helper functions to
```

scop:
To me, these things have always been just "completions", but I accept that might not be too descriptive to others. Maybe "completion recipes"? Or "completion implementations"?

We could also try and elaborate by adding another sentence about how after installing this, the shell's tab completion starts to provide better suggestions for many commands' options and arguments than before.

akinomyoga:
Thank you. *Recipes* sound good, while *completion implementation* sounds to me like an implementation of the mechanism of completing the word in the command line.  I've adjusted the description using *recipes*. 

> We could also try and elaborate by adding another sentence about how after installing this, the shell's tab completion starts to provide better suggestions for many commands' options and arguments than before.

I added a sentence in commit edd6cc616, but I'm not sure if that's what you assumed. You can actually rewrite the sentence.

README.md
akinomyoga: I applied.

scop (approved):
Thanks! I took a stab at simplifying it some more and use the word "recipe" more consistently in 220a30fb82010b9c6daab17362afcb92c9e07d23, good to go with this as far as I'm concerned.

akinomyoga:
Thank you!

PR #1683, opened 2026-07-08
feat(sudo): option arg (non-)completion improvements

scop:


completions-core/sudo.bash
```diff
@@ -23,7 +23,7 @@ _comp_cmd_sudo()
                 break
             fi
             # shellcheck disable=SC2254
-            [[ ${words[i]} == @(--@(close-from|chdir|group|host|prompt|chroot|role|command-timeout|type|other-user|user)|-${noargopts}[CDghpRrTtUu]) ]] &&
+            [[ ${words[i]} == @(--@(close-from|chdir|group|host|prompt|chroot|role|command-timeout|type|other-user|user|preserve-env)|-${noargopts}[CDghpRrTtUu]) ]] &&
```
akinomyoga: I think the ordering of the list of the long option names is roughly based on the ordering in the man page. Then, I think `preserve-env` should be inserted after `chdir` (based on the man page in sudo-1.9.15p5). For the same reason, I think `command-timeout` should be moved after `other-user` (or after `type` if one bases on `sudo --help` instead of the man page). Another option is to sort them in lexicographical order.

completions-core/sudo.bash
```diff
@@ -37,8 +37,9 @@ _comp_cmd_sudo()
             _comp_compgen -- -g
             return
             ;;
-        --close-from | --prompt | --role | --type | --command-timeout | \
-            -${noargopts}[CprTt])
+        --close-from | --help | --prompt | --role | --type | \
+            --command-timeout | --preserve-env | --version | \
```
akinomyoga: Similarly, `--preserve-env` should be inserted after `--close-from`.

akinomyoga (approved):
pre-approving with the following points.

completions-core/sudo.bash
scop: I have not compared the two, but the ordering seemed to be roughly that of the `--help` output, and `--preserve-end` is at the end with sudo-rs so it ended up there.

Reordered all of them in 80156e31e9e8df76fe9468abfd0236b496782875, merging with that.

PR #1686, opened 2026-07-09
feat(curl): add more option arg (non-)completions

scop:
From curl 8.18.0.


completions-core/curl.bash
scop, line 1: Will wrap long lines before merge, left undone for review ease.

yedayak (approved):

PR #1693, opened 2026-07-19
fix(make): keep plain target names that are also path prefixes

### LINKED ISSUE #1219: [Regression] make autocompletion skips install
## To reproduce

0. rename the attached file to Makefile (github won't let me upload it with it's correct name) and put it in a folder
1. cd to 'directory_that_contains
[Makefile.txt](https://github.com/user-attachments/files/15877833/Makefile.txt)
_attached_makefile'
2. Type 'make inst' and press the TAB key
3. See that it autocompletes to make install/

## Expected behavior

It should autocomplete to make install like it used to (given that there is an install target in the Makefile)
 
- [ ] Operating system name/distribution and version: ArchLinux
- [ ] bash version, `echo "$BASH_VERSION"`: 5.2.26(1)-release
- [ ] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.14.0

## Additional context

This used to work with bash completion 2.11

15daksh-2003:
The make completion collapses /-separated target names as filesystem paths, so a target that is also the prefix of others (install with install/local, X with X/fast) is dropped and only its /-suffixed siblings are offered. 

Fixes #1219 

Approach:

- In _comp_cmd_make__truncate_non_unique_paths, record the candidates make actually reported in an is_target map (keyed on the original candidate). In the emit loop, additionally offer a candidate's plain form when it is both a reported target and a path node with children. Synthesized ancestor nodes (pure prefixes make never reported) are not in is_target, so the recursive-Makefile collapsing from #544/#858 is unchanged.
- Coupled nospace fix: a plain target and its dir/ prefix can now coexist in an unspecified COMPREPLY order. Set compopt -o nospace when any candidate (not just the first) ends in /, so drilling into the prefix keeps working.
- Known limitation (pre-existing): a target whose name literally ends in / is still not surfaced.

### DIFF
--- a/completions-core/make.bash
+++ b/completions-core/make.bash
@@ -26,6 +26,12 @@ _comp_cmd_make__extract_targets()
 # "abc/" instead of generating both candidates directly.  When there is only
 # one candidate "abc/def", we generate the full path "abc/def".
 #
+# The phony targets (prefixed by "phony:") are offered as-is: a `/' in a phony
+# name is a naming convention (e.g. CMake's `install', `install/local'), not a
+# directory boundary, so they are never collapsed.  Only the non-phony targets
+# (prefixed by "file:"), which are real filepaths, go through the path
+# collapse.
+#
 # @var[in] cur
 # @var[in] mode
 # @var[in] targets - Array containing the target names.  This array is supposed
@@ -36,13 +42,24 @@ _comp_cmd_make__truncate_non_unique_paths()
     local prefix=$cur
     [[ $mode == -d ]] && prefix=
 
+    COMPREPLY=()
+    local nreply=0
+
     # collect the possible completions including the directory names in
     # `paths' and count the number of children of each subdirectory in
     # `nchild'.
     local -A paths nchild
     local target
     for target in "${targets[@]}"; do
-        local path=${target%/}
+        if [[ $target == phony:* ]]; then
+            # The phony targets are directly added to COMPREPLY without
+            # considering the pathname structures with slashes.
+            COMPREPLY[nreply++]=${target#phony:}
+            continue
+        fi
+
+        local path=${target#file:}
+        path=${path%/}
         while [[ ! ${paths[$path]+set} ]] &&
             paths[$path]=set &&
             [[ $path == "$prefix"*/* ]]; do
@@ -51,8 +68,6 @@ _comp_cmd_make__truncate_non_unique_paths()
         done
     done
 
-    COMPREPLY=()
-    local nreply=0
     for target in "${!paths[@]}"; do
         # generate only the paths that do not have a unique child and whose
         # all parent and ancestor directories have a unique child.
--- a/completions-core/make.bash
+++ b/completions-core/make.bash
@@ -18,54 +18,55 @@ _comp_cmd_make__extract_targets()
     _comp_awk -f "${BASH_SOURCE[0]%/*}/../helpers-core/make-extract-targets.awk"
 }
 
-# Truncate the non-unique filepaths in COMPREPLY to only generate unique
-# directories or files.  This function discards the files under subdirectories
-# unless the path is unique under each subdirectory and instead generate the
-# subdirectory path.  For example, when there are two candidates, "abc/def" and
-# "abc/xyz", we generate "abc/" instead of generating both candidates directly.
-# When there is only one candidate "abc/def", we generate the full path
-# "abc/def".
+# Truncate the non-unique filepaths in the "targets" array to only generate
+# phony targets, unique directories, or files in the COMPREPLY array.  This
+# function discards the files under subdirectories unless the path is unique
+# under each subdirectory and instead generate the subdirectory path.  For
+# example, when there are two candidates, "abc/def" and "abc/xyz", we generate
+# "abc/" instead of generating both candidates directly.  When there is only
+# one candidate "abc/def", we generate the full path "abc/def".
 #
 # @var[in] cur
 # @var[in] mode
-# @var[in,out] COMPREPLY
+# @var[in] targets - Array containing the target names.  This array is supposed
+#   to contain at least one element for the Bash-4.2 nounset bug.
+# @var[out] COMPREPLY
 _comp_cmd_make__truncate_non_unique_paths()
 {
     local prefix=$cur
     [[ $mode == -d ]] && prefix=
-    if ((${#COMPREPLY[@]} > 0)); then
-        # collect the possible completions including the directory names in
-        # `paths' and count the number of children of each subdirectory in
-        # `nchild'.
-        local -A paths nchild
-        local target
-        for target in "${COMPREPLY[@]}"; do
-            local path=${target%/}
-            while [[ ! ${paths[$path]+set} ]] &&
-                paths[$path]=set &&
-                [[ $path == "$prefix"*/* ]]; do
-                path=${path%/*}
-                nchild[$path]=$((${nchild[$path]-0} + 1))
-            done
+
+    # collect the possible completions including the directory names in
+    # `paths' and count the number of children of each subdirectory in
+    # `nchild'.
+    local -A paths nchild
+    local target
+    for target in "${targets[@]}"; do
+        local path=${target%/}
+        while [[ ! ${paths[$path]+set} ]] &&
+            paths[$path]=set &&
+            [[ $path == "$prefix"*/* ]]; do
+            path=${path%/*}
+            nchild[$path]=$((${nchild[$path]-0} + 1))
+        done
+    done
+
+    COMPREPLY=()
+    local nreply=0
+    for target in "${!paths[@]}"; do
+        # generate only the paths that do not have a unique child and whose
+        # all parent and ancestor directories have a unique child.
+        ((${nchild[$target]-0} == 1)) && continue
+        local path=$target
+        while [[ $path == "$prefix"*/* ]]; do
+            path=${path%/*}
+            ((${nchild[$path]-0} == 1)) || continue 2
         done
 
-        COMPREPLY=()
-        local nreply=0
-        for target in "${!paths[@]}"; do
-            # generate only the paths that do not have a unique child and whose
-            # all parent and ancestor directories have a unique child.
-            ((${nchild[$target]-0} == 1)) && continue
-            local path=$target
-            while [[ $path == "$prefix"*/* ]]; do
-                path=${path%/*}
-                ((${nchild[$path]-0} == 1)) || continue 2
-            done
-
-            # suffix `/' when the target path is a subdiretory, which has
-            # at least one child.
-            COMPREPLY[nreply++]=$target${nchild[$target]+/}
-        done
-    fi
+        # suffix `/' when the target path is a subdiretory, which has
+        # at least one child.
+        COMPREPLY[nreply++]=$target${nchild[$target]+/}
+    done
 }
 
 _comp_cmd_make()
@@ -154,12 +155,12 @@ _comp_cmd_make()
         #     mode=-d # display-only mode
         # fi
 
-        _comp_split COMPREPLY "$(LC_ALL=C \
+        local targets
+        _comp_split targets "$(LC_ALL=C \
             $1 -npq __BASH_MAKE_COMPLETION__=1 \
             ${makef+"${makef[@]}"} "${makef_dir[@]}" .DEFAULT 2>/dev/null |
-            _comp_cmd_make__extract_targets "$mode" "$cur")"
-
-        _comp_cmd_make__truncate_non_unique_paths
+            _comp_cmd_make__extract_targets "$mode" "$cur")" &&
+            _comp_cmd_make__truncate_non_unique_paths
 
         if [[ $mode != -d ]]; then
             # Completion will occur if there is only one suggestion
--- a/helpers-core/make-extract-targets.awk
+++ b/helpers-core/make-extract-targets.awk
@@ -3,6 +3,8 @@
 # This AWK script is used by the function `_comp_cmd_make__extract_targets` in
 # `completions-core/make`.  This script receives the output of `make -npq' as
 # the input file or stdin and outputs the list of targets matching the prefix.
+# The phony targets are prefixed by "phony:", and the others are prefixed by
+# "file:".
 #
 # @env prefix         Specifies the prefix to match.
 # @env prefix_replace Specifies the string that replaces the prefix in the
@@ -14,6 +16,7 @@ BEGIN {
   prefix = ENVIRON["prefix"];
   prefix_replace = ENVIRON["prefix_replace"];
   is_target_block = 0;
+  is_phony_target = 0;
   target = "";
 }
 
@@ -40,6 +43,7 @@ NR == 1, /^# +Make data base/ { next; }
   #
   # https://github.com/scop/bash-completion/pull/1577
   is_target_block = 0;
+  is_phony_target = 0;
   target = "";
   next;
 }
@@ -55,17 +59,34 @@ is_target_block == 0 { next; }
 
 /^# +File is an intermediate prerequisite/ { # cancel the block
   is_target_block = 0;
+  is_phony_target = 0;
   target = "";
+  next;
+}
+
+# The comment sections after the .PHONY targets include the following line,
+# which was confirmed in make-3.80 (2002) and make-4.4.90 (master b380278,
+# 2026):
+#
+# #  Phony target (prerequisite of .PHONY).
+#
+/^# +Phony target/ {
+  is_phony_target = 1;
   next;
 }
 
 # end of target block
 /^$/ {
-  is_target_block = 0;
   if (target != "") {
-    print target;
+    if (is_phony_target) {
+      print "phony:" target;
+    } else {
+      print "file:" target;
+    }
     target = "";
   }
+  is_target_block = 0;
+  is_phony_target = 0;
   next;
 }
 
--- a/test/fixtures/make/test3/Makefile
+++ b/test/fixtures/make/test3/Makefile
@@ -1,0 +1,22 @@
+# CMake "Unix Makefiles" style: phony targets whose names use `/' as a naming
+# convention, not a directory boundary.  `make -p' marks them phony, so the
+# completion offers them as-is instead of collapsing `install' into `install/'
+# or dropping the plain `MyProgram' in favor of only `MyProgram/fast'.
+#
+#   - single-child case: `MyProgram' alongside `MyProgram/fast'
+#   - multi-child case:  `install' alongside `install/local' and `install/strip'
+
+all: MyProgram
+.PHONY: all
+
+MyProgram MyProgram/fast:
+	@echo $@
+.PHONY: MyProgram MyProgram/fast
+
+clean clean/fast:
+	@echo $@
+.PHONY: clean clean/fast
+
+install install/local install/strip:
+	@echo $@
+.PHONY: install install/local install/strip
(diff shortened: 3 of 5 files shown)


akinomyoga:
#1219 discusses a possibility of using the information of whether the name is a `.PHONY` target. I haven't carefully read the changes in this PR to understand it, and I'm not even sure if it is possible, but the present changes don't seem to consider it.

15daksh-2003:
Few things on it @akinomyoga (if I am getting it right as to what you are referring to) :

I did weigh .PHONY. The PR keys on a different bit: the collapse stage previously didn't know which names make actually reported as targets (the stage separation you noted in #1219). It now carries that across — an is_target map marks the extracted names, and the collapse stage re-offers a candidate's plain form only when it's both a reported target and a path node with children. Pure synthesized prefixes (the recursive subdir/… case) are never in is_target, so #544/#858 collapsing is unchanged.

On .PHONY specifically: it is extractable (the -p dump marks blocks with #  Phony target (prerequisite of .PHONY).), but as the criterion, I feel it's both insufficient and unnecessary here:

a. Insufficient — a plain install: rule not declared .PHONY is still a real, user-facing target that hits the same collapse; a phony-only rule keeps dropping it, i.e. only partially fixes #1219:

install:            # not .PHONY -> phony-only STILL drops it
echo install
install/local:
install/strip:

With the approach in PR, both this and the CMake .PHONY case complete to install + install/.

b. Unnecessary — CMake marks its targets .PHONY, so is_target already covers the reported cases.

The two criteria only differ for a non-phony target that is also a path prefix and is genuinely directory-prep, e.g. sub: ; mkdir -p sub with sub/a sub/b: | sub. For the approach taken here, it offers sub alongside sub/; a phony-only rule would offer just sub/. I think offering sub is acceptable: make sub is valid, it matches pre-2.12 behavior, and it's statically indistinguishable from the legitimate non-phony install above — the only signal separating them is .PHONY, which misclassifies that case. It also needs no change to the portability-sensitive extraction awk.

akinomyoga:
I don't know how seriously I should respond to the details of AI-generated text these days, but I guess short responses would suffice for LLM, which has an excellent ability of capturing the context.

> a. Insufficient

Isn't that just the Makefile is broken?

> b. Unnecessary

Yes, it's unnecessary to satisfy your purpose. However, it breaks another behavior, as indicated in the next paragraph in your response.

> I think offering sub is acceptable:

This is the key point. First of all, there is a reason that subdirectory names are dropped, i.e., to be consistent with the normal pathname completions. That's independent of the present issue, where the non-subdirectory `install` is missing, or suffixed by `/`. Basically, I think that whether we switch to generating subdirectory names should be discussed independently. An unrelated change shouldn't be introduced to justify a *possibly poor* implementation, unless it is really necessary. I need to carefully review the present change and also consider other approaches to conclude its necessity.

15daksh-2003:
The AI remark… genuinely not something you come across daily in communities

That said, I think I feel why it came across that way — and it's a genuine miss on my side, not going to dress it up.

I anchored on "make reported install as a target, so don't drop it" and never circled back to what that actually means: .PHONY. You're right that .PHONY is the thing that separates a logical install (no real directory — fine to surface) from a real sub/ that just happens to also be a target, and the is_target check can't tell those apart.

For what it's worth, the "statically indistinguishable" line was me reasoning past the obvious answer — I'd fixated on the CMake repro (all phony there) and generalized off the flat target list. And the non-phony install example was a weak justification anyway; it just gave me the nudge to go back and re-read the manual.

15daksh-2003:
as for carrying over .PHONY info to collapsing stage..if relied on awk parser, it seems it would degrade to today's behavior on non-gnu make, which I don't know how much doing it would solve the core idea for issue

15daksh-2003:
but yeah, the extracter is already only parses GNU's -p output it seems

akinomyoga:
> The AI remark… genuinely not something you come across daily in communities

It's unrelated to this PR, but I strongly feel that these days, being a maintainer of several projects. I've been thinking of what we got and what we lost in the OSS communications with the emergence of LLM. The LLM expands the human inputs into a large amount of text, which contains both something that makes sense and doesn't. When I directly chat with an LLM, I can simply ignore weird parts. The issue happens when it is used for real conversations. I can't tell which part of the weirdness arises from LLM and which part would be the OP's thoughts, so I am supposed to take a lot of time to respond to all of the points, including LLM quirks, which leaves me with an empty feeling.

I'm very comfortable with the style after you switch (though you might still use LLM to improve wording).

akinomyoga:
I'm now trying to see if there is another solution. I haven't yet tested it, but my naive idea is like f09a37cfb5c1b838d8c292ef92fd94233a532882 (after simple refactoring 2c121909b38c3ed140340d529a8e6149f94ac754).

> as for carrying over .PHONY info to collapsing stage..if relied on awk parser, it seems it would degrade to today's behavior on non-gnu make, which I don't know how much doing it would solve the core idea for issue
>
> but yeah, the extracter is already only parses GNU's -p output it seems

Yeah, e.g., BSD make doesn't support it. We already assume GNU make for our completion.

15daksh-2003:
I was just checking something...it seems upstream has introduced: make --print-targets( which I believe would fall on (4.4.90→4.5) ), which might mean moving enumeration off the -p dump. So reanchoring a new dependency on -p's phony prose...where we might move to print targets once it lands for enumeration....

15daksh-2003:
at least the first stage would be better moving away from awk parsing helper way project has today...

15daksh-2003:
not sure if .phony gating is that of a good idea anymore...only if we stick to the way of enumeration we are doing today for bash completion

akinomyoga:
> I was just checking something...it seems upstream has introduced: make --print-targets( which I believe would fall on (4.4.90→4.5) ), which might mean moving enumeration off the -p dump.

Thanks. Does that new option emit a machine-parseable format? That would be useful. Thank you for the information.

> So reanchoring a new dependency on -p's phony prose...

The format of `-p` for phony targets seems stable at least for 24 years since 2002. I haven't checked older versions because I don't have the working binary of older versions immediately.

> not sure if .phony gating is that of a good idea anymore...only if we stick to the way of enumeration we are doing today for bash completion

Even if the new make is shipped, we still need to support older versions of make at the same time. We may include the code to use `--print-targets`, but we still need a fallback of using the `-p` output.

15daksh-2003:
> Even if the new make is shipped, we still need to support older versions of make at the same time. We may include the code to use `--print-targets`, but we still need a fallback of using the `-p` output.

I agree to that. Beleiving maintainance also would only go up: with print targets on top of -p as fallback. Perhaps, i dont know but, something like --print-phony on upstream... idk if it has been discussed over there or if Paul had any opinions on it, but seems like wishful thinking at this point. I would check this there once. meanwhile, for now, gating on phony where we have -p dump seems the only way feasible at least

15daksh-2003:
> Does that new option emit a machine-parseable format?

"--print-targets will print all explicit, non-special targets defined in the makefiles, one per line, then exit with success. No recipes are invoked and no makefiles are re-built."

seems flat only, not structured like a json or something. seems would be trivially consumable by while read, grep, compgen -W

akinomyoga:
Yeah, I cloned the repository from Savannah, built the `master` branch, and was observing the behavior. As you described, the new option seems to simply print only the target names [with no headers or prefixes (like `phony:` or `file:` in f09a37cfb5c1b838d8c292ef92fd94233a532882), indicating the type of the targets]. Although it seems to drop some of the filename targets starting with `../`, it basically prints all non-trivial targets, including both phony and non-phony ones. So there isn't a way to tell which ones are phony, unless the suggested `--print-phony` would be implemented, or unless relying on the `-p` output.

I also checked the `-p` output of the master branch, which seems consistent with the traditional behavior.

15daksh-2003:
on your posted approach, a couple of things(from the commit):

{target#target:}  — shouldn't it be ${target#phony:}

The is_phony_target = 0 in the starts_with($0, prefix) rule.....which fires on every line when the prefix is empty (make <TAB>), wiping the flag the marker just set.

akinomyoga:
> {target#target:} — shouldn't it be ${target#phony:}

Ah, yeah. Right. Thank you.

> The is_phony_target = 0 in the starts_with($0, prefix) rule.....which fires on every line when the prefix is empty (make ), wiping the flag the marker just set.

Hmm, the existing implementation seems tricky. This means that an unrecognized line like `#  Phony target (prerequisite of .PHONY).` has been triggering `is_target_block = 1` in the current implementation, though the side effect is suppressed by `target = "";` after `is_target_block = 0;` in `/^# +File is an intermediate prerequisite/`.

15daksh-2003:
i tried testing above what u suggested. have modified the tests a bit, along with empty prefix thing and the other one

akinomyoga:
The refactoring commit should be separated. Otherwise, it's hard to review the substantial changes. I've separated back the original commit of refactoring and force-pushed.

completions-core/make.bash
```diff
@@ -19,13 +19,18 @@ _comp_cmd_make__extract_targets()
 }
 
 # Truncate the non-unique filepaths in the "targets" array to only generate
-# phony targets, unique directories, or files in the COMREPLY array.  This
+# phony targets, unique directories, or files in the COMPREPLY array.  This
```
akinomyoga, line 22: Ah, so this fix should rather be included in the refactoring commit.

completions-core/make.bash
```diff
@@ -19,13 +19,18 @@ _comp_cmd_make__extract_targets()
...
 # function discards the files under subdirectories unless the path is unique
 # under each subdirectory and instead generate the subdirectory path.  For
 # example, when there are two candidates, "abc/def" and "abc/xyz", we generate
 # "abc/" instead of generating both candidates directly.  When there is only
 # one candidate "abc/def", we generate the full path "abc/def".
 #
+# The phony targets (prefixed by "phony:") are offered as-is: a `/' in a phony
+# name is a naming convention (e.g. CMake's `install', `install/local'), not a
+# directory boundary, so they are never collapsed.  Only the non-phony targets
+# (prefixed by "file:"), which are real filepaths, go through the path collapse.
```
akinomyoga: This line doesn't fit in 79 columns (leaving one column for the continuation mark in terminal text editors).

15daksh-2003:
> The refactoring commit should be separated. Otherwise, it's hard to review the substantial changes. I've separated back the original commit of refactoring and force-pushed.

I am not aware much of conventions for commits here. I carried forward from cmake's conventions for one commit per PR

completions-core/make.bash
15daksh-2003: let me patch that up

completions-core/make.bash
```diff
@@ -163,9 +177,21 @@ _comp_cmd_make()
             _comp_cmd_make__truncate_non_unique_paths
 
         if [[ $mode != -d ]]; then
-            # Completion will occur if there is only one suggestion
-            # so set options for completion based on the first one
-            [[ ${COMPREPLY-} == */ ]] && compopt -o nospace
+            # A candidate ending in `/' is a prefix meant to be drilled into, so
```
akinomyoga: Is this change necessary? `compopt -o nospace` only takes effect when COMPREPLY contains a single suggestion, so checking the first element is sufficient.

completions-core/make.bash
akinomyoga: Ah, sorry. I've fixed it just now and force-pushed.

completions-core/make.bash
akinomyoga: I've also applied this change to the first comit.

akinomyoga:
> I am not aware much of conventions for commits here. I carried forward from cmake's conventions for one commit per PR

That doesn't apply to this project. The convention seems to largely vary depending on the project.

completions-core/make.bash
akinomyoga: > `compopt -o nospace` only takes effect

Actually, that is my understanding, so there might be an exceptional case that I'm unaware.

15daksh-2003: my thought here was scan complete could be needed for menu complete, where just checking first would not be sufficient

15daksh-2003: verified locally, with multiple candidates readline would insert only the common prefix with no trailing space, so nospace should be a no-op. we can drop it i think

akinomyoga: The code of checking only the first element for `nospace` is actually a widely used pattern in the codebase. We haven't seen a problem with it so far. If this needs to be fixed, they need to be changed altogether.

akinomyoga: Yeah, you are right. `menu-complete` is affected.

akinomyoga: If you want to contribute, a separate PR is welcome for fixing existing cases in all `completions-core/*.bash`.

15daksh-2003: for this here, i believe, we are fine for now. i pushed the changed to revert it back and the tests associated with it

15daksh-2003:
corrected for few over spills over 79 cols in test_make.py

15daksh-2003:
@akinomyoga  one question out of curiosity: i was seeing zsh way of doing things, and I believe they expose a mechanism with some flags for opt in "collapse" with different semantics offered, while I think the default remains to be something collapsing targeted  when implemented. what are your thoughts on that?

akinomyoga:
> a mechanism with some flags for opt in "collapse" with different semantics offered,

That would touch the central policy of this project, which I asked about on the very first day when I joined the maintainer group. The answer has been

https://github.com/scop/bash-completion/discussions/539#discussioncomment-830724

That is, we want to minimize the customizability unless necessary. There is a related argument in https://github.com/scop/bash-completion/issues/720#issuecomment-1092370272.

**edit**: I was forgetting, but I noticed that there has also been [a related sentence in `doc/api-and-naming.md`](https://github.com/scop/bash-completion/blob/c2850058a2faa0d496eab7fbdd0e3ceaf2f81d1a/doc/api-and-naming.md?plain=1#L76-L79).

akinomyoga:
I added adjustments in the push [`4cb0d6e..d047e2f`](https://github.com/scop/bash-completion/compare/4cb0d6e5a580f469ead4e40fbf91a3ff2e60ad93..d047e2ffa5302ad3746f340c488b184976084479). We now have confirmed the `-p` output with the latest `master` branch of `make`. `is_target_block = 0;` has been moved to be paired with `is_phony_target = 0;`.

I also adjusted the commit message to clarify that make 3.80 is merely the minimum version I checked; the format of `#  Phony target` in the `-p` output might have existed even before. I added `Co-authored-by` to both commits to mention the mutual contributions.

helpers-core/make-extract-targets.awk
```diff
@@ -42,17 +45,34 @@ is_target_block == 0 { next; }
...
-    print target;
+    if (is_phony_target) {
+      print "phony:" target;
+    } else {
+      print "file:" target;
+    }
     target = "";
   }
+  is_target_block = 0;
+  is_phony_target = 0;
```
15daksh-2003, line 89: @akinomyoga what is the reasoning behind moving this here? alignment? because funtionally I don't think anything changes here. If (target!="") block does not reference or mutate the is target block flag

15daksh-2003: Okay, didn't see the previous comment

15daksh-2003:
> Hmm, the existing implementation seems tricky. This means that an unrecognized line like `# Phony target (prerequisite of .PHONY).` has been triggering `is_target_block = 1` in the current implementation, though the side effect is suppressed by `target = "";` after `is_target_block = 0;` in `/^# +File is an intermediate prerequisite/`.

@akinomyoga  couple of things (both sort of non blocking I believe in the scope of this):
a. I was going through the manual, and it seems that not all special targets are in the list we have in target header block. Like .NOTINTERMEDIATE, which got added in make 4.4 or .EXTRA_PREREQS. 
b. the semantics around is_target_block, which get a bit twisted for empty prefix scenario...should we think of moving header rule above the gate? something like this:

```
/^[^#\t:%]+:/ {
  # A header line begins a new block; assume it is not a wanted target
  # until proven otherwise below.
  is_target_block = 0;
  is_phony_target = 0;
  target = "";
   ... special-target skips, unchanged ...
   ... hidden-target skip, unchanged ...
  # only process the targets the user wants
  if (!starts_with($0, prefix)) next;

  is_target_block = 1;
  target = $0;
   ...
   next;
 }

# skip every line outside a wanted target block
is_target_block == 0 { next; }
```

what are your thoughts on this? if we do decide to go with either or both, should this be a separate PR or this one only?

akinomyoga:
> a. I was going through the manual, and it seems that not all special targets are in the list we have in target header block. Like .NOTINTERMEDIATE, which got added in make 4.4 or .EXTRA_PREREQS.

I think the list in the present completion code hasn't been updated since it was added by b28d7108d3677c61bd01c51ccee8bb1cf9e3bfba (2012). Those special targets aren't intentionally excluded. They can be added to the list. Or another option may be to exclude all all-caps targets starting with a period.

> b. the semantics around is_target_block, which get a bit twisted for empty prefix scenario...should we think of moving header rule above the gate? something like this:

I also felt it while I was checking another PR #1577, but I haven't carefully thought about whether that is totally equivalent to the present handling. The code was originally contributed in the form of a sed script. The strange way of processing might be due to the limitations of the sed language. However, there might possibly be another reason for the present strange way.

If we can mathematically transform the present processing to the suggested one, then we can safely rewrite the processing in the suggested way. Otherwise, we want to think carefully about hypothetical inputs causing different results, and confirm that `make -npq` doesn't produce such a pattern. However, as long as it's working, we wouldn't bother with rewriting the code, which would be the reason that the present implementation hasn't been fixed.

You may modify the loop structure this time, with attention. Or you can simply add `/^[^#\t:%]+:/ && ` before `starts_with` in the present implementation, which is safer but uglier.

akinomyoga:
Since another PR #1577 for the make script was merged, I've rebased the PR head on top of the current `main` branch. Then, I modified the code added in #1577 to make it consistent with the change of this PR ([`2573ed2..b29d4ea`](https://github.com/scop/bash-completion/compare/2573ed28f2e34efd3d017658c2d691a56258312a..b29d4ea18c69439d66c8498ed36f64d0f3437a4a)).

15daksh-2003:
> > a. I was going through the manual, and it seems that not all special targets are in the list we have in target header block. Like .NOTINTERMEDIATE, which got added in make 4.4 or .EXTRA_PREREQS.
> 
> I think the list in the present completion code hasn't been updated since it was added by [b28d710](https://github.com/scop/bash-completion/commit/b28d7108d3677c61bd01c51ccee8bb1cf9e3bfba) (2012). Those special targets aren't intentionally excluded. They can be added to the list. Or another option may be to exclude all all-caps targets starting with a period.
> 
> > b. the semantics around is_target_block, which get a bit twisted for empty prefix scenario...should we think of moving header rule above the gate? something like this:
> 
> I also felt it while I was checking another PR #1577, but I haven't carefully thought about whether that is totally equivalent to the present handling. The code was originally contributed in the form of a sed script. The strange way of processing might be due to the limitations of the sed language. However, there might possibly be another reason for the present strange way.
> 
> If we can mathematically transform the present processing to the suggested one, then we can safely rewrite the processing in the suggested way. Otherwise, we want to think carefully about hypothetical inputs causing different results, and confirm that `make -npq` doesn't produce such a pattern. However, as long as it's working, we wouldn't bother with rewriting the code, which would be the reason that the present implementation hasn't been fixed.
> 
> You may modify the loop structure this time, with attention. Or you can simply add `/^[^#\t:%]+:/ && ` before `starts_with` in the present implementation, which is safer but uglier.

will push something in a bit

akinomyoga:
> will push something in a bit

I think those should be done in separate PRs.

15daksh-2003:
> I think those should be done in separate PRs.

sure. fine with me. is there anything else left for this PR to close?

akinomyoga (approved):
Nothing from me. Let's wait to see if others have something.

scop:
Nothing from me, quite frankly the make completion as a whole is already beyond something I'd like to touch or look at, but didn't catch anything alarming on a quick skim, and there are tests, and no existing tests needed modifications, so I'm fine. A quick LLM review didn't yield anything of substance that I can tell either.

15daksh-2003:
Hey @akinomyoga, if review is complete, can we try closing this?

PR #1701, opened 2026-07-28
fix(scp): Use less escaping for filenames when not needed

### LINKED ISSUE #1540: scp: remote filenames get escaped twice
## Describe the bug

[Since OpenSSH 9.0](https://www.openssh.org/releasenotes.html#9.0), scp uses the sftp protocol by default, so it only needs single escaping.

## To reproduce

1. Complete a remote filename destination which needs escaping, e.g. `my file.pdf`
2. Spaces in the remote filename get completed as `\\\ `, e.g. `remote:my\\\ file.pdf`
3. When executed, the local file gets copied to a remote file named `my\ file.pdf` instead of `my file.pdf`

## Expected behavior

Special characters in the remote filename get completed using a single backslash, e.g. `remote:my\ file.pdf`

- [ ] Operating system name/distribution and version: Arch Linux
- [ ] bash version, `echo "$BASH_VERSION"`: 5.3.8(1)-release
- [ ] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.17.0

## Additional context

Passing `-O` to scp uses the old method, thus still requiring double escaping.

## Debug trace

I don't think one is needed for this issue.

yedayak:
fix(scp): Use less escaping for filenames when not needed

OpenSSH 9.0 changed the default protocol the "scp" command uses to
sftp, which needs less escaping for special characters in remote
filenames. Check if the command has the option to use the legacy
protocol and do less escaping.

If the command line has the "-O" option to use the legacy protocol,
we revert to the old behaviour.

The extra escaping is necessary for older versions, so we can't do it
unconditionally.

Fixes: https://github.com/scop/bash-completion/issues/1540

### DIFF
--- a/completions-core/ssh.bash
+++ b/completions-core/ssh.bash
@@ -688,7 +688,26 @@ _comp_cmd_scp()
     case $cur in
         !(*:*)/* | [.~]*) ;; # looks like a path
         *:*)
-            _comp_xfunc_scp_compgen_remote_files
+            # if scp supports "-O", which forces legacy scp/rcp protocols,
+            # it means it uses the sftp protocol by default, so we need less
+            # escaping.
+            # The change was done in OpenSSH 9.0 [0], see [1]
+            #
+            # [0] https://www.openssh.org/releasenotes.html#9.0
+            # [1] https://github.com/scop/bash-completion/issues/1540
+            local arg legacy_scp=""
+            for arg in "${words[@]}"; do
+                if [[ $arg == -*O* ]]; then
+                    legacy_scp=set
+                    break
+                fi
+            done
+
+            if [[ $legacy_scp || ! $("$1" --usage 2>&1) =~ scp\ \[-[^]]*O ]]; then
+                _comp_xfunc_scp_compgen_remote_files
+            else
+                _comp_xfunc_scp_compgen_remote_files -l
+            fi
             return
             ;;
     esac
--- a/test/t/test_scp.py
+++ b/test/t/test_scp.py
@@ -1,4 +1,5 @@
 import os
+import re
 import sys
 from itertools import chain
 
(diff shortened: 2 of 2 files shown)


completions-core/ssh.bash
```diff
@@ -688,7 +688,18 @@ _comp_cmd_scp()
     case $cur in
         !(*:*)/* | [.~]*) ;; # looks like a path
         *:*)
-            _comp_xfunc_scp_compgen_remote_files
+            # if scp supports "-O", which forces legacy scp/rcp protocols,
```
akinomyoga, line 691: I haven't checked the behavior, but I have a naive question. When the user specifies the `-O` option to `scp`, does it change the number of required escapes? In such a case, should we use the legacy quoting even for the newer versions of `scp`?

yedayak: Yes, updated PR

test/t/test_scp.py
```diff
@@ -114,30 +115,57 @@ def test_remote_path_with_nullglob(self, completion):
...
+        else:
+            assert completion == r"\ in\ filename.txt"
 
-    def test_remote_path_with_backslash(self, bash):
+    def test_remote_path_with_backslash(self, bash, needs_extra_escaping):
         assert_bash_exec(
             bash, r"ssh() { printf '%s\n' 'abc def.txt' 'abc\ def.txt'; }"
         )
         completion = assert_complete(bash, "scp remote_host:abc\\")
+        completion_legacy_proto = assert_complete(
```
yedayak: I don't love this, but I think its better than having completely separate tests for `-O`. Open to other suggestions

completions-core/ssh.bash
```diff
@@ -683,12 +683,31 @@ _comp_cmd_scp()
             ;;
     esac
 
+    local arg legacy_scp=""
+    for arg in "${words[@]}"; do
+        if [[ $arg == -*O* ]]; then
+            legacy_scp=set
+            break
+        fi
+    done
```
akinomyoga: I think this should be put inside the relevant branch. The variable `legacy_scp` is used only inside the `*:*` branch of the case statement below.

test/t/test_scp.py
```diff
@@ -204,11 +232,21 @@ def test_local_path_ending_with_backslash(self, bash, tmpdir_backslash):
...
-    def test_remote_path_ending_with_backslash(self, bash):
+    def test_remote_path_ending_with_backslash(
+        self, bash, needs_extra_escaping
+    ):
         assert_bash_exec(bash, "ssh() { echo 'hypothetical\\'; }")
         completion = assert_complete(bash, "scp remote_host:hypo")
+        completion_legacy_proto = assert_complete(
+            bash, "scp -O remote_host:hypo"
+        )
+        completion = assert_complete(bash, "scp remote_host:hypo")
```
akinomyoga: This line appears to be a duplicate of the second previous sentence. Is there a reason that we run the same `assert_complete(bash, "scp remote_host:hypo")` twice?

test/t/test_scp.py
akinomyoga: Is there a reason that it's better not to separate? The tests seem independent. Is that because we want to reduce the number of calls for defining the mock `ssh() { ... }`?

akinomyoga (changes requested):
Thank you for updating.

completions-core/ssh.bash
yedayak: Agreed, moved

test/t/test_scp.py
yedayak: Nice catch, not relevant anymore since I split it up

test/t/test_scp.py
yedayak: Yeah thinking about it more it's not a good reason, split it up.

akinomyoga (approved):

completions-core/ssh.bash
akinomyoga: Thank you.

completions-core/ssh.bash
```diff
@@ -688,7 +688,26 @@ _comp_cmd_scp()
...
+            # [1] https://github.com/scop/bash-completion/issues/1540
+            local arg legacy_scp=""
+            for arg in "${words[@]}"; do
+                if [[ $arg == -*O* ]]; then
+                    legacy_scp=set
+                    break
+                fi
+            done
+
+            if [[ $legacy_scp || ! $("$1" --usage 2>&1) =~ scp\ \[-[^]]*O ]]; then
```
scop, line 706: I wonder if the logic with `$legacy_scp` here is the wrong way around.

I'm assuming the intent is to do less escaping if we found `-O` was given.

So with that, shouldn't we be doing this instead?
```suggestion
            if [[ ! $legacy_scp && ! $("$1" --usage 2>&1) =~ scp\ \[-[^]]*O ]]; then
```

...and if logically yes, we could flip the if/else around and remove the negations, so

```
            if [[ $legacy_scp || $("$1" --usage 2>&1) =~ scp\ \[-[^]]*O ]]; then
                # do the -l here
```

---

Also, I find this code with `$legacy_scp` variable name kind of hard to read, would be better if it had the `has_` prefix, but just adding that would kind of make the variable read unclear still. How about something like `$has_O_given`?

scop (changes requested):
Let's also mention _remote_ filenames in the PR title and commit message. I think I'd rephrase it something like "use less remote filename escaping where supported", as the "less ... not needed" reads odd to me.

completions-core/ssh.bash
scop: Er
> I'm assuming the intent is to do less escaping if we found -O was given.

...this assumption of mine is false and we want it the other way around, so the original code is fine after all. Will merge with the commit message tweaked.

PR #1706, opened 2026-07-29
feat(*sum): treat uppercase variant of *sums as checksum file

scop:


test/fixtures/sha256sum/SHA256SUMS
scop, line 1: Removed the lowercase `sha256sums` at it would have caused a case insensitivity naming conflict in the repo. The lowercase `sha256sums.txt` still provides a bit of coverage for that case.

akinomyoga (approved):

PR #1713, opened 2026-08-02
feat(dockerfmt): add fallback 3rd party completion loader

scop:
https://github.com/reteps/dockerfmt#usage

### DIFF
--- a/completions-fallback/.gitignore
+++ b/completions-fallback/.gitignore
@@ -60,6 +60,7 @@
 /dive.bash
 /dlv.bash
 /docker.bash
+/dockerfmt.bash
 /dpkg-deb.bash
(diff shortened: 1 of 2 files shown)


akinomyoga (commented):
The usage in README isn't clear, but nixpkgs seems to contain an example:

https://github.com/NixOS/nixpkgs/blob/59c057d92e7575ce35672a819552d2c22709bcbd/pkgs/by-name/do/dockerfmt/package.nix#L34

**edit**: I'm neutral to whether we should include the above example in the commit message. The PR may be merged as is or may be merged with the link added in the commit message. Either works.

akinomyoga (approved):

PR #1720, opened 2026-08-09
fix(apt-get): handle options after subcommands

### LINKED ISSUE #1690: apt-get: long "--" option completion (e.g. --no-install-recommends) unreachable after subcommands like install/remove/purge
### Description

In `completions/apt-get`, the `_apt_get()` function contains two separate
completion code paths:

1. A "special subcommand" branch, entered via `if [[ -v special ]]; then ... fi`,
   which fires whenever a subcommand such as `install`, `remove`, `purge`,
   `autoremove`, `source`, `build-dep`, `download`, or `changelog` is found
   anywhere earlier on the command line.
2. A generic long-option branch further down:
   ```bash
   if [[ $cur == -* ]]; then
       COMPREPLY=($(compgen -W '--no-install-recommends --install-suggests
           --download-only --fix-broken ... ' -- "$cur"))
   fi
   ```

The problem is that the `install)` arm inside the "special" `case` block ends
with `;;&` (fall-through-and-retest), so execution always continues into the
final catch-all arm:

```bash
*)
    COMPREPLY+=($(_xfunc apt-cache _apt_cache_packages))
;;
```

...which attempts **package name** completion, and the whole `if [[ -v special ]]`
block ends with an unconditional `return`.

As a result, whenever a subcommand like `install` is present anywhere on the
command line, the function returns before ever reaching the generic
`[[ $cur == -* ]]` branch further down — so **long options can never be
completed after a subcommand**, even though the option list clearly exists
in the script and is intended to work.

### Steps to reproduce

```bash
apt-get install --no<TAB><TAB>
```

**Expected:** completion offers `--no-install-recommends`, `--no-download`,
`--no-upgrade`, `--no-show-upgraded`, etc. (all present in the hardcoded
`compgen -W` list later in the function).

**Actual:** nothing is completed. The shell instead silently attempts
`_apt_cache_packages` (package-name completion) against `--no`, which
naturally matches no package, so `COMPREPLY` stays empty.

This affects **any** `apt-get` subcommand handled by the `special` case
(`install`, `remove`, `purge`, `autoremove`, `source`, `build-dep`,
`download`, `changelog`) — i.e. exactly the subcommands people most commonly
want to pass options like `--no-install-recommends`, `--purge`,
`--dry-run`, etc. to.

### Environment

- `bash-completion` version: `1:2.11-8` (also reproducible by inspection on
  upstream tags `2.9`, `2.10`, and `2.11`)
- `apt`/`apt-get` version: `2.8.3`
- OS: Linux Mint 22.3 (Cinnamon), based on Ubuntu 24.04 "noble"
- GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)

fudianchn:
AI assistance: written with AI assistance (Claude Code, model claude-opus-4-8). The code and tests were reviewed and tested locally before pushing.

## Problem

Global option handling was ordered after the "special" subcommand branches. Those branches return after completing command operands, so both option-name and option-argument completion could become unreachable after commands such as install and remove.

For example, apt-get install --<TAB> produced no option candidates, and apt-get remove --config-file <TAB> completed installed packages instead of files.

## Fix

Process completion in semantic priority order:

1. option arguments selected by prev;
2. option names selected by cur;
3. operands for special subcommands;
4. top-level subcommands.

The existing option handlers and completion candidates are unchanged.

Added test_options_after_subcommand and test_option_argument_after_subcommand covering both reported cases.

Closes #1690

### DIFF
--- a/test/t/test_apt_get.py
+++ b/test/t/test_apt_get.py
@@ -14,3 +14,11 @@ class TestAptGet:
     @pytest.mark.complete("apt-get build-dep ")
     def test_build_dep_dirs(self, completion):
         assert "dpkg/" in completion
+
+    @pytest.mark.complete("apt-get install --")
+    def test_options_after_subcommand(self, completion):
+        assert "--no-install-recommends" in completion
+
+    @pytest.mark.complete("apt-get remove --config-file ", cwd="apt-mark")
+    def test_option_argument_after_subcommand(self, completion):
+        assert completion == "example.conf"
--- a/completions-core/apt-get.bash
+++ b/completions-core/apt-get.bash
@@ -25,6 +25,48 @@ _comp_cmd_apt_get()
         fi
     done
 
+    local noargopts='!(-*|*[eoct]*)'
+    # shellcheck disable=SC2254
+    case $prev in
+        --error-on | --help | --version | --option | -${noargopts}[ehvo])
+            return
+            ;;
+        --config-file | -${noargopts}c)
+            _comp_compgen_filedir
+            return
+            ;;
+        --target-release | --default-release | -${noargopts}t)
+            # Prefer `apt-cache` in the same dir as command
+            local pathcmd
+            pathcmd=$(type -P -- "$1") && local PATH=${pathcmd%/*}:$PATH
+            _comp_compgen_split -- "$(apt-cache policy | command sed -ne '
+                s/^ *release.*[ ,]o=Debian,a=\([[:alnum:]]*\).*/\1/p
+                s/^ *release.*[ ,]o=Ubuntu,a=\([[:alnum:]]*\).*/\1/p
+            ')"
+            return
+            ;;
+    esac
+
+    if [[ $cur == -* ]]; then
+        _comp_compgen -- -W '--no-install-recommends --install-suggests
+            --download-only --fix-broken --ignore-missing --fix-missing
+            --no-download --quiet --simulate --just-print --dry-run --recon
+            --no-act --yes --assume-yes --assume-no --no-show-upgraded
+            --verbose-versions --no-list-columns --comment --host-architecture
+            --build-profiles --compile --build --ignore-hold --with-new-pkgs
+            --no-upgrade --only-upgrade --allow-downgrades
+            --allow-remove-essential --allow-change-held-packages
+            --force-yes --print-uris --purge --reinstall --list-cleanup
+            --snapshot --target-release --default-release --trivial-only
+            --mark-auto --no-remove --auto-remove --autoremove --only-source
+            --diff-only --dsc-only --tar-only --arch-only --indep-only
+            --allow-unauthenticated --no-allow-insecure-repositories
+            --allow-releaseinfo-change --show-progress --with-source --error-on
+            --update --help --version --audit --config-file --option
+            --cli-version --no-color --color'
+        return
+    fi
+
     if [[ $special ]]; then
         case $special in
             remove | auto?(-)remove | purge)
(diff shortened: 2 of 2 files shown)


test/t/test_apt_get.py
```diff
@@ -14,3 +14,10 @@ def test_2(self, completion):
     @pytest.mark.complete("apt-get build-dep ")
     def test_build_dep_dirs(self, completion):
         assert "dpkg/" in completion
+
+    @pytest.mark.complete("apt-get install --")
+    def test_options_after_subcommand(self, completion):
+        # Long options should be completed after a subcommand too (#1690);
+        # previously the option block was unreachable once a "special"
+        # subcommand like install/remove/purge was present.
```
yedayak: I don't think this comment is needed

yedayak (changes requested):
This fix is incomplete, since options with special handling like `--config-file` still aren't handled correctly with the "special" subcommands.
For example, `apt-get remove --config-file <TAB>` completes packages instead of files.

BTW, was this PR AI generated?

test/t/test_apt_get.py
fudianchn: Removed in 87631cd.

fudianchn:
@yedayak On `--config-file`: the first version only moved the option-name block (`if [[ $cur == -* ]]`) above the special-subcommand branch, and left the option-argument block (`case $prev`, which handles `--config-file`, `--target-release`, etc.) still below it. So `apt-get remove --config-file <TAB>` fell through to the package handler. I had overlooked the option-argument case in that version.

The current commit moves `case $prev` up as well, so the dispatch order is now: option argument (prev) -> option name (cur) -> special-subcommand operand -> top-level subcommand. `apt-get remove --config-file <TAB>` now completes files. I added `test_option_argument_after_subcommand`, which runs that exact command in the `apt-mark` fixture directory and asserts the result equals `example.conf`.

Verified locally:
- `python -m pytest -p no:cacheprovider -q test/t/test_apt_get.py`: 5 passed
- `bash -O extglob -n completions-core/apt-get.bash`: passed
- Local shfmt, shellcheck, Ruff, and direct mypy checks passed.
- GitHub Actions: pre-commit, Lint commits, and all distcheck jobs passed.

On the AI question: yes, AI assistance was used (Claude Code, model claude-opus-4-8). The code changes and tests in this revision were written with AI assistance, and AI also helped translate the description and commit message into English (my native language is not English) and polish the wording. I reviewed the resulting code and ran the tests and `bash -n` locally before pushing.

yedayak:
> @yedayak On `--config-file`: the first version only moved the option-name block (`if [[ $cur == -* ]]`) above the special-subcommand branch, and left the option-argument block (`case $prev`, which handles `--config-file`, `--target-release`, etc.) still below it. So `apt-get remove --config-file <TAB>` fell through to the package handler. I had overlooked the option-argument case in that version.
> 
> The current commit moves `case $prev` up as well, so the dispatch order is now: option argument (prev) -> option name (cur) -> special-subcommand operand -> top-level subcommand. `apt-get remove --config-file <TAB>` now completes files. I added `test_option_argument_after_subcommand`, which runs that exact command in the `apt-mark` fixture directory and asserts the result equals `example.conf`.

Thanks

> Verified locally:
> 
>     * `python -m pytest -p no:cacheprovider -q test/t/test_apt_get.py`: 5 passed
> 
>     * `bash -O extglob -n completions-core/apt-get.bash`: passed
> 
>     * Local shfmt, shellcheck, Ruff, and direct mypy checks passed.
> 
>     * GitHub Actions: pre-commit, Lint commits, and all distcheck jobs passed.
> 

> On the AI question: yes, AI assistance was used (Claude Code, model claude-opus-4-8). The code changes and tests in this revision were written with AI assistance, and AI also helped translate the description and commit message into English (my native language is not English) and polish the wording. I reviewed the resulting code and ran the tests and `bash -n` locally before pushing.

I think you should disclose the AI usage up front. All the verifications you write about don't are mostly just fluff, just running pre-commit will check everything automatically, no need to do it yourself or mention. 

The code change itself looks fine, will merge in a few days if no one has any other opinions on it.

fudianchn:
Updated the description: moved the AI disclosure to the top, and dropped the verification list (kept the two new tests).

PR #1722, opened 2026-08-10
test: retry sendintr/expect once on timeout (addresses #706)

edmundlod:
This MR is an attempt to address the Intermittent timeouts under high host load issue in #706.

It adds one retry on "Send Ctrl+C, and wait for the prompt to come back" for tests that raise a `TIMEOUT`. If the test failed due to high load, rather than due to a test error, then there is a good chance that it will pass on the second round.

This would then fix CI issues like the ones [Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1119845) is having with TIMEOUTs in the test suite.

Tested under deliberately induced CPU load (several runs with all cores saturated): a few sendintr/expect timeouts occurred, all recovered by the retry, so that the test suite did not error out.

_Claude was used in triaging this issue, but code and solution are mine._

### DIFF
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -182,6 +182,21 @@ def get_testdir():
     return os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
 
 
+def _sendintr_and_expect_prompt(bash: pexpect.spawn) -> None:
+    """Send Ctrl+C and wait for the prompt, retrying once on timeout.
+
+    A lone retry absorbs a transient scheduling/host-load stall without
+    masking a genuinely broken session: if the second attempt also times
+    out, the TIMEOUT still propagates as before.
+    """
+    try:
+        bash.sendintr()
+        bash.expect_exact(PS1)
+    except pexpect.exceptions.TIMEOUT:
+        bash.sendintr()
+        bash.expect_exact(PS1)
+
+
 @pytest.fixture(scope="session")
 def test_session_tmpdir(tmp_path_factory) -> Path:
     tmpdir = tmp_path_factory.mktemp("bash-completion.session.")
@@ -349,8 +364,7 @@ def bash(request, test_session_tmpdir, tmp_path_factory) -> pexpect.spawn:
             # Not exactly sure why, but some errors leave bash in state where
             # getting the env here would fail and trash our test output. So
             # reset to a good state first (Ctrl+C, expect prompt).
-            bash.sendintr()
-            bash.expect_exact(PS1)
+            _sendintr_and_expect_prompt(bash)
             diff_env(
                 before_env,
                 get_env(bash),
@@ -494,8 +508,7 @@ class bash_env_saved:
 
     def _safe_sendintr(self):
         try:
-            self.bash.sendintr()
-            self.bash.expect_exact(PS1)
+            _sendintr_and_expect_prompt(self.bash)
         except Exception as e:
             if self.noexcept:
                 self.captured_error = e
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -189,8 +189,8 @@ def _sendintr_and_expect_prompt(bash: pexpect.spawn) -> None:
     masking a genuinely broken session: if the second attempt also times
     out, the TIMEOUT still propagates as before.
     """
+    bash.sendintr()
     try:
-        bash.sendintr()
         bash.expect_exact(PS1)
     except pexpect.exceptions.TIMEOUT:
         bash.sendintr()


test/t/conftest.py
```diff
@@ -182,6 +182,21 @@ def get_testdir():
...
+    A lone retry absorbs a transient scheduling/host-load stall without
+    masking a genuinely broken session: if the second attempt also times
+    out, the TIMEOUT still propagates as before.
+    """
+    try:
+        bash.sendintr()
+        bash.expect_exact(PS1)
+    except pexpect.exceptions.TIMEOUT:
+        bash.sendintr()
+        bash.expect_exact(PS1)
```
yedayak, line 197: Maybe? makes it clear the the interrupt itself isn't raising the timeout
```suggestion
    bash.sendintr()
    try:
        bash.expect_exact(PS1)
    except pexpect.exceptions.TIMEOUT:
        bash.sendintr()
        bash.expect_exact(PS1)
```

akinomyoga:
I tried this patch, but it doesn't seem to improve the situation. In my system (Fedora 41, 2 cores (4 cores with hyperthreading)), the single test run doesn't experience the TIMEOUT failure, while I can reproduce the TIMEOUT failures with 5 parallel test runs. With the present `main` branch, 9 TIMEOUT failures in the first trial and 6 TIMEOUT failures in the second trial have occurred. With this patch, 10 TIMEOUT failures in the first trial and 5 TIMEOUT failures in the second trial. There isn't any significant difference.

> Tested under deliberately induced CPU load (several runs with all cores saturated): a few sendintr/expect timeouts occurred, all recovered by the retry, so that the test suite did not error out.

The details on the test aren't provided, and the explanation seems too plain, which makes me feel like this sentence was generated by AI. Can anyone confirm improvements in actual systems?

test/t/conftest.py
edmundlod: Thank you, yes, makes sense.

yedayak:
> I tried this patch, but it doesn't seem to improve the situation. In my system (Fedora 41, 2 cores (4 cores with hyperthreading)), the single test run doesn't experience the TIMEOUT failure, while I can reproduce the TIMEOUT failures with 5 parallel test runs. With the present `main` branch, 9 TIMEOUT failures in the first trial and 6 TIMEOUT failures in the second trial have occurred. With this patch, 10 TIMEOUT failures in the first trial and 5 TIMEOUT failures in the second trial. There isn't any significant difference.
> 
> > Tested under deliberately induced CPU load (several runs with all cores saturated): a few sendintr/expect timeouts occurred, all recovered by the retry, so that the test suite did not error out.
> 
> The details on the test aren't provided, and the explanation seems too plain, which makes me feel like this sentence was generated by AI. Can anyone confirm improvements in actual systems?

I can see significant improvements with this.
I tested this by running a bunch of separate pytest instances in parallel:
```
pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ & pytest -q test/t/unit/ &
``` 
and to make it more likely to fail I set the pexpect timeout to 1:
```
index dcd34f3e3751..f97ffe2d5381 100644
--- a/test/t/conftest.py
+++ b/test/t/conftest.py
@@ -300,6 +300,7 @@ def bash(request, test_session_tmpdir, tmp_path_factory) -> pexpect.spawn:
             # MAGIC_MARK).  Increase window width to reduce wrapping.
             dimensions=(24, 240),
             # TODO? codec_errors="replace",
+            timeout=1,
         )
         # See https://pexpect.readthedocs.io/en/stable/commonissues.html?highlight=delaybeforesend#timing-issue-with-send-and-sendline
         # We don't need the delay, and it makes running the tests much slower.
```
Without this patch, I get around 5 failures in all runs combined, some TIMEOUTS and some like this:
```
AssertionError: Error running "[[ ${OLDPWD-Re8SCgEdfN} == "${_comp__test_83_NEWVAR_OLDPWD-Re8SCgEdfN}" ]]": exit status=127, output="
```
With this patch I don't get any failures

akinomyoga (approved):
Thank you for the confirmation. Then, let's merge this. The second commit created by the GitHub UI is causing the error in the commit message format, but it can be squashed in merging.

edmundlod:
> I tried this patch, but it doesn't seem to improve the situation. In my system (Fedora 41, 2 cores (4 cores with hyperthreading)), the single test run doesn't experience the TIMEOUT failure, while I can reproduce the TIMEOUT failures with 5 parallel test runs. With the present `main` branch, 9 TIMEOUT failures in the first trial and 6 TIMEOUT failures in the second trial have occurred. With this patch, 10 TIMEOUT failures in the first trial and 5 TIMEOUT failures in the second trial. There isn't any significant difference.

I am sorry this didn't work for you with parallel runs. My desire was first of all to make this work on non-parallel runs, which is what Debian does in the CI (on purpose! since an error blocks the process).

For me this situation seemed to be very much improved.

> > Tested under deliberately induced CPU load (several runs with all cores saturated): a few sendintr/expect timeouts occurred, all recovered by the retry, so that the test suite did not error out.

> The details on the test aren't provided, and the explanation seems too plain

I apologise for not having been more detailed.

The baseline run (without the patch) was run as follows (I had already one test run by itself, so 4 total):

```sh
for i in 1 2 3; do
  echo "=== RUN $i ==="
  BASH_COMPLETION_TEST_BASH_COMPLETION=$PWD/bash_completion python3 -m pytest test/t -q -s 2>&1 | grep -E "RETRY-DIAGNOSTIC|passed|failed|error" | tail -10
done
```

Then, since I did not have a load producer installed in the VM (could have done so easily, I suppose), I ran:

```sh
trap 'kill $(jobs -p) 2>/dev/null' EXIT
for i in 1 2 3 4; do yes > /dev/null & done
sleep 1
BASH_COMPLETION_TEST_BASH_COMPLETION=$PWD/bash_completion python3 -m pytest test/t -q 2>&1 | tee /tmp/patched-load-run.log
kill $(jobs -p) 2>/dev/null
wait 2>/dev/null
...
...
...
=========================== short test summary info ============================
FAILED test/t/test_curl.py::TestCurl::test_interface_ipv6 - AssertionError: a...
FAILED test/t/test_javadoc.py::TestJavadoc::test_4_comp_dequote_incomplete - ...
FAILED test/t/test_pkg_config.py::TestPkgConfig::test_variable - AssertionErr...
FAILED test/t/test_pkgconf.py::TestPkgconf::test_variable - AssertionError: a...
FAILED test/t/test_rsync.py::TestRsync::test_local_path_with_spaces - Asserti...
FAILED test/t/test_scp.py::TestScp::test_capital_f_without_space_2 - Assertio...
FAILED test/t/test_scp.py::TestScp::test_capital_f_without_space_3 - Assertio...
FAILED test/t/test_scp.py::TestScp::test_explicit_legacy_proto - AssertionErr...
FAILED test/t/test_scp.py::TestScp::test_remote_path_with_spaces - AssertionE...
FAILED test/t/test_scp.py::TestScp::test_remote_path_with_backslash - Asserti...
FAILED test/t/test_scp.py::TestScp::test_remote_path_with_backslash_2 - Asser...
FAILED test/t/test_scp.py::TestScp::test_local_path_ending_with_backslash - A...
FAILED test/t/test_scp.py::TestScp::test_remote_path_ending_with_backslash - ...
FAILED test/t/test_scp.py::TestScp::test_local_path_mark_1 - AssertionError: ...
FAILED test/t/test_scp.py::TestScp::test_local_path_with_spaces_2 - Assertion...
FAILED test/t/test_scp.py::TestScp::test_local_path_backslash - AssertionErro...
FAILED test/t/test_sshfs.py::TestSshfs::test_local_path_suffix_1 - AssertionE...
FAILED test/t/test_sshfs.py::TestSshfs::test_remote_path_ending_with_backslash
FAILED test/t/test_wget.py::TestWget::test_3 - AssertionError: assert ':1' in...
FAILED test/t/unit/test_unit_compgen_ip_addresses.py::TestUnitCompgenIpAddresses::test_4
20 failed, 1791 passed, 436 skipped, 13 xfailed, 3 xpassed in 896.80s (0:14:56)
```

As you can see, 0 errors (the 20 fails are different kind of fails, not timeouts -- and I get the same 20 fails when I run it without load). Ran inside a VM running Debian Trixie, with 3 CPU cores.

I ran a loaded test on the unpatched version, and got 30 errors.

I hope this helps!

PR #1734, opened 2026-09-13
fix(tar): complete files for combined short options like -tf with bsdtar

### LINKED ISSUE #1728: `bsdtar` (libarchive) completion wrongly assumes that combining short options is not supported
## Describe the bug

The completion for `bsdtar` (or `tar` in general when `bsdtar` is used) incorrectly assumes that it does not support combining short options in GNU style, like `-tf`. As a result, filenames are not completed there.

## To reproduce

1. Install libarchive.
2. Type 'bsdtar -tf ' and try to tab-complete a path.
3. No completion occurs.

## Expected behavior

Path completion, same as for `gtar`.

- [x] Operating system name/distribution and version: Gentoo Linux amd64
- [x] bash version, `echo "$BASH_VERSION"`: 5.3.15(1)-release
- [x] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: 2.18.0

## Additional context

```
$ bsdtar --version
bsdtar 3.8.9 - libarchive 3.8.9 zlib/1.3.1.zlib-ng liblzma/5.8.3 bz2lib/1.0.8 liblz4/1.10.0 libzstd/1.5.7 liblzo2/2.10 openssl/3.6.3 libb2/0.98.1 libacl/2.4.0 
```

## Debug trace

[trace.txt](https://github.com/user-attachments/files/31220207/trace.txt)

fudianchn:
AI disclosure: this change was prepared with AI coding agents, reviewed and revised line by line by me.

## What
`bsdtar -tf <TAB>` and `star -tf <TAB>` now complete archive files. Combined short options ending in `f` or `b` complete their argument like the plain form. Plain `-f` is unchanged.

## Why
bsdtar and star parse combined short options like GNU tar does. The completion dropped the file argument for every combined form.

## How
`_comp_cmd_tar__posix` calls `_comp_cmd_tar__cleanup_prev` before matching the previous word, mirroring the GNU path.

## Root cause
`_comp_cmd_tar__posix_prev_handle` matches exact `-f` only. `_comp_cmd_tar__cleanup_prev`, which turns `-tf` into `-f`, ran in the GNU path only.

## Testing
- `test_tar.py#test_26` (`bsdtar -tf `): fails on main, passes with the fix.
- `test_27` (`bsdtar -f `) guards plain `-f`: passes on both.
- `test_28`-`test_30`: `star -tf ` positive, `-cf ` create-mode combo guard (directories only), `-tzf ` compression-extension guard.
- All 33 tar tests pass (test_tar 30 + test_bsdtar 2 + test_smbtar 1). shellcheck and shfmt clean.
- Not covered: `bsdtar -t <TAB>` and `-tv <TAB>` still complete nothing. Pre-existing gap in `_comp_cmd_tar__try_list_archive`, left untouched.

## Verification of the original issue
Issue reports no completion for `bsdtar -tf `. With the fix, the same line completes archives like `gtar -tf ` does.

Fixes #1728


akinomyoga:
The function modified in this PR is `_comp_cmd_tar__posix`, which is not specifically for BSD tar. [POSIX (SUSv2) tar](https://pubs.opengroup.org/onlinepubs/7908799/xcu/tar.html) doesn't support `-tf`. At the same time, the missing support for BSD tar is not limited to the option argument to `-f`; there are many BSD options we do not support.

fudianchn:
1. `_comp_cmd_tar__posix` is reached from the bsdtar/star registrations and from `_comp_cmd_tar`, which dispatches bare `tar` to it when `--version` does not report GNU. In practice that bare `tar` is bsdtar (macOS, BSD). #1728 reports both entries and expects them to complete archives like `gtar -tf `.

2. For a tar that strictly follows SUSv2 and rejects bundled short options, this change would complete files after `-tf `, a form that tar rejects; harmless, but wider than its syntax. I could not find a clean way to narrow it: enabling the normalization only for the bsdtar/star command names would miss the bare `tar` dispatch from #1728, and the dispatch itself already selects "not GNU" rather than "POSIX".

3. The BSD option surface (long options, -C, ...) is a larger gap. This PR only restores the argument completion of bundled short options, matching what the GNU path does with `_comp_cmd_tar__cleanup_prev`. I kept it to that; if the BSD options gap is worth tracking, I can open a separate issue.

I can rework this in a different direction if preferred.

PR #1739, opened 2026-09-21
fix(cd): do not offer the same directory twice with CDPATH

### LINKED ISSUE #1545: cd completion gives `dir` and `dir/` when CDPATH is set
## Describe the bug

This bug seems to require a few things to occur:
- CDPATH must be set
- The parameter being completed must be the full name of a directory in the current directory
- No other directory is a valid completion

## To reproduce

```
$ CDPATH=.
$ cd /
$ cd usr[TAB][TAB]
usr/  usr// 
```

## Expected behavior

Should have completed to `cd bin/`

- [ ] Operating system name/distribution and version: Fedora 43
- [ ] bash version, `echo "$BASH_VERSION"`: 5.3.0(1)-release
- [ ] bash-completion version, `(IFS=.; echo "${BASH_COMPLETION_VERSINFO[*]}")`: git commit ed0ae1053ca8f377aca5e3c622ef0dbbc62de015 and later

## Additional context

This bug is in git but not in the distribution package. It appears to be a regression from commit ed0ae1053ca8f377aca5e3c622ef0dbbc62de015 (the bug does not occur in its parent).

Commenting out this section appears to make the issue go away (although I don't understand what it is for):
```
    if ((${#_cdpaths[@]} == 1)); then
        _p=${_cdpaths[0]}
        if [[ $_p == "$cur" && $_p != */ ]]; then
            _cdpaths[0]=$_p/
        fi
    fi
```

ethanstoner:
## Summary

- With `CDPATH` set, `cd <dir><TAB><TAB>` on a directory of the current
  directory offered the same directory twice — `usr/` and `usr//` — instead
  of completing it.
- `_comp_cmd_cd__compgen_cdpath` suffixed a sole CDPATH match with a slash
  *before* `filedir` had contributed its own, bare spelling of the very same
  directory, so two spellings of one directory reached `COMPREPLY`.
- The generators' results are now collected in one array, de-duplicated, and
  only then considered for the sole-completion slash.

Fixes #1545

## The regression

`ed0ae10` ("refactor(cd): factorize generator for CDPATH") moved the
sole-completion adjustment from *after* `_comp_compgen -a filedir -d` to
*inside* the new CDPATH generator, which runs *before* it:

```diff
-    _comp_compgen -a filedir -d
-
-    if ((${#COMPREPLY[@]} == 1)); then
-        i=${COMPREPLY[0]}
-        if [[ $i == "$cur" && $i != */ ]]; then
-            COMPREPLY[0]="${i}/"
+    if ((${#_cdpaths[@]} == 1)); then
+        _p=${_cdpaths[0]}
+        if [[ $_p == "$cur" && $_p != */ ]]; then
+            _cdpaths[0]=$_p/
```

Before that commit the check saw the *final* candidate list, so a single
directory could never be counted as one candidate and then joined by a
second. Afterwards it only sees the CDPATH candidates: in `/` with
`CDPATH=.` and `cd usr`, the generator counts one candidate, rewrites it to
`usr/`, and `filedir -d` then adds `usr`. With `-o filenames` and readline's
`mark-directories` those render as `usr//` and `usr/`.

## On the fix suggested in the issue

The issue suggests deleting the `((${#_cdpaths[@]} == 1))` block outright. I
tried that first, and it does make the reported symptom disappear — but only
because `COMPREPLY` then holds `usr` twice and readline hides exact
duplicates. The duplication is still there, and the block is dropped along
with the case it was added for: a CDPATH directory that does *not* exist in
the current directory is not recognized as a directory by `mark-directories`
(and that variable may be off entirely), so without the suffix, completion
on a fully typed sole CDPATH directory cannot descend into it. Hence
de-duplication plus restoring the adjustment's original position, rather than
removal.

## Changes

- `completions-core/cd.bash`: `_comp_cmd_cd__compgen_cdpath` no longer
  suffixes the sole match. `_comp_cmd_cd` collects the CDPATH and `filedir`
  candidates into one array, drops duplicates, and suffixes the sole
  candidate only when CDPATH contributed it — matching the pre-`ed0ae10`
  condition, so plain `cd` completion without CDPATH is untouched.
- `test/t/test_cd.py`: `test_cdpath_dir_in_cwd_not_duplicated`, which fails
  on `main` with `['foo.d/', 'foo.d//']`.

## Validation

Run on macOS 15 (arm64) against Homebrew GNU bash 5.3.20, Python 3.14.7,
pytest 9.1.1, pexpect 4.9.0:

- `pytest test/t/test_cd.py` — 10 passed. The new test fails on `main`
  (`assert <CompletionResult ['foo.d/', 'foo.d//']> == '/'`) and passes with
  the fix.
- `pytest test/t/test_cd.py test/t/test_pushd.py test/t/unit/` — 531 passed,
  4 skipped, 1 xpassed; the same selection on unmodified `main` gives 530
  passed, 4 skipped, 1 xpassed (the one new test is the only difference).
- `prek run shellcheck shfmt ruff-check ruff-format update-test-cmd-list
  typos` on both touched files — all passed. `committed` passes on the commit
  message. `test/runLint` reports nothing for either touched file.
- Checked by hand, outside pytest, that `COMPREPLY` is correct for: the
  reported case; no `CDPATH` at all (unchanged, single bare candidate); a
  directory reachable only through `CDPATH`, with `mark-directories` on and
  off; a name that exists in both the current directory and a `CDPATH` entry;
  a `CDPATH` entry matching several directories; and a `cur` containing a
  slash. Also exercised under `set -u` and `set -o posix` with no errors.

Not run locally: the `distcheck` matrix (almalinux9, alpine, centos7,
debian10, debian10+BSD, fedoradev, ubuntu14), which runs the full suite over
many shells and userlands inside containers. I have no coverage for older
bash versions or for the BSD userland; the change uses only `local -A`,
`${assoc[key]-}` and `_comp_compgen` options already used elsewhere in the
tree, but that is reasoning, not a test result.

## Checklist

- [x] Tests passing
- [x] Test case added for the fixed issue
- [ ] Docs updated — not applicable
- [x] No breaking changes
- [x] Linked to issue

---

Generated with Claude Code on behalf of @ethanstoner.


akinomyoga:
This PR performs deduplication, which is usually unnecessary because it is supposed to be performed by Bash. The issue is that Bash's deduplication fails when one generator generates `usr` while the other generates `usr/`. The right fix should be to normalize the paths so that Bash's deduplication works properly, instead of performing the manual deduplication in a script. The issue appears to be simply fixed by the following one line fix:

```diff
diff --git a/completions-core/cd.bash b/completions-core/cd.bash
index 4b0307a1..74be57f9 100644
--- a/completions-core/cd.bash
+++ b/completions-core/cd.bash
@@ -75,6 +75,6 @@ _comp_cmd_cd()
     compopt -o filenames
     _comp_cmd_cd__compgen_cdable_vars
     _comp_compgen -ai cd cdpath
-    _comp_compgen -a filedir -d
+    _comp_compgen -a filedir -df
 }
 complete -F _comp_cmd_cd -o nospace cd pushd
```

**edit**: I tested the above change, but it seems the situation is more complicated. When we suffix `/` to the directory reachable from the current path, and when `compopt -o filenames` is specified, Bash duplicates the slash in the completion menu, although the inserted word preserves the single slash. However, if we remove `compopt -o filenames`, Bash stops quoting the special characters (such as a space) in the filename.

ethanstoner:
The `distcheck (centos7)` failure here is flaky and unrelated to this change — the same job on the same commit passed in a sibling run started two seconds later:

- failed: `distcheck (centos7)` in [run 35642880939](https://github.com/scop/bash-completion/actions/runs/35642880939/job/106476236221)
- passed: `distcheck (centos7)` in [run 35633762463](https://github.com/scop/bash-completion/actions/runs/35633762463/job/106476237942)

The error is `ERROR test/t/test_installpkg.py::TestInstallpkg::test_4 - pexpect.ex...`, alongside 2080 passed / 158 skipped / 27 xfailed / 3 xpassed. It's an error rather than an assertion failure, so it's the pexpect interaction timing out, not an expectation mismatch.

For what it's worth I also checked the one way this diff could plausibly have reached that test: `test_cd.py` sorts before `test_installpkg.py` in the same session, so a leaked `CDPATH` would be a real suspect. It isn't one — the new test sets `env=dict(CDPATH=".")` through the `bash_env_saved` context manager in `test/t/conftest.py:849-855`, which restores the variable on exit. Nothing outside `completions-core/cd.bash` and `test/t/test_cd.py` is touched.

Happy to rebase if you'd like a fresh run.

Written with Claude Code on behalf of @ethanstoner.

ethanstoner:
I switched to normalizing in d7c8438 and dropped the manual dedup. The mismatch comes from the sole-match block in `_comp_cmd_cd__compgen_cdpath`, which appends `/` even when the directory also exists in the current directory, while the loop above it already leaves such names bare. With `! -d $_p` added to that condition, both generators produce the same word and Bash's own dedup removes the copy.

`-o filenames` and `filedir -d` stay as they are, so the menu has no double slash and names with spaces are still quoted. There is a new test for the space case.

With `mark-directories` off, `cd usr<TAB>` under `CDPATH=.` now inserts `usr` without a slash, the same as with `CDPATH` unset.

